PR #3, opened 2015-09-29
setup.py script to install MXCuBE 3

mguijarr:
renamed mxcube.py to mxcube3, installs as a script
generated MANIFEST file (python setup.py sdist)
added requirements.txt


meguiraun:
Can we add the following to mxcube3 launcher file so we do not need to do an install to run the server:
```python
try:
    from mxcube3 import app, socketio
except:
    from src import app, socketio
```
Or anything more elegant if you prefer.

mguijarr:
We may want to rename `src` to `mxcube3` instead of doing the `try...except` thing ?
It has implications on the submodules (that would need to be reinitialized)
If you think it's a good idea I can do it otherwise if you prefer the `try...except` just do it :)

meguiraun:
let's go with the simplest solution, renaming is enough. thanks

mguijarr:
I did the change, I renamed `src` to `mxcube3`, and so the `mxcube3` script is renamed `mxcube3-server`. However I can't make the pull request while my other PR is pending. I will do it as soon as the previous PR gets accepted.

amilan:
PR accepted!! You are free to continue ;)

mguijarr:
Thanks :)
In fact, it's already all done, there is some magic I can't explain: the PR from yesterday got updated with latest master from my fork, and so latest changes we were talking about in this thread are
already merged.
Perfect, but weird... Anyway.

Issue #10, opened 2015-09-29
Sample Centring


meguiraun:
Manage the sample positioning. In a manual or automatic way.
* Manual centring:
  * Usage of javascript mouse events for retrieving the three click points
  * Link to diff for rotating sample after every click
  * During the centring procedure messages will be passed between the server and the interface (actually the user)
* Semi-automatic centring
* Automatic centring

meguiraun:
### Update 30/10/2015

(See also #9) 
The following features are implemented (and tested in 911-3):

##### Lucid centring
To be deeply tested because I am having an offset in the final centring position. It can be due to a mismatch of the image size?
The user only needs to click the launch button in interface.
 
##### 3-Click centring

The user clicks the launch button and supplies the three positions for the centring. No message is given to the user, it is assumed that the user knows how to center...


##### To be done
Mostly in #9...

Provide a really useful way of knowing the status of the centring, a modal, pop-up, on canvas text, etc...

meguiraun:
Closed as main features are implemented.

Issue #9, opened 2015-09-29
Sample Video


meguiraun:
Display the sample centring view. With the following features:
* Display the sample video from the diffractometer
* Ability to take snapshots of the image
* Ability to control the motors, light, etc.
* Tools to measure distance/angles and draw lines/angles (not needed in the beginning)
* Draw beam position
* It should also display the queue for the current mounted sample (getting the pre-configured experiment from the mxcube3-queue), but also must provide tools for modifying parameters, collections...
* All these features needs code in the server as well

Preliminary layout from https://github.com/meguiraun/WebLayoutTests:

![sampleview](https://cloud.githubusercontent.com/assets/12758327/10160034/8df0ceba-669b-11e5-9e39-bd2f3953b428.png)

Note: modified slightly after the discussion about the web layout and assigned to MaxIV (which means to me...) as we spoke in the meeting.

meguiraun:
### Update 30/10/2015

The following features are implemented (and tested in 911-3)

* Display the sample video, taking images from the udiff camera HO, the image sending rate is set in the camera HO itself. I removed the sse stuff and use the ```img``` html tag. The position/scale of the image/canvas must be adjusted properly, resizing the window causes weird views...

* Sample centring:
 * Lucid centring: to be deeply tested because I am having an offset in the final centring position
 * 3-Click centring

* Motor movement

* On canvas drawing:
 * Zoom text: mapping between the current zoom level (1, 2, ..., 9) and the real scale is missing. Currently 'Zoom x' is shown. Also, the scale bars might need to be re-scaled according to the zoom level.
 * Distance measurement: the distance in um between two clicked points is displayed over the image. The calculation does not take into account different pixelPerMm for the different zoom levels.
 * When 3-click centring the clicked point is displayed and kept in the canvas (perhaps I should implement a max number of points to display).

* Some interface improvements (tooltips, more icons, bug fixes in the queue, etc.)

* To implement:
 * Save centring positions, show/hide, move to the centring positions...
 * Download image
 * Display the most important messages more clearly: a small modal displaying the centring result?
 * Queue:
   *  Fill the queue parameters with the real ones and launch/stop the collection. So the link with the qmo is needed
 * Find appropriate icons

##### Description of the icons
From left to right

1. Save snapshot: it is a call to ```camera_hwobj.take_snapshot()```
2. Measureme distances: click and the measurement state is activated
3. Nothing
4. Same as 1
5. Automatic centring
6. 3 Click centring
[...]

Issue #6, opened 2015-09-29
User interface and server message handler


meguiraun:
Extend the existing messaging system to the web. Flask-socketIO (websockets) can be used to send the pydispatcher’s signals to Javascript functions. It is also possible to define different namespaces over a single websocket. Flask-SocketIO transparently downgrades itself for older browsers that don’t have WebSocket, SocketIO emulates the connection using one of the older solutions, the best one available for each given client.

Two kind of signals:
* Signal from HO to HO (hidden for the user, so no action is required)
* Signals from HO to User interface

In any case, whenever a signal is intercepted for sending to the ui, it must fulfill a generic structure (to be defined), in order to ease the data retrieval in the web interface.

Code example:
```python
from flask.ext.socketio import emit
#generic and unfiltered callback, it just will send all messages to the web interface
def signalCallback(signal, sender, *args): 
	socketio.emit('test',{'data':'A signal', 'signal': signal, \
		'sender':str(sender.__class__).split('.')[0]}, \
		namespace='/test')
#in the init code of the resolution hardware object (as example) we link the 'DeviceReady' message with the callback function
self.resolution.connect(self.resolution,'deviceReady', signalCallback)
```
We need to add `self.someHO.connect(hw_obj, 'aSignal', 'aCallback')` for each message we want to propagate to the user interface. For that, we need the lists of messages for each hw_obj, such as the following:

```python
MaxLabMicrodiff_signals = ['diffractometerMoved', 'minidiffReady', 'minidiffNotReady',
'phiMotorStateChanged' ......]

BL9113MultiCollect_signals = ['collectReady', 'collectImageTaken', 'collectReady',
'collectStarted', ....]
```
And add these lists to a config.py (or config/messages.py or whatever)

meguiraun:
See [wiki] (https://github.com/mxcube/mxcube3/wiki/User-interface-and-server-message-handler)

PR #36, opened 2015-11-05
Sample centring PR

meguiraun:
Hi, 
Latest stuff I did, sample centring is working (both lucid and 3click). And some minor additions, see #9 and #10. Also see #6  in the file ```routers.signals.py```.

We can review&discuss&(accept if ok) tomorrow in the meeting if you want.

cheers,

mikel


mxcube3-server
```diff
@@ -5,4 +5,4 @@ monkey.patch_all()
 from mxcube3 import app, socketio
 
 if __name__ == '__main__':
-    socketio.run(app, host='localhost', port=8081)
+    socketio.run(app, host='192.168.56.101', port=8081)
```
mguijarr: I think we cannot let an IP address like this, in fact yesterday I changed 'localhost' to '0.0.0.0' in order to
bind to all interfaces, and PR was accepted. Would it be ok for you too ?
I would recommend to merge your repository before redoing the PR in order to sort out this kind of details.

mxcube3/__init__.py
```diff
@@ -4,32 +4,42 @@
...
 ###Initialization of the HardwareObjects
-from .HardwareRepository import HardwareRepository 
+from .HardwareRepository import HardwareRepository, setLogFile
 HardwareRepository.addHardwareObjectsDirs([os.path.join(os.path.dirname(__file__), 'HardwareObjects')])
 # this is to allow Hardware Objects to do 'from HardwareRepository import ...'
-sys.path.insert(0, os.path.dirname(__file__)) 
+sys.path.insert(0, os.path.dirname(__file__))
 
 hwr_directory = cmdline_options.hwr_directory
+hwr_directory = './mxcube3/HardwareObjects.xml/'
```
mguijarr: Sorry to bother again... In my case I use existing xml files from another directory ; would it be ok if you start mxcube3-server with `-r` option instead of hardcoding the directory ?

mxcube3-server
meguiraun: Yes, I agree with yesterday's PR.
Ignore my changes. It was like that due to my vm configuration.

mxcube3/__init__.py
meguiraun: Yep, no problem again... It is me the one who should say sorries...

meguiraun:
I did not notice these changes I did during my tests. Reject this PR and I will merge first.

PR #38, opened 2015-11-09
Removed some hard-coded paths (see last PR) and added command line options

mguijarr:
Last committed code introduced some hard-coded paths, this PR removes that.

The solution for not having hard-coded paths can be the following:
- use of "-r" command line switch to specify HWR directory
- use of "-s" command line switch to specify beamline setup hardware object (defaults to "/beamline-setup"
- normally, diffractometer and other devices are all loaded by beamline setup so we can use this
object to get access to all others
- except "queue-model", so I added a "-q" switch

Otherwise be careful not committing hard-coded paths ;-) 
Anyway the command line options are needed so it's good to add them.

I also enabled absolute imports, it is the new Python standard (since 2.7 I think ?) ; also, it fixes a weird
bug I had with Hardware Repository package (I am still running Python 2.6 here...). Should not make any difference for you.

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -25,7 +25,7 @@ opt_parser.add_option("-l", "--log-file",
 opt_parser.add_option("-s", "--beamline-setup", 
                       dest="beamline_setup", 
                       help="Beamline setup HWR file", 
-                      default='/beamline_setup')
+                      default='/beamline-setup')
 opt_parser.add_option("-q", "--queue-model", 
                       dest="queue_model", 
                       help="Queue model HWR file", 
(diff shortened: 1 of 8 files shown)


meguiraun:
Sorry... I missed that one... the next time I will run ```git diff +caffe```...

mguijarr:
No problem don't worry it makes a joker for me next time I will miss something ;)

mguijarr:
If you agree with those changes, can we accept the PR ? So I can proceed with other changes like 2.1 version of HardwareObjects and then the webpack PR

mguijarr:
Thanks!!!

PR #42, opened 2015-11-11
Webpack + components + attempted MXCuBE 3 application layout

mguijarr:
I apologize in advance for the big PR with tons of changes... In addition to configuring webpack and
the whole tool chain, I worked on some components: SampleGrid (not finished yet), Login, Error notification. I tried to integrate existing work (Sample Centring etc) but I could not make it at the moment. Nevertheless I propose an architecture/schema for inserting "Data Collection", ie. sample centring and queue.
I would need some help from Mikel in order to integrate his work in the "new" development tree and code (if we all agree with my changes).

If we survive this PR, I can tell you we are on good rails ;)

It's bank holiday today here in France, I propose to get in touch with you all tomorrow maybe we could have a kind of meeting again ? If you have time to review all this before ?
We also need to think about Barcelona's meeting: what we can present to others, for example, as we need to help preparing the agenda.

I wonder if it would be a good idea for me to come to MAX IV for 2 days or so on the 23-24th ? Maybe it can wait our Barcelona meeting - but I have the feeling we need more real life pair programming and this moment of the project.


meguiraun:
Hej Matias,

You are scaring me ; ) … I will have a look at this today. We can meet tomorrow, not sure if I will be able to check everything but it is a good idea anyway, also I do not know if all of us are available. I will be back to you about the meeting tomorrow later on during the day. Is there any time you can’t do the meeting?

It is great if you come to us by November, we booked our beamline for those days so we can play before the Barcelona meeting. Let me first talk to Marjolein&Antonio about  it (she is traveling so it will not be a quick response)

cheers,

mikel

amilan:
Hi Matias, 
Are you available for a meeting on Friday at 10? This time we can try with Google Hangout instead of Skype.

About you coming to visit us, I think it's a great idea. At this moment of the project the communication is a key factor, so let's discuss about the details (maybe here is not the best place, let's continue in an email).

meguiraun:
Hi Matias, some comments:
* The web part is already compiled, could you remove the compiled files? There are some weird files in the static folder
* We tried to compile and we found that grunt is installed globally, could you add it to the package.json?
* ```npm start``` fails with "Cannot find module './backend_server.js'"

Regarding the layout, probably is best to discuss face to face like we did when we had the last meeting. So, no problem for me as you proposed, we can discuss later how to switch/jump/move between the views. In theory (cof cof) we can develop the components safely and adapt the main view ~easily. However, I still need to adapt the SampleCentring to the webpack stuff.

I still need to look at the server side, but a first question... do you use mockups for the lims, udiff, etc? Should we create a "HardwareObjects.xml"  folder for storing mockups? (to be sure all of uses the same) I am using the mxcube mockups but I made some modifications to the camera. (Of course, I also have tested everything in the real environment)

cheers,

mikel

meguiraun:
I do not see anything apart from the previous comment. If you can clean the precompiled files, fine for me. The idea is that everyone should compile locally.

* backend_server.js is named as backend_server.js.example, so actually it was there

mguijarr:
Hello :) Sorry I could not reply before.

Indeed, `backend_server.js` is supposed to be a file each one modifies to set the URL of the back end server ; I decided to provide an example file, that is part of the repository, but the final `backend_server.js` is not part of the repository (since everybody has to change it). The webpack config uses it, so each one has the same webpack config even if the backend server is different.
I tried to explain this [here](https://github.com/mxcube/mxcube3/wiki/Webpack-and-frontend-development-stack#latest-addition-to-webpackconfigjs-proxy-feature).

Is it ok ? Or do you have a better idea ?

About grunt: we do not depend on grunt, as far as I know. I guess some modules are depending on grunt, then. Do you know which one(s) ?
I just checked the list of modules I have, there is not grunt although the compilation works fine.

If it turns out grunt is needed, it's easy to add packages in `package.json`:
`npm install -D name_of_the_package`

About additional static files: I thought it was useful to have a version of the client files ready as soon as someone clones the repository in order to be able to start the server and actually see the web application. Moreover there is another use case: here at ESRF we tend to have old operating systems on beamlines, it's difficult to have a recent version of NodeJS so we can't run `npm` etc. we can't easily compile the files. So I develop on my computer, and precompile for testing on beamlines, and for this it's convenient if it's part of the git repository.

Of course it means sometimes files have to be updated, to keep static files with latest version of
the UI and the web application.
Do you still think those files should be removed or can we keep them ?

mguijarr:
It is a very good idea to have mockups for performing tests etc. directly configured with xml files within the repository I open a specific issue for this: #45.

meguiraun:
* backend_server.js: ok ok... I see now. It's fine for me.
* grunt... I redownload again your repo ... no grunt error messages... I do not understand... Ignore my previous comment regarding this. 
* I still think that precompiled files should be kept away, each of us should compile its own. With that we keep a cleaner repo. Anyway, we can discuss it tomorrow.

PR #64, opened 2015-11-27
fix Login

JieNanMAXIV:
Have tested with the maxlab ldap and ispyb env.


JieNanMAXIV:
Login, use the middleware thunk for async, tested with maxlab ldap and ispyb

PR #72, opened 2015-12-07
indentation error fix + usage of new inline function definition to avoid '.bind(this)'

mguijarr:
Keep on with the good work guys :)


meguiraun:
Gladly accepted.

PR #88, opened 2015-12-21
Use of a linting tool / code formatting

mguijarr:
I propose to use eslint to check our code automatically when doing "npm start". It gives warnings/advices on how to improve code and checks for mistakes. eslint understands ES6 and React, with the appropriate plugin.

In addition, I propose to use esformatter : it takes code and formats it in a nice way, it also understands the HTML in JSX files (plugin installed).

Do you agree ?

If so, don't forget to do `npm install` to get missing packages in your repositories otherwise it won't work.
The formatting tool can be started by doing `npm run esformatter [... options to esformatter ...]`.

I would be happy if one day we decide to do a __formatting only__ commit, i.e. transforming all our files at the same time using the esformatter tool, and if we try to use it regularly afterwards to keep nice code consistency in term of formatting for all our files. The same can be applied to Python files. We can even think of a git pre-commit hook to format things automagically... But since it would change code we are working on, I am not sure it is a good idea.


mguijarr:
Welcome back Mikel ! Hope you're better now - too bad you had flu during your holidays in Spain.
It's good to see you back accepting pull requests ;)

meguiraun:
Thanks! I am not fully recovered but I am back! My finger is on fire!

cheers,

mikel

PR #91, opened 2015-12-22
Manually mounted feature

mguijarr:
This PR adds the "Manually mounted" feature, it is made available thanks to a split button, instead of
having just "Check sample changer contents" and additional dropdown allows to select "Manual mount".
This disables ISPyB synchronization, and leaves the sample grid with only 1 sample, ie. the manually mounted one.

__Question__: the Queue has to be cleared, so I added a CLEAR_QUEUE action. I am not sure if this is the best way, since it just removes the queue on the UI side I guess it should be removed from the server too ? Or to switch to another queue ? I guess switching to another queue is better, if people want to come back to the previous queue they were doing before. Like, someone uses the sample changer, then puts a manually mounted sample, then comes back to using the sample changer... I think it is a legitimate use case.

### DIFF
--- a/mxcube3/routes/Login.py
+++ b/mxcube3/routes/Login.py
@@ -26,85 +26,20 @@ def convert_to_dict(ispyb_object):
 def login():
     beamline_name = os.environ.get("SMIS_BEAMLINE_NAME")
     content = request.get_json()
-    Proposal = content['proposal']
+    loginID = content['proposal']
     password = content['password']
-    proposal = "".join(itertools.takewhile(lambda c: not c.isdigit(), Proposal))
-    prop_number = Proposal[len(proposal):]
-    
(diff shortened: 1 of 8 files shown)


meguiraun:
I would say that the queue in the server should also be cleared in order to be synchronized all the time, we already have an api call for that:
```
"/mxcube/api/v0.1/queue/clear", methods=['PUT']
```
We also have the call for saving (now to a file in the server, retrieval is still missing). I would say for now it is enough with your proposal, i.e. clear existing queue and move forward, we can open an issue for the retrieval of saved queues later (from file or database).

I can not accept this PR as it says there are conflicts, can you have a look?

cheers,

mguijarr:
I fixed the conflicts - the PR can be merged now.

I will use the 'clear queue' API to clear the queue on the server in next commits...

PR #93, opened 2016-01-12
more with the queue

meguiraun:
a few more things for the queue/server api
- executing sample entry (added a 'checked' ~flag synchronized with the client)
- add some basic parameters to the methods' models
- fixes here and there

Notes:
- In order to avoid some dependencies in the queue_entry (_view and widget_color stuff) I added a Mock() so it can run and avoids rewriting the hardware obj
```
aQueueEntry._view = Mock() #associated text deps
aQueueEntry._set_background_color = Mock() #widget color deps
```
- The execution of an entry does not finished successfully with the current xml mockups (it fails on internal queue_entry.execute), I have a rogue dummy motor as well as missing DataAnalysis (no mockup available). You can launch the execution through the api anyway (not implemented in the client yet) but you will always get the 409 error message, meaning some error happened in the call.


mguijarr:
good idea to use Mock module

PR #94, opened 2016-01-13
Updated build process to work with babel 6 and node 4

bolmsten:
To make it easier for users to install MxCUBE I have updated the build process to work with the latest packages of babel and the newest version of node.


mguijarr:
Is it backward compatible (i.e does it work with current Node without changing anything ?)

meguiraun:
I tried with my setup (node v4.2.2 and npm 2.14.7), a clean clone from fredrik's fork and it works. A few warnings when ```npm install```, although we all know what to do with a warning...
```
npm WARN package.json Dependency 'babel-polyfill' exists in both dependencies and devDependencies, using 'babel-polyfill@^6.3.14' from dependencies
npm WARN package.json Dependency 'css-loader' exists in both dependencies and devDependencies, using 'css-loader@0.17.0' from dependencies
npm WARN package.json Dependency 'style-loader' exists in both dependencies and devDependencies, using 'style-loader@0.12.3' from dependencies
npm WARN deprecated lodash@0.9.2: lodash@<2.0.0 is no longer maintained. Upgrade to lodash@^3.0.0
npm WARN deprecated css-list@0.1.3: Deprecated.
npm WARN deprecated lodash@1.0.2: lodash@<2.0.0 is no longer maintained. Upgrade to lodash@^3.0.0
```

bolmsten:
I have noticed that the hot-reloading does not totally work so I will continue investigating

mguijarr:
Ok so let's wait a bit... By the way thanks a lot Fredrik for taking care of this I tried once in my corner
to pass to the new babel and it didn't work immediately so I postponed it.

My current Node version is... 0.12.7 ; do you think it still work for this version of Node ? Installing
Node can be painful on old architectures (sorry).

meguiraun:
I though we specified the versions for node... see [wiki] (https://github.com/mxcube/mxcube3/wiki/How-to:-Installation). It is written by me ... so I might be something unilateral...

I remember having some trouble with older versions of node ... I do not remember exactly the problem but some feature was not working... I will try with one of my virtual machines that has old node

meguiraun:
tested in Centos 6, node v0.12.7, npm 2.11.3. The behavior is the same, hot reloading does not fully work. So, once the hot reloading issue is fixed we can test again for all of our working versions (hopefully only two...).
For earlier versions of node (e.g. 0.10.36) does not work (this PR neither the original code).

mguijarr:
Thanks a lot for those tests Mikel. I think we can forget about pre-0.12 versions of Node, personally
I am happy if it works with Node 0.12.7 which seems to be the case so that's excellent.

Let's wait for the fix for hot reloading then we can probably merge this PR.

bolmsten:
I have solved the problem and a new PR request will be up tomorrow.

mguijarr:
Thanks for the PR.

However I have a little problem with it : it mixes the babel 6 and Node 4 work with other unrelated things (Queue refactoring, etc.)

Do you mind to create a local branch in your repository with only changes for babel 6 and node 4
(this is what this PR is about at first), then push this local branch to your remote repository on github,
then redo the PR from this branch to mxcube3/master, then do another one for your other changes if
needed ?

I am sorry to be a bit meticulous with this but I think it's important what do you think ?

PR #95, opened 2016-01-15
Babel 6

bolmsten:
This PR holds both babel 6 and the queue refactoring,  the reason it is in the same was that babel 6 required some refactoring to be done. I agree with you Matias that it could be broken up more but this time I think we will keep it as it is, because the amount of work would be quite large. 

This version has been tested by Mikel and seems to work. I have commented out eslint as it needs to be changed to babel-eslint inorder to work with ES7 functions.


mguijarr:
Grrr

PR #96, opened 2016-01-17
A Separate Login Page

JieNanMAXIV:
Add a separate login page as the start page, with the following features / changes,
- load the synchrotron_name and beamline_name from session HO (session.xml) and display on the login page
- depending on LoginType that is defined in dbconnection.xml, the placeholder of LoginID is either User or Proposal
- after submission, the form is disabled and a loader icon is displayed while waiting for the response from Server
- Proposal info and SignOut are added into NavBar props (removed from login props), for displaying proposal info and logging out.
- Login form can be submitted by pressing "Sign In" button or return key.
- Login error msg from Server is displayed on the top in the errorNotification panel.


mxcube3/ui/components/Login.jsx
```diff
@@ -1,50 +1,103 @@
...
-import React from 'react';
+import React, { Component, PropTypes } from 'react'
+import { Redirect, Router, Route } from 'react-router';
 import classNames from 'classnames';
 import "bootstrap-webpack!bootstrap-webpack/bootstrap.config.js";
-import { Nav, Input, ButtonInput } from "react-bootstrap";
+import { Input, ButtonInput, Form, Well } from "react-bootstrap";
+import './Login.css';
+import Main_login from './Main_login'
+import SampleViewContainer from '../containers/SampleViewContainer'
```
mguijarr: Why the Login component has to import SampleViewContainer ?
I think it is better if Login component is only for the login feature.
Moreover, this is a component not a container so I don't think it
should contain big parts of the application.

mxcube3/ui/components/Login.jsx
```diff
@@ -1,50 +1,103 @@
...
+  },
 
-    logOut() {
-        this.props.signOut()
-    }
+  signIn: function() {
+      let proposal = this.refs.proposal.getValue();
+      let password = this.refs.password.getValue();
+      this.props.signIn(proposal, password);
+      $("#submit").prop('disabled', true);
```
mguijarr: Direct DOM manipulation in an anti-pattern in a React component as far as I understand.
Use directives in the `render()` method to render conditionally, based on properties.

mxcube3/ui/components/Login.jsx
```diff
@@ -1,50 +1,103 @@
...
+        } else {
+        $("#submit").prop('disabled', false);
+        $("#ajax_loader").toggleClass('hidden',true);
+        login_input_form = (
+          <div>
+          <div className="row row-centered">
+          <div>
+           <img src="img/mxcube_logo20.png" className="img-logo"/>
+          </div>
+          <h3 >Welcom to  {beamline} at {synchrotron}</h3>
```
mguijarr: typo: Welcom -> Welcome

mxcube3/ui/components/Main.jsx
```diff
@@ -1,30 +1,21 @@
...
-import NavBar from './NavBar'
+import Login from '../containers/LoginContainer'
 import { ErrorNotificationPanel } from './Logging'
 import PleaseWaitDialog from './PleaseWaitDialog'
 import './Main.css'
 
 export default class Main extends React.Component {
     render() {
-        return (<div className="container-fluid"> 
+        return ( <div align="center">
```
mguijarr: I thought we could have weird layout problems if the main div in a Bootstrap based app is not a container class,
due to the margins or other properties of other classes like `row` for example.

mxcube3/ui/components/Login.jsx
```diff
@@ -1,50 +1,103 @@
...
-        let password = this.refs.password.getValue();
-        this.props.signIn(proposal, password);
-    }
+getInitialState: function() {
+    return {
+      login_info: {synchrotron_name: " ", beamline_name: " ", loginType: ""}
+    };
+  },
+  componentDidMount: function(){
+    $.ajax({
```
mguijarr: I am not sure but I think getting login info should go to the redux scheme: triggering an action, etc. instead of direct AJAX call ?
Instead of a query, another possibility to retrieve login info may be to have it set by login action through props.

JieNanMAXIV: Hi Matias,

Thanks for all the comments! I will fix them and make a new PR.

Best,
Jie

PR #107, opened 2016-01-27
A first version of session management

bolmsten:
This will retrieve the latest session from the server in case the client has crashed. This is as the title says a first version as the session is:

1. Not being saved to a particular user
2. The client sends the whole state to the server for saving on queue actions

In the future the server will save this data to the logged in user and it will also create the data needed to retrieve a session itself without the need for it to be sent from the client.


meguiraun:
sorry, will not work.. missing variables in the Queue

Issue #58, opened 2015-11-25
Session management


amilan:
The session history must be kept in the web server, in case the client crashes we could be able to redraw the client interface at the same state as it was. Note that this requires an study on the available and useful web tools.
It must be saved automatically and with the possibility of save it manually.

mguijarr:
Somehow linked with #90

mguijarr:
Solved by PR #333

PR #108, opened 2016-01-27
A first version of session management (bis)

meguiraun:
This will retrieve the latest session from the server in case the client has crashed. This is as the title says a first version as the session is:

Not being saved to a particular user
The client sends the whole state to the server for saving on queue actions
In the future the server will save this data to the logged in user and it will also create the data needed to retrieve a session itself without the need for it to be sent from the client.

Sorry, fixed missing vars

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,41 @@
+# MXCuBE 3 (web)
+
+MXCuBE stands for Macromolecular Xtallography Customized Beamline Environment.
+The project started in 2005 at [ESRF](http://www.esrf.eu), since then it has
+been adopted by other institutes in Europe. In 2010, a collaboration
+agreement has been signed for the development of mxCuBE with the following
+partners:
+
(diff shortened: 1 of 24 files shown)


mguijarr:
I didn't have a look at the code yet, but I wonder if you are using Flask built-in session management feature ? Or something like beaker ?
If not, is there a reason why not ?
I am volunteer to try to use the Flask session thing if you want.

meguiraun:
mmm, I was trying to use *flask-login*, but I am stuck with the flask context (or with something else) so I did not advance too much, also we would need to store the current internal queue (clearly needing a redesing as for #110 ) and switch from user to user. So I thought that it is quite a big task and left for after milestone Jan16. 

This PR kind of solves  the issue #58, as we redraw the interface, but we do not have any user login...

But of course, if you think that you could do something with the flask session go ahead!

PR #123, opened 2016-02-05
Reduxed SampleView with new GUI

bolmsten:
This PR contains the Reduxed SampleView with a new GUI. I have also made small changes to the rest of the CSS to correspond to the new GUI changes. 

This is a first implementation of the new SampleView and it uses the Fabric JS library to draw on the Canvas. This seems to be the best library for this use at the moment but it does not build correctly in npm so it was added manually. It seems that they are working on fixing this issue so it will be included into the npm build later.

At the moment you can right click on points and see what actions are available. You can also select two points and draw a line. All this corresponds to the discussed functionality in Barcelona.


mxcube3/ui/components/Login/Login.jsx
```diff
@@ -28,15 +24,17 @@ class Login extends React.Component {
   signIn(){
       let fields = this.props.fields;
       this.props.signIn(fields.username.value, fields.password.value);
-      this.setState({ show_loader: true });
+      $("#ajax_loader").toggleClass('hidden',false);
```
mguijarr: I guess there is a regression here... jQuery is back.
A merge failed maybe ?

mguijarr:
Well done !

I will merge it as soon as the jQuery problem is fixed (see my comment in code)

bolmsten:
It should be reverted now

bolmsten:
Good catch

mxcube3/ui/reducers/index.js
```diff
@@ -3,6 +3,7 @@ import login from './login'
 import queue from './queue'
 import samples_grid from './samples_grid'
 import methodForm from './methodForm'
+import sampleView from './sampleView'
```
JieNanMAXIV: should be sampleview, as the file sampleView.js does not exist.

mguijarr: Yes! You're completely right

PR #125, opened 2016-02-16
Sample Centring API 

meguiraun:
- Added or fixed api calls (centring abort, get positions of the motors)
- Draw centred positions, zoom, beam, over the sample image directly in the server (I do not like any kind of drawing in the server too much tough)
- and basically do not lose track with bolmsten

Note: tested in real environment.


mguijarr:
Indeed I think it's better if we add overlays on the client, using Fabric or another library. We will talk about that next week :) Thanks for this.

meguiraun:
yes, I agree with you. However, the server also needs the picture with the overlays... I also think that the server must also be fully featured. In any case, based on what we have now we can *easily* switch between server and client rendering.

Let's discuss next week!

PR #131, opened 2016-03-03
Signals

meguiraun:
some fixes from last week. I will split all I have in several pr-s


mguijarr:
So it this the 1st PR to merge ? Or are you going to split this one and then I should discard it ?

meguiraun:
this is the first one, if this one is accepted I do not plan to change this one (unless bug…)

PR #133, opened 2016-03-04
Clicking

meguiraun:
Some fixes with the way we pass the 3 clicks.
Now there is a spawn_later(3 s) just after the third click so the diff has time to go to the centred position, and then we store the position.

### DIFF
--- a/mxcube3/routes/SampleCentring.py
+++ b/mxcube3/routes/SampleCentring.py
@@ -11,7 +11,6 @@ import json
 import signals
 
 SAMPLE_IMAGE = None
-
 posId = 0
 
 # ##all drawing to be moved into ~shapehistory...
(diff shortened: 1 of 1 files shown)


mxcube3/routes/SampleCentring.py
```diff
@@ -511,6 +490,27 @@ def aClick():
     else:
         return Response(status=409)
 
+def waitForCentringFinishes(): 
+    if mxcube.diffractometer.centringStatus["valid"]:
```
mguijarr: From what I see, this doesn't wait : it assumes the centring is finished when it gets called.
As you said it is much better to use signals ; we can't rely on the 3 seconds for all diffractometers, this is not reliable.
I propose you to refactor it like this: create a global `CENTRED_EVENT` event (`gevent.event.Event object`),
connect the centring successful signal and centring failed signals to a global function that would just
do `CENTRED_EVENT.set()`.
You would also need to change the `saveCentringWithId` function to return a generator so Flask can handle other requests while waiting for end of centring:

    if len(clicks) == 3:
        CENTRED_EVENT.wait()
        # then put code like current waitForCentringFinishes()

Hopefully there is nothing else to do ; I know bottle much better than Flask, maybe Flask requires something to be able to handle requests while waiting for the centring event but I hope not since we use gevent it should be automatic.

meguiraun:
modified as you suggested, but I do not see the necessity of the Event(), just linking ```centringSuccessfull``` and ```centringFailed``` signals to the ```waitForCentringFinishes``` would work.  I commited&tested only with that.

Am I missing something regarding the usage of an event for this?

mguijarr:
I thought you wanted to wait for procedure to be done before continuing... If there is no need to wait then there is no need for event.

PR #134, opened 2016-03-07
 added front light

meguiraun:
and fixes with the get_status if a motor is not present. Note that I renamed the api calls for the On/off.

We slightly changed the names of the roles (wagobacklight, backlight, wagofrontlight, frontlight) which might have effect in your tests, see SampleCentring.py#L419

### DIFF
--- a/mxcube3/routes/SampleCentring.py
+++ b/mxcube3/routes/SampleCentring.py
@@ -422,7 +422,7 @@ def get_status():
     try:
         for mot in motors:
             motor_hwobj = mxcube.diffractometer.getObjectByRole(mot.lower())
-            if motor_hwobj not None:
+            if motor_hwobj is not None:
                 if mot == 'Zoom':
                     pos = motor_hwobj.predefinedPositions[motor_hwobj.getCurrentPositionName()]
                     status = "unknown"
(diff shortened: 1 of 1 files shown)


mguijarr:
Why role names have been changed? I don't like the 'wago' prefix, it seems too much specific for a kind of hardware

meguiraun:
there were changed to give them a more meaningful name, also to clearly differentiate back and front light md motors. Does it cause any trouble to you? 

We can remove the 'wago' prefix for something else.

meguiraun:
I think I misunderstood your objections... Will this PR be accepted if I switch back to the original role names? Or it is enough if I remove the 'wago' keyword?

mguijarr:
Hi Mikel,
Sorry I forgot about this PR... Indeed, I think it would be great to remove the "wago" prefix.
Then I would happily accept it ;)

meguiraun:
No sorries, too many things to do during a shutdown! 
I renamed as 'backlightswitch', etc. Hope it makes sense.

beteva:
Hello,
seems to me something have been lost, as I have changed wago to actuator some time ago. There is an in/out hardware object, which deals with that and the respective calls in some bricks.

meguiraun:
We are using the MicrodiffInOut hwobj, this was only related to the name of the role. Clearly a bad renaming since it has added more confusion... sorry about that.

PR #139, opened 2016-03-14
Work since... wait for it... last MAX IV workshop

mguijarr:
(subject line inspired by How I met your mother)

Here are my changes since last workshop, I rebased my code on top latest `mxcube3/master` in order
to follow latest commits including changes by Mikel and Frederik. Hopefully I didn't make mistakes,
and that everybody still find its things.

I made little changes to the CSS and layout, mainly trying to simplify a bit, removing unneeded divs,
making more use of Bootstrap classes, etc.
I chose to rename _Method_ to _Task_ everywhere, at least in front-end code, because I really think
the word _task_ is more appropriate ; indeed, a centring is not a data collection method, an energy scan neither ; _task_ is more general. I hope you agree with this.

I added buttons to add tasks (Characterisation, Data Collection) to queue for samples using the samples grid. Added tasks are "clickable", it opens the parameters window. I also added a close button, but it doesn't seem to work for the moment (sorry - there is an error in the server when
deleting an item).

I have to admit I had extreme difficulties finding my way in the server code :-(, with all the IDs and I find some parts of the code quite confusing. Probably it's due to the fact that I focused on the front-end
development those last weeks and I only tried to reuse existing code without really understanding it. But when I tried to understand, ouch. It hurts !

I have the feeling there are many features which need more love, in fact I often get error messages in the server which lead to bad behavior in the front-end. Before doing deeper debugging I would like to know what do you guys think about
the current state: what is ok for you, what is not ok, etc. ; maybe we should spend some time to
create issues for the problems in order to keep track of what has to be done in term of bug fixing and refactoring ? 

I found some problems with signing out : it keeps the old queue, it seems, or something like
that, as a result when signing out and signing in again I could not add samples to the queue because
they were already in. The queue is not flushed it seems. I tried to fix the problem by setting it to
`None` in the sign out route. But it seems I am missing something still :-(

I have a question: how do you test the sample view ? Do you test it with real hardware or with the
mockup ? Just starting MXCuBE 3 server with current distributed mockup files gave me an error
about missing motors and the sample view didn't show. So I changed the initial state to be more
tolerant to missing motors, but as far as I see there are motors in `minidiff-mockup.xml` so I don't
understand. Maybe your mockup files are different ?

I have the feeling most problems I have are due to misconfiguration of my xml files or some
misunderstandings from my side of how it should behave. I wish you could all be here: there is
nothing like a bit of pair programming to solve this kind of issues.

There is a good news from ESRF: Marcus just started his work on MXCuBE 3, at the moment we
are doing the _Beamline Setup_ part. In the next days/weeks we can probably take care of more
tasks.

I propose to organise a Skype meeting after Easter. And to talk about your next visit to ESRF
in April, as we said.

Your turn... The ball is in you court ;)


meguiraun:
looks like your sweet revenge!! hehe

You are absolutely right with the messy server code, I should update the gh-pages and add some code comments and perhaps extend the wiki... but yes, working with the id-s is a pain (and there still things to be added).  I will also carefully test for the bugs you mention.

We are testing the most recent socketio library with python2.7, we get a clear stability improvement.
(Flask v0.10.0, Flask-SocketIO v2.2, python-socketio v1.1).

We test the sample view with real hardware (the md2 we moved to our lab, lucky us). Probably we added things that the mockup did not like.

We will go through all of this hopefully today...

Welcome Marcus!

mguijarr:
Thanks for your quick reply Mikel.

I never managed to get newest socketio client library working with Flask so I am happy you gave
me the version numbers ! I will try that.

Let's update `requirements.txt` then ;)

Looking forward to hear from you,
Matias.

bolmsten:
Hi Matias, 

Great work! I have now installed it and started to test. One thing, I noticed that we are now using jquery and that causes it to fail on npm start. To fix this I needed to do "npm install jquery --save", it is possible that you have this installed globally so you did not get this error. 

I will continue to test and get back to you.

Br Fredrik

JieNanMAXIV:
We tried to test with both real hardware and mockups, and also for 2.2 and 2.1 branches. I just did a fresh setup, and I could only see phi motor was missing during initialization, which I can make a PR for the mini-diff.xml soon. For the sample video, the failure was from new_sample_video_frame_received in routes/SampleCentring.py, I think Mikel has better answers.

meguiraun:
Several of the issues you are facing are bugs introduced by us in the last PR:

* delete sample: ```mxcube.queue.queueList...``` does not exist anymore, so call failed
* camera: saving to file before sending works with the real image from the md2, directly sending no
* state issues:  needs a fix, I would suggest to do a quick fix (e.g. remove the session on login) so we do not have existing id-s. We can turn on this and properly fix later on?

In addition, there is also the jquery issue and the wrong motor names. The modal does not update parameters.

We can accept this PR as it is and we fix these issues in the next days.

mguijarr:
Thanks for your comments friends !

@bolmsten : I didn't know I was making any new use of jQuery... I have an old npm which is maybe installing dependencies automatically, which has been removed from newest versions AFAIK. Indeed, we depend on jQuery for some things (one I remember:  the x-editable widget, that we plan to use more extensively in Beamline Setup, more news later...)
@JieNanMAXIV , @meguiraun : what is the wrong motor names issue ? I don't understand. Any help would be appreciated. For the camera, well normally the new frame is a jpeg frame so we don't need to save it and open it again I have to admit I am lost too...

I will be busy in the next few days so I won't work on MXCuBE 3 again until next week. Marcus will make PRs I guess ;) We are focused on Beamline Setup for now, so don't hesitate to submit other
PRs or fixes in the meantime, it should not conflict with our changes.

Keep on the good work,
cheers,
Matias.

meguiraun:
sorry, it was about the _light_ motor names, since we added the front light we renamed _backlight_ and _frontlight_ to avoid collision and confusion. The client still needs some renaming.

For the camera, I think I am at the same point as you are... When I started working with the server side rendering I needed to save to a file (for later drawing the overlay on it), and it remains there... I will fix so it is no longer needed to save to a file. However, your fix did not work, something is missing.

cheers,

mikel

PR #140, opened 2016-03-14
Update of installation related files (requirements.txt and backend_server.js.example)

marcus-oscarsson:
Hello everybody !

So I'm back to work with MxCuBE, I hope you guys will enjoy it as much as I do :). I just got started and while installing MxCuBE I made some smaller updates to installation related files. I think the diff basically speaks for itself nothing that really needs to be explained

Thanks, and its nice to be back !
Marcus


meguiraun:
Hej Marcus! Good to see that you are onboard!

I am glad to accept your PR but... c&p from [here] (https://flask-socketio.readthedocs.org/en/latest/), section "upgrading to..." 

_If you wish to continue using gevent, then uninstall gevent-socketio from your virtual environment, as this package is not used anymore and may collide with its replacement, python-socketio._

We are already testing it,  and it does not cause any issue, so, could you remove gevent-socketio from the requirements? I should have added that on my previous comment, sorry ...

marcus-oscarsson:
Hi,

Was it you I met in Soleil two years ago ?

Great , so I will create a new PR with python gevent-socketio removed. For next time, its perhaps easier if you accept the PR and then do the necassary fixes after its accepted. I think its less work for everybody, unless you don't have any objections :).

Thanks for taking a look,
Marcus

meguiraun:
I have never been in Soleil... but we will meet soon at ESRF!

No need to close this PR and create a new one, you can make the changes and push to the repo ```marcus-oscarsson:master``` (the origin of this pr) and it will automatically added to the pr

marcus-oscarsson:
Allright, sorry missed that :), I removed gevent-socketio so everything should be fine now.

PR #150, opened 2016-03-17
Stability PR 

bolmsten:
This PR tries to fix a lot of the bugs that was introduced by the two big PR done by me and Matias. I have tried to merge our code and done a bit of cleanup of both code and HTML.

Please take a look and give feedback.


mxcube3/ui/reducers/sampleview.js
```diff
@@ -7,7 +7,7 @@ const initialState = {
   width: 0,
   height: 0,
   lightOn: false,
-  motors: { Light: {}, Phi: {}, PhiY: {}, PhiZ: {}, Sampx: {}, Sampy: {}, Zoom: {} },
+  motors: {},
```
mguijarr: I don't agree with this change - when testing code here it didn't work at all (sample view could not be displayed) because the `motors` object was empty

mxcube3/ui/components/SampleGrid/TaskButtons.jsx
```diff
@@ -4,28 +4,30 @@ import { Button, Glyphicon, ButtonToolbar } from "react-bootstrap"
 export default class SampleTaskButtons extends React.Component {
      constructor(props) {
          super(props);
-
-         this.showCharacterisationForm = this.props.showForm.bind(this, "Characterisation");
-         this.showDataCollectionForm = this.props.showForm.bind(this, "DataCollection");
+         this.showCharac = this.handleSubmit.bind(this, "Characterisation");
+         this.showDataCollec = this.handleSubmit.bind(this, "DataCollection");
```
mguijarr: why using short names ? "Charac" or "Collec" doesn't mean anything, and it's only a few letters less than the normal word.

mxcube3/ui/reducers/sampleview.js
bolmsten: This happens if "http://w-v-kitslab-mxcube-1:8081/mxcube/api/v0.1/sampleview" don't return all the necessary values. 

I think it is dangerous to implement dummy values as this does not expose that the server is not returning all necessary values.

mxcube3/ui/components/Tasks/Characterisation.js
```diff
@@ -1,137 +1,293 @@
+'use strict';
+
 import React from 'react'
 import {reduxForm} from 'redux-form';
-import { Modal } from "react-bootstrap";
+import Modal from 'react-modal';
```
mguijarr: What is "react-modal" ? How does it compare with the normal Modal window that comes with React ?
Normally I use "react-bootstrap" because it is the Bootstrap implementation for React, it is pretty convenient.
If react-modal doesn't bring any added value let's remove this dependency ?

mxcube3/ui/components/Tasks/Characterisation.js
```diff
@@ -1,137 +1,293 @@
...
+            optSAD: fields.optSAD.value,
+            spaceGroup: fields.spaceGroup.value,
+            crystMin: fields.crystMin.value,
+            crystMax: fields.crystMax.value,
+            omegaMin: fields.omegaMin.value,
+            omegaMax: fields.omegaMax.value,
+            stratComp: fields.stratComp.value
+
+            };
+
```
mguijarr: Is there a way to make all this more compact ? it's very repetitive code

mxcube3/ui/components/Tasks/DataCollection.js
```diff
@@ -1,137 +1,290 @@
...
+            left                       : '40px',
+            right                      : '40px',
+            bottom                     : '40px',
+            border                     : 'none',
+            background                 : 'none',
+            overflow                   : 'auto',
+            WebkitOverflowScrolling    : 'touch',
+            borderRadius               : '4px',
+            outline                    : 'none',
+            padding                    : '20px'
```
mguijarr: I don't see the difference of style between the normal Bootstrap Modal and your modified variant ? Please tell me what difference it makes.

package.json
```diff
@@ -28,6 +28,7 @@
     "history": "1.17.x",
     "isomorphic-fetch": "^2.2.0",
     "isotope-layout": "2.2.2",
+    "jquery": "^2.2.1",
```
mguijarr: Nice catch.

mxcube3/ui/reducers/taskForm.js
```diff
@@ -1,15 +1,16 @@
 const initialState = {
-  sample_ids: [],
+  sampleIds: [],
```
mguijarr: So we agree on camel case instead of underscores ?

mguijarr:
Thanks for all this Frederik. I put my comments inline ;) I wait for your answers and/or changes and then I will accept it.

mxcube3/ui/components/Tasks/Characterisation.js
bolmsten: It is a popular library for showing and controlling modals, I try to use it everywhere where modals exist. We could remove it but then I would have to recreate the functionality that it has such as show hide from props. 

We are going to use modals a lot so it is good if we have a standard way of using them

mxcube3/ui/components/SampleGrid/TaskButtons.jsx
bolmsten: I agree

mxcube3/ui/components/Tasks/DataCollection.js
bolmsten: isOpen={this.state.modalIsOpen}
onRequestClose={this.closeModal}
style={customStyles}

It is mostly the open and close I want and I thought it unnecessary to create a own modal component

mxcube3/ui/reducers/taskForm.js
bolmsten: Yes I try to use this everywhere but fail sometimes. Every time I refactor anything I will change it to camelcase, if you agree.

mxcube3/ui/components/Tasks/Characterisation.js
bolmsten: Good question, it might be possible to do this with a for-loop instead.

mxcube3/ui/reducers/sampleview.js
mguijarr: Agreed, but then it would be better to display sample view, then having an error notification than simply don't displaying it at all.

mxcube3/ui/components/SampleGrid/TaskButtons.jsx
mguijarr: Ok so you're going to change it ?

bolmsten: Yes

mxcube3/ui/components/Tasks/Characterisation.js
mguijarr: I agree we are going to use modals a lot, and that we should use a nice library for this. If you're happy with react-modal, ok then - I just want to make sure we don't have unneeded dependencies as we already have modals from bootstrap.

mxcube3/ui/reducers/taskForm.js
mguijarr: Ok. It's just a matter of agreeing. The Python standard is to have underscores, so I tend to write with underscores. But I don't mind having camel case in JS ; the only thing is that we should maybe try to add a rule if it exists in our linting tool, in order to see where we used bad names for variables in order to unify everything.

mxcube3/ui/components/Tasks/DataCollection.js
mguijarr: React Boostrap has exactly the same: look https://react-bootstrap.github.io/components.html#modals

bolmsten: Very interesting! I see your point and I agree, I should refactor all modal windows to use this. The question is if I do that in another PR?

mxcube3/ui/components/Tasks/Characterisation.js
mguijarr: In fact, in my PR I did it like that:

`let parameters = Object.assign({Type: "Characterisation"}, this.props.values);`

`this.props.values` is something that is defined in redux form, it contains field names + values for each field (see documentation)

Did you notice a problem with it ? Maybe it's converting all values to string, but I am not sure.

mxcube3/ui/components/SampleGrid/TaskButtons.jsx
mguijarr: Thanks ! :+1:

mxcube3/ui/components/Tasks/DataCollection.js
mguijarr: No you can continue with this PR I imagine it makes your life easier ? If it is not the case, let's do it in 2 PRs : you decide.

bolmsten:
I have now changed the things you pointed out. I will keep the refactoring to bootstrap modal to the next PR as I am already working on that in another repository. 

The error message handling we can talk about tomorrow as it will be a complex question.

PR #159, opened 2016-04-05
WIP: Beta of beamline setup

marcus-oscarsson:
Early version of the BeamlineSetup component.

The different attributes of the beamline like energy, transmission, resolution and so on are accessed
via an API very similar to the one already existing in the Beamline route. The only difference is that
we now are using set and get since that's slightly more abstracted than i.e move.

The idea is then to add a more abstracted way of getting and setting beamline attributes to the BeamlineSetup HO. This is still to be done and the "mockup" code in the Beamline route will then be removed.

We are also working on a way to display progress of the value change.


meguiraun:
hi, I have a few comments regarding the API, ```Beamline.py```. I know it was there before but it has never been used, so I blame myself in advance for some of these comments...

* Missing status response codes, see routes.{SampleCentring, Queue) but something like this should be enough (until we define custom codes...)
```python
try:
    data = sth done here...
    resp = jsonify(data)
    resp.status_code = 200
    return resp
except Exception:
     return Response(status=409)
```
* No need to add `move` keyword in the url, see [here] (https://github.com/mxcube/mxcube3/blob/master/mxcube3/routes/SampleCentring.py#L328), the `PUT` on a motor with a `new_position` implicitily asssumes a movement
* same for the ```xxx/beamline/<name>/set```, I would remove the `set` and change to `PUT`
* xxx/beamlie/status and xxx/beamline/info looks like the same thing, also I would say that doing a GET on xxx/beamline does the job (no status or info in the url)

bolmsten:
Hello Marcus,

I just have a few comments:

- Everything regarding /actions/beamline_setup.js, /reducers/beamline_setup.js  looks good, Mikel will probably have some comment regarding the API call but that is a minor change.

- Regarding  /components/SampleView/PropertyInput.jsx I think we should avoid using JQuery as much as possible as it is might interfere with React, Is it possible to do this with ordinary HTML/Javascript? 

(React is creating a virtual DOM and is responsible for keeping it up to date, JQuery is changing in the real DOM, this makes it impossible for React to keep track of the DOM)

-I think /containers/BeamlineSetupContainer.jsx should be a ordinary component called by SampleViewContainer, I say this because it is already called and having container calling another container is a bit strange. 

- The CSS located in /containers/beamline_setup_container.css can then be moved down to the SampleView component folder

- Please remove all console.logs and all other stuff giving warnings when doing a PR, In my latest PR I have almost managed to refactor all code giving warnings and I want to keep the repo clean.

Do not misunderstand me, I think you have done a great job! My comments are only small stuff.

marcus-oscarsson:
Hey  @meguiraun,

Don't worry, I started to suspect that Beamline.py was not used because there were some stuff in there which didn't make sense :). In anyways are you planing to use the functions you already have there or can I remove them ?

Sorry, I have to admit that I forgot to take a look at the SampleCentring source that we talked about on in the meeting. The idea you have for the REST API seems much more elegant to me, so I'll do it that way instead.

Thanks for your feedback,

meguiraun:
hej, just remove & rewrite as much as you need, you are the only one working on them right now. Regarding the api structure... if you find anything unclear or nonsense let me know and we discuss it.
cheers

marcus-oscarsson:
Good morning @bolmsten,

I hesitated a lot before I used JQuery because of the reasons that you are mentioning. We are using this X-editable that depends on JQuery. X-editable provides, as you perhaps already know these inline or popup inputs. We thought that would be a good way to present the input to the user and we wanted to try it out. The reason that we think this kind of control is nice is because it makes it possible to display the values in compact way and then provide more information to the user when she wants to change the value.

Maybe you know of a more react compatible way of doing this, I looked a bit at react-xeditable but it still seems to be in early beta. The aim would be to eventually get rid if the JQuery dependency. I think we managed to implement it in such a way that it does not have any strange side effects, so It works until we found a suitable replacement.

The react-redux way is a bit new to me so I might have miss understood something. With that said I think its completely alright to have components inside components. If we consider a container as a reusable entity it makes sense to be able to have components of components, at least as long as the data used by a child component is not used directly by its parents. Otherwise we risk just forwarding data that is not actually used by the parent, which maybe in itself is not really a problem, however it might create unnecessary glue code (especially when reusing a Container). What do you think ? 

However its maybe not the case of the BealineSetup, and I have nothing really against making it a component if it works better with what we have today :).

Thanks for the good feedback,
Cheers, Marcus.

marcus-oscarsson:
@meguiraun, Great thanks :) !

bolmsten:
Okay I understand your decision with JQuery and think we can move forward with that for now, maybe we can discuss and look at alternatives when we visit. 

Regarding Containers inside Containers I see your point, I have however never seen that before but that might just mean I looked at to trivial of example code. If you see the point of having it nested please move forward and we can talk more later on the subject, I am very happy to see new ways of doing things.

marcus-oscarsson:
Hello everybody, 

Started Monday morning with some git exercises to merge the latest changes :). I noticed that I cant
get the SampleViewContainer to render if I don't remove SampleControls and MotorControl. Do you
experience the same problem or did I miss-configure something ?

Just a short summary and motivation of my latest changes :)

Since we have said that we should not change the current hardware object implementation, something is needed to make it possible to add  functionality that is needed by the new web interface. A sub-package called ho_mediators (name most likely to be changed) have been added to deal with this problem rather than adding this functionality directly to the routes package.

The idea would be that we eventually include these changes into the hardware objects once we feel that the interface have been tested with the current hardware object implementation. It will in this way be easier to ensure stable operation of both web and qt interfaces.

Please let me know what you think about the idea, maybe you have already though of something ?

bolmsten:
Good solution with the overlay and popover from bootstrap-react. 

I think we are ready to accept this now, the only thing missing seems to be a gif (busy.gif) is this something that should be added?

bolmsten:
rootClose={true} can be replaced with rootClose only, this should get ride of some warnings as well

meguiraun:
hi, could you add back the following method:
```python
@mxcube.route("/mxcube/api/v0.1/beamline/beamInfo", methods=['GET'])
def getBeamInfo():
    etc etc
```
It was introduced by this commit 6f39d9ea488bd9a10ffe1f78eae484031572f06c. I did not check properly when I told you to `remove & rewrite as much as you need` ...

Other than that, happy to accept.

marcus-oscarsson:
Hey @bolmsten,

Thanks, I was really happy to remove the jQuery dependency. We also needed something that was 
a bit more customizable, especially for displaying progress information and various errors.

marcus-oscarsson:
Hola @meguiraun,

Wooops, sorry about that I just removed everything (since you said that the file was never used :) ).
In anyways, your missing function is back where it belongs.

meguiraun:
hej Marcus, I am still having a couple issues when ```npm start```
````
ERROR in ./mxcube3/ui/components/PopInput/PopInput.jsx
Module not found: Error: Cannot resolve module 'bluebird' in /Users/mikegu/Documents/MaxLab/MxCube/ReferenceCode/marcus-oscarsson/mxcube3/mxcube3/ui/components/PopInput
 @ ./mxcube3/ui/components/PopInput/PopInput.jsx 16:16-35

ERROR in ./mxcube3/ui/components/PopInput/PopInput.jsx
Module not found: Error: Cannot resolve 'file' or 'directory' ./style.css in /Users/mikegu/Documents/MaxLab/MxCube/ReferenceCode/marcus-oscarsson/mxcube3/mxcube3/ui/components/PopInput
 @ ./mxcube3/ui/components/PopInput/PopInput.jsx 22:0-22
```

marcus-oscarsson:
I renamed PropertyInput to PopInput for Popover Input and moved it so that its self contained. The dependency of jQuery was, as @bolmsten noticed, removed in favor of Popover and Overlay. This makes it easier to provide a more customizable component. The idea is that its going to be able to display a custom control/component while for instance loading data.

I consider the preliminary work done so you can merge if you don't have any further comments. There are still some features to be added but the basic functionality is there. 

Ive also made a PR for HardwareObjects with an improved EnergyMockup, that makes the testing slightly more realistic.

Thanks !

marcus-oscarsson:
@meguiraun 

Ive added the bluebird dependency and the missing style.css so I hope it works better now.

meguiraun:
I noticed one weird thing, this PR has 84 commits and changes in 57 files... while you branch has 42 commits and 14 files changed... I am a bit confusing...

marcus-oscarsson:
@meguiraun 

It think its because of the merge and rebase I did on master on Monday.

marcus-oscarsson:
I'll close this PR and create a new one so that it will be easier for you to merge.

PR #163, opened 2016-04-15
Re-indented webpack.config.js

marcus-oscarsson:
Hey !

I know, I know I'm a bit picky but I noticed it a while ago and the file have been on my staging
area asking for being committed some time now :).

Cheers !


marcus-oscarsson:
Good morning !,

I also adapted the eslint.rc to use the airbnb eslint config and added the necessary dependencies in
package.json

Marcus

PR #166, opened 2016-04-19
Socketio signal refactoring

meguiraun:
Some changes to make easier working with socketio signals, basically, now the event name will be: 'motors', 'task', etc... and the hwobj signal will be within the data sent to the client. A more structured way of sending/listening.
It is working as before, if new signals need to be sent, add to ```signals.py```, and link to the appropiate method (or create one)

image


meguiraun:
Some changes are also required in the client, fredrik is on it

mguijarr:
Sounds ok for me.

Should I wait for client changes by Fredrik or can I accept it now ?
The important thing is to keep everything coherent (= in an usable state).

meguiraun:
snif snif... We will need to wait until Fredrik's PR ... ; (
Since we changed the socketio event names both server and client need to be synchronized.... The next time I will push first to Fredrik's repo.

Issue #165, opened 2016-04-18
Javascript and React style guide


marcus-oscarsson:
Hello everybody !

We spoke about Javascript and React guidelines on our last meeting, so I looked around a bit and I found one used by Airbnb. I find it really complete and it goes a little bit further than just being a guideline and also provides some best practices. I purpose that we use this one for our project, we maybe have to make some smaller deviations of our own but I think its basically good as it is. I have made a fork for it on the MxCuBE organization so that it will be easier to discuss it and purpose changes. You can find it here : 

[JS and React style guide](https://github.com/mxcube/javascript)

Please let me know what you think,
Cheers !
Marcus

bolmsten:
Hi,

I think this is great, the more we can set a universal style the better!

Best regards Fredrik

marcus-oscarsson:
@bolmsten,

Yeah I really agree and they have made an excellent work of describing each rule. The only thing we need to do is to add what we consider is missing and maybe change some smaller things.

There are actually only three things that I would maybe like to change,
 
1. I prefer to use spaces after parenthesis in conditionals like if ( expr ) {}, 
    the style guide say no to that i.e use if(expr){}.

2. I prefer to use double quotes " instead of ' for everything since it makes life easier, they use single
    quotes for JS and double for JSX.

3. I prefer as short filenames as possible, with underscores to separate terms. They name the file
    after the class it exports (Like Java).

Point three is more important to me than one and two. However, these points does not really matter to much to me in the end. So its basically up to what you guys think ?

Cheers,
Marcus

bolmsten:
@marcus-oscarsson 

I have some initial feedback on the new linting rules:

1. Do we need to force proptypes? I see that this could be valuable for the smaller components but it adds a lot of overhead in the containers. 

2. Regarding naming, I prefer to use the AirBnB standard of:

"22.2 Use camelCase when naming objects, functions, and instances. eslint: camelcase jscs: requireCamelCaseOrUpperCaseIdentifiers"

"22.3 Use PascalCase only when naming constructors or classes. eslint: new-cap jscs: requireCapitalizedConstructors"

but this can of course be discussed. 


I will get back to you if I have more thoughts.

bolmsten:
We can also use eslint --fix to remove a lot of the errors for free

bolmsten:
I also don't like comma-dangle and want to change this to the opposite

marcus-oscarsson:
[...]

PR #167, opened 2016-04-19
SampleMotors  + Refactored logging to redux/react

bolmsten:
A new GUI for moving samples and I have refactored the logging to a react-redux version. 

With the new linting rules I get a lot of errors that I will try to fix In the coming days for my code.


marcus-oscarsson:
Hi @bolmsten,

Sorry for the linting errors, the new rules are just a proposal so please let us know what you think, if you like some changes or if you think they are to strict ?. We can discuss these things in issue #165.

I'm taking a look at your PR right now and I hope I get the time to merge this morning.

Marcus

marcus-oscarsson:
@blomsten,

Well done, the UI is starting to take shape now, its great to see :).

My idea with the PopInput was that it should be general enough to be used with most beamline "attributes" for example motors or something like energy. The benefit with the PopInput is that it also have the concept of state with corresponding graphical representation (i.e background color). The state (IDLE, MOVING, ON-LIMIT) is something that is very important for motors. Its also important to be able to abort a movement/action.

Did you consider using it or did you find limitations with it that made it impossible to use ?. In anyways I would suggest to generalize it so what we can use it for most of the beamline "attributes". What do you think ?

Something I thought about regarding the logging feature is that each entry is, as far as I understand, now stored in the data model of redux. I suspect that Matias had this in mind when he wrote it and that there might be a reason for that he did not use redux. I imagine that the number of log entries we will have for a session might be quite large,  and so will the size of the redux data model. Do you think that will be an issue, perhaps I'm overthinking it a bit ?

Cheers,
Marcus

bolmsten:
@marcus-oscarsson 

I tried initially to use the Popinput and I like the idea of having a general component for this. 

The problem I faced with your Popinput was that the functionality that you need and the functionality that I need was very different. The fields I have should be input fields with stepping functions onClick for immediate updating and In the future you should be capable of setting the step sizes (hardcoded at the moment). 

I could generalize the Popinput to get it's HTML from props.children but what I noticed then was that I ruined the purpose of the Popinput. I should probably have created a own version of Popinput component for my use case.

Regarding the logging, it will not be more work for the browser to use the Redux Framework compared to ordinary Javascript I think,  so far I have not noticed any problems.

marcus-oscarsson:
@bolmsten,

I see,  My intention is that it should also be possible to specify a custom control for the input, as a child or as a property. Just like its possible to specify a custom loading control. We could even Imagine the possibility to specify one control per state, even if I don't think that's immediately useful.

I'll update the PopInput to make it possible to specify the custom input control and then we can try to integrate that into the sample position control UI. I don't really see anything that directly prevents us from doing that, but I might be mistaken. What do you think ?.

Regarding the logging, it was just a concern for the size of the data model on the client that I was thinking about. Its perhaps as you say not a problem.

bolmsten:
@marcus-oscarsson 

I think it should be possible to do and If you want to try please do. Do you want to accept this PR now and then do a refactoring for it or should I wait for the updated PopInput?

Regarding the logging I think we can try this new version and if we get into trouble I will refactor it. The size of the data model would be the same for the browser I guess, in my case it is saved in the redux state and in the old it is saved on the window object.

marcus-oscarsson:
@bolmsten 

Sorry for the delay, Ive started to do the necessary changes to the PopInput It should be done by tomorrow. I'll leave the decision about the merging to Matias, its all fine by me :). I have notified Matias about the PR and he will a take a look as soon as he can (hes working on the beamlines today).

Good evening,
Marcus

mguijarr:
Hello :)

I propose to wait for the updated PopInput, then if you don't mind you can change the code to use
it, and then I would happily merge the PR.
Then we can make another PR afterwards for adding state information for motors (eg. using
background color with some color code for READY, MOVING, etc.).

Would it be ok like this ?

About the logging: I didn't want to *reduxify* it on purpose, because I didn't want the global redux
state to hold all log records (which can be a lot !), because I thought we may want to serialize
the redux state entirely to be able to restore it when needed (for example, if user closes the
web browsers, restarts it and logs in again).

I have to admit I don't know how redux internals work, if the different reducers states are separated
it's probably ok ? It's just that in doubt I preferred to keep log records away from the redux state.

By the way I like how you simplified the "log navigation", it's better to have only 1 label with page
number instead of many boxes :+1:

marcus-oscarsson:
@bolmsten, @mguijarr 

So Ive created a PR with the new more generic PopInput please see PR #168. I hope you find it useful :)

I agree with Matias the new log navigation is nice !

Cheers !

marcus-oscarsson:
Me and Fredrik did some work together and we merged this PR with  #168 (Attempt to create a more generalized version of the PopInput). So I'm closing this PR.

PR #168, opened 2016-04-21
Attempt to create a more generalized version of the PopInput

marcus-oscarsson:
Hi 

So Ive made an attempt to generalize the PopInput so that you hopefully can use it for the motor controls. I'm open to suggestions and we can change the way it works if there something you don't like or find cumbersome. Don't hesitate to let me know :).

The way it works right now: The PopInput can have two children that are used as input and busy controls. The PopInput itself is responsible for showing and hiding these depending on the state of the action performed. There is an example for the Energy input in BeamlineSetupContainer.jsx

    <PopInput
      ref="energy" name="Energy" pkey="energy" suffix="keV"
      data={this.props.data.energy} onSave={this.onSaveHandler}
      onCancel={this.onCancelHandler}
    >
      <DefaultInput
        key="input" dataType={this.props.dataType} inputSize={this.props.inputSize}
      />
      <DefaultBusy key="busy" />
    </PopInput>

The data object given to the pop input should look like:

    data: { value:'' , state:'' , msg: ''}
    
where state, for the moment, is one of: IDLE, BUSY and ABORTED.

I know that some people advise against passing objects in properties, I did it like that for the moment to avoid creating a lot of properties. The risk is of course that we by mistake  mutate this
object, but I don't believe its a problem.

Please let me know what you think, thanks !

Marcus


bolmsten:
@marcus-oscarsson 

I can't see how this will work with motors, how do I specify the steps and onClick for the input field? 

The big difference with the motors is that we use the steps in the input field and that it should update directly when the user clicks on the stepping button in the input field. 

Maybe I am missing something?

marcus-oscarsson:
@bolmsten 

I see, my example was maybe not very clear the idea for custom controls is that you need to provide your own control for input. so instead of DeafultInput you would maybe use MotorInput. The onClick handler would then call the save method of the PopInput, I think that would do the trick.

In anyways we have been reflecting a little bit about the controls for the sample position. The popover feature is nice for many of the beamline attributes However we believe that the user would
appreciate a more direct access to the sample position controls. So we would suggest to avoid
the popover for the sample control inputs. We could just simply display the control inline instead, this would make it easier for the user to access and change the sample position. 

Our suggestion would be that we add a mode to the PopInput so that the controls can be displayed inline. We would then reasonably need to change the name of PopInput to something more suitable.

Also some thoughts on step size, the default step size and limits should probably be configurable
on the server side. Some controls might give the user the possibility to specify or add additional preferred step sizes to a list.

Another thing to keep in mind is that we perhaps can't simply update the input directly after the user have changed the value, the movement of the motor takes an arbitrary amount of time. We need to show the user that the motor is moving and give them the possibility abort the action. So what I'm saying is that we need to have a control for the "busy" state.

My suggestion is that I implement the inline mode for PopInput we can then rename PopInput to
XInput (suggestions for a better name are welcome). I can also make a better and more clear example for a motor control, because I realize that my previous example was not very clear.

What do you think ?

bolmsten:
@marcus-oscarsson 

Ahh okay now I understand, thank you for the clarification! I think that is a great way of making it more flexible! 

Regarding the design I don't know. We have been using this version here with real hardware for a week and it has worked quite well.
 
I designed this version with Jie and of course we should discuss design but I would suggest to have that discussion face to face next week as it is a lot easier. If you agree we would keep everything as it is and then decide the exact layout next week? 

We are working on a lot of new functionality that we wanted to show you next week and we aren't that focused on design at the moment. 

Cheers, 

Fredrik

marcus-oscarsson:
@bolmsten 

Great :), Maybe your motors are much faster than ours then :). I agree it's a lot nicer to speak about these things when you come. Its going to great to see you guys,  we are looking forward to see the new features you have implemented ! :).

You do as you like, you can maybe keep at as it is until we've done the inline mode or decided otherwise.

Good evening,
Marcus

PR #173, opened 2016-04-29
Session

meguiraun:
first attempt to restore the saved session asking first the user


mguijarr:
We accept it like it is, but queue restore doesn't work (yet). The development will continue in next PRs.

PR #182, opened 2016-05-01
fix when trying to retrieve the saved queue but there is nothing

meguiraun:


mxcube3/routes/Login.py
```diff
@@ -49,7 +49,10 @@ def login():
 #        "local_contact": self.get_session_local_contact(todays_session['session']['sessionId']),
 #        "person": prop['Person'],
 #        "laboratory": prop['Laboratory']}
-        mxcube.queue = jsonpickle.decode(session.get("queueList"))
+	try:
+            mxcube.queue = jsonpickle.decode(session.get("queueList"))
```
mguijarr: Wouldn't it be enough to do like this:

```python
mxcube.queue = jsonpickle.decode(session.get("queueList","{}"))
```
instead of raising the exception ?

I guess the flask session obeys the `dict` semantics

PR #188, opened 2016-05-04
New Motor GUI and moved queue to the sampleview

bolmsten:
This PR moves the queue to the sampleview so you guys can work with the samplegrid. I have also implemented the motors with the design as discussed in the workshop. Please take a look.


marcus-oscarsson:
Hi Fredrik,

Wonderful looks really great :+1: 

Marcus

mguijarr:
Thanks Frederik :+1:

PR #190, opened 2016-05-04
Fixing the build process

bolmsten:
I have now managed to create a good build by running "npm run build".  

I know that I rejected the idea of having the build files in the repository but maybe it is a good idea? If people want to test then all they need to do is get the server up and running.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "scripts": {
     "start": "npm run dev",
     "dev": "./node_modules/.bin/webpack-dev-server --hot --inline --colors --host 0.0.0.0 --port 8090 --display-error-details --content-base mxcube3/ui",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "scripts": {
     "start": "npm run dev",
-    "dev": "./node_modules/.bin/webpack-dev-server --hot --inline --colors --host 0.0.0.0 --port 8090 --display-error-details --content-base mxcube3/ui",
+    "dev": "NODE_ENV=production ./node_modules/.bin/webpack-dev-server --hot --inline --colors --host 0.0.0.0 --port 8090 --display-error-details --content-base mxcube3/ui",
```
mguijarr: I don't think `NODE_ENV` should be set to `production` here... ?

bolmsten:
Good catch @mguijarr ! I struggled a bit to get it working so I tested a few things and this crept in.

PR #191, opened 2016-05-04
Popinput and Beamline setup

marcus-oscarsson:
Hello everybody !

Some smaller updates to the PopInput mostly to make to code a bit more elegant. I've also added the switch input for the "in/out" attributes like beamstop, safety shutter, fast shutter and capillary.

Ive still not used the states from AbstractMotor that we discussed earlier but that will be done as soon as possible. 

Have a nice Weekend,
Cheers,
Marcus


mguijarr:
Oups sorry this PR adds more commits than it should... We will look at the problem and redo it properly on Monday.

marcus-oscarsson:
Thanks Matias !

PR #192, opened 2016-05-09
Popinput and Beamline setup

marcus-oscarsson:
Hello everybody !

Some smaller updates to the PopInput mostly to make to code a bit more elegant. I've also added the switch input for the "in/out" attributes like beamstop, safety shutter, fast shutter and capillary.

I've still not used the states from AbstractMotor that we discussed earlier but that will be done as soon as possible.

Cheers,
Marcus


marcus-oscarsson:
Hola @amilan !

Merge conflict resolved :)

amilan:
Thanks!

Issue #172, opened 2016-04-28
login id for session should rely on proposal returned from ISPyB instead of user login ID


mguijarr:

hnaalas:
Salut Matias

could you precise exactly what you mean to do with this ticket? As you know, SOLEIL is very much complicating everything related to login, ISPyB and related, so any serious modifications to the actual scheme might become a catastrophe for us.

Cheers, leo

> On 28 Apr 2016, at 18:13, Matias Guijarro <notifications@github.com> wrote:
> 
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> 

-
Leonard Chavas
- 
Synchrotron SOLEIL
Proxima-I
L'Orme des Merisiers
Saint-Aubin - BP 48
91192 Gif-sur-Yvette Cedex
France
- 
Phone:  +33 169 359 746
Mobile: +33 644 321 614
E-mail: leonard.chavas@synchrotron-soleil.fr
-

mguijarr:
At the moment in MXCuBE 3, the "login id" (individual username -like at MAX IV- or proposal -like at ESRF-) is used to identify a "session". So for example, if a user is logged in on the MXCuBE 3 web
application, and she closes Firefox, then opens Firefox and do the login again, the state of the
application is restored (contrary to what happens with MXCuBE 2.x, where the interface restarts
from fresh). 

The bug to fix identified by this issue is the following: if 2 users or more are working on the same
proposal, each of them would get a different session with current implementation at MAX IV, because
each user has its own username. At ESRF it's not a problem (yet) because users identify themselves
with a proposal account.

The work to do is quite easy: let's always save a session by proposal.

This affects only MXCuBE 3.

hnaalas:
Thanks Matias for the explanation; things are clear now. SOLEIL is the same as ESRF in this regard (proposal numbers).

Cheers, leo

> On 29 Apr 2016, at 16:30, Matias Guijarro <notifications@github.com> wrote:
> 
> At the moment in MXCuBE 3, the "login id" (individual username -like at MAX IV- or proposal -like at ESRF-) is used to identify a "session". So for example, if a user is logged in on the MXCuBE 3 web
> application, and she closes Firefox, then opens Firefox and do the login again, the state of the
> application is restored (contrary to what happens with MXCuBE 2.x, where the interface restarts
> from fresh).
> 
> The bug to fix identified by this issue is the following: if 2 users or more are working on the same
[...]

PR #196, opened 2016-05-12
Improvements regarding server-side Queue management (fixes #187 and #172)

### LINKED ISSUE #187: mxcube.queue de-serialization when no previous queue has been stored
This is an issue to deal with the problem identified by Peter:

    File "/mnt/sync/pkeller/mxcube3/try3/mxcube3/mxcube3/routes/Login.py", line 52, in login
    # mxcube.queue = jsonpickle.decode(session.get("queueList", "{}"))

Although `jsonpickle.decode(session.get("queueList", "{}"))` gives a proper empty dictionary, it
prevents MXCuBE to run.

I guess this is because an empty Queue object has to be returned, not an empty dictionary...

mguijarr:
See detailed info in commit messages.

One remark : I am not so convinced any more that we should hold the queue state entirely in the
server, I started to experiment with `redux-persist` in order to have persistence on the client side (I created a branch for that, this is a work in progress). While doing this I ended up thinking we should almost only store on the client... So the opposite compared to what we are doing now ! I will come back on this later.

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -6,6 +6,7 @@ from optparse import OptionParser
 import os
 import sys
 import logging
+import jsonpickle
 import gevent
 
 opt_parser = OptionParser()
@@ -73,7 +74,7 @@ if not app.debug or os.environ.get("WERKZEUG_RUN_MAIN") == "true":
         app.diffractometer = app.beamline.getObjectByRole("diffractometer")
         routes.SampleCentring.init_signals()
         app.db_connection = app.beamline.getObjectByRole("lims_client")
-        app.queue = hwr.getHardwareObject(cmdline_options.queue_model)
+        app.empty_queue = jsonpickle.encode(hwr.getHardwareObject(cmdline_options.queue_model))
         routes.Queue.init_signals()
         app.sample_changer = app.beamline.getObjectByRole("sample_changer")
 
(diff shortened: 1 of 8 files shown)


meguiraun:
Thanks for the work, a much clearer code. 
A few comments:
* It is far more easier to understand working with a queue dict rather than with the mxcube.queue (in routes.Queue: when adding samples, executing entry, etc.). How do we know, e.g, if a sample has been already added if we only know the sample location string? In other methods it should be pretty easy to move to working only with the mxcube.queue, for a later PR? 
* mxlogin_required decorator: you are absolutely right, it is there as an old test, is it not in use
* why did you remove the ```dispatch(getInitialStatus)``` in ```actions/login.js```?

I still think that the server is the place for storing the queue... Storing in the client would not mean to have some potential issues when synchronizing? The client does not need to have the whole mxcube.queue but the corresponding dict, but the server needs the whole hwobj which would be managed/saved/whatever by the client. We can discuss it of course!

mguijarr:
Hi Mikel,

`getInitialStatus` is called from the `getLoginInfo` action now. I didn't remove it ?

At the beginning I thought exactly the same as you regarding the queue, I thought the state should
be kept in the server, just as we said during our last workshop.

But I started to have another idea, it's not ready yet so I prefer to talk about it when I would have time
to think about all cases and how it could work, and the more important for us : if it gives better
results in less  time for us than storing the queue on the server. More news later... ;)

I was also thinking that maybe we could generate new Queue objects every time we need it instead of
mutating one Queue from the beginning (in this case, we could parse our own JSON structure from the
client for example and make a new Queue on the fly each time). Well, I am not sure of this neither -
I am looking for ways of simplifying our code...

PR #201, opened 2016-05-24
merging my fork into light branch

meguiraun:
bug fixes and changes needed by the light branch


mguijarr:
Hi :)
Is it only for MXCuBE 3 "light" ?
From the diffs I do not see front end code changes, so it let me thinking it can also be applied
to master ?
What would be the potential problems with that ?

bolmsten:
Hi @mguijarr ,

This true but I will try to make a pull request for the light version today or tomorrow to merge it into master so we only have one branch.

PR #203, opened 2016-05-24
Minor fixes made to improve fault-tolerance

marcus-oscarsson:
Good evening !,

just some minor fixes made to improve fault-tolerance and beamline inter-operability.

Cheers,
Marcus and Matias

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -73,6 +73,10 @@ if not app.debug or os.environ.get("WERKZEUG_RUN_MAIN") == "true":
         app.session = app.beamline.getObjectByRole("session")
         app.collect = app.beamline.getObjectByRole("collect")
         app.diffractometer = app.beamline.getObjectByRole("diffractometer")
+        try:
+            routes.SampleCentring.init_signals()
+        except Exception:
+            sys.excepthook(*sys.exc_info())
         app.db_connection = app.beamline.getObjectByRole("lims_client")
         app.empty_queue = jsonpickle.encode(hwr.getHardwareObject(cmdline_options.queue_model))
         app.sample_changer = app.beamline.getObjectByRole("sample_changer")
--- a/mxcube3/routes/Login.py
+++ b/mxcube3/routes/Login.py
@@ -93,84 +93,84 @@ def get_initial_state():
         :statuscode: 200: no error
         :statuscode: 409: error occurred
     """
-    motors = ['Phi', 'Focus', 'PhiZ', 'PhiY', 'Zoom', 'BackLightSwitch','BackLight','FrontLightSwitch', 'FrontLight','Sampx', 'Sampy'] 
+    motors = ['Phi', 'Focus', 'PhiZ', 'PhiY','Sampx', 'Sampy'] 
     #'Kappa', 'Kappa_phi',
     data = {}
     data['Motors'] = {}
+    for mot in motors:
+       motor_hwobj = mxcube.diffractometer.getObjectByRole(mot.lower())
+       if motor_hwobj is not None:
+           try:
+               pos = motor_hwobj.getPosition()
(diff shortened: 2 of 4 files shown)


mxcube3/__init__.py
```diff
@@ -73,6 +73,10 @@ def complete_initialization(app):
         app.session = app.beamline.getObjectByRole("session")
         app.collect = app.beamline.getObjectByRole("collect")
         app.diffractometer = app.beamline.getObjectByRole("diffractometer")
+        try:
+            routes.SampleCentring.init_signals()
+        except Exception:
+            sys.excepthook(*sys.exc_info())
```
meguiraun: Does the lines 76-79 appear twice?

mxcube3/routes/Login.py
```diff
@@ -93,84 +93,84 @@ def get_initial_state():
...
-            'imageHeight':  mxcube.diffractometer.image_height,
-            }
+        motor_hwobj = mxcube.diffractometer.getObjectByRole('zoom')
+        data['Motors'].update({"Zoom": {"Status":motor_hwobj.getState(), "position": motor_hwobj.predefinedPositions[motor_hwobj.getCurrentPositionName()] }})
+    except Exception:
+        logging.getLogger('HWR').exception('[SAMPLEVIEW] could not get Zoom motor')
+    
+    for light in ('BackLight','FrontLight'):
+        hwobj = mxcube.diffractometer.getObjectByRole(light)
+        switch_state = 1 if hwobj.getActuatorState()=='in' else 0
```
amilan: Hi, 
We are using different hwobj for the actuator and motor, and we have seen that this commit is braking the mockups.
And therefore we should reverse this merge. 
Sorry for the inconvenience

mxcube3/ui/serverIO.js
```diff
@@ -34,17 +32,13 @@ export default class ServerIO {
           break;
       }
     });
-
-    energy.on('value_change', (data) => {
-      this.dispatch(beamlinePropertyValueAction(data));
+    
     });
```
amilan: This line is breaking the client

meguiraun:
this kind of fixes the ```routes.Login.get_initial_status```:
```python
if hasattr(hwobj, "getActuatorState"):
    switch_state = 1 if hwobj.getActuatorState()=='in' else 0
else:
    hwobj_switch = mxcube.diffractometer.getObjectByRole(light+'Switch')
    switch_state = 1 if hwobj_switch.getActuatorState()=='in' else 0
```

mxcube3/__init__.py
mguijarr: Hello,
Indeed it is appearing twice... Sorry we are going to clean this asap

mxcube3/ui/serverIO.js
mguijarr: Sorry for the bad PR... !

mxcube3/routes/Login.py
mguijarr: We should agree on which hwobj to use. Personally I find it good to separate motors
and things like backlight and frontlight. Zoom is also a bit different so that's why we excluded it
from the Motors list.
So, which hwobj are you using for the lights?
The lights are both an actuator (in/out) and a "motor" (with an intensity value).

mguijarr:
Thanks Mikel, indeed for the moment it solves this problem

mxcube3/routes/Login.py
meguiraun: hej,
I think that what you propose is great, easier to follow the code and find any potential issue.
In our case we use “MicrodiffLight" for the light ~motor and “ MicrodiffInOut” for the light ~actuator. We could switch to a single hwobj but I do not know if the whole community can use the same hwobj?

PR #207, opened 2016-05-25
Second try with this PR: Minor fixes made to improve fault-tolerance

mguijarr:
I am sorry of the inconvenience with the nasty typos and bugs introduced with previous PR.
Hopefully this time we have it right...

Please give it a try :)

For `get_initial_state` : we have to agree on which hardware objects to use for front light and back
light, in order to have a common set of referenced hardware objects in the diffractometer object xml. We always had a specific hwobj acting both as an actuator and as a motor AFAIK, I think we should
continue with something like that. I have no time at the moment but I will try to have a look very
soon, I thought `MicrodiffLight` was doing the job. Which hardware objects are you using for the
front/back lights ? At the moment, I use Mikel's suggestion and I guess it is ok, it should work with
our both cases.


amilan:
Thanks Matias!
Sorry for the extra work.

PR #208, opened 2016-05-26
Menu GUI update

bolmsten:
This is PR with the new GUI update for the menu. The design was decided on the November workshop and this is a first version of that.

### DIFF
--- a/mxcube3/ui/actions/sampleview.js
+++ b/mxcube3/ui/actions/sampleview.js
@@ -8,20 +8,11 @@ export function setCurrentPhase(phase) {
   };
 }
 
-export function setImageRatio(ratio) {
+export function setImageRatio(clientWidth) {
   return {
-    type: 'SET_IMAGE_RATIO',
-    ratio: ratio
+    type: 'SET_IMAGE_RATIO', clientWidth
   };
 }
-
-export function setCanvas(canvas) {
-  return {
-    type: 'SET_CANVAS',
-    canvas: canvas
-  };
-}
-
 
 export function setAperture(size) {
   return {
@@ -245,6 +236,7 @@ export function sendZoomPos(level) {
 
 export function sendLightOn(name) {
   return function (dispatch) {
+    dispatch(setLight(name, true));
     fetch('/mxcube/api/v0.1/sampleview/' + name + 'lighton', {
       method: 'PUT',
       credentials: 'include',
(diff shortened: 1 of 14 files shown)


mguijarr:
Hej Frederik,

It would be more clear to split this PR in two at least: one with commits 9101e69 and 90277f7 and
another one with only the new design (b8e5b1c).

It is fairly easy to do:
* create a new branch (`git checkout -b fixes`)
* do `git rebase -i master` (maybe it's more `origin/master` or `upstream/master` it depends on what
your local `master` is following or not)
* remove the commit for the new design (`b8e5b1c`)
* save and finish the rebase
* push this branch, in order to be able to create a PR for it
* go back to `light-menu` branch
* `git rebase -i master` again
* remove the two unwanted commits (`9101e69` and `90277f7`)
* save and finish the rebase
* push with `-f` in order to update the `light-menu` branch on github

In fact it would even be better to split in 3, with one specific PR for changes in `SampleQueueContainer.js` and `SampleGridContainer.jsx` files I think there are not related to the menu update ?
This is more difficult to do, so I understand if you don't have time for it but it would be great if you
could.

Thanks for the hard work... I hope it's going alright at MAX IV - do not hesitate to ask if you need
us to work on something specific to help you making progress faster.

marcus-oscarsson:
Hey Fredrik,

Alot of things are happening on the repository right now, that's great ! :+1: . I've looked at the menu, as Ive said earlier I also prefer the menu to the left. I think having the menu to the left makes better usage of the screen space and also gives a more web app like feel.

I think its important with on mouse over overlays that display the menu option title, since Icons easily can mean different things to different people.  Maybe you already considered that ? I also think that we should remove the top menu and put the user info in the browser title. The logout button could then for instance be placed on the bottom left.

Less important at the moment, but I thought ill mention it while we are on the topic, we should maybe use a white (or light) background for the mxcube logo. Much less important, maybe I have some Volvo design sickness, but I think the icons in the menu bar should be less round. Most of the elements in the UI are more "square-y" and the round icons creates a bit of visual "noise".  Perhaps I'm alone with that opinion, in that case leave it as it is :)

Cheers,
Marcus

bolmsten:
@marcus-oscarsson I agree with you on many points and I consider this only as a first step. This menu is just following the design proposal that was agreed upon from before, I wanted to implement that exact version so that the discussion about enhancements can begin, as it now has!

PR #211, opened 2016-05-26
Need signals for getting points from server

bolmsten:
Do not know why this are removed but they are needed.


mguijarr:
We removed that because we suspected it was causing problems but it's not.

PR #214, opened 2016-05-26
Working with signals

meguiraun:


mxcube3/routes/SampleCentring.py
```diff
@@ -24,12 +24,22 @@ def init_signals():
...
+    except Exception:
+        logging.getLogger('HWR').exception('[SAMPLEVIEW] front light hwobj error')
+
+    try:
+        backlight_hwobj = mxcube.diffractometer.getObjectByRole('backlight')
+        backlight_hwobj.connect(backlight_hwobj, 'positionChanged', signals.motor_event_callback)
+        backlightswitch_hwobj = mxcube.diffractometer.getObjectByRole('backlightswitch')
+        backlightswitch_hwobj.connect(backlightswitch_hwobj, 'actuatorStateChanged', signals.motor_event_callback)
+    except Exception:
+        logging.getLogger('HWR').exception('[SAMPLEVIEW] back light hwobj error')
```
mguijarr: We need to do the same kind of trickery as in `Login.py`, `get_initial_state()` : we (at ESRF) do not have 2 separate objects for backlight and frontlight (I wonder why you separated them, by the way... ?). I can have a look and propose a change, if you want ? It's going to be something like what you proposed yesterday for the front/back light in `Login.py`.

meguiraun:
I agree with you, sorry that I missed that one in this file, let's see if now is ok

PR #220, opened 2016-05-27
Fixes

bolmsten:
The big change in this PR is the parsing of values from task forms, before all values were sent as strings something the server could not handle directly. Now I check all values submitted and if they are not in the stringlist I cast them to number. 

This is a bit of a hassle but as javascript is not strongly typed we get this problem here.

Other additions include CSS and that I now clear the queue on login as we not have session sync implemented .

### DIFF
--- a/mxcube3/ui/actions/login.js
+++ b/mxcube3/ui/actions/login.js
@@ -1,7 +1,7 @@
 import fetch from 'isomorphic-fetch';
 import { getInitialStatus } from './general';
 import { showErrorPanel, setLoading } from './general';
-import { synchState } from './queue';
+import { synchState, sendClearQueue } from './queue';
 
 
 export function doLogin(proposal, password) {
@@ -33,6 +33,8 @@ export function doSignOut() {
 
 export function getLoginInfo() {
   return function (dispatch) {
+    dispatch(getInitialStatus());
+    dispatch(sendClearQueue());
     fetch('mxcube/api/v0.1/login_info', {
       method: 'GET',
       headers: {
@@ -45,8 +47,7 @@ export function getLoginInfo() {
             dispatch(setLoginInfo(loginInfo));
             if (loginInfo.loginRes.Proposal) {
               dispatch(afterLogin(loginInfo.loginRes));
-              dispatch(getInitialStatus());
-              dispatch(synchState(loginInfo.queue));
+              //dispatch(synchState(loginInfo.queue));
             }
           }, () => {
             throw new Error('Server connection problem (getLoginInfo)');
--- a/mxcube3/ui/components/SampleQueue/app.less
+++ b/mxcube3/ui/components/SampleQueue/app.less
@@ -107,6 +107,7 @@
 .queue-root{
   font-size: 1em;
   margin: 10px;
+  line-height: 33px;
 }
 
 ::-webkit-scrollbar {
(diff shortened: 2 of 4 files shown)


mguijarr:
Hello,

What would be really good is a kind of "validation" of the input fields: if a value that is supposed to
be a number is filled with a string, or with an out of range value, the field could be displayed with
red background and the form would not be allowed to be sent to the server...
Validation rules would need to come from the server (like, the server sends a list of fields with
type and limit values that the client can interpret for validation).

I think this is really important - especially if, later, we decide to not send new tasks to the server each
time it's created but only once in one go. Well, even in this case, validation can happen on the
server and errors can be reported to the client in a confirmation dialog box...

Otherwise, if it's just for a convenience to the server side code, I prefer the previous behavior: the
client is only sending strings, the server has to do the type conversions. It puts the logic on the
server only.

amilan:
Hi Matias, 
I think it's ok if we do some validation in the client, but I think it should be validated in the server as well. The server is the one providing the API, therefore the client should follow the API and if not, the server should complain about this.
My point is, I think we must have this validation/conversion in the server. Also it's good to have it in the client, but it can be added in a different feature or Pull Request and this could be accepted.

mguijarr:
(sorry I made this comment in the PR commit log... I copy it again to make the conversation easy
to read)

Ok.
To make it more clear: I think the server has to provide the validation rules to the client, but validation should
be performed on the client first (giving immediate feedback to the user).
I agree on a 100% server-side validation (but then, why the client should make type conversions ?), but I am
afraid it's not good for responsiveness : imagine you need to ask the server each time a value is entered on the form... It's going to be lagging a lot.
That's why I propose a mixed approach, ie. validation rules are given from the server to the client, validation
really happens on the client with immediate visual feedback.

marcus-oscarsson:
Hi both,

I think that all values definitely needs to be validated on the server before the are passed on. One could imagine that the server provides a validation service for the attributes that we can ask for validation. The client could then query this service for validation. The client and server could in this way utilize the same code to validate values. We could add some extra validation in the client to create a more responsive application, but I don't think that would be necessary in most cases.

Speaking about data types and the API provided by the server, the client should send exactly the types that the server expects. We should perhaps avoid tedious type conversions, as Fredrik mentioned it might become even more tedious because of Javascripts type system.  A question that comes to my mind is that: does html input elements return the value in the type of that of the control.
That is does number input return number values it normally does does it not ?.

Ciao,
Marcus

marcus-oscarsson:
Ahh,

I really don't like that GitHub does not have interactive updates ! 

I don't think server side validation necessarily have to be slow, but its true that we have to keep it mind. Just take search keyword complete as an example, it usually works well and updates are fast. However its true that there might be some exceptions.

bolmsten:
@marcus-oscarsson 

Good question Marcus.

First of all I agree with you about the validation but I think it is a future feature and was not meant to be implemented in this PR.

 I agree with you Marcus that the server should get the exact types it expects this is what I tried to fix with this PR. Unfortunately the form always returns strings from all Input fields and therefore I needed to make the conversions "manually"

marcus-oscarsson:
Hey again,

I completely agree with you that validation is a feature to be implemented in the future. Its annoying that the form returns strings :)

Keep up the great work guys :+1: 

Have a nice WE
Marcus

PR #221, opened 2016-05-27
Removing right padding, adjusting label position and height

marcus-oscarsson:
Hello,

We are preparing a small demo version on ID30a-3 for some internal testing and to get some feedback from our scientists (using the real thing :) ). I noticed that the cancel button of the motor input ended up on a new row with the latest changes, so I made a simple fix for it. I also took the opportunity to adjust the position of the label and add some spacing between the inputs.

Maybe you already noticed this and fixed it, in that case you can just discard this PR. Not something very urgent but I thought Ill make a PR since I already made a fix for it :).

Cheers,
Marcus

### DIFF
--- a/mxcube3/ui/components/SampleView/MotorControl.js
+++ b/mxcube3/ui/components/SampleView/MotorControl.js
@@ -15,7 +15,7 @@ export default class MotorControl extends React.Component {
 
             <div className="row">
 
-                <div className="col-sm-12">
+                <div className="col-sm-12 motor-input-container">
                     <p className="motor-name">Omega: </p>
                     <MotorInput
                       save={save}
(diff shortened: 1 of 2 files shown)


bolmsten:
@marcus-oscarsson ,

This is dependent on the screen size I guess? What this PR is doing is removing the padding that bootstrap adds to cols. I think it will look worse on a big screen and better on a smaller one. 

What screeen size are you using and I can see what we can fix

marcus-oscarsson:
Hey,

Its a normal widescreen monitor, I guess that its our video stream that somehow pushes the content a bit (but I'm not sure yet, I have no possibility to test because I'm not at the beamline). It looked good yesterday on the same screen so I assumed that it was some other element that had grown a bit. In anyways I'm happy for any fix. 

Cheers,
Marcus

marcus-oscarsson:
Hi again,

I removed the padding css rule that I added to prevent the "cancel action" button from sliding down. Remaining is the rules that align the title to the actual control and adds some spacing between each control.

Cheers,
Marcus

PR #224, opened 2016-05-30
PR Linting

bolmsten:
I am trying to remove as many linting errors as possible as it would be good to have a "clean" state for development. 

There is still a lot to do but this brings it down a bit.


mguijarr:
Thanks for this.

PR #225, opened 2016-05-30
Collection signal fix

meguiraun:
a fix related to the signals sent during a data collection, there was an issue with some signals being connected to internal methods in ```queue_entry``` and colliding with the ones in ```routes.signals```

### DIFF
--- a/mxcube3/routes/Queue.py
+++ b/mxcube3/routes/Queue.py
@@ -6,7 +6,7 @@ import time
 import logging
 import collections
 import gevent.event
-import os
+import os, sys
 import json
 import queue_model_objects_v1 as qmo
 import QueueManager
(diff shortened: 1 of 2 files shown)


mxcube3/routes/Queue.py
```diff
@@ -309,6 +311,7 @@ def executeEntryWithId(nodeId):
                         mxcube.queue.queue_hwobj.execute_entry(childEntry)
                         time.sleep(1) # too fast to synch properly signals, it should not be a problem with real stuff
                     except Exception:
+                        print sys.exc_info()
```
mguijarr: what about removing this `print` and replacing `logging.error` below with `logging.exception` ?

mguijarr:
I think it would be a good idea to do like Frederik did last time with `git merge —squash` (or similar)
in order to have a much cleaner git history for this PR

PR #226, opened 2016-05-30
Bug fixing with the light actuators

meguiraun:
- in the light/off fix the string type
- Note that this might only work properly with the hwobj we have here, one for the level, another one for the actuator. Can you test on your side for your combined hwobj? You will probaly need to add sth...


mguijarr:
I think it would be a good idea to do like Frederik did last time with `git merge —squash` (or similar)
in order to have a much cleaner git history for this PR

amilan:
Hi Matias, 
I agree with you, history should be much cleaner, so tomorrow we will fix it.
Thanks for the feedback.

PR #230, opened 2016-05-31
Refactoring

mguijarr:
Some refactoring to not repeat code between `get_initial_status` and `get_status_of_id` functions,
see commits to have more information.

NOTE: `backlightswitch`,`frontlightswitch`,`frontlight` and `backlight` roles in XML file *have to* be in lower case

PS: this PR also removes a print statement and an unused function


meguiraun:
hej, great refactoring! Could you do the same as in ```routes.Login``` [here] (https://github.com/mxcube/mxcube3/blob/refactoring/mxcube3/routes/SampleCentring.py#L451) please?

mguijarr:
Ok I will take care of this

mguijarr:
Done !

Finally I made a bigger refactoring, see commit message for more info.

mxcube3/routes/Beamline.py
```diff
@@ -84,25 +84,93 @@ def beamline_get_attribute(name):
...
+    except Exception:
+        logging.getLogger("HWR").exception("Failed to get beam info")
+        return Response(status=409)
+    else:
+        resp = jsonify(ret)
         resp.status_code = 200
         return resp
+
+
+@mxcube.route("/mxcube/api/v0.1/diffractometer/movables/state", methods=['GET'])
```
meguiraun: * why not in the ```routes.Diffractometer``` file?
* I would put just ```/mxcube/api/v0.1/diffractometer/movables``` (removing 'state'), the call assumes that a GET methods to the movables will give their position/status and relevant info (I know, it's a bit picky...)

mxcube3/routes/Beamline.py
```diff
@@ -84,25 +84,93 @@ def beamline_get_attribute(name):
...
+        ret.update(Utils.get_movable_state_and_position(movable))
+
+    ret.update(Utils.get_light_state_and_intensity())
+
+    resp = jsonify(ret)
+    resp.status_code = 200
+    return resp
+
+
+@mxcube.route("/mxcube/api/v0.1/diffractometer/info", methods=['GET'])
```
meguiraun: * same as before, remove 'info'?

mxcube3/routes/Beamline.py
```diff
@@ -84,25 +84,93 @@ def beamline_get_attribute(name):
...
+        ret['current_phase'] = mxcube.diffractometer.current_phase
+    except AttributeError:
+        ret['current_phase'] = 'None' # in case the diff does not have this implemented
+
+    resp = jsonify(ret)
+    resp.status_code = 200
+    return resp
+
+
+@mxcube.route("/mxcube/api/v0.1/sampleview/camera/info", methods=['GET'])
```
meguiraun: * same as before, remove 'info'?

meguiraun:
Apart for the minor comments in the code, should we add a general debug call that returns a ~big dict with all the relevant info? (perhaps out of the scope of this PR) Something like the ```get_initial_status``` but with more info. We usually use that call not only for the initial launch but in the middle of the operation for the debugging purposes

mxcube3/ui/actions/general.js
```diff
@@ -18,22 +18,48 @@ export function showErrorPanel(show, message = '') {
...
+        'Accept': 'application/json',
+        'Content-type': 'application/json'
       }
-      return response.json();
-    })
-    .then(function (json) {
-      dispatch(setInitialStatus(json));
     });
 
+    motors.then(response => { state.Motors = response.json() });
```
bolmsten: This will set state.motors equal to a promise as response.json() returns that

mxcube3/ui/actions/general.js
```diff
@@ -18,22 +18,48 @@ export function showErrorPanel(show, message = '') {
...
+    beamInfo.then(response => { state.beamInfo = response.json() });
+    sampleVideoInfo.then(response => { state.Camera = response.json() });
+    diffractometerInfo.then(response => { Object.assign(state, response.json()) });
+    
+    Promise.all([motors,beamInfo,sampleVideoInfo,diffractometerInfo]).then(() => {
```
bolmsten: I see what you are trying to do here but it seems that Promise.all only cares about the promises on the first level. 

I played around a bit and I think this might be what you want:


export function getInitialStatus() {
  return function (dispatch) {

    let motors = fetch('mxcube/api/v0.1/diffractometer/movables/state', {
      method: 'GET',
      credentials: 'include',
      headers: {
        'Accept': 'application/json',
      }
    });
    let beamInfo = fetch('mxcube/api/v0.1/beam/info', {
      method: 'GET',
      credentials: 'include',
      headers: {
        'Accept': 'application/json',
      }
    });
    let sampleVideoInfo = fetch('mxcube/api/v0.1/sampleview/camera/info', {
      method: 'GET',
      credentials: 'include',
      headers: {
        'Accept': 'application/json',
      }
    });
    let diffractometerInfo = fetch('mxcube/api/v0.1/diffractometer/info', {
      method: 'GET',
      credentials: 'include',
      headers: {
        'Accept': 'application/json',
      }
    });
    
    Promise.all([motors, beamInfo, sampleVideoInfo, diffractometerInfo]).then((responses) => {
      Promise.all(responses.map((response) => response.json())).then((values) => {
        dispatch(setInitialStatus(Object.assign({}, ...values)));
      });
    });
  };
}

This will first wait for all http promises then it wait for the json promises. There is one modification that needs to be done on the server for this to work and that is to wrap all motors in a Motors object first and the same for the rest.

mxcube3/ui/actions/general.js
mguijarr: Thanks for this - I didn't know. One more thing I find weird in this fetch api !
I am going to fix it asap.

mxcube3/routes/Beamline.py
mguijarr: We talked here and agreed to also remove `movables` ; ok for moving the code to `Diffractometer.py`

mguijarr: Ok to move the code to `routes.Diffractometer`

For the moment we prefer to keep routes as it is - I suggest to talk about all routes again soon, we can make another PR for route name changes.

mxcube3/routes/Beamline.py
mguijarr: I like the `info`... See my previous comment

mguijarr:
I updated my PR. Thanks for the help with the promises Frederik ; I didn't do exactly as you suggested but something in the same spirit.

bolmsten:
Hi @mguijarr ,

We have tested this now and it looks fine!! The only thing I would want you to do is to clean up actions/general.js linting errors if it is possible?

mguijarr:
I am sorry Frederik I have no time for this, but I can assign myself an issue for later (soon) to fix the linting errors

bolmsten:
Okay @mguijarr that is fine by me, just resolve the conflicts and we will merge

mguijarr:
Done

PR #235, opened 2016-06-03
WIP: Tasks

bolmsten:
Linting of Tasks and added the rootPath where the file is stored

### DIFF
--- a/mxcube3/routes/Login.py
+++ b/mxcube3/routes/Login.py
@@ -4,7 +4,7 @@ from mxcube3.routes import Queue, Utils
 import logging
 import os
 import types
-
+import time
 
 def convert_to_dict(ispyb_object):
     d = {}
--- a/mxcube3/routes/Login.py
+++ b/mxcube3/routes/Login.py
@@ -172,6 +172,12 @@ def get_initial_state():
     except AttributeError:
         data['current_phase'] =  'None' # in case the diff does not have this implemented
 
+    try:
+        today = time.strftime("%Y%m%d")
+        data['rootPath'] = '/ramdisk' + today + '/RAW_DATA/'
+    except Exception, e:
+        logging.getLogger('HWR').error("Error retrieving data path")
+
     resp = jsonify(data)
     resp.status_code = 200
     return resp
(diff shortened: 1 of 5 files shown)


mxcube3/routes/Login.py
```diff
@@ -172,6 +172,12 @@ def get_initial_state():
     except AttributeError:
         data['current_phase'] =  'None' # in case the diff does not have this implemented
 
+    try:
```
marcus-oscarsson: This block of code should be in a beamline/site specific file, I think its in the beamline setup hardware 
object today. Perhaps you already thought about it and its just a temporary fix ?

mxcube3/ui/components/Tasks/Characterisation.js
```diff
@@ -1,9 +1,7 @@
-'use strict';
-
 import React from 'react';
 import { reduxForm } from 'redux-form';
 import { Modal } from 'react-bootstrap';
-
+/* eslint camelcase: 0 */
```
marcus-oscarsson: I understand that this is because of the back end that exports variables with _. Perhaps its also a temporary fix ?. Id rather see that we do some kind of conversion/mapping than disabling the style guide.

Another thing I noted is that you guys seem to mix upper and lower camel case, like point and Type in the code below. It does not matter to me which one we use as long as we are consistent. We should however use lower camel case if we want to respect the style guide.

marcus-oscarsson:
Hey Fredrik, Mikel, Jie and Antonio, ( @bolmsten @amilan @meguiraun @JieNanMAXIV )

I hope everything is going alright with the implementation of the inauguration version !. We understand that you guys are under some stress right now, we know a bit how it can be :).
We are also really curious to see the results of all the great work you guys are doing !.

Thanks for the PR,  I made some comments on the code above. Nothing critical for me, I don't mind accepting it as it is. I understand that the variable case can be difficult to change right now, we can wait with that until a later moment. We should perhaps decide on which camel case to use to minimize work in the future.

The path thing is maybe more important and it's really a site/beamline specific implementation. I understand that you guys are in hurry to finish something so again we can maybe let that pass for now. I'll let Matias have the final word on that.

Good luck and let us know if there is something we can do to help you out !.

Cheers,
Marcus

mguijarr:
I mark this one as a WIP since it has the hard-coded `rootPath` problem. I don't think we should merge it with hard-coded paths.

meguiraun:
the hardcoded path will be removed by https://github.com/mxcube/mxcube3/pull/239 (if everything is ok), and then we adapt this one.

PR #239, opened 2016-06-07
retrieve data path

meguiraun:
in https://github.com/mxcube/mxcube3/pull/235 there was a ugly hardcoded path, the current pr provides a call for retrieving the data path.


marcus-oscarsson:
Hey Mikel !

That's excellent !

Marcus

PR #245, opened 2016-06-08
Motor states and position updates while moving

mguijarr:
This PR fixes the display of states (associating proper "palette" with hardware state)

It also updates the Hardware Objects (2.2 branch) to the latest version, introducing a new `MotorMockup` class with simulated move, which allows to test state and position changes.

### DIFF
--- a/mxcube3/ui/components/SampleView/MotorInput.js
+++ b/mxcube3/ui/components/SampleView/MotorInput.js
@@ -2,6 +2,7 @@ import React from 'react';
 import cx from 'classnames';
 import { Button } from 'react-bootstrap';
 import PopInput from '../PopInput/PopInput';
+import './motor.css';
 
 export default class MotorInput extends React.Component {
 
@@ -46,10 +47,10 @@ export default class MotorInput extends React.Component {
     const { value, motorName, step, suffix, decimalPoints } = this.props;
     const valueCropped = value.toFixed(decimalPoints);
     let inputCSS = cx('form-control input-sm', {
-      active: this.props.state === 4,
-      success: this.props.state === 2,
-      error: this.props.state === 3,
-      warning: this.props.state === 1
+      'motor-moving': this.props.state === 4 || this.props.state === 3,
+      'motor-ready': this.props.state === 2,
+      'motor-fault': this.props.state <= 1,
+      'motor-onlimit': this.props.state === 5
     });
 
     let data = { state: 'IMMEDIATE', value: step };
--- a/mxcube3/ui/components/SampleView/MotorInput.js
+++ b/mxcube3/ui/components/SampleView/MotorInput.js
@@ -66,6 +66,7 @@ export default class MotorInput extends React.Component {
                 step={step}
                 defaultValue={valueCropped}
                 name={motorName}
+                disabled={this.props.state === 4 || this.props.state === 3 || this.props.state <= 1}
               />
               {this.props.saveStep ?
                 <span>
--- a/mxcube3/ui/components/SampleView/SampleView.css
+++ b/mxcube3/ui/components/SampleView/SampleView.css
@@ -34,26 +34,6 @@
     margin-top: 10px;
 }
 
-.motor-input-container{
-    margin-bottom: 1em;
-}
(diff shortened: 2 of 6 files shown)


mxcube3/ui/components/SampleView/SampleView.css
```diff
@@ -82,25 +62,6 @@
...
-.motor-button{
-    margin: 15px !important;
-}
-
-#zoom-control{
-    display: inline;
-    width: auto;
-    position: relative;
-    top: 3px;
-}
```
bolmsten: This is needed

mxcube3/ui/components/SampleView/MotorInput.js
```diff
@@ -46,10 +47,10 @@ export default class MotorInput extends React.Component {
     const { value, motorName, step, suffix, decimalPoints } = this.props;
     const valueCropped = value.toFixed(decimalPoints);
     let inputCSS = cx('form-control input-sm', {
-      active: this.props.state === 4,
-      success: this.props.state === 2,
-      error: this.props.state === 3,
-      warning: this.props.state === 1
+      'motor-moving': this.props.state === 4 || this.props.state === 3,
+      'motor-ready': this.props.state === 2,
+      'motor-fault': this.props.state <= 1,
```
bolmsten: Better to show red then yellow on motor fault?

mxcube3/ui/components/SampleView/MotorInput.js
```diff
@@ -65,6 +66,7 @@ export default class MotorInput extends React.Component {
                 step={step}
                 defaultValue={valueCropped}
                 name={motorName}
+                disabled={this.props.state === 4 || this.props.state === 3 || this.props.state <= 1}
```
bolmsten: I did this originally but it does not work, the reason is that if you disable OnMove the cursor moves away from the input box and the user needs to reselect it for additional movement

mxcube3/ui/components/SampleView/SampleView.css
mguijarr: Oups sorry I am going to add it back

mxcube3/ui/components/SampleView/MotorInput.js
mguijarr: I don't understand could you elaborate on this ?

In any case we need to find a way to prevent user from typing in this input while the motor is moving, or if it is
in fault state.

mxcube3/ui/components/SampleView/MotorInput.js
mguijarr: Yes ! I forgot to commit this change (I already did it here in my version)

mguijarr:
I think it's ready for merge now. Thanks for the review.

mxcube3/ui/components/SampleView/MotorInput.js
bolmsten: This is actually done already in the onClick and onKey functions

mguijarr: Er... I tried this morning, first with this code (with the `disabled`) and I could not find any problem with it and then I removed the `disabled` attribute, I could type in the input field for the motors in fault state (Kappa ones) and I could type while motor was moving.
I am sure I am missing something ?

bolmsten: The problem I have found is that when the field is disabled it moves the cursor away from it and you will have to reselect it to move it again. 

Yes you can type in new data in the field while it is moving but it will be ignored and overwritten by the new value of the motorposition.

A possible solution would be to reselect the field for the user after the disable is removed..... This is probably better then the solution I implemented before.

mguijarr: I didn't experience the same problem as you describe, maybe it has something to do with the browser: I am using Firefox here I guess you're running Chrome ?

It's not very nice that it is possible to type on a "logically" read-only field.

In any case I don't see the fact that focus moves away to be very problematic in this particular situation. I advocate for accepting this PR and solving this problem in another one.

Issue #242, opened 2016-06-08
Sample Grid Prototype


marcus-oscarsson:
Hi all,

We thought it would be good share the latest development of the Sample Grid, so that you guys can follow whats happening and even try it out. There is still plenty of work to be done overall, from design to functionality. So be prepared that its in a early stage :). The branch is called samplegrid-prototype.

Its currently possible to move sample items around in the grid by selecting it and pressing the little cross in the upper left corner. (Its currently not possible to move several items at once). It's further possible to multiple select and pick items for collection.

The design have changed slightly from what we purposed when you were here last time. The Pipeline mode and collect button have been moved above the sample grid itself. All the controls above the sample grid is also contained in a "sticky" container so that they scroll with the screen when there are alot of samples.

All of these things are of course under development and evaluation so they might change during the process. We are really grateful  for any input on anything from design to code details, although keep in mind that its in a quite early stage :). We also understand it might not be the best time for you guys to check this thing out right now.

I'm going to spend the next couple of days to do some refactoring and pay pack some of the "messy code" debt that we have :). We would like to continue with the work on the queue after that, hopefully at the end of this week or beginning of the next. It would be great to know were you are on the topics we discussed when you were here last time, since this impacts the queue behavior ?

Enjoy,
Marcus

amilan:
Hi Marcus, 
Thanks for sharing this development. We understand that is not the final version and it's still under heavy development, so no problem with that, is always good to have it available in order to provide you better feedback.
We have scheduled a review here at Max IV for tomorrow (Friday) morning, so we will inform you soon.

Feedback is coming ... 

Thanks,
Antonio.

marcus-oscarsson:
Hey Antonio,

That's great to hear, happy to know that its useful for you. We are in particular interested in your thoughts about the queue and the concept in general, at least this stage at least :).

Ciao !
Marcus

mguijarr:
In order to make it easier to understand this proposal, I can clarify some terms a bit:

* Selecting a sample *= clicking on the sample card, or dragging the mouse over multiple cards*
  * to **add a task** on the sample
    * e.g characterisation (**pipeline**)
  * to **move** the sample
    * using arrow keys or the buttons that appear on the sample card
  * to **pick the sample** for data collection
    
When a sample is selected, a thick dark border is drawn around the sample card.

* Picking a sample *= clicking on the checkbox of sample card, or using the __Pick__ buttons*
  * to **add to the queue**
  
[...]

PR #246, opened 2016-06-08
Samplegrid

marcus-oscarsson:
Hi,

NOTE: Please make sure to merge #244 first.

This is only the sample grid not to be confused with the samplegrid-prototype branch I published earlier today (issue #242). The sample grid it self is fairly self contained so it should not be a problem merge it with master. It would be great if we could get this into master as soon as possible since I would like to continue with some refactoring that might effect some other parts than the sample grid. 

Its also important for keeping the PRs small, and making it easier apply future fixes.

Cheers, 
Marcus

### DIFF
--- a/mxcube3/ui/actions/SamplesGrid.js
+++ b/mxcube3/ui/actions/SamplesGrid.js
@@ -3,14 +3,16 @@ import { sendClearQueue, sendRunSample, sendAddSample, sendMountSample } from '.
 import { showTaskForm } from './taskForm';
 import { setLoading, showErrorPanel } from './general';
 
+
 export function doUpdateSamples(samples_list) {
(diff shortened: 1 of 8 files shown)


mxcube3/ui/actions/SamplesGrid.js
```diff
@@ -223,3 +243,50 @@ export function sendDeleteSampleTask(parent_id, queueID, sampleID) {
     });
   };
 }
+
+
+export function doReorderSample(sampleOrder, key, targetPos){
+  console.log(`REORDER ${key}->${targetPos}`);
```
mguijarr: To be removed

marcus-oscarsson: Wopsie dopsie, fixed now

PR #252, opened 2016-06-14
Samplegrid

marcus-oscarsson:
Hello all,

I wanted to make a PR out of this before it gets to big. What I did was simply to clean up the "interface"
exposed by the action/SampleGrid.js. I also took a look at the action names and tried to give them a more well defined meaning, as far as that is possible, so it will be easier to debug.

I would like to take a deeper look at the logic in the reducer and maybe clarify/simplify the logic a bit. Ive noticed that the logic in the reducers quickly becomes quite messy and large (maybe its just me ?), I would like to do something about this at least for the SampleGrid reducer, what I have in mind is simply to extract certain things to functions and then adding some documentation. Perhaps someone
have a better idea ?

I think it also would be really valuable to document the data model, store, so that its clear what it should
contain and what each property does. Maybe this is good in general for reducers ?

Marcus

### DIFF
--- a/mxcube3/ui/actions/SamplesGrid.js
+++ b/mxcube3/ui/actions/SamplesGrid.js
@@ -4,19 +4,19 @@ import { showTaskForm } from './taskForm';
 import { setLoading, showErrorPanel } from './general';
 
 
-export function doUpdateSamples(samples_list) {
-  return { type: 'UPDATE_SAMPLES', samples_list };
+export function updateSampleListAction(sampleList) {
+  return { type: 'UPDATE_SAMPLE_LIST', sampleList };
 }
 
 
-export function doGetSamplesList() {
+export function sendGetSampleListRequest() {
   return function (dispatch) {
     dispatch(setLoading(true));
     fetch('mxcube/api/v0.1/sample_changer/samples_list', { credentials: 'include' })
             .then(response => response.json())
             .then(json => {
               dispatch(setLoading(false));
-              dispatch(doUpdateSamples(json));
+              dispatch(updateSampleListAction(json));
             }, () => {
               dispatch(setLoading(false));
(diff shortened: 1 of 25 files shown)


mguijarr:
I propose you to squash all commits into one before it gets merged

marcus-oscarsson:
Yes, your right its probably nicer. Finally there are two commits since i needed to fix some conflicts.

bolmsten:
It seems that you have changed the "ADD_METHOD" to "ADD_TASK" , this is fine and good but you have forgotten to change what the queue reducer and task reducer listen to.

marcus-oscarsson:
Hi Fredrik !,

Yes I changed from METHOD to TASK in several places because it seems that both TASK and METHODS were used. I thought it would be better to be consistent so I changed to TASK everywhere. I simply chose task since its more consistent with the terminology we use when we speak about the queue. For instance we say that we are adding tasks to the queue, those might in turn be data collection methods. 

I tried to make changes that were directly linked to the samplegrid not to disturb you guys to much. It seems like I missed to change in queue and taskForm. 

I agree with the that SET_ALL_ATTRIBUTES are a bit to vague Ill change that when I take a another look at the beamlinesteup. But its not a part of this PR, since I'm focusing on the samplegrid right now.

marcus-oscarsson:
Hey, Ive done the changes in queue.js and taskForm.js

bolmsten:
@marcus-oscarsson 

I have tested it again and now I see that you changed parent_id to parentID. I totally agree with this change but it is not changed down in the actions of queue so it crashes. 

You can se this by trying to add a task to a sample

marcus-oscarsson:
Sorry yes, I noticed that too. Its fixed now :)

PR #253, opened 2016-06-15
Fixes after experiment

meguiraun:
These are just some minor fixes, they do not stop the experiment but make life a bit easier.

*** Just a note: the queue does not set the entry as 'executed' (I do not know why...), so I manually do it after the execution, not sure if it is the best solution.


mxcube3/routes/Queue.py
```diff
@@ -678,10 +672,12 @@ def addDataCollection(id):
         task1Entry.set_data_model(taskNode1)
 
         colNode.acquisitions[0].acquisition_parameters.set_from_dict(params)
-        colNode.acquisitions[0].path_template.directory = params['path']
+        colNode.acquisitions[0].path_template.directory = mxcube.session.get_base_image_directory() + '/' + params['path']
```
mguijarr: should use `os.path.join`

PR #256, opened 2016-06-17
detector dist

meguiraun:
Added:
* Try/except because we do not have all those hwobj in place yet (and perhaps the fast shutter should go out of here?)
* added the detector distance motor

### DIFF
--- a/mxcube3/ho_mediators/beamline_setup.py
+++ b/mxcube3/ho_mediators/beamline_setup.py
@@ -31,30 +31,30 @@ class _BeamlineSetupMediator(object):
 
 
     def getObjectByRole(self, name):
-        ho = self._bl.getObjectByRole(name.lower())
         try:
-            if name == "energy":
-                return self._ho_dict.setdefault(name, EnergyHOMediator(ho, "energy"))
-            elif name == "resolution":
-                return self._ho_dict.setdefault(name, ResolutionHOMediator(ho, "resolution"))
-            elif name == "transmission":
-                return self._ho_dict.setdefault(name, TransmissionHOMediator(ho, "transmission"))
-            elif name == "fast_shutter":
-                return self._ho_dict.setdefault(name, InOutHOMediator(ho, "fast_shutter"))
-            elif name == "safety_shutter":
-                return self._ho_dict.setdefault(name, TangoShutterHOMediator(ho, "safety_shutter"))
-            elif name == "beamstop":
-                return self._ho_dict.setdefault(name, BeamstopHOMediator(ho, "beamstop"))
-            elif name == "capillary":
-                return self._ho_dict.setdefault(name, InOutHOMediator(ho, "capillary"))
-            elif name == "dtox":
-                return self._ho_dict.setdefault(name, DetectorDistanceHOMediator(ho, "dtox"))
-            else:
-                return ho
+            ho = self._bl.getObjectByRole(name.lower())
         except Exception:
             logging.getLogger("HWR").exception("Failed to get get object with role: %s" %name)
 
-
+        if name == "energy":
+            return self._ho_dict.setdefault(name, EnergyHOMediator(ho, "energy"))
+        elif name == "resolution":
+            return self._ho_dict.setdefault(name, ResolutionHOMediator(ho, "resolution"))
+        elif name == "transmission":
+            return self._ho_dict.setdefault(name, TransmissionHOMediator(ho, "transmission"))
+        elif name == "fast_shutter":
+            return self._ho_dict.setdefault(name, InOutHOMediator(ho, "fast_shutter"))
+        elif name == "safety_shutter":
+            return self._ho_dict.setdefault(name, TangoShutterHOMediator(ho, "safety_shutter"))
+        elif name == "beamstop":
+            return self._ho_dict.setdefault(name, BeamstopHOMediator(ho, "beamstop"))
+        elif name == "capillary":
+            return self._ho_dict.setdefault(name, InOutHOMediator(ho, "capillary"))
+        elif name == "dtox":
+            return self._ho_dict.setdefault(name, DetectorDistanceHOMediator(ho, "dtox"))
+        else:
+            return ho
+       
 
     def dict_repr(self):
         """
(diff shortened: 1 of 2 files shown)


mguijarr:
I wonder if it is really needed to add detector distance as an element of the `BeamlineSetup` hardware object, since we already have the `Resolution` hardware object in it. The resolution is calculated from the energy and the detector distance.

Why can't you use the resolution hardware object to get detector distance ?

Why do you suggest the fast shutter should go out of here ?

mxcube3/ho_mediators/beamline_setup.py
```diff
@@ -32,44 +32,80 @@ def __init__(self, beamline_setup):
 
     def getObjectByRole(self, name):
         ho = self._bl.getObjectByRole(name.lower())
```
mguijarr: Don't you think the try...except should just be around this line instead of the whole block ? I prefer small try...except blocks, in general.

meguiraun:
I said about the fast shutter because the MD manages itself the fast shutter.
Regarding the detector distance, we would like to have it, it is useful during commissioning. We started looking at the motor position rather than at the resolution.

mguijarr:
It's fine to have detector distance (it is present in current MXCuBE 2.x), but there is no need to add
an extra hardware object in the beamline setup for that since the motor can be accessed via resolution hardware object.

For the fast shutter: it's important to keep it because people often need to see the beam, and operate the shutter themselves.

meguiraun:
you mean something like this?
Fine with the fast shutter.

mguijarr:
Yes, it's better because it doesn't change the xml file. It's ok for now.

hnaalas:
Dear all

Leo (SOLEIL) speaking. Personally, I would love to get the fast-shutter status visible, but for no reason I want to have the users being able to open the fast shutter by themselves. It is too dangerous and could clearly damage the detector if opened in wrong conditions. Would there be any possibilities to get the optional 'check beam position' button (or whatever you call it) link to this fast shutter, rather than giving direct access to the fast shutter opening?

Cheers, leo

-
Leonard Chavas
- 
Synchrotron SOLEIL
Proxima-I
L'Orme des Merisiers
Saint-Aubin - BP 48
91192 Gif-sur-Yvette Cedex
France
- 
Phone:  +33 169 359 746
Mobile: +33 644 321 614
E-mail: leonard.chavas@synchrotron-soleil.fr
-

mguijarr:
Hi @hnaalas :)
We talked about your remark this morning during our skype meeting, in fact fast shutter control has always been present in MXCuBE and both Daniele and Marjolein don't think it should be removed. However, you're right about potential risks for the detector - at ESRF we have a cover that goes down in front of the detector to protect it. There is also the beamstop.
It is fairly easy to put the safety guards in hardware objects, if you want - like preventing fast shutter
from opening if something is not safe, hence doing the same thing as a "Check beam position"
button at a lower level.
The best would be to have some kind of hardware interlock.

IvarsKarpics:
Hi @hnaalas , @mguijarr  and other,

Not sure how it is implemented in other sites, but in our case fast shutter is available just in beam location phase. Beam location phase for MD2/MD3 is used to visualize beam on scintillator or read flux from photodiode. Otherwise the control is disabled. The phase check is in hardware object level.

beteva:
Dear @hnaalas,
adding to Matias´s comment it might be a better idea to implement  any check you want in the corresponding Hardware object, which makes it more matching your needs. Adding buttons might overload the already crowded view.

hnaalas:
Dear all

I guess we all have our views on why buttons should stay or not, and I fully encourage this indeed. All in all, although we perform the same kind of work, we all have different ways to manage the beamlines for one, and (mostly) we all have different kinds of equipments. Now, my point was that I would strongly encourage and see advantages in getting as flexible as possible, wherever possible.

Reading back your comments, I can understand that things are more or less straight forward with MD2/3. As you certainly know, we (PROXIMA-1) are not on this MD2/3, so we need to code quite a bit to get your same safety procedure working well at our level. My approach, as I believe it should be, is to get the main code working on our beamline, with all the 'beamline specific features' not crashing with the main branch. As I can see things right now: the main code implies getting opened to the users the access to the Fast-shutter, which has caused problems at our PROXIMA-1 in the recent past.

The solution of Ivars sounds round: to get the button somehow 'hidden' to classical users, linked to an alignment procedure not fully used by all the users. (to go further here: I would personally not give users access to any alignment-related procedures, but again, this is on PROXIMA-1...). Yet, I think we could try to make this part a bit more 'exportable' to all the various beamlines.

My proposal would be as follows: would it be possible to have the fast shutter, or whatever shutter, sub-divided in three different parts; i.e. status, open, and close? Again very pragmatically, I see no advantage at PROXIMA-1 in getting the users opening the fast shutter, however closing it might be necessary. Similarly, I see no need for the users to close the front-end shutter, while opening it might become required. Safety shutters could get benefits from both opening and closing...

In the above scheme and for users willing to (and knowing how and why to do so) open the fast shutter, the 'Expert' mode or whatever super-user options could just make all of fast shutter control possible.

Cheers, leo

-
Leonard Chavas
- 
Synchrotron SOLEIL
Proxima-I
L'Orme des Merisiers
Saint-Aubin - BP 48
91192 Gif-sur-Yvette Cedex
France
- 
Phone:  +33 169 359 746
Mobile: +33 644 321 614
E-mail: leonard.chavas@synchrotron-soleil.fr
-

IvarsKarpics:
Hi all,

Before going more deep. I would suggest to add this topic to the agenda of mxcube meeting.
Depending from time we can add it to the GUI session, developers session (the one where steering committee and developers separate) or to developers workshop.
I think it is better to gather all opinions, show some existing or possible implementations and agree on a solution.

Ivars

hnaalas:
Great idea! PROXIMA-1 will be off, not being able to join the meeting, however PROXIMA-2A will be present and shall 'defend our crusade' ^^

Cheers, leo

-
Leonard Chavas
- 
Synchrotron SOLEIL
Proxima-I
L'Orme des Merisiers
Saint-Aubin - BP 48
91192 Gif-sur-Yvette Cedex
France
- 
Phone:  +33 169 359 746
Mobile: +33 644 321 614
E-mail: leonard.chavas@synchrotron-soleil.fr
-

PR #258, opened 2016-06-20
Samplegrid - improvements after feedback

marcus-oscarsson:
Hello all,

I hope you all spent a great weekend !.

Here comes a small update after the feedback we had from you and the discussions we have had here. There is still plenty of things to do, we hope to be able to integrate with the queue and ISPyB soon. I hope these small improvements can provide a fairly good idea of how it could be when its finished.

The order of the sample in the queue is now only displayed on picked samples. The order is displayed in the bottom right corner, the layout of the cards have not been fixed yet so this might change (ideas are very welcome). All tough it kind of makes sense that the cards are moved to the top left most available position, we think this might be disturbing to the user. We believe that this is something that will be disturbing since the cards will be moving back and forth as you pick them. The risk is further that the user gets a bit lost when the cards are moved automatically, especially when there is many cards. That being said, its fairly easy to add this feature if we later discover that its something that the users want.

We discussed the possibility to have a "picked" sample filter the last time you where here, so we have made a preliminary implementation of that.  You can for the moment use the keyword "is" with a property like "picked". For example typing "is:picked" in the filter box shows only the picked samples. The only combination that works for the moment is "is:picked" :). The exact implementation of the filter functionality have not been decided yet.

Apart from this there are some small layout updates, there is a new little "target" icon on the sample cards which is for working in "sample-by-sample" mode. This icon is a bit too similar to the move icon, so we might need to change one of them. Or maybe come up with a smarter way of accessing the "mount sample directly" action :).

Thanks again for the feedback, and we are really happy to know that you think. I'm sure your going to talk about it next week. I'm really sorry that I cant meet you all to discuss it :(.

Cheers,
Marcus

### DIFF
--- a/mxcube3/ui/actions/SamplesGrid.js
+++ b/mxcube3/ui/actions/SamplesGrid.js
@@ -113,6 +113,7 @@ export function sendManualMountRequest(manual) {
         dispatch(showErrorPanel(true, 'Could not toogle manual mode'));
       } else {
         dispatch(sendClearQueue());
+        dispatch(updateSampleListAction({}));
         dispatch(setManualMountAction(manual));
         if (manual) {
           dispatch(showTaskForm('AddSample'));
--- a/mxcube3/ui/actions/queue.js
+++ b/mxcube3/ui/actions/queue.js
@@ -182,7 +182,6 @@ export function sendClearQueue() {
       }
     }).then(() => {
       dispatch(clearAll());
-      dispatch(updateSampleListAction({}));
     });
   };
 }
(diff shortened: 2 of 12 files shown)


marcus-oscarsson:
Small update on the filter functionality,

We originally thought that have the possibility to use keywords in the filter, like is:picked. However its maybe better to present the user with a list of filter options. Like a list of check boxes that turns on and off a certain filter. Its in this way easy to combine filters without the need of typing a long string of keywords. The filter text box will be kept so that its possible to filter on Acronym, sample name and location. The idea is that the little drop down to the right of the textbox will show a bigger grid/table with filter options that can be combined.

Marcus

PR #259, opened 2016-06-20
Using unified definition for input background colors

marcus-oscarsson:
Hey,

I wanted to update the color scheme used for the beamline setup inputs so that they corresponded to the rest of the UI. Doing this I thought that it was better to define common css classes for the various states. So I added those to Main.css so that its easier to reuse.

I also added a css transition since I was using this for the beamline setup, I honestly don't know if thats very nice/good. Let me know what you think, I don't mind removing that. I initially though that it was nicer than just changing colors, but the intermediate grey-brown is not that nice either :).

Cheers,
Marcus


bolmsten:
This is a great idea! I have been thinking about cleaning up the css as well.

marcus-oscarsson:
God morning, Fredrik !

:), Cleaning up the css is a good idea too

PR #266, opened 2016-06-29
Add `redux-persist` for UI state client-side persistence

mguijarr:


bolmsten:
Hi @mguijarr ,

This has some linting rules to be fixed. 

If I understand correctly this is only saving locally and checking the proposal ID in local storage? It has no synchronisation with the server? 

/ Fredrik

mguijarr:
Linting errors fixed.

Indeed, this is only the local storage part. This code is in the same state as after MXCuBE meeting.

We need to continue with implementing more advanced stuff once this gets accepted.

PR #268, opened 2016-06-29
[wip] Marcus enhancements

mguijarr:
Latest code from Marcus before holidays, see commit messages for details

### DIFF
--- a/mxcube3/ui/actions/SamplesGrid.js
+++ b/mxcube3/ui/actions/SamplesGrid.js
@@ -76,11 +76,20 @@ export function sendSyncSamplesRequest(proposalId) {
 export function addTaskAction(sampleQueueID, sampleID, task, parameters) {
   return { type: 'ADD_TASK',
            taskType: task.Type,
-           index: sampleID,
+           sampleID,
            parentID: sampleQueueID,
            queueID: task.QueueId,
            parameters
          };
+}
+
+
+export function addTask(sampleID, parameters) {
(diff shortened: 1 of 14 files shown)


bolmsten:
I find this very interesting and I like the idea of cleaning up the queue but I get this when trying to add a sample:

" action @ 13:09:18.400 ADD_TASK 
main.js:128355  prev state Object
main.js:128359  action Object
main.js:128363  error TypeError: tasks.concat is not a function(…)
main.js:128367  next state Object
main.js:127042 Uncaught TypeError: tasks.concat is not a function
"

Should this be renamed as WIP ? And discussed further

mguijarr:
I will check when I am back from holidays.

mguijarr:
I confirm this is a problem - it happens when adding tasks from right-click in Sample View (not from the grid). 

This is because in this PR `state.queue`  is supposed to be a list (`[]`) instead of a dictionary-like object (`{}`), and the change was half-done.

It's ok for me if it stays as a WIP for a while - it will be part of a proposal for a simplification of the data
structures.

marcus-oscarsson:
Hi all,

Thanks for taking care of this Matias and sorry for the problems is caused you Frederik. I'll start to take a look at this later today.

Cheers !

marcus-oscarsson:
Hi,

Its better to divide the final result of this work in several PRs to make it easier to merge. So I will close this one.

Marcus

PR #271, opened 2016-07-04
Minor fixes and linting

bolmsten:
This PR contains a first version of a userlog for the sampleview. Matias you mentioned that Marcus had done a lot of good work for filtering what messages to send and show, where is this located? 

I also did some linting fixes, I am trying to get down to zero. 

For the rest please check the commits.


mguijarr:
Will check all this next week.

mxcube3/ui/components/SampleView/SampleControls.js
```diff
@@ -15,6 +17,7 @@ export default class SampleControls extends React.Component {
     this.setApertureSize = this.setApertureSize.bind(this);
     this.toogleFrontLight = this.toogleLight.bind(this, 'front');
     this.toogleBackLight = this.toogleLight.bind(this, 'back');
+    this.toogleCentring = this.toogleCentring.bind(this);
```
mguijarr: `toogle` should be `toggle`

mxcube3/ui/components/SampleView/SampleControls.js
```diff
@@ -1,3 +1,5 @@
+/* eslint no-unused-expressions: ["error", { "allowTernary": true }] */
```
mguijarr: Is this changing a rule for this file only ?

I think it's better if we follow the rules without exceptions.

mxcube3/ui/serverIO.js
```diff
@@ -32,6 +37,10 @@ export default class ServerIO {
       }
     });
 
+    this.hwrSocket.on('beam_value_change', (data) => {
```
mguijarr: `beam_value_change` is a bit confusing, I would prefer something like `beam_position_change`

mguijarr:
The branch has conflicts, so let's solve those first - I took the opportunity to put comments

mxcube3/ui/serverIO.js
bolmsten: It is not only position. It is also beam size and shape,  so we choose to use something more generic.

mxcube3/ui/components/SampleView/SampleControls.js
bolmsten: It is complaining about this:

"    clickCentring ? sendAbortCentring() : sendStartClickCentring() " 

I can add it globally but I think we should be allowed to use "?" operator, do you agree?

mxcube3/ui/serverIO.js
mguijarr: What about `beam_change` then ? `value` adds nothing here

mxcube3/ui/components/SampleView/SampleControls.js
mguijarr: Personally I have nothing against ternary operator.

But we decided to comply with some linting rules, so... I propose to either replace it with the more
verbose version:

' ' '
if (clickCentring) {
  sendAbortCentring()
} else {
  sendStartClickCentring()
}
'''

Or to globally change the rules to add something like [no-unneeded-ternary](http://eslint.org/docs/rules/no-unneeded-ternary), if it works in this case.

bolmsten:
Fixed comments now

PR #274, opened 2016-07-06
Listen to beam postion change

bolmsten:
With this PR mxcube changes the beam position and size depending on the signal sent from the server,


mxcube3/ui/serverIO.js
```diff
@@ -32,6 +32,10 @@ export default class ServerIO {
       }
     });
 
+    this.hwrSocket.on('beam_value_changed', (msg) => {
```
mguijarr: Is this a kind of conflict with PR #271 ? It seems to me the same lines are added twice

PR #278, opened 2016-07-11
hotifx

meguiraun:
the last PRs have introduced some bugs... I opened this PR as **WIP** so we are aware of it.

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -67,10 +67,10 @@ if not app.debug or os.environ.get("WERKZEUG_RUN_MAIN") == "true":
     import logging_handler
     #root_logger = logging.getLogger()
     #root_logger.setLevel(logging.DEBUG)
-    hwr_logger = logging.getLogger("HWR")
+    #hwr_logger = logging.getLogger("HWR")
+    hwr_logger = logging.getLogger("user_level_log")
     custom_log_handler = logging_handler.MX3LoggingHandler()
     custom_log_handler.setLevel(logging.DEBUG)
-    #root_logger.addHandler(custom_log_handler)
     hwr_logger.addHandler(custom_log_handler)
     app.log_handler = custom_log_handler
 
--- a/mxcube3/routes/Diffractometer.py
+++ b/mxcube3/routes/Diffractometer.py
@@ -94,7 +94,7 @@ def get_movables_state():
     ret = {}
     for movable in mxcube.diffractometer.centring_motors_list:
         motor_info = Utils.get_movable_state_and_position(movable)
-	if motor_info[movable]['position']:
+	if motor_info[movable]['position'] is not None:
             ret.update(motor_info)
 
     ret.update(Utils.get_light_state_and_intensity())
--- a/mxcube3/routes/Diffractometer.py
+++ b/mxcube3/routes/Diffractometer.py
@@ -93,7 +93,9 @@ def md_in_plate_mode():
 def get_movables_state():
     ret = {}
     for movable in mxcube.diffractometer.centring_motors_list:
-        ret.update(Utils.get_movable_state_and_position(movable))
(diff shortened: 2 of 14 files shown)


meguiraun:
removed wip flag, ready to click merge...

mxcube3/routes/Diffractometer.py
```diff
@@ -93,7 +93,9 @@ def md_in_plate_mode():
 def get_movables_state():
     ret = {}
     for movable in mxcube.diffractometer.centring_motors_list:
-        ret.update(Utils.get_movable_state_and_position(movable))
+        motor_info = Utils.get_movable_state_and_position(movable)
+	if motor_info[movable]['position']:
```
mguijarr: I do not understand why we need this test. Moreover, if `motor_info[movable]['position']` is 0, the `ret` dict won't be updated, which is certainly not what we want ?

mxcube3/routes/signals.py
```diff
@@ -132,12 +132,14 @@ def motor_event_callback(*args, **kwargs):
     motors_info = dict()
     # the centring motors are: ["phi", "focus", "phiz", "phiy", "zoom", "sampx", "sampy", "kappa", "kappa_phi"]
     for name in mxcube.diffractometer.centring_motors_list:
-        motors_info.update(Utils.get_movable_state_and_position(name))
+	motor_info = Utils.get_movable_state_and_position(name)
+	if motor_info[name]['position']:
```
mguijarr: Same question as above

meguiraun:
my fault, if we send ```None``` as motor position the client does not work.

Fixed the condition.

mxcube3/__init__.py
```diff
@@ -67,10 +67,10 @@ def exception_handler(e):
     import logging_handler
     #root_logger = logging.getLogger()
     #root_logger.setLevel(logging.DEBUG)
-    hwr_logger = logging.getLogger("HWR")
+    #hwr_logger = logging.getLogger("HWR")
+    hwr_logger = logging.getLogger("user_level_log")
```
mguijarr: Sorry to come back on this now... Can you explain why there is a huge improvement
in speed by using this logger instead of HWR ?
I understand there is much less traffic on this one, but it's not the same : one is a general purpose
log (HWR), the other one is just for showing some messages to the user (user_level_log).
With this change we don't process a lot of messages that were displayed in MXCuBE 2 in the
System tab. Is it what we want ?
There is still the idea to use Sentry for example for storing HWR log, then in this case we don't need all the log messages in MXCuBE 3... However I didn't install it here yet, it's a more complex task than just having log messages in the application. Did you give it a try ?

meguiraun:
We saw that sending all the hwr messages introduced a long delay in the client (several seconds after a centring is finished to get the updated positions of the motors for example), so we reduce the number of sent messages by this change, as a first approach to an optimisation. The client can not digest all the messages quickly enough.

I know that with this change we are receiving less info on the client, but as you say, something like sentry would solve the problem. But we have not done anything with that. So, no, we have not tested any alternative for getting the hwr messages to the web, but the client was almost unusable otherwise. I am sorry because it looks as I sneak this commit without proper discussion...

I think that at this stage of the development, this could be a solution. Do you think that we need hwr messages in the client now? In that case we would have to start optimising (both server and client?).

mguijarr:
I imagine you experienced the performance problem in development mode, right ? (still using webpack-dev-server ?). Webpack-dev-server doesn't know how to deal with websockets so when we develop we are always in a degraded state (socket.io degrades 'gracefully').
I don't think there are a lot of messages, so if you tested it in production and the problem is there we have a serious performance issue !
The solution for development will be to switch to webpack 2.0, as the problem with websockets has been fixed. But this release is not ready yet AFAIK.
So, for now it's ok but we should keep an eye on all this.

PR #279, opened 2016-07-14
Making changes to the client to make persist work

bolmsten:
I have blacklisted a number of reducers that redux-persist should nut rehydrate and I also changed two reducers to make the merge between rehydration and initialisation work.

### DIFF
--- a/mxcube3/ui/main.jsx
+++ b/mxcube3/ui/main.jsx
@@ -45,7 +45,7 @@ export default class App extends React.Component {
     this.serverIO = new ServerIO(store.dispatch);
     this.serverIO.listen();
 
-    const persistor = persistStore(store, {}, () => {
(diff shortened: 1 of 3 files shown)


mguijarr:
Are we sure about not persisting `sampleview` reducer ?
Of course we don't want to keep the motor positions, but what about the other things ?
I thought we can always persist everything, then on rehydration we synchronize (for example,
for motor positions etc.).
I am just wondering... If it's clear for you I will accept the PR.

bolmsten:
Hi @mguijarr ,

I am sure at this moment, I have tested it and you get all the necessary information for the sampleview from the initialisation call. 

I think we should keep the localstorage as lean as possible for many reasons:

1. It takes some effort to save on every action down to local storage
2. All this information will in the future need to be synced with the server
3. The less information we save the less risk of synchronisation problems that need to be solved

If things change we can add more information to the persist of course.

PR #282, opened 2016-07-18
Last linting possible!

bolmsten:
I have now gone through and fixed all possible linting errors.

We have only two errors left :

main.js:721 ./mxcube3/ui/reducers/index.js

/Users/frebol/Desktop/master/mxcube3/mxcube3/ui/reducers/index.js
  36:5  error  Assignment to function parameter 'state'  no-param-reassign

/Users/frebol/Desktop/master/mxcube3/mxcube3/ui/components/SampleQueue/TaskItem.js
  67:5  error  Assignment to property of function parameter 'monitor'  no-param-reassign

The only way to fix these are to do exceptions in the header of the file.  If you agree I will do this and we then can have a totally clean repo?

### DIFF
--- a/mxcube3/ui/actions/general.js
+++ b/mxcube3/ui/actions/general.js
@@ -16,20 +16,16 @@ export function showErrorPanel(show, message = '') {
   };
 }
 
-function checkStatus(response) {
+function parse(response) {
   if (response.status >= 200 && response.status < 300) {
-    return response;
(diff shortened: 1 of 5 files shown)


mguijarr:
It's ok to add exceptions, however I also have no problem if those 2 errors remain. As you want.

Thanks for the great job of solving the linting errors !

PR #283, opened 2016-07-19
Real Characterisation

meguiraun:
(**Removed WIP**)

Here it is the characterisation stuff. Tested in Biomax.

Basically a callback listens to the `child_added` and it checks if it is  a _DataCollection_, and if so, checks if it has _Diffraction plan_  in the name, and if so it creates the missing entries (DataCollectionQueueEntry and TaskGroupQueueEntry) and enqueues them.

Finally it emits `add_task` signal through the socketio.

Sorry for this long commit history, there have been several bug fixes here and there during the development and testing. 

What is left (and I left out for now)
* A way of displaying results: a link to ispyb? new tab displaying the edna result html?

### DIFF
--- a/mxcube3/routes/Queue.py
+++ b/mxcube3/routes/Queue.py
@@ -26,6 +26,7 @@ def init_signals(queue):
                            signals.collect_oscillation_failed)
     mxcube.collect.connect(mxcube.collect, 'collectOscillationFinished',
                            signals.collect_oscillation_finished)
+    queue.connect(queue, 'child_added', qutils.add_diffraction_plan)
 
     queue.queue_hwobj.connect("queue_execute_started",
                               signals.queue_execution_started)
--- a/mxcube3/routes/qutils.py
+++ b/mxcube3/routes/qutils.py
@@ -352,12 +352,12 @@ def set_dc_params(model, entry, task_data):
     acq.path_template.base_prefix = params['prefix']
 
     full_path = os.path.join(mxcube.session.get_base_image_directory(),
-                             params.get('path', 'dummy_path'))
+                             params.get('path', ''))
 
     acq.path_template.directory = full_path
 
     process_path = os.path.join(mxcube.session.get_base_process_directory(),
-                                params.get('path', 'dummy_path'))
+                                params.get('path', ''))
     acq.path_template.process_directory = process_path
 
     # If there is a centered position associated with this data collection, get
--- a/mxcube3/routes/qutils.py
+++ b/mxcube3/routes/qutils.py
@@ -12,6 +12,7 @@ import queue_entry as qe
 from flask import jsonify
 from mock import Mock
 from mxcube3 import app as mxcube
+from mxcube3 import socketio
 
 
 class PMock(Mock):
@@ -46,7 +47,11 @@ def node_index(node):
         task_groups = sample_model.get_children();
         group_list = [group.get_children() for group in task_groups]
         tlist = [task for task_list in group_list for task in task_list]
-        index = tlist.index(node)
+        # there is probably a better fix...
+        try:
+            index = tlist.index(node)
+        except Exception:
+            print 'node not in list: ', node
 
     return {'sample': sample, 'idx': index, 'queue_id': node._node_id}
 
@@ -344,11 +349,16 @@ def set_dc_params(model, entry, task_data):
 
     acq.acquisition_parameters.set_from_dict(params)
     acq.path_template.set_from_dict(params)
+    acq.path_template.base_prefix = params['prefix']
 
     full_path = os.path.join(mxcube.session.get_base_image_directory(),
                              params.get('path', 'dummy_path'))
 
     acq.path_template.directory = full_path
+
+    process_path = os.path.join(mxcube.session.get_base_process_directory(),
+                                params.get('path', 'dummy_path'))
+    acq.path_template.process_directory = process_path
 
     # If there is a centered position associated with this data collection, get
     # the necessary data for the position and pass it to the collection.
@@ -516,18 +526,17 @@ def add_diffraction_plan(parent, child):
             # then we do know that we need to add the entry here
             # Create a new entry for the new child, in this case a data collection
             dc_entry = qe.DataCollectionQueueEntry(PMock(), child)
+            dcg_entry = qe.TaskGroupQueueEntry(PMock(), parent)
+
+            sample = parent.get_parent()  # mxcube.queue.get_model_root()
+            sample_model, sample_entry = get_entry(sample._node_id)
+            # TODO: check if the parent entry exits in case multiple diff plans
+            sample_entry.enqueue(dcg_entry)
+            enable_entry(parent._node_id, False)
+
             # Add the entry to the newly created task group, brother to the characterisation
-            parent_entry.enqueue(dc_entry)
-            enable_entry(child._node_id, True)
-
-            # Tell client that we added a new entry for the diffraction plan
-            params = child.as_dict()
-            msg = {'sampleID': parent.get_parent()._node_id,
-                   'task': {
-                        'params': params,
-                        'Type': 'DataCollection'
-                            }
-                   }
-            # log.info("Diffraction plan added to the queue: %s" % str(msg))
-
+            dcg_entry.enqueue(dc_entry)
+            enable_entry(child._node_id, False)
+            msg = _handle_dc(sample._node_id, child)
+            # TODO: add saved centring pos id, centred_position is removed in _handle_dc
             socketio.emit('add_task', msg, namespace='/hwr')
--- a/mxcube3/routes/qutils.py
+++ b/mxcube3/routes/qutils.py
@@ -511,7 +511,7 @@ def add_diffraction_plan(parent, child):
         parent_model, parent_entry = get_entry(parent._node_id)
         # the parent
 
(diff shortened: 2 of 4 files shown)


mxcube3/routes/qutils.py
```diff
@@ -43,10 +44,14 @@ def node_index(node):
     elif node.get_parent():
         sample_model = node.get_parent().get_parent()
         sample = sample_model.loc_str
-        task_groups = sample_model.get_children();
+        task_groups = sample_model.get_children()
         group_list = [group.get_children() for group in task_groups]
         tlist = [task for task_list in group_list for task in task_list]
-        index = tlist.index(node)
+        # there is probably a better fix...
```
marcus-oscarsson: As you state, I'm not sure that this is a fix. 

node_index(node) gets the index of node that is in the queue, it simply does not work on nodes that are not in the queue. So the question should be why is it called on nodes that are not in the queue. This try except basically only hides the problem, well we see it because there is the print but there is a risk of missing that. Maybe returning -1 or None for as index would be better.

But the question is really why are we trying to get the index in the queue of something that is not in the queue ?

mxcube3/routes/Utils.py
```diff
@@ -95,12 +97,142 @@ def my_execute_entry(self, entry):
         foo = ['collectOscillationFinished', 'collectOscillationFailed', 'warning']
         mxcube.collect.emit(random.choice(foo))
     elif isinstance(entry, qe.CharacterisationGroupQueueEntry):
+        logging.getLogger('HWR').info('[QUEUE] Executing CharacterisationGroupQueueEntry entry: %s' %entry)
```
marcus-oscarsson: I'm wondering in general, without having thought to much about it, why we need this method my_execute_entry ? It as I understand it redefines the way entries are executed ?. 

I'm also wondering why there are some time.sleep calls, do we have some kind of synchronization problems ?

mxcube3/routes/Queue.py
```diff
@@ -939,13 +939,15 @@ def serialize_queue_to_json():
             if child['py/object'].split('.')[1] == 'TaskGroup': #keep going down
                 for grandChild in child['_children']:
                     if grandChild['py/object'].split('.')[1] == 'TaskGroup': #keep going down one more time for the Char
+                        """ the following loop is not needed
```
marcus-oscarsson: Are you still using this method ?

mguijarr:
Thanks for all the work @meguiraun :)

It seems your branch for the PR is on top of Frederik's changes for helical.
Could you please clean your PR to only have changes related to characterization ?

bolmsten:
An alternative is to check if my PR is okay, though mine is clean and only contains Helical

marcus-oscarsson:
I think its the case too, this PR seems to be somehow rebased on top of Fredriks they share some commits. Its perhaps better to rebase it on master ?

mguijarr:
@bolmsten: ok but we didn't start reviewing yours yet, if we have changes to propose it would make things more difficult because the two branches are not independent. Anyway, it's unclean - there is no reason to have mixed branches. 

Moreover it's easy to fix: rebase on top of the current master, remove all commits that are not part of the characterization feature and push -f.

mxcube3/routes/qutils.py
meguiraun: thats tricky, at least to my brain... It was crashing during the char, I guess because there was a task without being properly enqueued into the queue, yet

mxcube3/routes/Utils.py
meguiraun: Sorry for not being clear... No, it is not used, it was used only in the beginning. You can ignore `my_execute_entry` completely.

I used this in the beginning when we did not have all the hwobj in place, so at least we could run and test client-server interaction without worrying about any hwobj. The sleep calls were for simulating an execution that takes time, otherwise we could not see the client switching task's colors.

mxcube3/routes/Queue.py
meguiraun: nope, where this came from? jarrll my friend git...

meguiraun:
snif snif... I will clean the commits...

mxcube3/routes/qutils.py
```diff
@@ -344,12 +349,17 @@ def set_dc_params(model, entry, task_data):
...
     acq.path_template.set_from_dict(params)
+    acq.path_template.base_prefix = params['prefix']
 
     full_path = os.path.join(mxcube.session.get_base_image_directory(),
                              params.get('path', 'dummy_path'))
 
     acq.path_template.directory = full_path
 
+    process_path = os.path.join(mxcube.session.get_base_process_directory(),
+                                params.get('path', 'dummy_path'))
```
marcus-oscarsson: What is 'dummy_path' used for ? Is it to prevent a KeyError exception, in this case 'dummy_path' is not much better ?

mxcube3/routes/Queue.py
```diff
@@ -383,6 +384,12 @@ def add_centring(id):
     return resp
 
 
+@mxcube.route("/mxcube/api/v0.1/results/<int:number>")
+def serve_char_result(number):
```
marcus-oscarsson: Its maybe better to have a function that just returns the result as HTML or JSON, instead of serving the file statically.  Because the results will later be stored in LIMS.

mxcube3/routes/Utils.py
marcus-oscarsson: I think its better to remove code that is not used, I understand that It might be good to have at some point but you can always get it back with git.

mxcube3/routes/Queue.py
marcus-oscarsson: Okey, I was afraid that the queue_to_json in qutils did not work as you wanted. So please if you are not using this method remove it.

marcus-oscarsson:
We can help you out to clean the commits if you like, do the changes that you want first and let us know.

mxcube3/routes/qutils.py
marcus-oscarsson: I see, you are right its not so easy, but you agree that its strange to ask for the index of the node that is not in the queue, and we should ask our self's why that's happening. Because the problem might also have other side effects :)

marcus-oscarsson: By the way the code itself is fine, we can maybe remove the print as some point.

mxcube3/routes/Utils.py
meguiraun: I started this pr before your big refactoring, so looking commit by commit you could see that, but it was not in the final code.

mxcube3/routes/Queue.py
meguiraun: I started this pr before your big refactoring, so looking commit by commit you could see that, but it was not in the final code.

mxcube3/routes/Utils.py
meguiraun: the git cleanup has already remove a lot of it

mxcube3/routes/Queue.py
meguiraun: as I said, how to present the result is left out of this pr, even if those lines were there. In any case I removed those

meguiraun:
Let's see now. Since I started with this before marcus pr there outdated commits.
Also, #338 has some bug fixes that might prevent the characterisation from running, we can accept that first if you prefer.

mxcube3/routes/qutils.py
meguiraun: would this be better:
```python
full_path = os.path.join(mxcube.session.get_base_image_directory(),
                    params.get('path', 'raw_data'))
process_path = os.path.join(mxcube.session.get_base_process_directory(),
                           params.get('path', 'proccessed_data'))

```

marcus-oscarsson:
Wow that's great Mikel, that simplified alot !

meguiraun:
I just hope that I have not left anything out!

mxcube3/routes/qutils.py
marcus-oscarsson: Well each institute have their schema appending raw_data and processed_data might work in most cases. The risk is that its added twice.

But why append anything at all if path does not exist ?, and path should always exist. The user cant add a collection
without a valid path so this will later be validated before the user adds the collection.

I know that this dummy_path thing have been there for while, we can fix that at later state, we can let it pass 
for now. I should maybe have removed that myself when I did the refactoring but I forgot. But if you have the time
simply remove the second argument to get, or set it to ''.

meguiraun: done!

marcus-oscarsson:
:), me too. Does it run at your side ?, I did not test it here.

In anyways I'm happy if you are happy, so I think we can merge if it runs like before.

Marcus

meguiraun:
mmm... I will re-test again in the beamline tomorrow morning. Hold on, but I think I would need some fixes from #338

marcus-oscarsson:
Okey, but I think its fine for us to merge it in anyways. Ill give the chance to Matias to give us his opinion.

mguijarr:
Looking at the code I don't see why this PR would need fixes from #338 ; I noticed one commit (` 5bbd22`) is present here and in Frederik's branch so I asked Frederik to rebase his branch (and also to remove the merge commits which make the history quite difficult to read). 

Did you have time to test this branch ? If you're happy with it I can merge it.

meguiraun:
tested here and working.

PR #287, opened 2016-07-19
Sample changer

mguijarr:
Draft version of the sample changer control panel - inspired by `SCBrick` from MXCuBE 2.

Not working yet !


mguijarr:
JFYI we will continue the work on this one tomorrow since our beamlines are free

mguijarr:
Continued to work on the SC control.

This can be merged - it offers basic functionality. Tested on ID30A3.

TODO:
- clicking on the data matrix icon for the whole sample changer (root of the tree) has to trigger a full recursive scan
- more testing on loading / unloading samples
- improving UI

meguiraun:
Thanks for the work! We are going to review and test this during today.

mguijarr:
Do you already have your sample changer ?

This code is meant to work with the implementations of GenericSampleChanger.py

mguijarr:
By the way we already have some idea to improve the UI : it would be great to put the sample changer control in the same page as the sample grid. We could get rid of the 'Check sample changer contents' button, then. I am thinking of something that could slide-in from the left (or from the right) displaying the sample changer tree view, and a big arrow button to feed the sample grid for example. Well, something along those lines... 

It would remove a tab and would probably make things clearer for users.

Devil is in the details, like always... But what do you think ?

mguijarr:
Another TODO: there are now two routes to load / unload samples. This has to be sorted out - there should only be one. Normally the routes I added yesterday are the ones to keep, since it is just doing basic functionality. The routes that were there before have to be removed ; sample loading/unloading should be done directly from the queue somehow.

meguiraun:
Unfortunately we do not have our sample changer yet... But we could review and test as much as we can

bolmsten:
I fixed some linting but everything else looks good!

PR #288, opened 2016-07-20
After pep8

meguiraun:
fixes after the pep 8 refactoring

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -82,7 +82,9 @@ if not app.debug or os.environ.get("WERKZEUG_RUN_MAIN") == "true":
         app.session = app.beamline.getObjectByRole("session")
         app.collect = app.beamline.getObjectByRole("collect")
         app.diffractometer = app.beamline.getObjectByRole("diffractometer")
-        if getattr(app.diffractometer, 'centring_motors_list') is None:
+	try:
+	    app.diffractometer.centring_motors_list
+	except AttributeError:
             # centring_motors_list is the list of roles corresponding to diffractometer motors
             app.diffractometer.centring_motors_list = app.diffractometer.getPositions().keys()
         app.db_connection = app.beamline.getObjectByRole("lims_client")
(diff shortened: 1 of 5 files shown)


mxcube3/__init__.py
```diff
@@ -82,7 +82,7 @@ def complete_initialization(app):
         app.session = app.beamline.getObjectByRole("session")
         app.collect = app.beamline.getObjectByRole("collect")
         app.diffractometer = app.beamline.getObjectByRole("diffractometer")
-        if not hasattr(app.diffractometer, 'centring_motors_list'):
+        if getattr(app.diffractometer, 'centring_motors_list') is None:
```
mguijarr: This is not good since `getattr` raises `AttributeError` exception if the designated
member does not exist. If you want to keep `getattr` put `try...except AttributeError` or just use `hasattr` ;) What's wrong why pep8 and `hasattr` by the way ?

meguiraun: Nothing wrong with pep8 and hasattr... but it was not working as intended:
 ```hasattr(app.diffractometer, 'centring_motors_list')``` and ```hasattr(app.diffractometer, 'whateveryouputhere')``` were returning ```True``` in both cases.
In addition, the ```getattr``` was returning ```None``` for the mockup, no exception raising

I think I am missing something...

mguijarr: The mockup is wrong. There is a `__getattr__` that always returns `None` if attribute is
missing, see those lines in `DiffractometerMockup.py`:

```
    # to make it compatibile
    def __getattr__(self, attr):
        if attr.startswith("__"):
            raise AttributeError(attr)
        else:
            if attr == "currentCentringProcedure":
                return self.current_centring_procedure
            if attr == "centringStatus" :
                return self.centring_status
```

It would be great to add `raise AttributeError(attr)` at the end.

Meanwhile, the solution could be to replace `hasattr` test by `try...except AttributeError`
like this:

```
    try:
        app.diffractometer.centring_motors_list
    except AttributeError:
        # my code in the 'if'
        ...
```

What do you think ?

meguiraun:
I see it now. Modified as you suggest

mguijarr:
Thanks !

PR #289, opened 2016-07-20
the attr was still None

meguiraun:
the previous pr did not solve er... anything... let's see now.  ```app.diffractometer.centring_motors_list```		will always be None

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -83,10 +83,11 @@ if not app.debug or os.environ.get("WERKZEUG_RUN_MAIN") == "true":
         app.collect = app.beamline.getObjectByRole("collect")
         app.diffractometer = app.beamline.getObjectByRole("diffractometer")
 	try:
-	    app.diffractometer.centring_motors_list
+	    if getattr(app.diffractometer, 'centring_motors_list') == None:
+	        app.diffractometer.centring_motors_list = app.diffractometer.getPositions().keys()
 	except AttributeError:
-            # centring_motors_list is the list of roles corresponding to diffractometer motors
-            app.diffractometer.centring_motors_list = app.diffractometer.getPositions().keys()
+	    # centring_motors_list is the list of roles corresponding to diffractometer motors
+	    logging.getlogger('HWR').error('Error getting diffractometer centring motors')
         app.db_connection = app.beamline.getObjectByRole("lims_client")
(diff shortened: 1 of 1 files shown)


mxcube3/__init__.py
```diff
@@ -83,10 +83,11 @@ def complete_initialization(app):
         app.collect = app.beamline.getObjectByRole("collect")
         app.diffractometer = app.beamline.getObjectByRole("diffractometer")
 	try:
-	    app.diffractometer.centring_motors_list
+	    if getattr(app.diffractometer, 'centring_motors_list') == None:
```
mguijarr: I also have my brain on holidays ;) Of course with the broken Mockup we have it always
return `None`.
But this code is not good neither (sorry): when used with a normal diffractometer object, it would never be equal to `None` and it would never do `app.diffractometer.centring_motors_list = app.diffractometer.getPositions().keys()`.
So, I suggest to use the third argument to `getattr`, to specify `None` by default.

Code would be:

```
    if getattr(app.diffractometer, 'centring_motors_list', None) is None:
        app.diffractometer.centring_motors_list = app.diffractometer.getPositions().keys()
```

There is no need for `try...except` anymore, then.

Could you please try it and change your PR accordingly if it works ?

Thanks in advance !

meguiraun: but with a normal diff the ```centring_motors_list ``` will always contains the motor list, coming from the generic diff, so we do not need the ```keys()``` stuff. I tested this code with real and mockups.

mguijarr: I added this code for backward compatibility. We are not on 2.2 here, and there is no `centring_motors_list` in our diffractometer objects. I need this code to be able to continue with my tests and development on the ESRF beamlines.

In its current state I cannot merge the PR because it is missing the third argument to `getattr` : at the moment it would raise the `AttributeError` exception on a diffractometer object without `centring_motors_list`.

meguiraun: I added the third argument. I was not thinking on your non 2.2 version, sorry!!

mguijarr:
Thanks Mikel. 

I hope it is going to fix this mess now.

I am in a rush with many tasks so I really appreciate you took care of this.

meguiraun:
A pleasure! We are a team! Hope you still can enjoy the summer.

PR #292, opened 2016-07-22
mouse wheel usage

meguiraun:
super cool feature!!
With the mouse on the canvas, scrolling the wheel does:
* Zoom in/out
* If ```CTRL``` key is pressed, rotate omega by the step size defined in its box


IvarsKarpics:
Wow, nice feature!
Sorry, but I will steal the idea and implement it also in Qt4.

meguiraun:
ooh nooo! then web version will never be like qt!!! hehehe
Great you like it!

PR #293, opened 2016-07-25
Queue.py Fixes

marcus-oscarsson:
The problem is that sample_id is a number when a new sample is added manually and there is nothing that prevents us from using the NEW SAMPLE control in the queue when use_sc is true.

We consequently try to split a number, which of course will fail.

This should maybe be handled differently and the original problem will disappear if/when the NEW SAMPLE control is removed from the queue. I simply but a simple log message to notify us about a potential problem.

### DIFF
--- a/mxcube3/routes/Queue.py
+++ b/mxcube3/routes/Queue.py
@@ -270,8 +270,8 @@ def execute_entry_with_id(node_id):
                     mxcube.queue.last_queue_node.update({'id': elem['QueueId'],
                                                          'sample': queue[node_id]['SampleId']})
                     # mxcube.queue.queue_hwobj.execute_entry = types.MethodType(Utils.my_execute_entry, mxcube.queue.queue_hwobj)
-                    mxcube.queue.queue_hwobj.execute_entry(childEntry)
-                    childEntry.set_enabled(False)
+                    mxcube.queue.queue_hwobj.execute_entry(child_entry)
+                    child_entry.set_enabled(False)
                 except Exception:
                     logging.getLogger('HWR').exception('[QUEUE] Queue error executing child entry with id: %s' % elem['QueueId'])
     else:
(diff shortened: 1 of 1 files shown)


mxcube3/routes/Queue.py
```diff
@@ -234,7 +230,7 @@ def execute_entry_with_id(node_id):
         (any further error might still happen)
         :statuscode: 409: queue entry could not be executed
     """
-    last_queue_node = mxcube.queue.last_queue_node
+    # last_queue_node = mxcube.queue.last_queue_node
```
marcus-oscarsson: Hey

I noticed this when I skimmed trough the file, its maybe something that remains after the pep8 work.  It has of course no impact but thought it was better to let you know.

mxcube3/routes/Queue.py
```diff
@@ -535,6 +535,7 @@ def add_centring(id):
         409 something bad happened. Plus:
        data ={ "CentringId": newId}
     '''
+    params = request.get_json()
```
marcus-oscarsson: Perhaps also from pep8 work, params was undefined I assumed that was the request parameters we want. But maybe not ?

mxcube3/routes/Queue.py
```diff
@@ -715,9 +716,9 @@ def update_method(sample_id, method_id):
     """
     params = request.data
     params = json.loads(params)
-    sample_node = mxcube.queue.get_node(int(sample_id))
+#    sample_node = mxcube.queue.get_node(int(sample_id))
```
marcus-oscarsson: Same as above :)

mxcube3/routes/Queue.py
meguiraun: it has been here since the first crystal... good catch!

mxcube3/routes/Queue.py
meguiraun: The client will never be able to add a centring ~directly, i.e., it will be done through a datacollection. You are right about the missing ```params```, but perhaps we can remove completely ```add_centring``` method?

PR #294, opened 2016-07-26
SampleView Improvements

bolmsten:
This PR contains a few shortcuts in operating the SampleView to make the life of a user a bit easier.

We have added a few listeners on top of the SampleView:

1. Scroll
2. Double Click

With the scroll you now have three actions when over the SampleView:

1. Scroll to Zoom in/out
2. Hold Ctrl and Scroll to rotate omega (Using stepsize set by user for the motor)
3. Hold Alt and Scroll to focus (Using stepsize set by user for the motor)

And with the double click we have implemented goToPosition, this function can also be accessed by the context menu.

### DIFF
--- a/mxcube3/routes/SampleCentring.py
+++ b/mxcube3/routes/SampleCentring.py
@@ -623,9 +623,7 @@ def pos_from_x_y():
             }
     # new positions:
     mot_y = mxcube.diffractometer.getObjectByRole('phiz')
-    print "ALING Z go to pos:  ", mot_y.getPosition() + data['Y']
     mot_x = mxcube.diffractometer.getObjectByRole('phiy')
-    print "ALING Y go to pos:  ", mot_x.getPosition() + data['X']
-    mot_y.moveRelative(data['Y'])
-    mot_x.moveRelative(data['X'])
+    mot_y.moveRelative(data['Y'])  # -1 md3
+    mot_x.moveRelative(-1*data['X'])  # -1 md2 and md3
     return Response(status=200)
--- a/mxcube3/routes/SampleCentring.py
+++ b/mxcube3/routes/SampleCentring.py
@@ -601,8 +601,31 @@ def accept_centring():
 
 @mxcube.route("/mxcube/api/v0.1/sampleview/centring/reject", methods=['PUT'])
 def reject_centring():
-    """
-    Reject the centring position.
-    """
+    """Reject the centring position."""
     mxcube.diffractometer.rejectCentring()
+    return Response(status=200)
+
+
+@mxcube.route("/mxcube/api/v0.1/sampleview/moveto", methods=['PUT'])
+def pos_from_x_y():
+    """Go to given x, y."""
+    params = request.data
+    params = json.loads(params)
+    click_position = params['clickPos']
+    logging.getLogger('HWR').info("A point submitted, x: %s, y: %s"
+                                  % (click_position['x'],
+                                     click_position['y']))
+    pixels_per_mm_x, pixels_per_mm_y = mxcube.diffractometer.get_pixels_per_mm()
+    center_x = mxcube.diffractometer.image_width / 2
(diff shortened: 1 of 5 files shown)


meguiraun:
Please have a look at the server call. It is working for us in the MD2 (and with minor modifications in the MD3). I am wondering if that will still work for you (without using the centring-math etc).

meguiraun:
removed WIP tag.

mxcube3/routes/SampleCentring.py
```diff
@@ -601,8 +601,29 @@ def accept_centring():
...
-    """
-    Reject the centring position.
-    """
+    """Reject the centring position."""
     mxcube.diffractometer.rejectCentring()
     return Response(status=200)
+
+
+@mxcube.route("/mxcube/api/v0.1/sampleview/moveto", methods=['PUT'])
+def pos_from_x_y():
```
mguijarr: It seems to me this method is to move a (x, y) position on sample video to beam position. Right ?

There are a number of problems:
- the beam might not be at the center (it is the case for MD2M, MD2, MD2S, MD3, ...) but believe me it can be elsewhere (on PX1 at Soleil maybe... Or any other diffractometer that doesn't assume beam on the centre of sample video)
- this assumes (x, y) positions from video corresponds to phiy and phiz roles, which we cannot know for all diffractometers

In fact there is a `moveToBeam` method (in MiniDiff.py) or `start_move_to_beam` (in GenericDiffractometer.py) that should be used, to delegate all the job to the Hardware Object not the Python server.

meguiraun: kind of... but it only tries to move to the center of the image, it ignores anything related to the beam position, just to the center.
Regarding phiy and phiz roles, well, that's my concern, I see that the CentringMath defines those, so we can pick from there, but you are not using it yet.

What would be needed in order to accept this PR? Make clear that this is not _movetobeam_ but _movetocenter?_

I will have a look at those methods.

mguijarr: Why would we need to move to center of sample video ? Normally we want to move to beam (which is the center in your case). Maybe there is something I don't see please tell me if we need to move to center ?

meguiraun: In order to see and visually analyse the sample, easier to find crystals where you do the centring, a quick way of moving a potential good position to an easier to manage image area.

On the other hand, implementing the _moveToBeam_ is different in:
* MiniDiff: basically what we have done in the PR but with beam-pos, the calc is the same, only with two motors.
```
            beam_xc = self.getBeamPosX()
            beam_yc = self.getBeamPosY()
            self.centringPhiz.moveRelative((y-beam_yc)/float(self.pixelsPerMmZ))
            self.centringPhiy.moveRelative((x-beam_xc)/float(self.pixelsPerMmY))
```
* GenericDiff: it calls ```get_centred_point_from_coord```, which gives you a centringPoint (?) and takes into account more motors (and raises _NotImplementedError_). But, Ivars already implemented it in EMBL diff so we could _steal_ the code which does all the calcs, but this will not be compatible with 2.1 I guess.

If you do not see the usefulness of a _move_to_center_ we can change to a _move_to_beam_ (it would have come sooner or later), and call to the code in the Minidiff (we would then implement those lines in our MD2/MD3 for now). Switching to the CentringMath hwobj approach would require a bit more of work. Would that be ok?

mguijarr: I think calculations in general should be delegated to underlying hardware objects, when a similar functionality exists, in order to keep a clean code in the Python server.

So I would do something like this (pseudo code, not tested):

    try:
        mxcube.diffractometer.moveToBeam()
    except AttributeError:
        # 2.2 and above
        mxcube.diffractometer.start_move_to_beam()

meguiraun:
No trace of go to center, the server only calls the hwobj method. Let's see now...

Issue #309, opened 2016-08-03
Cant test with latest master, wrong version of hardware objects ?


marcus-oscarsson:
Dear all,

I tried to test my latest development against the latest master but I cant seem to get it to work. I thought that it might be because of incompatible versions of hardware objects. So i started to look into it ! 

I have been using hardware objects 2.2 because it was as I understood it what you are using ?. There are two commits from you guys on the master branch more recent than 2.2. So I tried to use master instead but it did not change anything. I even tried to cherry-pick your commits from master into 2.2, with no success. So which version of hardware objects are you using ? Maybe having the right one will resolve the problems I'm experiencing. Or its perhaps some configuration that I'm unaware of ?

So here is what I get using latest master of mxcube3 and hardware objects 2.2.

I cant center and I get the following message:

`    Error: Server refused to add point`

A centered positions is sometimes added when I've tried to center and then refresh the page. Saving the position takes ages but works, if you have some patience :). Then adding a task is not possible because server refused to add to the queue, giving me:

```
      File "/home/oscarsso/projects/mxcube/mxcube3/mxcube3/routes/Queue.py", line 722, in add_method
    return add_data_collection(node_id)
  File "/home/oscarsso/projects/mxcube/mxcube3/mxcube3/routes/Queue.py", line 675, in add_data_collection
    col_node.acquisitions[0].acquisition_parameters.centred_position = qmo.CentredPosition(cpos['motorPositions'])
KeyError: 'motorPositions'
```
I started to debug it a bit but it seems like I'm either missing some commits or not having the right version of something so I prefer to ask you before i dig any deeper :). 

Thanks,
Marcus

meguiraun:
sorry marcus, consequences of the pep8 I guess...

see #307, motorPositions -> motor_positions

marcus-oscarsson:
Hey Mikel !

Not sure that's the problem, Ive tried to reset the branch to the commit before that but it does
not change anything.  Ive tried resting to d8f5522 ( Merge pull request #306 ... ) which works better
but its impossible to save the centered position, the option is not present in the context menu.

Does the latest master work for you ?, which version of hardware objects are you using ?

Cheers,
Marcus

marcus-oscarsson:
Infact it does work in d8f5522, it was just me who did not save a sample.

marcus-oscarsson:
Looking into more in detail, the latest commit that work for me is 7b523 (do not move if not ready, and send back 406).

[...]

Issue #305, opened 2016-07-29
Store organisation


marcus-oscarsson:
Good morning,

I thought it would be a good idea to separate "server-side data" and "presentation data" in the store. With "server-side data" I mean data that represents something on the server like motors, the queue, etc and with "presentation data" something that is related to the state of a component.

I think this will lead to a clearer organization of the data where its easier to share data between different parts of the application. This will remove the problem of potential discrepancies between data. I'm as you know currently working on unifying the queue used by the SampleGrid and the SampleQueue. However the same also applies to the list of samples and motors.

My suggestion would be to use the beamline store for everything related to beamline hardware and the queue store for the queue and list of samples.

This means that the things related to the state of the SampleQueue components needs to be moved from the queue store to a new store called for instance SampleQueueView. Further, the motors needs to be moved from sampleview to beamline.

What do you think about this idea ?

Regards,
Marcus

marcus-oscarsson:
I will close this since we discussed on skype

PR #310, opened 2016-08-03
Store reorganization

marcus-oscarsson:
Dear all,

So here is the re-organization of the store that I described in issue #305. I have not yet moved motors from sampleView to beamline.

The list of samples (sampleList) and manulMount  have been moved from SamplesGrid to queue. The data structure of the queue was partially changed by a previous merge of a incomplete PR. This broke the queue related features as Fredrik noticed, we are sorry for this glitch. The queue data structure have been fixed and should now be working fine.

I have tested as much as I can here, adding samples and tasks. Moving the Tasks around, editing and removing them. These changes are reflected both in the SampleView and in the SampleGrid so everything seems to work fine.

However, testing with the latest master is impossible because the centering and adding tasks does not work properly. See issue #309 for more details.

I would have liked to move the motors and clean up the queue reducer a bit before doing this PR. But I would like to sort out the issues in #309 before so that I can be certain that what I did works with the latest master. So Ill mark this as WIP.

Cheers,
Marcus


mguijarr:
Issue #309 is probably solved now thanks to latest PR by Mikel.

I am leaving on holidays so Marcus will accept PRs for the next 2 weeks for ESRF.

marcus-oscarsson:
Hi, 

So I'm finally finished with this I have tested against the latest master and it seems to work fine. I finally moved the motors to the beamline store as well. Doing this I noticed that pixelsPerMm is communicated through the motors update, but I kept pixelsPerMm in the sampleview since the rest of the camera related things are there. It somehow makes more sense to keep pixelsPerMm in sampleview, especially for debugging. Please let me know if you are of another opinion ? 

In anyways I noticed that the pixelsPerMm is sent with the Camera object at SET_INITIAL_STATUS in
/reducers/sampleview.js. Why is it that its sent with the camera in the beginning and then with the
motors ?.

Marcus

bolmsten:
Great work, I see that this is a big one! 

I will begin testing today but as a first thing I noticed that "UPDATE_TASK" does not work anymore it seems to have gotten lost in the refatoring between samplegrid and queue?

bolmsten:
"Why is it that its sent with the camera in the beginning and then with the
motors ?" 

This is to reduce the amount of actions when updating from the socketIO, it probably should be refactored someway.

marcus-oscarsson:
Okey, I see so its for optimization then. I noticed that motors are already present in action.data 
so I though there was another issue more related to the server. Just a question mostly out of curiosity, I guess its fine as it is.

marcus-oscarsson:
Good catch, it seems like UPDATE_TASK reducer was lost. Maybe it went on 
vacation with Matias :). Ill make a fix asap.

marcus-oscarsson:
Ive added the missing UPDATE_TASK reducer and made the necassary changes to the API. Updating a task should work fine now :)

Good evening :)

bolmsten:
Hi we have continued testing and see that when changing tasks and going back the button wrongly says "add task" instead of "change". 

We also had some problem running before but are having problems recreating that issue

bolmsten:
Minor linting errors also appears

marcus-oscarsson:
Thanks for the testing I know that its tedious, and its Friday too :)

Okey, Ill take a look at the problem with "add task" instead of "change". How did you get that ?, did you have some errors in the Console ?

Marcus

marcus-oscarsson:
Okey, so the latest commit should resolve the "add task" displayed instead of "change" problem.

bolmsten:
@marcus-oscarsson 

Good, I see that you have fixed it now. 

We will do one last test today to see that everything is working but by looking at the code it seems to me that "ADD_TASK_RESULT" should fail as it uses the old structure, do you agree? As I said we did not manage to test this the last time and will do so later today.

bolmsten:
Also regarding the new structure do you want to save task not on the samplelist but instead on the queue? 

If I understood your suggestion from before everything should be on the samplelist? This would make it easier to take the state from the server and give the client, presuming that the server outputs in the same form.

marcus-oscarsson:
Good morning !

Your right the ADD_TASK_RESULT might not be updated to the new structure of the queue. Ill take a look at it.

The new structure of the queue is a mix of what I had in the SampleGrid and what you were using in the queue and might honestly not be the final structure.  The main goal of this PR was the reorganization of the stores. I did some compromises to make the PR easier to handle and to at get it to work. For instance the queue is still accessed by queue ids which I would like to avoid in the future, I'm sure we will continue to exchange ideas to converge to something great :). 

The main idea with this representation of the queue is that its easy to read and thus to debug and hopefully to extend. We also hope that we can reduce the logic needed to manipulate the queue although this of course depends on how the queue is exchanged between the server and client. I know that it costs us time and effort now (and its definitely not the most fun thing to do) but I hope that it will save us time later.

The separation of samplelist and queue is done by purpose since we might have samples which are not necessarily in the queue. I see the queue as something that we want to collect on and the samplelist is just the list of potential samples to collect on. But there is perhaps a better structure ?

Cheers,
Marcus

bolmsten:
Great! 

I totally agree with your suggestion about taking it in steps and this would be the first one.  
I think the next step would be to refactor all node_ids to sampleID both in client and API.

marcus-oscarsson:
:)

That's a good idea, your right I think its enough to send sampleIDs between the client and server. I'm looking into the next refactoring and I have not decided exactly how to do it yet. I'm still working on the idea of sending the entire queue but its not finished yet, I have some code but its far from finished. But the sampleID refactoring is maybe a good start.

M

marcus-oscarsson:
So the last commit should fix ADD_TASK_RESULT, it seems to work fine here as well. I first though that was handled by UPDATE_TASK so I completely missed it, sorry about that.

PR #312, opened 2016-08-03
[WIP ] ImageIO (MultiClient picture)

bolmsten:
In this PR I have changed the way the client fetches the picture from the server.

Before we used "multipart/x-mixed-replace" served by flask as described by this tutorial:

http://blog.miguelgrinberg.com/post/video-streaming-with-flask

This does not work for multiple clients and as stated in the link above the most efficient ways of sending a picture is by websockets. 

Additionally the technology we are using is also slowly losing support:

https://bugs.chromium.org/p/chromium/issues/detail?id=249132

The problem I have encountered is that FireFox seems to handle the new approach quite badly and this is the reason I keep this as WIP. 

(Chrome works great no matter how many clients attached)

### DIFF
--- a/mxcube3/routes/SampleCentring.py
+++ b/mxcube3/routes/SampleCentring.py
@@ -1,8 +1,8 @@
 from flask import Response, jsonify, request
-
+import base64
 from mxcube3 import app as mxcube
 from mxcube3.routes import Utils
-
+from mxcube3 import socketio
 import copy
 import logging
 import gevent.event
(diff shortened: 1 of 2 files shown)


mguijarr:
At the moment I am against using websockets for the video transfer, because I prefer the
simplicity offered by the `x-mixed-replace` approach as it also takes care of images display optimization directly in the browser (no need for `requestAnimationFrame`...). The support will never drop, I think - it's used extensively by millions of cheap CCTV cameras... What may be removed from Chrome is the use of `x-mixed-replace` for something else than images and it is not our use case.
We can quite easily make it work for multiple clients if needed, so I really see no point changing anything now unless I see a clear benefit.

PR #313, opened 2016-08-04
Making image work with multiple clients

bolmsten:
This is now done in the same way as before but with multiple client support


mguijarr:
Well done :)

PR #316, opened 2016-08-04
Fixed hover on objects in fabric js

bolmsten:


marcus-oscarsson:
Thanks Fredrik !.

Looks much better :)

Marcus

Issue #304, opened 2016-07-28
Queue representation / handling


marcus-oscarsson:
Hi all,

Matias told me that he spoke with you about our ideas for queue handling and representation during the MxCuBE meeting. So we thought we would present our rough ideas.

We are thinking about a functional approach (like redux) for the queue, where a new queue representation is generated for each action performed on the queue. The queue representation is the very same one used by the queue store (JSON). This representation is then, surprise surprise :), exchanged between the client and the server. The server creates a new queue from the queue representation when a "SET SERVER QUEUE" request is sent . (Keep in mind that a queue can not be updated while its running.) The server in turn can send a "SET CLIENT QUEUE" request over a websocket, the only thing the client needs to do is to update the queue store. 

The advantage with this approach is that the API to handle the queue becomes very simple you either
do SET_SERVER_QUEUE or SET_CLIENT_QUEUE, and for the client in some cases GET_QUEUE. There will be no need for keeping track of node ids on the client, which can be error prone and difficult to maintain/debug. All in all it will be much easier to debug and the server side API is dramatically reduced.

The currently running task might have, if needed, and additional API to update for instance number of images or other progress. We however don't think this will be needed but we don't see it as a problem
to handle this separately if needed.

I have already written the necessary code to perform the serialization and queue construction. The format  I used are almost identical with the one that is used today.

Representation of queue:
```
{
    "1:01": [
        {
            "Type": "DataCollection", 
            "label": "Data Collection", 
            "parameters": {
                "detector_mode": "", 
                "energy": 0, 
                "exp_time": 0.0, 
                "first_image": 0, 
                "inverse_beam": false, 
                "kappa": 0.0, 
                "kappa_phi": 0.0, 
                "num_images": 0, 
                "num_passes": 0, 
                "osc_range": 0.0, 
                "osc_start": 0.0, 
                "overlap": 0.0, 
                "path": "/tmp/visitor/local-user/test/20160728/RAW_DATA/dummy_path", 
                "prefix": "", 
                "resolution": 0.0, 
                "run_number": 0, 
                "shutterless": false, 
                "snapshot": null, 
                "transmission": 0.0
            }, 
            "sampleID": "1:01", 
            "state": 0
        }
    ], 
    "1:02": [ ....
```
The construction of the queue from this representation is done on the server and all the client needs to do is to display it by updating the store.

[...]

PR #318, opened 2016-08-10
[WIP]: Queue API

marcus-oscarsson:
Hello !

Here is the first take of the Queue API refactoring that we discussed in issue #304. I wanted to share this preliminary work with you guys to get your ideas on it. The general idea with this first work is to decouple server internals like nodeId and to simplify the logic we use on the client. I then hope to be able to develop a simplified API for exchanging the queue. 

I've removed the mapping between server side nodeID's to samples and tasks (so far on the client side). The two lookup tables "lookup" and "queueID_lookup" have consequently been removed. The components have been updated to reflect this change. I'm currently using the sampleIDs and the position of a task in the queue to identify samples and tasks. 

The "display data" that were stored in queue state collapsedSamples and checked have been moved to sample and task data objects. This might not be the best solution but its easier to access this data directly rather than going through an extra state variable. I was thinking of creating a new state variable called displayData, with all the display data like label, collapsed and checked. I might do that in the end if we there is good reasons for it, for the moment I cant really see any. Let me know what you think !?.

Ive also removed the server calls for ADD, REMOVE and CHANGE tasks, which later will be replaced by the SET_QUEUE (not done yet). The mount sample route have been refactored to operate on sampleIDs instead of nodeIDs. 

Finally, the UI should work as before but no changes to the queue are sent to the server. So its of course not possible to run the queue yet.  The next thing to do is to implement the queue exchange functionality basically SET_QUEUE.

If anyone have some time over, take a look and let me know what you think. All input is welcome :).

Cheers,
Marcus

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -7,7 +7,7 @@ from optparse import OptionParser
 import os
 import sys
 import logging
-import jsonpickle
+import cPickle as pickle
 import gevent
 import traceback
 
@@ -90,7 +90,7 @@ if not app.debug or os.environ.get("WERKZEUG_RUN_MAIN") == "true":
 	    # centring_motors_list is the list of roles corresponding to diffractometer motors
 	    app.diffractometer.centring_motors_list = app.diffractometer.getPositions().keys()
         app.db_connection = app.beamline.getObjectByRole("lims_client")
-        app.empty_queue = jsonpickle.encode(hwr.getHardwareObject(cmdline_options.queue_model))
+        app.empty_queue = pickle.dumps(hwr.getHardwareObject(cmdline_options.queue_model))
         app.sample_changer = app.beamline.getObjectByRole("sample_changer")
         try:
             routes.SampleCentring.init_signals()
--- a/mxcube3/routes/Queue.py
+++ b/mxcube3/routes/Queue.py
@@ -135,7 +135,7 @@ def queue_unpause():
     return Response(status=200)
 
 
-@mxcube.route("/mxcube/api/v0.1/queue/clear", methods=['PUT'])
+@mxcube.route("/mxcube/api/v0.1/queue/clear", methods=['PUT', 'GET'])
 def queue_clear():
     """
     Clear the queue.
--- a/mxcube3/routes/Queue.py
+++ b/mxcube3/routes/Queue.py
@@ -14,9 +14,7 @@ import Utils
 from mxcube3 import app as mxcube
 from mxcube3 import socketio
 
-#for mocking the view of the queue, easier than adding sth like if not view:
-from HardwareRepository.BaseHardwareObjects import Null as Mock
-
+from Utils import PickableMock as Mock
 
 qm = QueueManager.QueueManager('Mxcube3')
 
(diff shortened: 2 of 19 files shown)


bolmsten:
I took a quick look and can give some initial feedback. 

Saving the state of "collapse..." on the queue and samplelist would make it harder to replace these with new information coming from the server without having to merge.

I like the idea of decoupling the client from server but I would like to take it in steps. So instead of removing the server calls for ADD, REMOVE and CHANGE tasks I would modify them to work with the new architecture and have something working immeaditely. 

The PR is already quite big and adding the new complexity for SET_QUEUE synk will make it even bigger.

marcus-oscarsson:
Hey Fredrik,

I have the same thoughts about the states of collapse, its maybe a good idea to separate them.

I know, I'm sorry for this big PR its not as clean as I would have like it to be. I would also like to have something that is smoother to integrate. I already have some initial work done for the SET_QUEUE so I will commit it here and then Ill figure out together with you guys how to split it into more manageable pieces :). Ill start to take a look at how to do that tomorrow.

marcus-oscarsson:
Hi again,

So the last two commits illustrate the main idea of the SET_QUEUE API. The JSON queue is sent from the client to the server via the ../queue/ route. The server constructs a new queue from the JSON data and responds with a serialized version of the result. Re-serializing the result is mostly for testing, but its perhaps not such a bad thing to do actually.

It should be possible to construct a queue and run the entire thing, running individual nodes will not work for the moment  :).

As we discussed earlier is perhaps a better idea to split this PR into two or more pieces. I wanted to share the latest development with you guys so that its easier to discuss.

Good evening,
Marcus

meguiraun:
hej Marcus,

The cleanup of the sampleIDs looks as a great idea... but we would really appreciate if you could split in two PRs. As we discussed earlier, we have some concerns about the {set, get}_queue new api, so it would be much easier to test if this current PR is split in several PRs. Thanks in advance.

On the other hand, now that the swedish summer is over, the Maxiv mxcube team will start its holidays and try to enjoy an extra summer somewhere else. So, unfortunately you will not have too many news from us. Back in the beginning of September.

marcus-oscarsson:
Hej,

Ahh, vacation that's always nice :). I wish you all a wonderful time with friends and family. Enjoy and see you back in September !

Yes, again sorry for this fairly complicated PR, it was easier for me to do everything in one shot but its true that its better to split it into several PRs. I was a bit reluctant do do that in the beginning since it will require much more work, some of which will never be used later. I would propose something like three PRs, one with the queueID/sampleID work, one with simplified serialization on the server, and finally one with the new queue API. What do you think about that ?

I'm curious to know what your concerns regarding the queue API are. There might be things that we didn't consider so I would be very great full to continue the discussion on this.

Marcus

bolmsten:
Hi,

We still feel a bit uncertain regarding the whole "out of sync with server architecture"  and that the server is not always the "single point of truth". One question that has arisen while working on "mxcube multiple clients" is that the server in that case still needs to relay all data changes to the other clients which means that there is no reason for the "out of sync" part.

Also always sending the whole queue for updating provides some benefits such as easier integration and more security regarding sync but also requires a lot more data being sent and computation effort (We will see if this is a problem).

Of course it is good to clean up and simplify the API but at the moment we have a proven and working API so our focus at the moment is not refactoring of that part. 

We clearly see the benefits of cleaning up queueID/sampleID and serialization of the server state that matches the client structure to easily initialize a client, these needs to be done no matter what and are independent of the last step which is more "experimental".

marcus-oscarsson:
Hi,

Thanks for your input Fredrik you raise some good and important issues :). Ill try to answer your questions. The idea might seem controversial in the beginning but I don't think it is actually the case.

Ill try to give you my view of the idea and maybe that can give you some more context. We would like to decouple the server and the client as much as possible. Not because this is a goal in it self but because we think this leads to easier maintenance in the future. We would also like to get rid of the temporal aspect of building the queue, that is built in "real time" as the user performs the corresponding actions. We think this will make it easier to handle unexpected situations were for instance the server crashes. It makes it much easier to guarantee that whats on a client is also what we are running. 

Sending and computing, especially computing this data should not be a problem. Consider also that this is done before we run the queue, we can accept to have a short delay before the data is sent and processed. The user should perhaps even be presented with a summary dialog with whats going to happen before the collect. The queue is probably already sent and processed before that's done.

The single point of truth is a bit difficult to apply here, I'm not even sure its possible to do so. In our case I would still like to say that the server is the "single point of truth" but its only relevant when the queue is executing. The interaction between client and server of course needs to be well defined. The client can for instance not send a queue when the server already is running one (this would make little sense). The server on the other hand can tell the client whats happening with the queue. In this case however we mostly (except for workflows) need to send which task currently running and the progress of that task.

I don't think we should couple the implementation of the queue exchange with the multi client feature to much. Its of course good if the multi client feature comes for free. However there are things on the client which is never sent to the server, we might also need to synchronize and share that. Depending on what kind of experience we want to give the listening clients. Only listening on server events will not
solve the entire problem. So a completely different solution is perhaps needed for the multi client feature. Sharing the "redux state" or using a dedicated solution similar to Surfly could perhaps be used. Perhaps you have some good ideas on how to do this already ?

All solutions of course have their advantages and disadvantages, we think this will lead to a software which is easier to maintain and extend because of the reduced logic. I hope that my explanation somehow can make you more confident in this kind of solution.

I appreciate your input Mikel and Fredrik it helps me think about these things a bit more in detail :)

Cheers,
Marcus

marcus-oscarsson:
Replaced by #325, so closing this one

Issue #317, opened 2016-08-09
React Issues in SampleGrid


bolmsten:
@marcus-oscarsson 

When I go to the samplegrid I get: 

main.js:11321 Warning: Failed propType: Required prop `title` was not specified in `DropdownButton`. Check the render method of `SampleGridContainer`.
Warning: Failed propType: Required prop `toggleActiveItem` was not specified in `SampleGrid`. Check the render method of `SampleGridContainer`.
Warning: Unsupported style property margin-bottom. Did you mean marginBottom?
......

I can fix it but I know that you are working with this component so maybe it is easier for you?

marcus-oscarsson:
Hey Fredrik,

Yes I noticed this to, I though Id make a separate PR with those fixes later since its not really blocking.
Ill take care of that after I have finished the first  refactoring of the queue, I'm currently taking care of the sample and queue IDs :).

Thanks,
Marcus

marcus-oscarsson:
Fixed by #319

PR #319, opened 2016-08-10
Fixing react warnings

marcus-oscarsson:
Fixing the react warnings Fredrik mentioned in #317


marcus-oscarsson:
I'm closing issue #317 since this should solve the problem.

PR #320, opened 2016-08-11
New logger component for sampleview

bolmsten:
This component is more in sync with the rest of the GUI


marcus-oscarsson:
Nice :)

PR #326, opened 2016-08-23
One PR containing #322, #323, #324, #325

marcus-oscarsson:
Dear all,

Since we have done alot of changes lately, we thought that it would perhaps be easier to follow the development and merge if there was one PR containing all the changes. They previous PRs
#322, #323, #324, #325 all basically build on each other and it would in the end be rather pointless
to not consider them as one. The exact details can be found in the previous PRs, (Ill close them for now).

So here you go, we hope it makes the merge job a bit easier :).

Cheers,
Marcus


bolmsten:
Hi @marcus-oscarsson ,

I will start testing this today but I expect it will take me some time. I will get back to you with feedback as fast as possible.

mxcube3/ui/reducers/queue.js
```diff
@@ -145,17 +145,18 @@ export default (state = initialState, action) => {
...
       return Object.assign({}, state,
         {
-          todo: { ...state.todo, nodes: state.todo.nodes.concat(action.queueID) },
-          queue: { ...state.queue, [action.queueID]: [] },
-          lookup: { ...state.lookup, [action.queueID]: action.sampleID },
-          lookup_queueID: { ...state.lookup_queueID, [action.sampleID]: action.queueID },
-          collapsedSample: { ...state.collapsedSample, [action.queueID]: true },
-          manualMount: { ...state.manualMount, id: state.manualMount.id + 1 },
-          sampleList
+          displayData: { ...state.displayData, [action.sampleID]: { collpased: false,
```
bolmsten: Change to collapsed. There seems to be some problems creating a new sample when one is already in place, I get a error message containing collapsed but only changing this did not seems to solve it.

mxcube3/ui/actions/queue.js
```diff
@@ -238,73 +257,80 @@ export function sendStopQueue() {
...
-    }).then((response) => {
-      if (response.status >= 400) {
-        throw new Error('Server refused to mount sample');
-      } else {
-        dispatch(mountSample(queueID));
-      }
-    });
-  };
+export function setQueueAction(queue) {
+  return { type: 'SET_QUEUE', queue };
```
bolmsten: Is this ever called? Should  the client not get the state from server on start? I guess this would be one of the big benefits of this is to exchange queue between server and client?

mxcube3/ui/components/SampleQueue/HistoryTree.js
```diff
@@ -20,14 +20,17 @@ export default class HistoryTree extends React.Component {
...
                 <div className={bodyClass}>
-                    {this.props.list.map((id, i) => {
-                      let sampleData = this.props.sampleInformation[this.props.lookup[id]];
+                    {this.props.list.map((sampleId, i) => {
+                      let sampleData = this.props.sampleInformation[sampleId];
                       return (
                         <HistorySampleItem
                           data={sampleData}
-                          show={this.props.collapsedSamples[id]}
+                          show={this.props.displayData[sampleData.sampleID].collapsed}
```
bolmsten: I think this should be sampleData.id to work

mxcube3/ui/actions/queue.js
marcus-oscarsson: The action SET_QUEUE Its not used at the moment but as you note yourself there are situations when it perhaps should be used. Initialization could be one such case, that's why I added the action :). However I wanted to wait with that logic until we have decided how to, if at all exchange the state in local storage.

mxcube3/ui/reducers/queue.js
marcus-oscarsson: Its not a problem that I get here, what is the message that you get ?

mxcube3/ui/components/SampleQueue/HistoryTree.js
marcus-oscarsson: You are right it should be sampleData.id, infact I'm using just sampleId here. That's a bit strange I might have missed that in a commit somehow.

marcus-oscarsson:
Hey Fredrik,

Welcome back, I hope you had a nice vacation !?

Let me know if there is something that I can do to make the merge job easier, we can be available on Skype if there is something you would like to discuss in more detail. I'm currently working on integrating the samplegird and the sampleview so I will be a bit busy with that, but I'm otherwise available to continue on this so we can get it merged.

Marcus

mxcube3/ui/components/SampleQueue/HistorySampleItem.js
```diff
@@ -6,17 +6,17 @@ import './app.less';
 export default class HistorySampleItem extends React.Component {
   constructor(props) {
     super(props);
-    this.collapseSample = props.collapseSample.bind(this, props.id);
+    this.collapseSample = props.collapseSample.bind(this, props.data.clientID);
```
bolmsten: This change seems to have broken the functionality of collapsing samples. Reverting to the old code fixes it

bolmsten:
It was great, thanks for asking.

I will continue my testing and get back to you regarding the Skype meeting.

mxcube3/ui/components/SampleQueue/HistorySampleItem.js
marcus-oscarsson: Wow wow, the data.clientID should not be there Ill better take a look at that.

Thanks

marcus-oscarsson: I'm really sorry about that it seems like I missed a commit. Ill see if I can find it and add it :)

marcus-oscarsson:
Ive added the small fixes missing to HistoryTree and HistorySampleItem in commit e05af6a

mxcube3/routes/Queue.py
```diff
@@ -242,155 +234,222 @@ def queue_load_state():
...
-        :statuscode: 409: sample could not be added, possibly because
-            it already exist in the queue
-        :example request:   * POST http://host:port/mxcube/api/v0.1/queue
-                            * Content-Type: application/json
-                            * {"SampleId": "1:07"}
-    '''
-    params = request.data
-    params = json.loads(params)
-    sample_id = params['SampleId']
+def queue_set():
```
bolmsten: This is a O(n) operation (instead of O(1) as it was before) and it seems to be following the linear formula (y=kx) where y is milliseconds and x is number of tasks such as datacollection. So in other words with every task that is added there will be an additional X milliseconds of wait. 

This could be quite bad, I have done some testing with mxcube in build mode and it becomes quite unusable when the queue grows. I have talked to Jie and it is not at all impossible to have several hundred tasks in a queue after a long shift.

marcus-oscarsson: I see, its strange that you experience such big delays when setting the queue. You are right sending and setting the entire queue is intrinsically a O(n) operation. 

We can perhaps not assume that adding a task is in the order of milliseconds, it should reasonably be less. Adding a task is basically appending an object to a list. I looked into the time it takes to add one data collection task since it seems you experience some unreasonably long delays. I timed it to 0.0007s, so adding 1000 tasks should take 0.7s, excluding the HTTP request.

Also keep in mind that inserting instead of appending n data collections in the queue is also a O(n) operation, plus the time it takes to find where to insert the tiem so that would result in 2*O(n) operation

We have not worked with large number of data collections in MxCuBE3 yet so we have not really noticed the delay, but the delay is there regardless of method used. At some point we have to add large number of data collections in one shot regardless of how its done. Speaking about delays, we don't see it as a big problem if the user have to wait 1-2 seconds before the queue starts. However it could be perturbing for the user if they have to wait while adding tasks.

It is possible that we have some performance issues when adding tasks, its even likely that we can optimize it (0.0007s seems a bit long). We have to look in to that but the fact that sending queue is an O(n) operation should not be seen as a problem.

That being said we of course need to find were this delay that you experience come from, but we need to do that regardless of method used.

mguijarr:
Hello,

I would like to know how far we are from merging this PR ?

Albeit the fact there might be some small glitches, my suggestion is to move on and have the new code
in the project as soon as possible since it brings a lot of improvements. Above all it makes things
simpler regarding queue management, as we discussed during our last (small) Skype meeting, and this is really added value.

We feel a bit uncomfortable because we would like to give a boost to the development by coding and testing the whole chain from sample changer + samples grid to queue ; we see as an important step that previously discussed topics are taken for granted in our collaboration and thus in code,
also to be ready for the next PRs to come.

@bolmsten :  regarding your last comment, thanks for pointing this. Finally we identified the function that takes 2-3 ms to execute, the culprit is `check_for_path_collisions` in file `HardwareObjects/QueueModel.py`. It can be improved, for sure. Unfortunately we could not see your video, if it's still relevant could you please make it available somewhere on the web so we can have a look ? Maybe you found another bug we would be happy to fix. That said, even with this time consuming function, I do not agree the overall performance is degraded from a user point of view, since in our new design the queue building happens only at certain times where user is expecting to wait... In our proposal the queue is defined in the client by the user. This is our 'single point of truth'. Then, it is sent to the server when needed and the new queue replaces the old one. We can assume this operation cannot fail, since it just consists in transforming a json representation to Python objects (no beamline hardware is involved) - if something gets wrong (whatever it could be) user will realize when queue execution starts. Which is fine. I hope this allays your fears.

bolmsten:
Hi @mguijarr ,

You can find the video here:

https://drive.google.com/file/d/0B_r8ullVomn_SDg2SVBOcVFLWlk/view?usp=sharing

We will email you tomorrow a document with concerns so that we can discuss.

marcus-oscarsson:
Hey,

I noticed that I forgot to commit some changes related to removeTask, the call in TaskItem used an the old API. Should work better now though. Sorry about that.

Marcus

marcus-oscarsson:
Dear all,

I will update this PR to include the API calls for ADD, CHANGE, DELETE and MOVE item. The API wont be exactly as before since I would like to take advantage of what Ive done in this PR. But the function will of course be the same. I will also take the opportunity to make some improvements for instance to make the API work better when adding multiple items.

Cheers,
Marcus

marcus-oscarsson:
Hi All,

So Ive added the API calls for ADD, UPDATE, DELETE, and SWAP queue items. I have tried to take advantage of what I previously did in this PR so the calls are not exactly as they were before but the function should be the same. I also added some simple error handling for the "optimistic" rendering. I hope we can minimize the usage of these error handlers and instead prevent the user from doing the wrong things.

The logger was also changed slightly because the log messages were displayed twice
The queue utility functions have been moved to a file called qutils.py

We hope that the latest changes will make it possible to accept this PR, we are still not convinced about this API but we prefer to continue that discussion while we are advancing with this PR and the project in general.

Marcus

bolmsten:
Great! 

Me and Mikel are starting testing this now and will get back to you soon.

marcus-oscarsson:
Thanks,

I noticed that I forgot to remove some of the methods that were moved from Utils.py to qutils.py so Ill push a commit that removes those double definitions.

Marcus

bolmsten:
@marcus-oscarsson  

As you see we have accepted this. Both me and Mikel think the new changes are great and will make future development a lot easier!

meguiraun:
thanks a lot for all of these huge improvements!!

PR #327, opened 2016-09-01
Samplegrid improvement and beginning of queue int

marcus-oscarsson:
Dear all,

This PR was a bit painful to finish since it was continuation of the famous queue API PR. I had to rebase it a couple of times and finally merge with the queue redesign that Fredrik made. So my apologies in advance if this PR is a bit difficult to follow, I think it should be all right though. Ill explain below what I did just incase:

There is a early version of the pipeline mode with a waiting screen for when the sample is loading, the
dialog that asks the user to save a position is still missing but the underlying logic works. That is the
"pipeline" continues when you save a position. That's basically it for the pipeline integration for the moment, I wanted to wait with the rest because I realized that it might get tricky to merge otherwise.

The MOUNT and UNMOUNT sample have been replace with a SET and CLEAR current sample (on the UI only) We still send MOUNT_SAMPLE to the server but we set the CURRENT_SAMPLE in the UI. Its maybe not very important but it makes more sense when we are working without SC, since we cant really say that we have mounted a sample then. CURRENT_SAMPLE is somehow always valid in both cases.

I removed the notion of PHASE from the UI. The phase should only be used when working with diffractometers that have the concept. It should as far as possible be transparent to the user.

I removed the notion of PICKED sample in the SampleGrid reducer a sample that is in the queue is
automatically considered picked.

The sample grid interaction have been improved slightly.

I would like to continue the samplegrid - queue integration but I prefer to wait a bit and discuss the next step with you guys.

Cheers,
Marcus

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -100,6 +100,7 @@ if not app.debug or os.environ.get("WERKZEUG_RUN_MAIN") == "true":
 
         try:
             SampleCentring.init_signals()
+            SampleChanger.init_signals()
             Beamline.init_signals()
         except Exception:
             sys.excepthook(*sys.exc_info())
--- a/mxcube3/routes/Queue.py
+++ b/mxcube3/routes/Queue.py
@@ -1,3 +1,4 @@
+import time
 import json
 import logging
 import signals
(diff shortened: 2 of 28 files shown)


mguijarr:
Thanks Marcus !

meguiraun:
Thanks for the work! We are going to review and test this during today.

bolmsten:
This looks good!

PR #328, opened 2016-09-13
Fixes

bolmsten:
This will fix the reorder problem and the add characterisation problem. I have used the unique ID for a sample as : type + run number


marcus-oscarsson:
Thanks guys :)

PR #330, opened 2016-09-14
Minor GUI changes to the sampleview

bolmsten:
Removed history from sampleview and added todo. I also added some small summary to each task and a progressbar at the top. I am still struggling with the GUI of the queue in the sampleview, mostly because it needs to handle both manual/semi automatic/automatic mode and I need to find a design that incorporates all these

### DIFF
--- a/mxcube3/ui/components/SampleQueue/CurrentTree.js
+++ b/mxcube3/ui/components/SampleQueue/CurrentTree.js
@@ -85,7 +85,7 @@ export default class CurrentTree extends React.Component {
     return (
       <div className="m-tree">
           <div className="list-head">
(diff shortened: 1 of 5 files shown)


mguijarr:
Thanks for this - we are going to have a look how it looks like now :)

For the design, feel free to share some sketches if you want ; we can open an issue for that and continue discussion other there, so we can give you some inputs/help/feedback if needed.

We are still thinking it's unnecessary complex to have the todo list on the sample view page since the information is already in the sample grid. I understand you don't want to have users changing page from
sample view to sample grid ; then maybe a button that would open a modal window with the to-do list
could do the trick ?

PR #332, opened 2016-09-14
Redesign of queue in sampleview

bolmsten:
I looked into Marcus WIP for SampleGrid and took some design inspiration on how we can merge samplegrid and sampleview.

The idea is to keep the queue in the sampleview looking the same no matter what mode you are in (semi automatic/ manual). 

Please take a look and give feedback.

### DIFF
--- a/mxcube3/ui/actions/queue.js
+++ b/mxcube3/ui/actions/queue.js
@@ -370,7 +370,7 @@ export function addSample(sampleData) {
 
     // Its perhaps possible to not even sendMountSample at this point,
     // does it even make sense ?
-    dispatch(sendMountSample(sampleData.sampleID));
+    //dispatch(sendMountSample(sampleData.sampleID));
   };
 }
 
--- a/mxcube3/ui/actions/queue.js
+++ b/mxcube3/ui/actions/queue.js
@@ -367,10 +367,6 @@ export function addSample(sampleData) {
 
     sendAddQueueItem([data]);
     dispatch(addSampleAction(data));
-
-    // Its perhaps possible to not even sendMountSample at this point,
-    // does it even make sense ?
-    //dispatch(sendMountSample(sampleData.sampleID));
   };
 }
 
--- a/mxcube3/ui/components/SampleQueue/CurrentTree.js
+++ b/mxcube3/ui/components/SampleQueue/CurrentTree.js
@@ -83,7 +83,7 @@ export default class CurrentTree extends React.Component {
       <div className="m-tree">
           <div className="list-head">
               {queueOptions.map((option) => this.renderOptions(option))}
-              <p className="queue-root" onClick={this.collapse}>Current: {sampleData.sampleName}</p>
+              <p className="queue-root" onClick={this.collapse}>Current Sample: {sampleData.sampleName}</p>
               <hr className="queue-divider" />
           </div>
           <div className={bodyClass}>
--- a/mxcube3/ui/components/SampleQueue/CurrentTree.js
+++ b/mxcube3/ui/components/SampleQueue/CurrentTree.js
@@ -17,20 +17,19 @@ export default class CurrentTree extends React.Component {
     this.state = {
       options: {
         QueueStarted: [
-        { text: 'Stop', color: 'danger', action: this.props.stop, key: 1 },
-        { text: 'Pause', color: 'warning', action: this.props.pause, key: 2 },
+        { text: 'Stop', class: 'btn-danger', action: this.props.stop, key: 1 },
+        { text: 'Pause', class: 'btn-warning pull-right', action: this.props.pause, key: 2 },
         ],
         QueueStopped: [
-        { text: 'New Sample', color: 'primary', action: this.showForm, key: 1 },
-        { text: 'Unmount', color: 'danger', action: this.unmount, key: 2 },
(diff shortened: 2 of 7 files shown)


marcus-oscarsson:
Hey !

Thanks for your PR Fredrik, the guys here are quite busy so I will try to answer this with a small reservation that what I say might not be 100% accurate. However I think they agree, the one that knows this best is Daniele. I prefer to say something as soon as possible because I imagine that you are working on this right now. I think that we share the same ideas here in anyways so if something changes it will be in the details.

We think that the only thing we need to have on the right is the queue for the current sample. Its after all this sample we are working on. Its nice to have the progress of the entire queue but we think that this simply can be displayed as Sample 1 of N somewhere in a button or as text. 

The tasks should auto expand when they are collected and collapse when they are done. A task is colored coded according to the status of the collect SUCCESS, FAIL, or IN_PROGRESS. Each task should have some progress information we added a progress bar per task to display for instance images collected. This information is really just to show the user that something is happening.

The name of the next sample or maybe even samples and perhaps a summary of whats going to happen with them should appear when hovering over the Next Sample button. Its a good idea to at least be able to know which the next sample is and whats going to happen to it. The sample grid can be used to get a better overview of whats going to happen, or to make decisions of what to do with the sample since this is where alot of the results are going to be displayed.

We should probably avoid to have New Sample on the sample view, or it should only appear when not using the sample changer. 

I don't know if it helps but here's a picture of the very first layout draft we made when you visited last time. You can see some of the ideas in the picture. 
queue

We can make a new more accurate layout if you like, or we can have a short Skype meeting about it. we are going to meet Daniele tomorrow so we will take a look at this together then. But I think these are roughly our ideas.

Marcus

bolmsten:
Hi Marcus!

You guys had a lot of great ideas that I will work with. Regarding the design we will discuss here as well and then we can iterate and merge our ideas.

At the moment this is what I have implemented:

screen shot 2016-09-14 at 16 32 33


This is implemented in the code now. It is based on the latest design discussed with you and is not finished yet.

What I needed to do was some hacks in the samplegrid to start adding things to the queue. I will change this PR to WIP until I we have done a better solution.

bolmsten:
I have a first version of the new queue with a very simple integration with the samplegrid. I think this is a good first step and if you agree we could accept this PR. 

There are still some GUI problems and enchantments that need to be fixed:

1. Progress per task? Should this exist and if so how is it calculated

2. Todo list, as per our latest discussion I have a todo list in the sampleview but no history. The todolist concept needs to be discussed and I want to have this very basic version as a first version to show some of the design decisions that still needs to be done.

3.  I have a overall queue progress that I think works fine for the moment but may need to be refined.

4. The integration with samplegrid is very basic but works, I assume Marcus will have a more complete version in the future

mguijarr:
Just a quick comment:

1. Progress per task: really important to have - the easiest is to have it based on frames to be collected for the task (number of images), then there is a `collectImageTaken` signal emitted from `AbstractMultiCollect.py` we can rely on to indicate progress

ddesanctis:
Hi all,

to help out this discussion and the design, some thoughts on the Sample queue are pasted below (nothing revolutionary or very different from what we have been always discussing)

Sample queue thoughts:

- linking positions to data collection also in sample queue is great (ex. P1 Data Collection)
- We need a logic sorting, so a Data Collection Plan that is coming
  from a Characterisation should be immediately after it, so in case
  there are more that one characterisation user can tell what is
  coming from what
- IMO we do not need Change - Delete - Results
  - this can be dealt as in MXCuBE2: 
  - Change: you can click on a data collection and have access to the parameters to change them
    (although we have to see whether this might be a duplication of
    what can be done on sample view, shall we have two modes?)
  - Results: Same as before user click or double click on the Data
    collection tile and have the results if the Data collection
    already happened
  - Delete: Maybe a cross button in one corner is sufficient? I agree
    a bit dangerous, but we can optionally add a warning
- Parameters summary: Another option would be a Rollover? We should
  not over use these effects, but for something that important may be
  a good way to show information without filling the screen
- Progress bar: Total Progress is great IMO, but I can imagine users
  (or even staff) wanting to know at what stage each data collection
  is. Currently I think we receive the image number from the detector
  device server, it could be a way to do it (If you have ever used
  Windows you do not trust progress bar! :-) )
- Other essential functionalities (I think these are in the drawings
  we made and Marcus posted already):
  - Stop: We always need a fast way of aborting a Data collection
    and/or the whole Queue (= Panic Button)
  - Run (of course)
  - Pause
  - Next sample (when finished with current sample and user want to
    move to next one in the selected list (from sample grid)
  - Name of "Current sample" should be the ISPyB name of the sample
  - On top of what Marcus sent, last week or so we also made a rough
    sketch of what it might look like
office lens 20160826-111018

- Some additional consideration on the Sample View:
  - We also have to include beam optimization buttons (realign, center
    beam) Maybe a logic place would be in the beamline setup? Though
    this can make the risk of moving the Queue further down (or be
    very short on small screens?)

- All I could think of, of course these things are better discussed
  around a table and making drawings, rather than writing, so if there
  is anything not clear let me know, if I find the time I'll make some
  better sketches.

mguijarr:
In order to make progress I accept this PR, to get the new code merged no matter what is decided afterwards. Let's continue the discussion...

PR #333, opened 2016-09-15
Foundations for Remote Access

mguijarr:
I took the time to implement some of the ideas from my last email, to demonstrate the technical solution better than with words I think :
- the UI state changes are monitored with redux-persist, but instead of storing state in localStorage we
store *on the server*. This puts back the server as the single source of truth while taking advantage of
redux to have an automatic way of saving the UI state

This enables the "screen duplication" feature, as you can see.
It was very easy to do, I am quite confident we can use this work as the foundation for the remote access feature.

For the moment, the first logged in user is master ; others are slaves. There is a new icon in the navigation bar, which is white if user is in control (master), otherwise it's grey. Both users need to be logged in with the same account. 

Current limitations:
- there is no way to change role (yet)
- it is possible for the slave user to use the application, but the final implementation will prevent this of course ! I just had no time to continue with this work for now.

If you have questions we can discuss here, I can explain more details of the implementation if needed.

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -1,6 +1,5 @@
 from __future__ import absolute_import
-from flask import Flask
-from flask import request
+from flask import Flask, request
 from flask.ext.socketio import SocketIO
 from flask.ext.session import Session
 from optparse import OptionParser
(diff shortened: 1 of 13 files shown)


bolmsten:
I will take a look at this now and get back to you.

PR #337, opened 2016-09-27
LIMS integration

marcus-oscarsson:
Hey !

Smaller improvements of the LIMS integration and started to use EXI. I will have a discussion with the EXI developer next week on howto reuse the components they have made for displaying results. I prefer get this merged in the meantime. 

There is so far display of collection progress and sample information. Display of the actual data collection and processing results will come next, we just need to find the best way to share this with EXI in order to share as much as possible.

There is currently a seperate LIMS client for EXI called ISPyBRestClient created by Ivars which we are using. Its maybe a bit strange to have two LIMS clients and I would personally prefer to have one, the idea is that the REST client eventually will replace the old SOAP based one.

Marcus


bolmsten:
Hi Marcus,

Will this work with the old ISPyB (4.0.9) or is it dependent on EXI. I do not know the state of EXI at the moment but I guess it is not ready for release?

bolmsten:
And can I test this? It seems that I need some new HO like lims-rest?

marcus-oscarsson:
Actually the REST API is part of ISPyB so you should be able to use the ISPyBRestClient without any problem. EXI is actually just the interface so its a bit incorrect of me to use that term. The staff here is using EXI today so its released for internal usage, but the users still use the old ISPyB interface.

marcus-oscarsson:
Yes you should be able to test it, the hardware object ISPyBRestClient.py as I mentioned above already exists. You need a configuration file
for your setup. I added an [example](https://github.com/marcus-oscarsson/mxcube3/blob/a9a4b54df5902507e41b2a739dad7311f6d00783/test/HardwareObjectsMockup.xml/lims-rest.xml), you just have to change our settings with yours.

IvarsKarpics:
I would recommend to remove login credentials from xml example...

marcus-oscarsson:
They are not in the xml example, thats just dummy data !

IvarsKarpics:
No, I did not mean to remove in principle the properties: ws_username and ws_password, because of course we need then to authenticate ispyb. What I mean is to remove real user name "ispyb**" and password of ISPyB from your example. Otherwise I can use these credentials to do nasty things.

marcus-oscarsson:
:) good. Actually we recently changed that here and I thought that was obfuscated data but its a real account so I have removed it from the file :). My understanding was that there were no clear text credentials in the file any more, but is not the case :). Thanks for pointing that out.

marcus-oscarsson:
Just updated the XML files with some examples they are a bit ESRF specific, but I hope thats okey

mguijarr:
@marcus-oscarsson I think the branch has to be rebased since it has conflicts.

bolmsten:
Hello Marcus, @marcus-oscarsson 

Great job with this PR. Me and Mikel spent much of Friday trying to get it going with our environment and had some success. 

We can get it running and it creates a session for us but we have not managed to sync the samples yet. It seems to be connected to the beamline name.

I have some small questions for you:

1. Am I correct that the ISPyB HO contacts a ldap directly and then contacts ISPyB? And isn't ISPyB contacting the same LDAP again when conforming credentials?

2. It seems that on init their is a connection made to ISPyB with the credentials given in the XML, if this is only a check to see if ISPyB is alive I guess it could be made without credentials. 

3. When this is accepted it is no longer possible to run mxcube "out of the box" without the ispyb connection? 

Br Fredrik

marcus-oscarsson:
Hey Fredirk,

That's good news the sample list is mapped to a proposal which in turn is associated with a beamline. This association is made in the ISPyB interface.

1:
We are using the ISPyB server to authenticate which in turn is using LDAP, but we are not explicitly authenticating with LDAP from MXCuBE. Jie made some changes around a year ago to make it possible to use LDAP authentication directly from the ISPyB HO. I guess that's because you at MAX IV can have several proposals per login where as we (for the moment) have a one to one mapping (the login and the proposal are the same. So I guess she wanted to separate the login from LIMS, she has to correct me if Im wrong though :) @JieNanMAXIV.

As it is today, with the 1-1 mapping, the login method of the ISPyB HO returns the valid proposal directly. This is of course not possible if you are using a 1-many mapping without, either in ISPyB map a current proposal to a login or add a interface for the user to select the proposal they want to use. That's something that we need to add but its not a part of this PR.

2:
These credentials are used by the collect routines to be able to write data to ISPyB. It was previously not possible to write to ISPyB with all accounts, so that default account have been used because of that. This have changed according to the developer of ISPyB so you should be able to use the user accounts
to write data, although that would probably require some smaller changes to the ISPyBClient HO.

3:
MXCuBE should work with any LIMS inlcuding ISPyB (provided that it provides some basic features). We should also be able to use MXCuBE without LIMS if necessary. Its just a matter of providing the right configuration for handling login and so on ...

PR #338, opened 2016-09-27
Helical Scan first version

bolmsten:
This PR is for creating lines and making Helical scans. It also has some minor GUI fixes to the todo queue

### DIFF
--- a/mxcube3/routes/qutils.py
+++ b/mxcube3/routes/qutils.py
@@ -24,6 +24,7 @@ class PMock(Mock):
 def node_index(node):
     """
     Get the position (index) in the queue, sample and node id of node <node>.
+
     :returns: dictionary on the form:
               {'sample': sample, 'idx': index, 'queue_id': node_id}
     """
@@ -59,6 +60,7 @@ def node_index(node):
 def queue_to_dict(node=None):
     """
     Returns the dictionary representation of the queue
+
     :param TaskNode node: Node to get representation for, queue root used if
                           nothing is passed.
     :returns: dictionary on the form:
@@ -81,6 +83,7 @@ def queue_to_dict(node=None):
 def queue_to_json(node=None):
     """
     Returns the json representation of the queue
+
     :param TaskNode node: Node to get representation for, queue root used if
                           nothing is passed.
     :returns: json str on the form:
@@ -105,6 +108,7 @@ def queue_to_json_response(node=None):
     """
     Returns the http json response object with the json representation of the
     queue as data.
+
     :param TaskNode node: Node to get representation for, queue root used if
                           nothing is passed.
     :returns: Flask Response object
@@ -157,6 +161,7 @@ def queue_to_json_rec(node):
     """
     Parses node recursively and builds a representation of the queue based on
     python dictionaries.
+
     :param TaskNode node: The node to parse
     :returns: A list on the form:
               [ {sampleID_1: [task1, ... taskn]},
@@ -186,6 +191,7 @@ def queue_to_json_rec(node):
 def get_entry(id):
     """
     Retrieves the model and the queue entry for the model node with id <id>
+
     :param int id: Node id of node to retrieve
     :returns: The tuple model, entry
     :rtype: Tuple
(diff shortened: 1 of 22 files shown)


mxcube3/routes/qutils.py
```diff
@@ -359,6 +359,23 @@ def set_dc_params(model, entry, task_data):
                 _cpos.index = int(params['point'])
                 acq.acquisition_parameters.centred_position = _cpos
 
+    if params["helical"]:
```
marcus-oscarsson: Hm, this was is already added by Mikels PR #283 are you sure that there is not a problem with the PRs ?

bolmsten: Good point, mine is before Mikels and maybe should be reviewed first

marcus-oscarsson:
Hey

I noticed some linting errors. I dont really mind its jut max-len you can perhaps fix them if you have time. Otherwise we will fix them later ...

Cheers,
Marcus

bolmsten:
I will try to fix them later on as I am planning a refactoring of the shape class

marcus-oscarsson:
Thanks Fredrik !,

The code is general clean and easy to read.

The modifier you are using right now seems to be Shift I think its maybe more common to use CTRL
for selecting multiple items ?

I also remembered that we would avoid to have the draw line step explicitly done by the user. But that
a line is drawn when two points are selected and helical chosen as method. We might have changed our minds and I'm simply not aware of it ? The reason was simply to save the user from drawing the line.

These are just reflections I think the code is fine in general, ill give Matias a chance to take a look before we merge.

bolmsten:
Thank you @marcus-oscarsson ,

I honestly do not remember the conversation regarding the line. I have briefly discussed with Mikel now and we feel like doing it like that is mixing styles if compared to creating a point. 

To add datacollection or charaterisation one first creates a point then adds tasks, our idea would be to do the same with the helical. This is in order to keep the styles the same but we can discuss best solutions.

Regarding the ctrl vs shift: I agree with you but the shift is built in the library of fabric js so I did no changes at all. We could look into how to change this.

marcus-oscarsson:
Okey,

I'm not so sure that it would be mixing styles, the actions in the drop down for a set of selected positions should be Add Collection, Add Characterization, Add Helical and Remove. So it makes sense that you can add a helical directly since you already can add collections and characterizations. The line is currently only used by the helical and explicitly adding it makes little sense. That's at least how we thought, but maybe you have something else in mind.

meguiraun:
Hej, I thought that we _operate_ on geometric elements, some actions are related to points and others to lines. In any case, could we accept this PR as it is now and let our product owners decide about right click on points/lines/whatever? We can modify this behavior later on, it should only affect helical.

mguijarr:
At least one commit in this branch is also present in Mikel's branch for characterisation : see [this one](https://github.com/meguiraun/mxcube3/commit/5bbd22a1e51f09bbfd99a97542cfe60be3e82209).
Could you please rebase your branch on master to have a cleaner PR ? It is difficult to follow.
Thanks in advance !

mguijarr:
@bolmsten, @meguiraun : the behaviour Marcus described above comes from our proposal last time you were at ESRF, see page 8 of the pdf document. If you don't have it I can send it to you again.

We talked about it at that time. This design is validated by Daniele of course. Do Marjolein have any strong opinion against it ?

As you propose Mikel we can accept the PR as it is regarding the 'right click' behavior, but it has to be fixed afterwards.

meguiraun:
no no no... It was just my opinion. I looked at the doc, my head can not contain much info... We can adapt to it later on.

mguijarr:
Thanks for removing the server io thing. But you added a commit to remove it, whereas it should not have been there at first... In addition you didn't rebase so we still have those ugly 'merge commits'.

bolmsten:
Rebased and implemented the way of adding helical as suggested

mxcube3/routes/qutils.py
```diff
@@ -23,7 +24,6 @@ def __reduce__(self):
 def node_index(node):
     """
     Get the position (index) in the queue, sample and node id of node <node>.
-
```
marcus-oscarsson: I would prefer if we keep the new lines, it makes the doc strings more readable

mguijarr:
Thanks a lot !

Issue #334, opened 2016-09-19
Filtering redux-persist


bolmsten:
There is a need to blacklist in more detail then a whole reducer. A example is sampleview, a lot of things in that reducer should be in sync with slave but pixelsPerMm is calculated for each client and should not be synced. 

I have not looked into redux-persist very much but Matias maybe you have a answer on how to do this?

mguijarr:
Yes I will take care of this when I will have some time. 

Basically it's as easy as writing a custom reducer for REHYDRATE action I think.

bolmsten:
The pixelsPerMm is not fixed yet but could be fixed in the same way I was doing with the points

mguijarr:
I am looking at this now... There are 2 aspects:
* `pixelsPerMm(Y|Z)` in the diffractometer object, which depends on the zoom level
* pixelsPerMm in the UI code, used for calculating distances or angles, which depends on the scaling factor
applied to the video image 

The first one doesn't depend on the client.
The second one has to be calculated from the first one when needed. I do not think we already have the feature for calculating distances and angles from points in the video image ?

It seems to me only the first one is kept in the state, which is ok.

So I am going to close this issue.

PR #339, opened 2016-09-29
Fixes ( includes performance update for redux-persist)

bolmsten:
This will be the first out of several small PR's focusing on bug fixing and performance.  

By breaking done reducers into smaller pieces code by comes more manageable and we can blacklist things that should not be updated by redux-persist. This includes points that are already sent out on the socketIO.


mguijarr:
Thanks !

I suppose this fixes issue #334 ?

The approach you took with an additional reducer we can blacklist is better than a custom REHYDRATE action as I was suggesting, because in this case it makes points management clearer IMO.

Issue #340, opened 2016-10-03
Mutating the redux state


bolmsten:
I observed today that some of the new code regarding the queue on the client is mutating the client state, look [here] (https://github.com/mxcube/mxcube3/blob/master/mxcube3/ui/reducers/queue.js#L184)

A way of testing this is to add some task to a sample and you will notice that according to the log the sample has always had that task. 

We should probably move over to immutable js or what do you think?

mguijarr:
I think we should be more careful.

What are the implications of moving to immutable.js ? What does it involves in term of time to spend to do it and in term of code ?

I am in the middle of several things, so I did not pay too much attention, but `Array.from` is making a copy so why do you say the state is mutated? (maybe I am overlooking something obvious ?)

bolmsten:
The Array.form is making a shallow copy of arrays with objects. It is very hard to do deep copy in javascript with complex types. 

I think we should fix this issue the ordinary way and have a discussion next time we meet regarding libraries that help with mutating the state.

marcus-oscarsson:
Solved by #343

PR #343, opened 2016-10-05
Deep copy

marcus-oscarsson:
Hey !

A very simple fix for issue #340, an alternative would be to use the node package [deepcopy](https://www.npmjs.com/package/deepcopy) or other similar solution. A third option could also be to use immuatable js as suggested by Fredrik. This should
however solve the problem for the time being. 

I know its maybe not the fastest way but I think it works well enough.

Marcus

### DIFF
--- a/mxcube3/ui/reducers/queue.js
+++ b/mxcube3/ui/reducers/queue.js
@@ -36,7 +36,7 @@ function deepCopy(obj) {
  *
  */
 function initSampleList(samples) {
-  const sampleList = Object.assign({}, samples);
+  const sampleList = deepCopy(samples);
 
   for (const key in sampleList) {
     if (key) {
@@ -58,7 +58,7 @@ function initSampleList(samples) {
  *
  */
 function recalculateQueueOrder(keys, gridOrder, state) {
-  const sampleList = Object.assign({}, state.sampleList);
+  const sampleList = deepCopy(state.sampleList);
   const sortedOrder = Object.entries(gridOrder).sort((a, b) => a[1] > b[1]);
 
   let i = 0;
@@ -85,7 +85,8 @@ export default (state = initialState, action) => {
     }
     case 'APPEND_TO_SAMPLE_LIST': {
       const sampleData = action.sampleData;
-      const sampleList = { ...state.sampleList, [sampleData.sampleID]: sampleData };
+      const sampleList = deepCopy(state.sampleList);
+      sampleList[sampleData.sampleID] = sampleData;
 
       return Object.assign({}, state, { sampleList });
     }
--- a/mxcube3/ui/reducers/queue.js
+++ b/mxcube3/ui/reducers/queue.js
@@ -1,6 +1,7 @@
 import { omit } from 'lodash/object';
 import { without } from 'lodash/array';
 import update from 'react/lib/update';
+import clone from 'clone';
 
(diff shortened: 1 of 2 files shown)


bolmsten:
Have you tested this solution? For me it still does not work.

marcus-oscarsson:
Of-course, what is it that does not work ?.

bolmsten:
If you look at the log before and after you add a task you will see the task is already their in the before.

marcus-oscarsson:
I see, you are right but I'm not sure that the line you identified is the source of the problem because the
deep copy routine works. So it has to be something else, ill try to see if I can find the actual source of this problem.

marcus-oscarsson:
So I found the problem, the problem was actually [line 188](https://github.com/mxcube/mxcube3/blob/master/mxcube3/ui/reducers/queue.js#L188).

I don't know why but I assumed that spread and Object.assign did a deep copy which is not the case. Its clearly stated in the Object.assign documentation:
```
    Warning for Deep Clone
    For deep cloning, we need to use other alternatives. This is because Object.assign() copies the 
    property  reference when the property being assigned is an object.
```
So actually we only copied the references. Actually we do still have a problem with the Array.from code
but it only shows when doing update. Since we otherwise only add and delete references. Ill take a quick look to see if there are more places to update.

Cheers,
Marcus

marcus-oscarsson:
So I'm using the deepCopy routine where I believe its needed, Its possible that I have missed to change in some places and I have only updated the queue reducer. This problem basically only occurs when we have key value pairs where the value is a reference which is, correct me if Im wrong, in the queue and the sample list.

bolmsten:
`      const queue = {
        ...state.queue,
        [sampleID]: {
          ...state.queue[sampleID],
          tasks: [
            ...state.queue[sampleID].tasks,
            action.task
          ]
        }
      };

      const displayData = {
        ...state.displayData,
        [sampleID]: {
          ...state.displayData[sampleID],
          tasks: [
            ...state.displayData[sampleID].tasks,
            { collapsed: false, state: 0 }
          ]
        }
      };`

bolmsten:
`
      const queue = {
        ...state.queue,
        [sampleID]: {
          ...state.queue[sampleID],
          tasks: [
            ...state.queue[sampleID].tasks,
            action.task
          ]
        }
      };

      const displayData = {
        ...state.displayData,
        [sampleID]: {
          ...state.displayData[sampleID],
          tasks: [
            ...state.displayData[sampleID].tasks,
            { collapsed: false, state: 0 }
          ]
        }
      };
`

bolmsten:
const queue = {
        ...state.queue,
        [sampleID]: {
          ...state.queue[sampleID],
          tasks: [
            ...state.queue[sampleID].tasks,
            action.task
          ]
        }
      };

      const displayData = {
        ...state.displayData,
        [sampleID]: {
          ...state.displayData[sampleID],
          tasks: [
            ...state.displayData[sampleID].tasks,
            { collapsed: false, state: 0 }
          ]
        }
      };

bolmsten:
A bit of code showing how to copy only the part of the queue that is changed

marcus-oscarsson:
Thanks for the code,

Ive have applied the principle used in the snippet to replace the clone call. The clone packages
has also been removed. 

Cheers,
Marcus

PR #345, opened 2016-10-11
WIP: Fix loggers and logging handlers

mguijarr:
This PR fixes the exceptions eaten up by the root logger, which only had 'custom_log_handler' handler: exceptions were not printed on stdout, which made debugging a lot hardier.

It also comes with logging enhancements:
* custom_log_handler level is now INFO (instead of DEBUG).
* use same formatter for all log handlers
* use TimedRotatingFileHandler for (optional) log file, which aggregates all log messages
* all log handlers log to stdout and emit messages to web clients.

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -6,6 +6,7 @@ from optparse import OptionParser
 import os
 import sys
 import logging
+from logging import StreamHandler, NullHandler
 import cPickle as pickle
 import gevent
 import traceback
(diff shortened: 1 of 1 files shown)


mguijarr:
Hi Mikel,

This was a WIP. So not intended to be merged immediately because I wanted to do more testing on it.

That said I think it works fine - did you test it ?

meguiraun:
jaarrrll!!!!!! sooooorryyy for my crazy finger!! I did not test the pr this time... I went through the code.

I can re-open if you prefer, it is an improvement as it is anyway

mguijarr:
Ok so I remove WIP.

PR #347, opened 2016-10-13
New queue gui with tabs

bolmsten:
A implementation where the queue now has tabs and the appearance of a task is a bit more inline with the rest of the gui


marcus-oscarsson:
Thanks Fredrik !

Issue #346, opened 2016-10-12
Running without ldap and lims by default


bolmsten:
Marcus latest PR was a great first step with getting mxcube and ispyb connected but it is no longer possible to run mxcube out of the box without configuring the ldap and lims xml files. 

I suggest that we go back to mockups for the lims,ldap for the default

marcus-oscarsson:
Hey !

Its simply a matter of the default configuration that we would like to provide. I wanted to provide a configuration that was working with LIMS with the PR I made. Your right that its maybe makes more sense
to use the mockup versions by default.

Marcus

mguijarr:
+1 on going back to mockups for lims and ldap by default

PR #348, opened 2016-10-13
Continuing LIMS and queue integration

marcus-oscarsson:
Dear all,

I wanted to divide this PR into two PRs, one for queue integration and one for LIMS but I finally decided that it was easier to just have one. In anyways, that's why there are both LIMS and queue related features in the same PR :). Ill try to avoid this in the future ...

I started to take care of the sample order, since the JS object is non ordered. Ive created a structure called sampleOrder and refactored the grid-ordering slightly for that purpose. The next thing to do is to use this in the queue and finally of course on the back end so that the samples are executed in order :). Exactly how to use this have to be determined it could perhaps be replaced by or replace the existing todo structure.

The "execution state" of a task was moved from dislpayData to queue. There are also some smaller improvements to the grid, like disabling certain controls while executing.

For the lims part, Ive started to prepare for having the data collection results being shown (LIMS collection ID being sent as part of the result) . The prefix is now taken from the Session object. There is also a new UI for filtering (although not completely done yet).

The mockup objects for lims are also used by default :)

Ive not yet had the time to take care of the proposal list for the "one login" to "many proposal" mapping. Ill try to do it next week (when our shutdown is over).

Cheers,
Marcus

### DIFF
--- a/mxcube3/routes/SampleChanger.py
+++ b/mxcube3/routes/SampleChanger.py
@@ -21,7 +21,7 @@ def get_samples_list():
         sample_dm = s.getID() or ""
         sample_data = {"sampleID": s.getAddress(),
                        "location": ":".join(map(str, s.getCoords())),
-                       "sampleName": "Sample %s" % s.getAddress(),
+                       "sampleName": "Sample-%s" % s.getAddress().replace(':', ''),
                        "code": sample_dm,
                        "type": "Sample"}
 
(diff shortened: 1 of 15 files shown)


bolmsten:
Thank you!

mguijarr:
This solves #346 it seems ?

Is there really a need for this: proposal list for the "one login" to "many proposal" mapping ? It's for you guys at MAX IV, as far as I know you are using user login, then proposal for ISPyB ?

bolmsten:
Yes it does!

We will need something like that but it does not have priority at the moment.

mguijarr:
Ok. So I add an issue for this, I let you give the priority you want and assign someone on your side or we will do it later.

PR #350, opened 2016-10-14
queue fixes and task expand on run

bolmsten:
Can copy the prefix and cursor is now a pointer of parameters. Task expand and close when queue is running


mguijarr:
Thanks :)

Have a nice weekend !

PR #351, opened 2016-10-17
First version of color selection points

bolmsten:
This PR gives colors of selected points in this case red (to be discussed). While developing this I encountered a number of questions that need to be answered, so I will put them here for disucssion. 

1. What color should a selected point be? 
2. Will it be the same color for a point that is not saved but selected as a saved point.
3. Can one create a group with a saved point and a tmp point? And how does one see the difference between this and a group with only saved points.

Test out the PR if you can and give feedback, there is a lot of cases to consider and I have probably missed a few.

### DIFF
--- a/mxcube3/ui/components/SampleView/SampleImage.jsx
+++ b/mxcube3/ui/components/SampleView/SampleImage.jsx
@@ -227,6 +227,7 @@ export default class SampleImage extends React.Component {
       fabricSelectables.forEach((shape) => {
         if (groupIDs.includes(shape.id)) {
           selectedShapes.push(shape);
+          this.setColorPoint(shape);
         }
       });
       this.canvas.setActiveGroup(
@@ -241,9 +242,12 @@ export default class SampleImage extends React.Component {
       fabricSelectables.forEach((shape) => {
         if (shape.id === selection.id) {
           this.canvas.setActiveObject(shape);
+          this.setColorPoint(shape);
         }
       });
     }
+    console.log(this.canvas.getObjects());
+    this.canvas.renderAll();
   }
 
 
--- a/mxcube3/ui/components/SampleView/SampleImage.jsx
+++ b/mxcube3/ui/components/SampleView/SampleImage.jsx
@@ -61,7 +61,7 @@ export default class SampleImage extends React.Component {
       shape.forEachObject((p) => {
         const point = p;
         if (point.type === 'SAVED' || point.type === 'LINE') {
-          const color = selection ? 'green' : point.defaultColor;
+          const color = selection ? '#88ff5b' : point.defaultColor;
           const width = selection ? 4 : 2;
           point.stroke = color;
           point.text.stroke = color;
@@ -73,12 +73,13 @@ export default class SampleImage extends React.Component {
       this.canvas.getObjects('SAVED').concat(
       this.canvas.getObjects('LINE')).forEach((p) => {
         const point = p;
-        const color = point.active ? 'green' : point.defaultColor;
+        const color = point.active ? '#88ff5b' : point.defaultColor;
         const width = point.active ? 4 : 2;
         point.stroke = color;
         point.text.stroke = color;
         point.text.fill = color;
         point.hasControls = false;
+        shape.hasBorders = false;
         point.strokeWidth = width;
       });
     }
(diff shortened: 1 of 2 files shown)


marcus-oscarsson:
Hey Fredrik,

Ahh, great ! 

1. Today the points become green and are made "bold" (thicker lines) when selected, 
    and a point is only displayed when its saved. So we dont have the exact same beahviour
    in MxCuBE 2. Since we now, in MxCuBE 3 display unsaved positions in white.

    However we can keep a similair approach, we can display a selected point in "bold" or use
    individual selection markers for each point. The points still keep their originial color (white 
    and yellow) but with a individual selction box, or bold, when selected. We can like this sill
    distinguish between saved an unsaved positions. Its like this also easier to select and
    deslect other points and still see whats selected, without the big selection box masking
    the entire selection.

2. I think the the anweser to question 1 also anwsers this one (if I didnt missunderstand you).
    Like today White for unsaved and yellow for saved, we can also use other colors if you have
    good suggestions. The colors are orginially chosen to be easy to see on for instance dark
    displays, like green and blue. Yellow is actually not so good because the light is sometimes
    a bit yellow and can mask the points a bit. But I think its okey for the moment to keep white
    and yellow.
    
3. Same here, the answer to 1 also ansers this one. The colors are kept but we use either
    bold or indivudial selection markers on selected points.

The points should also be labled with the number and a prefix like Pn, Ln, Gridn and so on.

I hope this it helps !

Marcus

bolmsten:
Hi @marcus-oscarsson,

please test it out a bit. I tried removing the color for selection but I think it is better to have something. 

Please play around a bit with it, all you have to change is a few variables in the function setColorPoint.

mguijarr:
Thanks Frederik we are going to test it this afternoon and will get back to you :)

marcus-oscarsson:
Hi Fredrik,

Sorry we were to busy to get back to you yesterday afternoon. Its already really good as it is I have a few ideas but maybe you guys are of another opinion.

It would be nice if its possible to get rid of the most outer selection anchor/box because its a bit in the way and it masks the actual selection.

The colors could be a bit brighter I know that it does not look very good with "florescent" colors but
they are easy to see on the video screen, especially when its dark. The selection box/anchor could also
be colored to be a bit more visible. The color scheme (green, blue and yellow) you have today is great 
its more the "luminescence" (if that's the right word) I'm thinking about.

I'm using firefox on windows and the fonts I get are a bit blurry so its a bit difficult to see the labels, what fonts are we using ?

Cheers,
Marcus

bolmsten:
Thanks for the feedback. 

The anchors of elements I can remove but I would keep them for groups as it becomes hard to see where to click to deselect or get a context menu for a group if removed. 

Regarding the colors I totally agree and I will try to make them "pop" more. 

I will fix and get back to you.

bolmsten:
I updated the colors and changed the border but there is more to do regarding interaction I think. But at the moment I will leave it as this and let people discuss.

marcus-oscarsson:
Thanks for the fast updates,

I noticed that you have removed the boxes around the individual elements I though that was quite a
nice way to mark something as selected. I still think that the "group box" is more in the way than it helps, especially when deselecting certain items. Deselecting items could basically be done by clicking on a none selected item and the context menu would apply to all elements when an (any item of the selection) item is clicked. The group box, in my opinion, only adds one more element to interact with.

I'm of-course happy to discuss the interaction further I'm happy to merge this PR as it is, apart from that I noticed something that might be a bug. Ill explain what I found and I hope you will let me know if its intentional or not. Or maybe that's what you meant with more to do regarding interaction ?

I selected two points and a line with the "rubber band" / mouse drag, the points are marked green as they should. However deselecting them does not mark them as deselected they remain green, but selection bounding box at the same time behaves as it should.

Also selecting two points and then trying to select the line between them does not seem to work, it seems like the line is masked by the bounding box and I cant seem to click it, or its simply not marked as selected.

Let me know if you prefer to fix possible bugs now or if you rather do it later, and Ill merge this PR.

Cheers,
Marcus

bolmsten:
I have written a lot of code to make it work like I think you want @marcus-oscarsson @mguijarr , please take a look and give feedback

marcus-oscarsson:
Hey !

The changes you made are really nice. We consider the canvas to be one of our core components and to make the interaction with it as easy as possible important not to make our users frustrated. That's why we are a bit picky when it comes to these things. You have already done a great job so far and I'm sure that you will keep it going :+1: .

Some final thoughts, is it possible to remove the control handles for the selection boxes ? Could we make the selection box color be a bit more neon blue so that its easier to see ? It would be great if
its possible to remove the group box. I'm in anyways happy to merge this since its a nice improvement.

Thanks,
Marcus

PR #356, opened 2016-10-20
Queue execution/integration (mainly automatic mode)

marcus-oscarsson:
Hi all,

So we have continued the work of the queue integration and execution. We are now able to run "semi automatic" data collections. What I mean with semi-automatic is that we run the queue automatically loading samples but we center with three click centering. There is in theory nothing that prevents us from using automatic centering but there are some additional work associated with feedback to the user when centering automatically.

There are still some work to be done at the level of user interaction but also some glitches in execution process to be fixed.

We have also made some more technical changes and most notably we have gone back to the famous
set queue API. This change is hopefully alright with you now, we can discuss this further on Tuesday if you think its necessary. I will in the meantime mark this as WIP although it is finished to be merged, feel free remove the WIP flag and merge if the api is no longer an issue for you.

Cheers,
Marcus

### DIFF
--- a/mxcube3/ui/actions/queue.js
+++ b/mxcube3/ui/actions/queue.js
@@ -39,7 +39,7 @@ export function sendClearQueue() {
       headers: {
         Accept: 'application/json',
         'Content-type': 'application/json'
-     }
+      }
     }).then((response) => {
       if (response.status >= 400) {
         throw new Error('Server refused to clear queue');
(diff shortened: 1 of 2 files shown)


meguiraun:
hej Marcus,

Thanks for the work, it's great you can launch semi auto collections!

However, we have never accepted the famous set queue API, so I am a bit surprised to see it again, specially being a part of a PR and not as an standalone PR. How about removing that part from the present PR so we can test the semi auto operation?

marcus-oscarsson:
Hej Mikael,

Well. we can perhaps discuss the API on Tuesday if you like, I believe Matias mentioned that we wanted to use this API in the last email and you didn't seem to have anything against it then. Sorry if we misinterpreted your reply. The reason that we changed to the set API now is that we started to see the problems that we thought we would have with the per operation API. We simply think that using the set API will save us a lot of problems and make the implementation easier.

Yes you are maybe right that I should have done two PRs. I just wanted to supply one PR that was working since I also did some changes that built on API changes. It would actually have been a bit complicated to provide something easy to use without providing three PRs. I simply found this the easiest. So I would really like to avoid to split this PR into several ones.

In the meantime you should be able to test the code in anyways, regardless of the API.

Marcus

mguijarr:
I didn't receive any reply on this particular point in my last email ("Do you see it as a problem if we
propose an implementation of our idea ?"), so I thought you finally had no problem in letting us
implementing the 'set queue' API, at least to give it a try.

So we did it (again) and tested it (again), and we are (again) very happy with this change - exactly as we
expected. If you really want we can have another technical discussion on this, but honestly it starts to be
embarrassing.
 
You can perfectly test this code as it is - we would be happy to get your feedback. 

We won't split the PR - we do not have time for this additional work, I think you can understand.

That said, the code is not carved in stone ; it's always possible to change it later if you can prove
something is inherently wrong with the 'set queue' API... Meanwhile, I can assure you it is better as it is now than before.

PR #357, opened 2016-10-20
Added sample controls in overlay

marcus-oscarsson:
Hello all,

Ive added the video controls in an overlay over the sample video as we did in some of the sketches. The zoom and light controls perhaps needs some improvements in order to look a bit better.

Let me know what you think.

Thanks,
Marcus


bolmsten:
Hi Marcus,

I will try to ask Marjolein for her input as this is more a design decision and get back to you.

/Fredrik

marcus-oscarsson:
Hi !

Okey, sounds good. I noticed that it might need some more tweaking on really light displays. We could also consider the possibility to hide it for users that are disturbed by it.

Marcus

PR #368, opened 2016-10-27
change phase in the samplecontrol bar

bolmsten:
This takes the phaselist from the server and gives the user a selection box for changing the phase.

I have only been able to test this with mockups and there it works.

### DIFF
--- a/config.gui.prod.js
+++ b/config.gui.prod.js
@@ -1,0 +1,3 @@
+module.exports = {
+	phaseControl: true
+}
--- a/mxcube3/routes/Diffractometer.py
+++ b/mxcube3/routes/Diffractometer.py
@@ -159,9 +159,15 @@ def get_diffractometer_info():
         ret['useSC'] = False  # in case the diff does not have this implemented
 
     try:
-        ret['current_phase'] = mxcube.diffractometer.current_phase
+        ret['currentPhase'] = mxcube.diffractometer.current_phase
     except AttributeError:
-        ret['current_phase'] = 'None'  # in case the diff does not have this implemented
+        ret['currentPhase'] = 'None'  # in case the diff does not have this implemented
+
+    try:
+        ret['phaseList'] = mxcube.diffractometer.get_phase_list()
+    except AttributeError:
+        ret['phaseList'] = []  # in case the diff does not have this implemented
+
 
     resp = jsonify(ret)
     resp.status_code = 200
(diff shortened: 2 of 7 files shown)


mguijarr:
Would it be possible to hide the phase selection control if the equipment doesn't need this?

bolmsten:
How can we determine if the equipment doesn't need it?

bolmsten:
I agree that having a list with features to hide or to show could be good way of customising the interface. The other way would be to keep the containers very clean and people then only modify them to change the interface. There is benefits to both, maybe we will have multiple components doing the "same thing" such as kryocontrol and people need to choose which too use.

mguijarr:
*Reply on 'How can we determine if the equipment doesn't need it?' :*
The first thing that comes to my mind is to make a specific route, in this route we can read a site-specific configuration file to determine what is there or not, the route would reply something like:
```{ "aperture": False, "phase": False, "kappa": False, ... }```, that would be taken into account in the UI to hide/disable corresponding components.
The configuration file could be a xml file (corresponds in spirit as existing Hardware Objects), or another format like Windows .INI files.

marcus-oscarsson:
Hm just some ideas, we could perhaps read a config and define variables on the fly on the process. So that we could use that to determine which component to use if there are specific components or other configuration options.

`for instance if (process.config.aperture) { .... }`

I dont know if it is possible to do some expansion in the JSX so that we somehow could do.

`<process.config.DiffractometerComponent ...>`

Otherwise we could perhaps just do some simple redefinition somewhere so that DiffractometerComponent always points at the specific component we would like to use.

for instance:

`<GenericComponents.DiffractometerComponent ...>`

mguijarr:
I find Marcus'idea to be better than mine if it is possible to do it. `process` is a global Node thing, that could be read when building the assets and then the result bundle could contain those components or not. In this case, the configuration file would a javascript file...

beteva:
I like the idea of the DiffractometerComponent always pointing to a specific implementation. The problem will come not only with the phase, but also with other diffractometer specific features  (like plate grippers for example or like Matias pointed out - apperture, minikappa, scintillator... )

bolmsten:
@mguijarr @marcus-oscarsson  I have done a implementation of what was mentioned above, please take a look.

mguijarr:
@bolmsten Looks great to me. Thanks.

PR #369, opened 2016-10-27
Lims integration - collection results

marcus-oscarsson:
Hello again,

This PR provides results for data collections, I have not done the results for characterizations and helical yet. The LIMS results are for the moment added when the collection results are sent to the UI (in the collection related callbacks). I'm not sure it will stay like this, it depends a little bit on our conclusions regarding the websocket discussion and to handle lost signals.

Ive also added the a simple functionality in login to handle the many proposals per login (as mentioned earlier). This does not return anything for us since we are not using this feature. I hope I got it right because, I cant really test it here. So please let me know if you have difficulties to login and Ill either fix the problem or remove the feature (since you said you solved it in some other way)

Cheers,
Marcus

### DIFF
--- a/mxcube3/routes/Login.py
+++ b/mxcube3/routes/Login.py
@@ -14,52 +14,62 @@ MASTER = None
 def login():
     """
     Login into mxcube application.
-        :form proposal: proposal as it appears in duo
-        :form password: corresponding password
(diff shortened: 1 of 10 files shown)


amilan:
Thanks, Fredrik already test it and it looks nice.

PR #370, opened 2016-10-28
Queue serialization and persistent storage (in redis)

marcus-oscarsson:
Hey !

Added some missing pieces to the queue serialization to for instance handle helical scans. Also added the routines necessary to load a queue from its serialized format. Which at the moment is a pickled dictionary (the one returned by queue_to_dict). We could use any queue_to method if we consider one representation being better than the other.

A side note, there is now a parameter in the parameters dict of the data collection named 'helical' its perhaps possible to renamed it to experimentType (or similar). This would match the internal model better and also makes it possible to easily add other types of data collection without adding an extra parameter.

Ive also disabled the loading for persistent storage, for the moment, since its causing us some problems with adding things to the queue. Not only because of PMock but also because the queue is loaded on login and not synchronized with the UI. We thus sometimes try to add the same samples twice which produces an error. 

The queue is currently saved on logout, the idea is also to save it on any other client disconnect event. And restore it on login if there is no "active" session.

Also note that restoring a queue from persistent storage also depends on the state of the hardware at the time of the save so it might not be as straightforward as just performing a load queue. So I prefer not to load the queue for the time being, until we have defined what to restore when ...

Cheers,
Marcus

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -91,7 +91,8 @@ if not app.debug or os.environ.get("WERKZEUG_RUN_MAIN") == "true":
 
     ### Importing all REST-routes
     from routes import (Main, Login, Beamline, Collection, Mockups, Utils,
-                        SampleCentring, SampleChanger, Diffractometer, lims)
+                        SampleCentring, SampleChanger, Diffractometer, Queue,
+                        lims, qutils)
 
     ### Install server-side UI state storage
     from mxcube3 import state_storage
@@ -110,6 +111,7 @@ if not app.debug or os.environ.get("WERKZEUG_RUN_MAIN") == "true":
         app.empty_queue = pickle.dumps(hwr.getHardwareObject(cmdline_options.queue_model))
         app.sample_changer = app.beamline.getObjectByRole("sample_changer")
         app.rest_lims = app.beamline.getObjectByRole("lims_rest_client")
+        app.queue = qutils.new_queue()
 
 
         try:
--- a/mxcube3/routes/Login.py
+++ b/mxcube3/routes/Login.py
@@ -46,6 +46,16 @@ def login():
 
         LOGGED_IN_USER = loginID
 
+        # Create a new queue just in case any previous queue was not cleared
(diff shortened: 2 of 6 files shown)


mxcube3/routes/Login.py
```diff
@@ -80,15 +94,11 @@ def loginInfo():
             return make_response("", 409)
 
         # auto log in
-        loginInfo["loginRes"] = limsutils.lims_login(loginID, loginInfo["password"])
+        #loginInfo["loginRes"] = limsutils.lims_login(loginID, loginInfo["password"])
```
mguijarr: Why is this removed in this PR?

mxcube3/__init__.py
```diff
@@ -110,6 +111,7 @@ def complete_initialization(app):
         app.empty_queue = pickle.dumps(hwr.getHardwareObject(cmdline_options.queue_model))
         app.sample_changer = app.beamline.getObjectByRole("sample_changer")
         app.rest_lims = app.beamline.getObjectByRole("lims_rest_client")
+        app.queue = qutils.new_queue()
```
mguijarr: What is the difference between app.empty_queue and qutils.new_queue? Why would we need both?

mxcube3/routes/Login.py
marcus-oscarsson: Because it was as I understood it not needed, the only thing lims_login does is to login to LIMS. We don't really need to do that if the user is already logged in. We, should, only need to login to LIMS when the users logs in. The auto login feature should just keep the session information for the "current" user. It seems unnecessary to login to LIMS twice, but maybe I'm mistaken.

mxcube3/__init__.py
marcus-oscarsson: Hm I agree with you that it seems strange. The app.empty_queue is a serialized version of the empty queue that is used to initialize a new queue, Its perhaps better to just instantiate a QueueModel Object directly. Or do you have another suggestion ? I thought it was done in that why for a reason so I didn't question it to much.

Issue #366, opened 2016-10-26
Wrong state of queue if client disconnects at run time


bolmsten:
If the client disconnects when the queue is running (even for a short while like seconds) it can lose information about a running sample. And be forever in this state:

![screen shot 2016-10-26 at 11 19 17](https://cloud.githubusercontent.com/assets/6403388/19720360/192ddea2-9b6e-11e6-9dcb-ca0d63b3bc6e.png)

mguijarr:
See PR #372

PR #372, opened 2016-11-03
Fix for #366

mguijarr:
This is a fix for #366, and more generally it adds a `safe_emit` function that can be used to ensure the Master client will receive updates from server. It uses the 'callback' feature from socketio: events (= signals from Hardware Objects) are appended to a double-ended queue, and delivered in FIFO order. When callback is executed, it means the Master client has received the message so the event is deleted and the next one can be sent.
It is enough to send events to the Master client to get all clients up-to-date thanks to the mechanism already in place for Remote Access.
This PR also brings some little improvements to Remote Access, like a dedicated Python module and a new `setRaMaster` event which will also be used later when switching from one Master client to another.
It has been tested here and seems to work beautifully.

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -94,7 +94,7 @@ if not app.debug or os.environ.get("WERKZEUG_RUN_MAIN") == "true":
                         SampleCentring, SampleChanger, Diffractometer, lims)
 
     ### Install server-side UI state storage
-    import state_storage
+    from mxcube3 import state_storage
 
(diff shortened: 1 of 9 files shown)


mguijarr:
I just rebased this one to remove the merge conflict (introduced by changes yesterday). So it can be merged.

amilan:
Hi Matias, 

Thanks for the work. Just as a comment, for the next time, could you split it in more than one commit? It will make things easier.

Thanks in advance.

PR #373, opened 2016-11-03
Rate limit motor position updates

mguijarr:
This PR improves motor position updates:
* limit to 3 events per motor per second (using token bucket algorithm)
* position updates have a dedicated callback, just transmitting to clients the new position instead of
doing all the stuff from `signals.motor_event_callback`


bolmsten:
Looks good!! Could help with performance!

PR #374, opened 2016-11-03
Add Machine Information component in mxcube3 based on MachInfoMockup

vrey01:


mguijarr:
I noticed this would set Hardware Objects submodule to commit a5ef714, which corresponds to master branch... It should be commit 32a7e37e4585fb1ed340cd944760a86f61cfabec (HEAD on 2.2 branch).
Could you please fix the submodule reference ? (you just need to adjust it properly in your repository, and do another 'push' to update the PR)

mguijarr:
I didn't do the code review for the rest yet, will do when the submodule will be fixed

vrey01:
Yes. I considered that.  
The question is part of a reflexion on how to deal with beamline or site specific items.  

Machine info can be general (if reduced to its minimum) and it is coded with generality in mind. I will add it to Beamline now. 

But the reflexion is opened. There are request already here for things that are perfectly reasonable to show and at the same time only applicable to this case.   We will have an ocassion to talk about other than in this thread.
Before reviewing, please wait for a new commit, with the MachInfo included in Beamline

mguijarr:
Sorry but it seems to me vrey01/mxcube3/HardwareObjects points to 75c21fc, whereas it should be 32a7e37... You're missing one 'pull' in your Hardware Objects to get the merge commit since I accepted your PR.

mguijarr:
For the ideas regarding site-specific components you might be interested to look at this conversation we had, with PR #368.
Marcus proposed to use the `process` object, and to set some variables depending on what we want to display or not in the UI. So when we will build the production bundles, we would get a different MXCuBE at each site depending on this configuration.
I tend to agree with this proposal... What do you think ?
We should stick with generic components and generic hardware objects as much as we can, but of course there are some differences which cannot be made generic so you're perfectly right we need a solution for those cases too.

mguijarr:
Thanks a lot :)

We will update the hardware objects submodule next time (it is 1 commit behind).

PR #379, opened 2016-11-08
Queue execution/integration (mainly automatic mode) 

marcus-oscarsson:
Hi all,

Rebased version of #356 that was accidentally removed due to a git mistake. See PR #356 for 
full discussion

Cheers,
Marcus


amilan:
Hi, 
As we agreed yesterday in our meeting, the easiest way to get the functionality that we need from this PR is merging it and then some extra work will be required in order to merge/solve conflicts with the #385. 

Cheers.

Issue #381, opened 2016-11-09
motor signal rate causing long delay


meguiraun:
https://github.com/mxcube/mxcube3/pull/373 introduced a limit on the number of signals sent per second. However, if there are many signals in a very small period there will be all delayed and the client will still receive signals long after the movement is finished. 

Rough test with our diff: 90deg rotation -> ~10 signals/second, and the movement takes ~1 sec

Could we discard the signals arriving inside the delay interval? So if we have 3 sig/sec as limit the client will only receive 3/4 signals, and all will arrive ~synchronized with the movement.

mguijarr:
Of course it's my mistake - it should be like you suggest here, my intent was to discard those events. We just did a fix, see #383

PR #383, opened 2016-11-09
Fix for issue #381

mguijarr:
* rate limitation decorator now drops events received in the delay interval
* motor states are handled individually (like motor positions since #373)
* as a consequence, `signals.microdiffSignals` and `motor_signals` have been purged from
position and state signals
* CSS transition duration for motor states is 300 ms now instead of 1 s

We tested this on ID30A3 just now.

### DIFF
--- a/mxcube3/routes/SampleCentring.py
+++ b/mxcube3/routes/SampleCentring.py
@@ -24,25 +24,28 @@ def init_signals():
     mxcube.diffractometer.savedCentredPos = []
     mxcube.diffractometer.savedCentredPosCount = 1
     for signal in signals.microdiffSignals:
-        if signal in signals.motor_signals:
-            mxcube.diffractometer.connect(mxcube.diffractometer, signal,
-                                          signals.motor_event_callback)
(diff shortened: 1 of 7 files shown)


meguiraun:
Wow, that was fast! We will test tomorrow morning! Thanks

meguiraun:
almost there. It has been improved but now we drop the last signal as well, so it happens that the final position is never updated (until any future emit on such motor).

mguijarr:
Ah ah. Sorry.

When I first tried to fix it there was still the big thing that sends all signals at a state change, and I relied on this behavior to make sure last position was up-to-date.

But then I decided to further optimize states too and I removed `motor_event_callback` for state updates, and it leads to the bug you describe.

I am going to fix it pretty soon :) Sorry for the inconvenience.

By the way, this is what I called "handling motors individually" in the list of issues I sent last time.

mguijarr:
Can you try now, Mikel ?

meguiraun:
sorry, beamline busy.... I will test tomorrow morning. Thanks.

PR #385, opened 2016-11-11
[WIP] Init client from queue HO

bolmsten:
This PR makes the client init from the queue ho. As discussed before we needed a way of getting the queue HO from the server down to the client. 

Me and Mikel have done a first implementation of this now and it seems to work. 

Please test it out.


mxcube3/ui/main.jsx
```diff
@@ -70,7 +70,7 @@ export default class App extends React.Component {
   componentWillMount() {
     const persistor = persistStore(store,
            { blacklist: ['remoteAccess', 'beamline', 'sampleChanger',
-                         'form', 'login', 'general', 'logger', 'points'],
+                         'form', 'login', 'general', 'logger', 'points', 'queue'],
```
mguijarr: We still don't agree on that (see all the previous discussions...).

mguijarr:
It's bank holiday today here, so we will continue this discussion on Monday I think.

However, I would be happy to know which problem exactly is solved by this PR ? (which scenario ?).

When will a client need to receive queue state from the server using this code ? The use cases are unclear to me at the moment.

bolmsten:
This makes the application a lot more robust as we always get the true state from the HO on init.

A instance we encountered here is that we got a error before trying to send a socketIO signal in the server and therefore the state differed. We could fix this and reload the client and have no data loss because of this feature.

mguijarr:
Could you please elaborate a bit on the problem you encountered? Which error? How did you fix it without restarting the server?

mxcube3/routes/qutils.py
```diff
@@ -140,14 +140,140 @@ def queue_to_json_response(node=None):
...
+
+        sample_list.update({name: sample_data})
+
+    if mxcube.diffractometer.use_sc:
+        if len(sample_list) != 0:
+            sample_list = samples
+        else:
+            sample_list = {}
+        try:
+            basket, puck = mxcube.queue.mounted_sample.split(':')  #sample_changer.getLoadedSample().split(':')
```
mguijarr: - where does `mounted_sample` comes from in `mxcube.queue` ? I had a quick look but I could not find it
- is it important to keep the current loaded sample when saving queue state ? Only the sample changer knows (what if user removes the sample? What if user changes the sample before the queue is restored?)
- If `mounted_sample` is `1:1:1` the code above breaks when doing the `split()`

marcus-oscarsson:
Dear all,

I have not yet looked at the code, Ill start as soon as I can. Thanks for the work you guys did plenty things :). Its great that you also take the time to remove extra white spaces, I really appreciate that. I personally like clean source files. Ive mostly, historically, tried to use PEP-8 as much as I can but I have not been to strict on it. Ive noticed that some of the lines are maybe a bit long, I don't know what you think but id suggest 79 cols has PEP-8. Maybe not very important, in anyways I prefer to share my opinion like this instead of leaving a line comment.

mxcube3/routes/qutils.py
```diff
@@ -157,14 +283,18 @@ def _handle_dc(sample_id, node):
     parameters.pop('acquisitions')
     parameters.pop('acq_parameters')
     parameters.pop('centred_position')
- 
+    queueID = node._node_id
+    enabled, state = get_node_state(queueID)
     res = {"label": "Data Collection",
            "type": "DataCollection",
            "parameters": parameters,
            "checked": node.is_enabled(),
```
marcus-oscarsson: Be a bit careful here you seem to be assigning "checked" twice here and on line 295

mxcube3/routes/qutils.py
mguijarr: I just looked at `HardwareObjects/sample_changer/GenericSampleChanger.py`, so:
* the `sample` argument passed to `.load()` can be a string, that represents an 'address', the 'address' being in the form: `container_index:...:sample_index` (like `1:1` or `1:1:1`) or a `Sample` object (see `sample_changer/Sample.py`)
* as far as MXCuBE is concerned, I think we always use `.load()` with a sample string so it's something to split on `:`
* I still don't see why we would want to keep the mounted sample as something that belongs to the queue (it's even more true if it is manually mounted)

mxcube3/ui/main.jsx
mguijarr: By the way for some reason this commit appears twice (with 2 different SHA-1 ids)

mxcube3/routes/qutils.py
meguiraun: this new call gives you:
```
{
"history": [],
"loaded": "1:01",
"sample_list": [ sampleID_1, sampleID_2 ...],
"todo": [ sampleID_1, sampleID_2, ...],
"queue": {sampleID_1: [task1, ... taskn]
                        .
                        .
                        .
                  sampleID_n: [task1, ... taskn] }
"sample_list": {......}
}
```
It is in fact the `queue_dict` plus some list populated with info from the queue hwobj. We also need the currently mounted sample in this call so it is easier to manage by the client, it does not need to do several calls to retrieve the info. I added to `mxcube.queue.mounted_sample` for convenience and also because it is somehow related, I can not add to the sample_changer in case there is not a sc, there will always be a queue...

I will improve somehow the address stuff.

meguiraun:
I also try to follow pep8, but I try to avoid to many changes together with other commits, but sometimes I can not avoid deleting those extra spaces... For the long lines, I usually tend to ignore unless they are really big.
We have had several pep8 PRs already, perhaps it is time for a new one (not with the current one of course)

rhfogh:
Hi All,

I am not committing anything (yet), but the GlobalPhasing strategy 
calculation code should sooner or later be integrated with MXCube, so I 
would like to fit with the house style.

So, can I ask you for a bit of advice?

PEP 8, with four-space indentation and 79 character limit, can leave you 
short of space pretty quickly.

How would you recommend that one should indent this (which goes into 
column 86)?


class Payload(object): """Payload - top level message object""" _intent= None def __init__(self): # This class is abstract intent= self.__class__._intent
         if intentnot in MessageIntents: if intentis None: raise RuntimeError( "Attempt to instantiate abstract class 
Payload" ) else: raise RuntimeError( "Programming error - Payload 
subclass %s has incorrect intent: %s" % (self.__class__.__name__, intent) )

Or this one, which goes to line 106 (the endless dotted expressions are 
pretty unavoidable when you are effectively calling java code):


py4jElement= self.gateway.jvm.co.gphl.beamline.v2_unstable.domain_types.ChemicalElement.valueOf( anomalousScatterer.chemicalElement
     ) py4jEdge= self.gateway.jvm.co.gphl.beamline.v2_unstable.domain_types.AbsorptionEdge.valueOf( anomalousScatterer.absorptionEdge
) return self.gateway.jvm.co.gphl.beamline.v2_unstable.domain_types.AnomalousScatterer( py4jElement, py4jEdge
)


Yours,

Rasmus

mguijarr:
Hi Rasmus,

For the first one, I would just let it go over the 79 chars limit - I see it more as
a recommended limit, but nothing prevents us from writing longer lines ; I don't
think we are PEP8 zealots. Moreover there are plenty of code in MXCuBE that
doesn't respect any coding style still...

For the second example, I guess it is possible to use intermediate variables ?
Something like:

    bl_domain =  self.gateway.jvm.co.gphl.beamline.v2_unstable.domain_types
    py4jElement = bl_domain.ChemicalElement.valueOf(anomalousScatterer.chemicalElement)

I find it a good compromise.

marcus-oscarsson:
Hey Rasmus !,

As Matias said we are not so strict on PEP8, readability should always go first. Its in the end up to the author to judge what he or she thinks is more readable.  I think 79 columns is a good recommendation, it makes it easier to read and compare files when you have them next to each other (for instance when diffing).

I think Matias suggestion above is more readable than the original code, and I would probably do the same.

Marcus

amilan:
So, 
Finally #379 it's merged, now it's time to fix that one.

Cheers.

bolmsten:
@mguijarr can you add the issues you found when testing this PR here?

mguijarr:
Yes I am going to have a look at it now

mguijarr:
So... I made more tests and there is a good news : I was a bit quick yesterday because I believed some bugs
I found were related to this PR, whereas it's not the case - I realized bugs or usability issues were already
present before...

Then I created a bunch of issues to address those problems.  Ideally, they should be fixed in separate PRs.

@amilan : how do you deal with issues in the scrum ? I think priority should be given on fixing issues before adding new stuff. There is a small work to check what we really want with 'Manual Mount' mode - for me it should not have individual samples, it makes things complicated for no real benefit (correct me if I am wrong). 

Finally the only major regression I see with the new code in this PR is about Remote Access, as we talked about that yesterday. The other problem is about the AJAX calls business ; we also talked about that yesterday and there is a plan, so I am not worried anymore :)

meguiraun:
I must say I am a bit relieved :-) !! thanks for the issues

amilan:
Hi @mguijarr, 

I was writing here an answer about how can we deal with bugs, but it's quite extensive and I think it could be beneficial for the future to have it as a wiki page, so I have done that, you can find it in here:

https://github.com/mxcube/mxcube3/wiki/Dealing-with-bug-in-our-project

I think that could be a good approach for our project. Please let me know your opinion. 

Antonio.

amilan:
Mmm ... thinking a little bit, I don't see any effective way of reviewing a wiki page (i.e. adding comments ...) If you have a good method, let me know, if not, you can always edit the page with your comments and we review it later on, or we open another issue just to discuss the topic, or we move to emails ... choose whatever you prefer.

BTW, Thanks for reviewing and adding the issues, good to read that they are independent of this PR.

marcus-oscarsson:
Hello,

Thanks for the wiki page !, Ive read it trough quickly. Ill get back to you at a later time with
feedback.

Cheers,
Marcus

marcus-oscarsson:
Hey again !,

Ill start to look at putting back the per operation API, please let me know if there is something
that you would like me to consider while doing this. I'm thinking about if there is something that will
help you in your continued work with this PR. 

My idea to proceed with this would be to start from this branch and resolve some the API conflicts,
Ill make a PR with those changes, provided that you are not already doing this ?

Also let me know how its going with the "safeSend" ideas, is there perhaps already some library that
can do something like that. I might have some alternative ideas if it turns out that its complicated to implement or for some other reason turns out not to work.

Marcus

bolmsten:
@marcus-oscarsson  Hi Marcus I am already working on the merge so please wait with that work. Regarding the "safeSend" we can talk on skype

mguijarr:
Hello @bolmsten and @marcus-oscarsson :)
In fact I would like to participate to your conversation, if you don't mind - so let's use something else than Skype. What about using Slack? I am sending invitations now... If you don't agree with Slack just discard the mails.

mguijarr:
It seems like ESRF proxy doesn't allow websockets, so finally the Slack option is not good for now, as it really needs websockets... :(

amilan:
Ooohh, that's a pity, I was also in favor of using Slack. Well, let us know what other alternatives are available for you network.
Cheers.

PR #386, opened 2016-11-21
Context log

marcus-oscarsson:
Hey everybody

Initial draft of a context based log, (user message component). Its simply a component that can be attached to an element and be displayed next to it (top, bottom, left or right). The messages can be filtered in two different ways either by target and or with a custom filter function.

The messages are taken from the data structure userMessages in the general store, this can of course be changed if there is a more appropriate place.

The design of the messages needs to be improved a bit but I thought it was better to give you guys an idea of how this could work/look like before finishing it.

Cheers,
Marcus

### DIFF
--- a/mxcube3/ui/actions/general.js
+++ b/mxcube3/ui/actions/general.js
@@ -1,5 +1,40 @@
 import fetch from 'isomorphic-fetch';
 
+export function addUserMessage(record, target) {
+  let duration = undefined;
+  let level = 'INFO';
+  const message = record.message;
+  const details = record.stack_trace;
+  const meta = record.logger;
+
+  if (record.severity === 'ERROR') {
+    duration = 5000;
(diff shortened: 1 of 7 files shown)


marcus-oscarsson:
Hi !

So I fixed the CSS problem for chrome, so you should now be able to actually see the text :). Its a bit unnecessary to use this log as it is used today (to show queue execution), which is already done by the queue UI itself. It in anyways illustrates the concept, we can remove this "context log" at a later moment. It could possibly be useful for the queue when something goes wrong or external events manipulate the queue and the user needs to be made aware of such event.

The idea is that this can be extended so that we can have messages without popover just a simple overlay that is shown without the little arrow. The idea would also be to add the possibility to add controls to the message such as Accept, Cancel.

Cheers,
Marcus

PR #411, opened 2016-11-28
[WIP] - API fixes for PR #385

marcus-oscarsson:
Hi,

This is the PR that goes back to the per operation API based on PR #385.

Ive started to test this here, please let me know, if you like to continue with this one
or use #385.

Marcus

### DIFF
--- a/mxcube3/routes/qutils.py
+++ b/mxcube3/routes/qutils.py
@@ -14,6 +14,7 @@ from flask import jsonify
 from mock import Mock
 from mxcube3 import app as mxcube
 from mxcube3 import socketio
+from . import limsutils
 
 
 def node_index(node):
@@ -182,7 +183,7 @@ def get_node_state(node_id):
 
 def get_queue_state():
     """
-    Return the dictionary representation of the queue state.
+    Return the dictionary representation of the queue state plus some extra lists.
 
     :returns: dictionary on the form:
               {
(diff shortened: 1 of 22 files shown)


mxcube3/ui/main.jsx
```diff
@@ -70,7 +70,7 @@ export default class App extends React.Component {
   componentWillMount() {
     const persistor = persistStore(store,
            { blacklist: ['remoteAccess', 'beamline', 'sampleChanger',
-                         'form', 'login', 'general', 'logger', 'points'],
+                         'form', 'login', 'general', 'logger', 'points', 'queue'],
```
mguijarr: As stated during the meeting at MAX IV, I think this should not be blacklisted

mxcube3/routes/qutils.py
```diff
@@ -140,14 +140,140 @@ def queue_to_json_response(node=None):
...
+
+        sample_list.update({name: sample_data})
+
+    if mxcube.diffractometer.use_sc:
+        if len(sample_list) != 0:
+            sample_list = samples
+        else:
+            sample_list = {}
+        try:
+            basket, puck = mxcube.queue.mounted_sample.split(':')  #sample_changer.getLoadedSample().split(':')
```
mguijarr: Sorry to come back on this... As already said in previous review, we can't split in 2 mounted sample location, since location can have any number of containers.

mxcube3/ui/main.jsx
bolmsten: I know but I will leave this as it is until we have fixed the rest of PR. The reason being that this could hide issues with the init queue_ho.

meguiraun:
The reason for that previous line was to reformat the sample string from "2:5" to "2:05", to keep compatible. The modified lines do:
```
"1:31:3" -> "1:31:03"
"2:04:1:31:3" -> "2:04:1:31:03"
"1:03" -> "1:03"
```

mxcube3/ui/main.jsx
mguijarr: Ok.

mxcube3/routes/qutils.py
```diff
@@ -248,8 +248,8 @@ def get_queue_state():
         else:
             sample_list = {}
         try:
-            basket, puck = mxcube.queue.mounted_sample.split(':')  #sample_changer.getLoadedSample().split(':')
-            loaded = basket + ":{:0>2d}".format(int(puck))  # "1:2" -> "1:02"
+            res = mxcube.queue.mounted_sample.split(':')  # no matter how many ~containers
+            loaded = ':'.join(res[:-1]) + ":{:0>2d}".format(int(res[-1]))  # "2:1:2" -> "2:1:02"
```
mguijarr: Looks much better. Thanks :)

Issue #400, opened 2016-11-28
Queue stopped after disconnection


amilan:
**User story**
As a user, I want to get my queue stopped after 60 seconds of disconnection, so that will prevent for further issues.

**Acceptance criteria**
Give a timeout, 
When the client is disconnected, 
Then the queue is stopped and the client should receive a message log.

meguiraun:
Closing this one, #418 was accepted.

PR #418, opened 2016-11-30
Queue stopped after disconnection #400

marcus-oscarsson:
Hey all !

For feature #400

So Ive started to implement this one, The queue is stopped when connection to the server is lost. This is
done on the socket.io disconnect callback. A dialog is shown to the user when reconnecting informing her
that the queue was stopped. The user then needs to take the appropriate action either running the queue
or clearing it.

A second dialog (entirely client driven) for connection loss to server have also been added.

Cheers,
Marcus

### DIFF
--- a/mxcube3/remote_access.py
+++ b/mxcube3/remote_access.py
@@ -1,10 +1,16 @@
+import logging
+
 from flask import request, session
 from mxcube3 import socketio
 from collections import deque
 
+from mxcube3 import app as mxcube
+
+
 MASTER = None
 MASTER_ROOM = None
 PENDING_EVENTS = deque()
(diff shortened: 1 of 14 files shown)


mguijarr:
:+1:

bolmsten:
looks good

marcus-oscarsson:
Thanks !,

Its done now, this time Ive tested in both Chrome and Firefox

Ciao !

amilan:
👏 
Nice. We'll test it again tomorrow morning.

Thanks!

bolmsten:
The countdown is only there for show? It seems that you check continues for reconnection?

PR #419, opened 2016-11-30
do not start centring if no sample is mounted

meguiraun:
just display the error message modal and do not launch the centring.

### DIFF
--- a/mxcube3/ui/actions/sampleview.js
+++ b/mxcube3/ui/actions/sampleview.js
@@ -142,7 +142,6 @@ export function toggleCinema() {
 
 export function sendStartClickCentring() {
   return function (dispatch, getState) {
-    dispatch(showErrorPanel(true, 'dfjsadn.fjksa.'));
     const { queue } = getState();
     if (queue.current.node) {
       fetch('/mxcube/api/v0.1/sampleview/centring/start3click', {
--- a/mxcube3/ui/actions/sampleview.js
+++ b/mxcube3/ui/actions/sampleview.js
@@ -141,22 +141,27 @@ export function toggleCinema() {
 }
 
 export function sendStartClickCentring() {
-  return function (dispatch) {
-    fetch('/mxcube/api/v0.1/sampleview/centring/start3click', {
-      method: 'PUT',
-      credentials: 'include',
-      headers: {
-        Accept: 'application/json',
-        'Content-type': 'application/json'
-      }
-    }).then((response) => {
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
Hey, 

Thanks Mikel !

Cheers,
Marcus

mxcube3/ui/actions/sampleview.js
```diff
@@ -141,22 +141,27 @@ export function toggleCinema() {
...
-    }).then((response) => {
-      if (response.status >= 400) {
-        throw new Error('Server refused to start 3click');
-      } else {
-        dispatch(startClickCentring());
-      }
-    });
-  };
+  return function (dispatch, getState) {
+    dispatch(showErrorPanel(true, 'dfjsadn.fjksa.'));
```
marcus-oscarsson: Whats 'dfjsadn.fjksa.' ?

marcus-oscarsson:
Hi again,

In which context is this method used, I can imagine that its useful to be able to center even if there is no sample mounted. It should perhaps not be possible to collect but to center.

They scientists sometimes put special pins on the goniometer, these pins can for instance be used to find the beam or do other diagnostics. So they would probably like to center these pins, but I might be wrong.

Its a question for our respective scientists ...

Marcus

mxcube3/ui/actions/sampleview.js
meguiraun: I am learning swedish...

meguiraun:
At this moment you are able to center, and we you try to add a task you realize it is not possible since you have not mounted a sample. For me this is spending too much time for ending in nothing (you need to add a sample and then repeat everything).

I think we need the concept of sample, as a named and identifiable element. If a scientist puts a special pin they will be also able to easily find the associated data if we consider it as a sample (or whatever).

In any case, the aim of this pr is not to solve if we need a sample or not, we are facing this annoying problem everyday and this is a way to decreasing the annoyance level.

mguijarr:
We need to talk about this concept of sample in 'Manual mount' when there is no sample changer. I let this to the scientists. What we are missing is the 'free-pin' mode, which let people collect data quickly without having to create a 'sample'. For me Manual mount was free pin, in fact, but it's a different beast for you guys so we need an agreement from the product owners.

For the moment the code seems fine in this PR so it gets merged :)

meguiraun:
thanks, I think is good to let the product owners fight on this ;)

PR #420, opened 2016-11-30
Queue lockdown

### LINKED ISSUE #415: Queue API calls (continuation of #411)
Now that PR #411 has been merged, implementation of the solution for the AJAX requests agreed at MAX IV last week has to be done:

* once a call regarding queue operation is made, UI is "blocked" (Queue part is disabled / or there is some busy indicator)
* when the call is finished (ACK reply from server received), the UI is released for user operation

I would say priority is quite high, in order to finish the work started with PR #411.

bolmsten:
This PR aims to fix #415 . It shows a overlay over the queue while updating,deleting, adding, moving, this makes it impossible for the user to send more request.

### DIFF
--- a/mxcube3/ui/actions/queue.js
+++ b/mxcube3/ui/actions/queue.js
@@ -7,6 +7,9 @@ export function setSampleListAction(sampleList) {
   return { type: 'SET_SAMPLE_LIST', sampleList };
 }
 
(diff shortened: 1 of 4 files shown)


mguijarr:
Thanks Fredrick. Will test asap.

PR #422, opened 2016-12-01
Issue 398: parameter validation

meguiraun:
I am quite happy now!

* routes for getting limits on motor positions (physical or pseudos) and acquisition parameters
* form validation: if any parameter in forms {datacollection, char, helical} is wrong, the problematic input field is set as warning and the "run now" and "add to queue" buttons are disabled until the user types a correct value

Notes:
* missing motor inputs limits on the client: the current implementation already prevents the user for going of limits so I have not modify anything in this regard
* long time without writing javascript so the code quality can be improved
* I am not sure if it is fully compatible with all possible hwobj
* Not tested in real world: and will not be able to test soon...

### DIFF
--- a/mxcube3/ho_mediators/beamline_setup.py
+++ b/mxcube3/ho_mediators/beamline_setup.py
@@ -5,7 +5,7 @@ import logging
 
 BEAMLINE_SETUP = None
 
-# Singleton like interface is needed to keep the same referance to the
+# Sinleeton like interface is needed to keep the same referance to the
 # mediator object and its corresponding hardware objects, so that the signal
 # system wont cleanup signal handlers. (PyDispatcher removes signal handlers
 # when a object is garabge collected)
@@ -33,11 +33,11 @@ class _BeamlineSetupMediator(object):
     def getObjectByRole(self, name):
         try:
             if name == 'dtox':
-                ho = self._bl.getObjectByRole('resolution') # we retrieve dtox through res_hwobj
+                ho = self._bl.getObjectByRole('resolution')  # we retrieve dtox through res_hwobj
             else:
                 ho = self._bl.getObjectByRole(name.lower())
         except Exception:
-            logging.getLogger("HWR").exception("Failed to get get object with role: %s" % name)
+            logging.getLogger("HWR").exception("Failed to get object with role: %s" % name)
 
         if name == "energy":
             return self._ho_dict.setdefault(name, EnergyHOMediator(ho, "energy"))
--- a/mxcube3/ho_mediators/beamline_setup.py
+++ b/mxcube3/ho_mediators/beamline_setup.py
@@ -1,7 +1,11 @@
 # -*- coding: utf-8 -*-
 from mxcube3 import socketio
+from mxcube3 import app as mxcube
+
 from .statedefs import (MOTOR_STATE, INOUT_STATE)
 import logging
+from numpy import arange
+import math
 
 BEAMLINE_SETUP = None
 
@@ -501,9 +505,53 @@ class ResolutionHOMediator(HOMediatorBase):
     def stop(self):
         self._ho.stop()
 
-
     def state(self):
         return MOTOR_STATE.VALUE_TO_STR.get(self._ho.getState(), 0)
+
+    def _calc_res(self, radius, energy, dist):
+        current_wavelength = 12.3984 / energy
+
+        try:
+            ttheta = math.atan(radius / dist)
+            if ttheta != 0:
+                return current_wavelength / (2 * math.sin(ttheta / 2))
+            else:
+                return 0
+        except Exception:
+            logging.getLogger().exception("error while calculating resolution")
+            return 0
+
+    def get_lookup_limits(self):
+        energy_ho = BeamlineSetupMediator(mxcube.beamline).getObjectByRole('energy')
+        e_min, e_max = energy_ho.limits()
+        limits = []
+        x = arange(e_min, e_max, 0.5)
+
+        radius = self._ho.det_radius
+        det_dist = BeamlineSetupMediator(mxcube.beamline).getObjectByRole('dtox')
+
+        pos_min, pos_max = det_dist.limits()
+
+        for energy in x:
+            res_min, res_max = self._calc_res(radius, energy, pos_min),\
+                self._calc_res(radius, energy, pos_max)
+            limits.append((energy, res_min, res_max))
+
+        return limits
+
+    def dict_repr(self):
+        """
+        :returns: The dictionary representation of the hardware object.
+        """
+        data = {"name": self._name,
+                "value": self.get(),
+                "limits": self.get_lookup_limits(),
+                "state": self.state(),
+                "msg": self.msg(),
+                "det_radius": self._ho.det_radius
+                }
+
+        return data
 
 
 class DetectorDistanceHOMediator(HOMediatorBase):
--- a/mxcube3/ho_mediators/beamline_setup.py
+++ b/mxcube3/ho_mediators/beamline_setup.py
@@ -5,7 +5,7 @@ import logging
 
 BEAMLINE_SETUP = None
 
-# Sinlgeton like interface is needed to keep the same referance to the 
+# Sinlgeton like interface is needed to keep the same referance to the
 # mediator object and its corresponding hardware objects, so that the signal
 # system wont cleanup signal handlers. (PyDispatcher removes signal handlers
 # when a object is garabge collected)
@@ -37,7 +37,7 @@ class _BeamlineSetupMediator(object):
             else:
                 ho = self._bl.getObjectByRole(name.lower())
         except Exception:
-            logging.getLogger("HWR").exception("Failed to get get object with role: %s" %name)
+            logging.getLogger("HWR").exception("Failed to get get object with role: %s" % name)
 
         if name == "energy":
             return self._ho_dict.setdefault(name, EnergyHOMediator(ho, "energy"))
@@ -57,18 +57,18 @@ class _BeamlineSetupMediator(object):
             return self._ho_dict.setdefault(name, DetectorDistanceHOMediator(ho, "dtox"))
         else:
             return ho
-       
+
 
     def dict_repr(self):
         """
         :returns: Dictionary value-representation for each beamline attribute
-        """        
+        """
 #        capillary = self.getObjectByRole("capillary")
 
         data = dict()
-        
+
         try:
-            energy =  self.getObjectByRole("energy")
+            energy = self.getObjectByRole("energy")
             data.update({"energy": energy.dict_repr()})
         except Exception:
             logging.getLogger("HWR").exception("Failed to get energy info")
@@ -208,7 +208,7 @@ class HOMediatorBase(object):
         """
         data = {"name": self._name,
                 "value": self.get(),
-                "limits":self.limits(),
+                "limits": self.limits(),
                 "state": self.state(),
                 "msg": self.msg()}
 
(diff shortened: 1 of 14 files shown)


mxcube3/ho_mediators/beamline_setup.py
```diff
@@ -5,7 +5,7 @@
 
 BEAMLINE_SETUP = None
 
-# Sinlgeton like interface is needed to keep the same referance to the 
+# Sinlgeton like interface is needed to keep the same referance to the
```
mguijarr: Typo: Sinlgeton -> Singleton

mxcube3/ho_mediators/beamline_setup.py
```diff
@@ -282,6 +282,20 @@ def stop(self):
...
 
+    def limits(self):
+        """
+        :returns: The energy limits.
+        """
+        try:
+            energy_limits = self._ho.getEnergyLimits()
+        except (AttributeError, TypeError):
+            energy_limits = (0, 50)
+            logging.getLogger("HWR").exception("Failed to get get object with role: %s" % name)
```
mguijarr: Typo: get get -> get

mxcube3/ui/actions/general.js
```diff
@@ -100,6 +108,14 @@ export function getInitialStatus() {
         'Content-type': 'application/json'
       }
     });
+    const beamlineLimits = fetch('mxcube/api/v0.1/beamline', {
```
mguijarr: As far as I understand it, this is not only getting limits but the whole dictionary representation of the `BeamlineSetupMediator` object. So maybe a more appropriate variable name like `beamlineSetup` would be better ?

mxcube3/ui/actions/general.js
```diff
@@ -152,11 +176,15 @@ export function getInitialStatus() {
     const pchains = [
```
mguijarr: As I am making the review, I looked at all those calls... I think we should do some refactoring because over the internet an AJAX call is never less than 100 ms (more like 200 ms in fact) and this initialization will take too much time in my opinion. The idea could be to group some of those ; what about having `motorLimits` with `motors`, and `acqParametersLimits` + `dcParameters`  with `beamline(Limits/Setup)`, or `dataPath` with  `queue` (this is only suggestions).

mguijarr:
In addition to my previous remarks, there is one little problem : resolution limits are sent from server to client once at initialization, so far so good, except that resolution limits depends on energy ; resolution limits update as to be implemented after energy changes too.

mguijarr:
Apart from that, it looks good to me :) Thanks Mikel.

I didn't look at the JS part, I let Marcus have a look !

mxcube3/ui/actions/general.js
meguiraun: ok. Just one issue: the diff motor info is done through two different calls for the position and the limits, whereas the beamline give position and limits with a single call. So, later in the js code there is a motorLimits var which contains diff motors and beamline limits plus beamline positions.

mxcube3/ui/actions/general.js
meguiraun: damm boy-scout rule...

marcus-oscarsson:
Looks fine to me too, great work !.

I had a smaller problem when I tried to run it though. I get props.motorLimits.energy.limits is Null
am I perhaps missing some value in a XML file ?

Marcus

meguiraun:
the energy hwobj mockup shoud give you the limits, but, only after one of the latest additions from Ivars, check the method please.
This pr also updates the mockups xmls

marcus-oscarsson:
I think I found the problem, I'm using the latest 2.2 of hardware objects and the method

[get_energy_limits(self)](https://github.com/mxcube/HardwareObjects/blob/2.2/EnergyMockup.py#L42)

returns None

Could that be the issue ?

meguiraun:
yes, it is only added in master

marcus-oscarsson:
Aha, that's why then. So we need to "backport" the EnergyMockup to 2.2 before accepting this PR.

Can I leave that up to you ?, actually most additions on mockups to master should also have a corresponding pr for 2.2. If its not braking already existing functionality of course.

Marcus

meguiraun:
yep, I will do that after addressing all your comments. Not sure if it will be today... beamline is calling...

mguijarr:
The idea behind branches was to have stable code to rely on ; so I am not very found of backporting anything from `master` to `2.2`. The only code that should be added in `2.2` is for **bug fixes**. We also said we won't change hardware objects while developing MXCuBE 3 ; those changes will happen at a second stage (like, MXCuBE 3.1...). It may be frustrating but this is the key to a successful deployment and adoption of MXCuBE 3 at places which are already using MXCuBE 2, since it guarantees the existing interface can still be used in case of problem with the new version and it limits the potential bugs.
We already diverged from this rule, but I am not happy with it at all since it makes the development more difficult ; adding stuff without removing older code in a branch is *tolerated* because sometimes we have no choice, but it should stay an exception not a rule and provided it doesn't break anything.

Originally MXCuBE 3 should have relied on `2.1` Hardware Objects *as they were*. It was decided by MAX IV to change to `2.2`, ok, but now please use `2.2` objects as they are without making more changes. 

The new code we write should work on `2.2` as it is today.

mxcube3/routes/Utils.py
```diff
@@ -72,24 +89,60 @@ def get_movable_state_and_position(item_name):
                 else:
                     pos = None
             else:
-		pos = hwobj.getPosition()
+                pos = hwobj.getPosition()
                 # invalid code: no arbitrary rounding, precision depends on motor
 		#try:
```
amilan: +1

meguiraun:
It was meaning only to return something in the mockups when getting the energy limits, I consider it as a bug that was only fixed in master. If you do not agree with this I can add some logic to this pr to avoid errors if there is no limits.

meguiraun:
regarding the resolution limits dependency on the energy, the limits are only updated when a new energy has been applied. So, I would need to implement the calculations that are done in the resolution hwobj in the validation js code, and pass the energy that the user wants to apply. But, I will need to send a bit more information in the beamline dict such as the det_radius.
Would this be ok? Do you have any other suggestion for this?

mguijarr:
@meguiraun, about resolution limits: the way you suggest would work, but calculations in js code should be avoided as much as possible because it introduces business logic in the UI, which is bad. My advice would be:
* make motors sending their limits in the state update event, this would automatically adjust limits for resolution motor when energy is moved (indeed, resolution has the same state as energy, so when energy goes from MOVING to READY, the same goes for the resolution)

mxcube3/ui/actions/general.js
mguijarr: We can add another issue for that, if you want - but as you were adding more calls yourself... Do as you prefer for now, but if it is not solved in this PR we can add another issue for this I think.

meguiraun:
Yep, I do not really like the idea of calculations in the client either. 

The problem that I see is that on the modal, the user will set an energy that (potentially) will be different compared to the one in the beamline, but the resolution limits will be always referring to the one in the beamline and not to the one being written in the modal.

I can do as you suggest, but there will be a discrepancy anyway when the user configures a data collection with a new energy.

mxcube3/ui/actions/general.js
meguiraun: I would rather open another issue and keep this one focused on its topic. I can take care myself of course.

mguijarr:
@meguiraun, well spotted... :) Damned. In fact this bug exists in MXCuBE 2 as far as I can tell...
For this specific one, then, in addition to what has already been said I would make an extra validation step when user clicks on 'Add', with a round trip to the server. The calculation stays on the server.

mxcube3/ui/actions/general.js
mguijarr: What do you suggest to make all this more clean?

meguiraun: I can add new route only for the limits, but thinking on it it is better to add the limits on the motors dictionary as well. Returning always {pos, state, limits} no matter if it is a diffractometer motor or a beamline movable.

marcus-oscarsson:
Hey Mikel,

There is only a small problem with that and that is that the concrete implementation does not have the method get_energy_limits. There is however a method called [getEnergyLimits ](https://github.com/mxcube/HardwareRepository/blob/master/HardwareObjects/Energy.py#L85). So its all fine and can be considered a bug fix if you call the method getEnergyLimits :)

Its not so easy to know all this, the current situation with the hardware objects seems a bit complicated :).

The work on the abstract classes have been on going for a while now and was unfortunately not completely finished when 2.2 was created. Furthermore many of the abstract classes only have a concrete implementation one institute (if any). We need to have more exchange before we can consider the abstract classes completely stable. In fact there does not event seem to be an abstract definition of the energy.

I created the ho_mediators module to be able to handle this situation in someway. The idea is basically to gather compatibility logic, some of which will later be moved to 2.3 (at the time 2.2 because we were still using 2.1). The motivation is as Matias said to keep 2.2 (2.1) stable until the UI is ready. We are of course aware that there are things that could be things done in a better way and we would really like to improve the hardware objects as well, but everything has its time and place :).  So you could also use ho_mediators, or let me know if you have another idea ?.

However, in this case,  I think its enough to rename the method to get getEnergyLimits.

Marcus.

package.json
```diff
@@ -50,7 +50,8 @@
     "socket.io-client": "1.4.5",
     "style-loader": "0.12.3",
     "underscore": "^1.8.3",
-    "x-editable": "1.5.1"
+    "x-editable": "1.5.1",
```
mguijarr: I think `x-editable` here is a mistake, it has been removed some time ago if I remember well.

meguiraun: woowow, that was fast!
mmm, we have it in master
https://github.com/mxcube/mxcube3/blob/master/package.json#L53

meguiraun:
Well, I cannot say that I am extremely happy with this implementation (although it works). In the beginning I was more in favor of client-side calculation but this lookup-table alternative seems to work fine, and it was a funny exercise.

The server creates the lookup table on init, taking energy limits and calculating corresponding resolution limits. At this moment the step size of the array is of 0.5 keV. 
Then, the validation of the modal does the interpolation calculation on each change of any input (that's how the redux-form work), it looks a bit overkill but from a user experience point of view, I did not see any lag. The javascript code could move the array mapping somewhere else and do it once... but I do not know where&how :(

In addition, as a surprise cof cof, this will not work on *any* mockups... since the resolution mockup does not have the `radius` implemented.

meguiraun:
I removed the WIP in the title. Let me know your thoughts about this pr!
We can talk about the missing code in the 2.2 mockups during our meeting this week?

marcus-oscarsson:
Hey !

Ive already made a PR for the missing ISPyBRestClientMockup

Marcus

package.json
mguijarr: Ok so I didn't remember well, then. I will remove it since we don't use it.

marcus-oscarsson:
Hey !, 

Ive added the missing ISPyBRestClientMockup

Marcus

mguijarr:
@meguiraun I accepted your PR in HardwareObjects, so now the radius issue is solved ?

mxcube3/ho_mediators/beamline_setup.py
```diff
@@ -1,11 +1,15 @@
...
+
 from .statedefs import (MOTOR_STATE, INOUT_STATE)
 import logging
+from numpy import arange
+import math
 
 BEAMLINE_SETUP = None
 
-# Sinlgeton like interface is needed to keep the same referance to the 
+# Sinleeton like interface is needed to keep the same referance to the
```
marcus-oscarsson: :)

meguiraun: arrrgggggghhhhhhh!!!!!!!!!!!!!!!!

mxcube3/routes/qutils.py
```diff
@@ -285,6 +285,10 @@ def _handle_dc(sample_id, node):
     parameters.pop('centred_position')
     queueID = node._node_id
     enabled, state = get_node_state(queueID)
+    try:
+        lims_data = mxcube.rest_lims.get_dc(node.id)
```
meguiraun: I can remove, but it is also removed from  #424  (done just now)

meguiraun:
With the radius in the mockup this PR should work out of the box!

PR #424, opened 2016-12-02
Small fixes

bolmsten:
When testing down in the beamline I did some hotfixes that should now be implemented into the master. See commit history for more info.

### DIFF
--- a/mxcube3/routes/qutils.py
+++ b/mxcube3/routes/qutils.py
@@ -286,11 +286,6 @@ def _handle_dc(sample_id, node):
     parameters.pop('centred_position')
     queueID = node._node_id
     enabled, state = get_node_state(queueID)
-    try:
-        lims_data = mxcube.rest_lims.get_dc(node.id)
-    except Exception, e:
-        lims_data = ''
-
     res = {"label": "Data Collection",
            "type": "DataCollection",
            "parameters": parameters,
@@ -299,7 +294,7 @@ def _handle_dc(sample_id, node):
            "queueID": queueID,
            "checked": enabled,
            "state": state,
-           "limstResultData": lims_data,
+           "limstResultData": mxcube.rest_lims.get_dc(node.id),
            }
 
     return res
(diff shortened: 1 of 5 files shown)


marcus-oscarsson:
Thanks !

mxcube3/routes/qutils.py
```diff
@@ -286,6 +286,11 @@ def _handle_dc(sample_id, node):
     parameters.pop('centred_position')
     queueID = node._node_id
     enabled, state = get_node_state(queueID)
+    try:
+        lims_data = mxcube.rest_lims.get_dc(node.id)
```
marcus-oscarsson: There should probably not be a try except here since get_dc should return '' or similar
when we cant get data from lims. I guess you get an attribute error now because the
method does not exist on your lims object. ?

marcus-oscarsson:
I just have the small comment above, I don't mind accepting this PR as this but the call to get_dc should
not produce an exception. Exceptions should in my opinion be handled within get_dc, or what do you think ?

mxcube3/routes/qutils.py
meguiraun: the mockup lims hwobj does not have such method, so we added the try/except

mguijarr: @marcus-oscarsson I think the mockup lims hwobj should be of the 'ISPyBRest' type (correct me if I am wrong). This is why it is `mxcube.rest_lims`. So the `try...except` should not be needed, provided the underlying objet is the right one. So what are the steps to take ? Adding a new `ISPyBRest` mockup ? Fixing the current mockup to have a `get_dc()` ?

marcus-oscarsson: I understand that you might want to handle it in this way, but the real problem is that the mockup is missing a method. So this is something that should not happen, we should add a new mockup. Ill do that as soon as I can since I already said that I would do this.

So ill add a mockup that we can use that have the method.

meguiraun: removed the exception handling!

mguijarr:
New ISPyBRestClient has been added by Marcus in HardwareObjects, with `get_dc()`.

mguijarr:
It would be wonderful to add the submodule update in this PR, and the fix in xml files in order to have `mxcube.lims_rest` pointing to the right object.

meguiraun:
ok, gonna update those

meguiraun:
let's see now.

Issue #398, opened 2016-11-28
Prevent wrong parameters to be applied


amilan:
**User story**
As a user, I don't want to be able to apply wrong parameters for beamline specifics, so that will be a cause of problems.

**Acceptance criteria**
Given a configuration file with parameters, 
When I write wrong values, 
Then the interface will not allow me to write them and will notify me.

beteva:
Please consider in the acceptance criteria that the parameters might be in the hardware configuration (e.g. motor limits) and not only in the configuration file.

meguiraun:
Closing this one since #422 was accepted and merged. See the PR discussion for more details, as a summary:
* if any parameter in forms {datacollection, char, helical} is wrong, the problematic input field is set as warning and the "run now" and "add to queue" buttons are disabled until the user types a correct value
* limits come from hwobjs (moveables and parameters)

PR #430, opened 2016-12-08
Issue 425: optimize initialization calls

meguiraun:
I removed three calls from the last PR (#398). Merging the following:
* Motors and limits
* beamline attributes and datapath
* Default parameters and param limits

Any further improvement might start messing around with the redux-state or with the return dict of current api... so I am a but cautious with that. Possible enhancements:
* Add the saved centred positions into the queue_state
* beam_info with the beamline

You can either accept this PR as it is, or we keep it on hold and I ask Fredrik for help with the last two enhancements (delaying this pr a few days)
______
Solving https://github.com/mxcube/mxcube3/issues/425


mguijarr:
Thanks a lot Mikel it is much nicer now than before :) We will see if we need further optimizations.

meguiraun:
Thanks!

Issue #394, opened 2016-11-25
Manual mount: Javascript blocked when adding sample in grid


mguijarr:
I noticed once every 3 attempts that Firefox becomes unresponsive and shows the 'Stop script' dialog when adding a new sample in Manual Mount mode.

mguijarr:
This cannot be reproduced in Chrome.
This cannot be reproduced when using production bundles.

mguijarr:
Fixed in #437

Issue #393, opened 2016-11-25
Manual mount: last sample automatically selected


mguijarr:
I think 'Manual mount' should not have samples (see #389). However at the moment there are samples.

When adding a sample, the last one is automatically selected in queue - it doesn't show as 'Upcoming sample'.

Issue #387, opened 2016-11-25
Task final state is not updated in Sample View page (queue)


mguijarr:
Once a task is executed, it turns "blue" ; however, the final state is not properly reported in the queue. It should be "green" if successful, or "red" (light red) if an error occurred.

With PR #385, the internal state on the server seems to be updated, because reloading the client shows executed tasks as "green". However, it is always green (even if the task could not be performed, or had an error).
With current master, the task state does not seem to be reported at all ; I guess it is missing a signal somehow.

meguiraun:
The failure with the last signal could be due to a crash in the `signals.collect_oscillation_finished/failed` methods, I saw if crashing if `mxcube.rest_lims.get_dc` does not exists. thus not reporting the final state.

I will check the state when reloading (#385), it should not be always green...

mguijarr:
Thanks Mikel.

I also saw this `.get_dc` problem - let's see if I can find where it comes from...

meguiraun:
I think we started using the rest_lims recently, mockups xml files not updated?

marcus-oscarsson:
Hey !

Yeah, that's probably my bad ! Because there is no proper mockup for the ISPyBRestClient so we are using the ISPyB2ClientMockup by default. We added the method get_dc fairly recently so it have been working fine until now. It will of course still work fine if we are using the "real" hardware objects configured properly.

I will take the time to make a ISPyBRestClientMockup to handle this !

Ciao !

mguijarr:
More news on this : I tried yesterday on a beamline, with the right ISPyBRestClient - the problem is still there. So it doesn't come from `.get_dc`.

mguijarr:
Fixed in #437

PR #432, opened 2016-12-12
[Do not merge ! Contains things to discuss] Fixes (#393, #394, #387) and questions

mguijarr:
Hi all,

While working on the snapshots feature on ID30A3, I stumbled into issues I try to fix in this PR. Maybe some things are not bugs but features, although I find some things very confusing. 

So please look at my changes and commits and let's review all this :)

The more important questions I have are related to Manual Mount mode: how is the Manual Mount
mode supposed to work ? How is it supposed to work with the Grid ? What a user is expected to do ?
My biggest concern is that I found Manual Mount being something in-between a manual (human) sample
changer implementation within the queue, and the 'free-pin' mode in MXCuBE 2.
Then, I started to have a deeper look at `dffractometer.use_sc` ; my goal is to remove this since it is unclear
what it is supposed to do.
Finally I looked at the sample grid, I realized it was being updated far too much times and I fixed it, at least temporarily (see last commit).


JieNanMAXIV:
Hi, 

Regarding manual mount and sample changer, in case we have different opinions on how it should work, below is how we think,
There should be two modes, Manual mount and Sample Changer, and both should in theory support pin and plate operation, depending on the sample changer, diffractometer and etc.  UI should automatically display either the normal pin interface or the plate manipulator interface, depending on the beamline hardware setup (in our case the diffractometer gonio head).
 
For manual mount, 
1. normal pin. Sample Grid is a great feature for the overview, but for collection experiments, users should mainly work on the SampleView interface, as there is no need to switch back and forth between the two views. Users can simply change to the next sample by click "next sample" (of course they need to mount the sample manually), which will clear all points, lines, collection tasks from the SampleView interface. The collection tasks from previous samples can still be reviewed in the SampleGrid.  
Users can use the queue feature to launch several collection tasks for the mounted sample (run sample).
2. plate, to be decided. But in general, the interface for sample grid could be similar to the sample changer - pin interface, maybe with a presentation of the plate!?. On the sampleView, we should have a plate navigator for users to move to a selected drop position. 

For sample changer,
1. normal pin, as it is now, which we already have a consensus.
2. plate, to be decided, which should allow users to change plate with Sample changer.

Perhaps we should reach a consensus first.  If not, we can continue the discussion next Tuesday on Skype.

mguijarr:
I find it contradictory to have 'Next sample' for Manual Mount - wouldn't it be better to call it 'Clear queue' ?
What I find disturbing for users is that 'Next sample' refers to the idea of a list of samples, and unless I thought the set of available samples were coming from the Grid. 
Moreover, there is the dialog box that prompts for sample name and protein name: this can turn users into thinking the sample will appear in ISPyB, but it's not true : it's just a free pin sample mounted manually (right ?).
So, I would consider those options:
- either we do a human sample changer thing, with samples data in ISPyB and standard sample changer management (loading / unloading etc)
- or we simplify the Manual Mount : there can be only 1 sample, whatever is mounted on the gonio, and tasks are added to it - people can clear the tasks whenever they want.
I think the second option makes things simpler for users.

Regarding plates, I agree with you. But there is one thing that puzzles me: are you going to mount plates manually ?

PR #434, opened 2016-12-14
Issue 416

meguiraun:
solving https://github.com/mxcube/mxcube3/issues/416, when adding a task, if the sample is not in the queue already we add it. Also, better addition of multiple tasks/samples

### DIFF
--- a/mxcube3/routes/qutils.py
+++ b/mxcube3/routes/qutils.py
@@ -512,10 +512,10 @@ def queue_add_item(item_list):
 
     Each item (dictionary) describes either a sample or a task.
     """
-    current_queue = queue_to_dict()
 
     for item in item_list:
         item_t = item["type"]
+        current_queue = queue_to_dict()
 
         # Is the item a sample, then add it and its tasks. If its not, get the
         # node id for the sample of the new task and append it to the sample
--- a/mxcube3/ui/actions/queue.js
+++ b/mxcube3/ui/actions/queue.js
@@ -511,7 +511,6 @@ export function addTaskAction(tasks) {
 
 export function addTask(sampleIDs, parameters, runNow) {
   return function (dispatch, getState) {
-
     const tasks = sampleIDs.map((id) => (
                   { type: parameters.type,
                   label: parameters.label,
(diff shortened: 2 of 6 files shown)


marcus-oscarsson:
Hey !

Thanks great work this really makes it easier to use the grid.  I noticed a couple of things here the first one a bit more serious than the other.

- I just added a some tasks on a couple of tasks and then used the pick samples checkbox to pick some others. When doing a browser refresh the order of samples, the number in the bottom right corner, was not at all the one same as before the refresh.

- There are some linting errors to fix :)

- Maybe not directly related to this PR, but it seems take a really long time to remove samples, do you experience the same ?

The first one seems a bit more serious and could perhaps be treated as a separate bug, like fixes after #411. What do you think ?
 
Thanks
Marcus

meguiraun:
I do not understand your first issue, or perhaps I do not see the same.

It is true that it takes time to remove a bunch of samples, it is because for each call to remove, a call to the server is done. At this moment, the server and client are not prepared for removing 'items' in a batch mode, only batch addition.

marcus-oscarsson:
Hm, ill try to give you some screenshots.

I noticed this by coincidence when I was testing and it was fairly easy to reproduce so you should be able to see it. Ah, one thing Im running the client and the server on two different computers, which might make some difference. 

Let me know if you are able to reproduce it otherwise ill try to provide you with some screenshots later today.

meguiraun:
Hej @marcus-oscarsson, I see it now.
When adding a sample (pick checkbox) it is added at the end of the queue, when recovering on refreshing the number order is defined by `sample_order` list, which is not the same as the bottom number before refreshing.
I was expecting that the order of each sample on the queue corresponds to the order on which were added, which does not correspond with the bottom number.

marcus-oscarsson:
Hej,

Wow, I was just about to make the screenshots so its good that you noticed. 

We are a little bit between two solutions now since #411 so representation of the order might be duplicated. The numbering of the samples and sample_order should match, since the numbering is the order in the queue, I guess ?

In anyways, the number in the bottom right corner is the order of the sample in the queue. This number is changed when the sample is moved around in the grid. The numbering starts from the top left and goes to the bottom right and should not be dependent on picking order. I realize that this is maybe not so evident so let me know if there is something I can help you with.

Marcus

meguiraun:
(sorry for coming to this late)
I would expect that the number is dependent on the picking order, and the order is given by the `sample_order` list. Perhaps we need a `add_sample_at_index` and implement the reorder methods as we did for the tasks.

In any case, I would raise this issue to our product owners and let them decide together will the auto/manual operation discussion, I think is related discussion.

PR #436, opened 2017-01-09
Removal of `.use_sc` flag

mguijarr:
(continuation of the work started in the -now closed- PR #432)

This PR removes an unused route, and gets rid of the `use_sc` flag, as we discussed briefly during our
last video conference before xmas.

Please find more details in the commit messages.

### DIFF
--- a/mxcube3/routes/Diffractometer.py
+++ b/mxcube3/routes/Diffractometer.py
@@ -56,18 +56,6 @@ def set_phase():
     return Response(status=200)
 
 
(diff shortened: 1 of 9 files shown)


mguijarr:
I think we should merge this, and deal later with the `mounted_sample` question.
We are soon going to send another PR which is based on this one and the one that have been merged this morning.

meguiraun:
Ok, just for this time ;) ... and we fix together the next week.
It breaks the following: when the client request the state of the queue it will see as if no sample is mounted event it is, I will open an issue so we do not forget.

PR #438, opened 2017-01-10
Unittesting server

meguiraun:
hoooooray!!!

A very minimal set of tests: add/get/delete sample, hardcoded sample data dict, and those three tests are related to each other and act on the same queue... at least is something!

Be careful: check the path modification in the `__init__.py`, I do not know if the change affects you ... if so I can find alternatives

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -16,7 +16,7 @@ opt_parser = OptionParser()
 opt_parser.add_option("-r", "--repository",
                       dest="hwr_directory",
                       help="Hardware Repository XML files path",
-                      default=os.path.join(os.path.dirname(__file__), 'HardwareObjects.xml/'))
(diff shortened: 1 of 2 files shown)


amilan:
Yuhuuuuuuuuu! It's a beginning! 👍 
I will have a look later on.
Thanks!

mguijarr:
It's a pleasure to see unit tests make you so enthusiastic guys ;)

meguiraun:
enthusiastic, enthusiastic...

mguijarr:
@meguiraun : if we are going to be serious about unit tests, maybe we should consider using [pytest](http://doc.pytest.org/en/latest/). It's a good start :) Now we need to write more test cases...

meguiraun:
thanks, gonna have a look at that

PR #440, opened 2017-01-13
(big) refactoring and sample grid+queue layout redesign

mguijarr:
Here is the final step to the work started a few days before xmas. Marcus and I did some refactoring
and UI layout changes, we hope you'll enjoy :)

On the overall it disambiguates the grid and queue roles at the code level and removes a whole bunch
of unused/dead code. It gives more clear names to some actions/functions. It also renames some files,
and adds some constants instead of free numbers or strings both on the client and server side.
See commits for more details.

Regarding the UI layout, we would like to propose some changes to make it better for users ; in
particular, when dealing with manually mounted samples vs. sample changer. The changes have been
approved or inspired by Daniele, just to say we have a strong opinion to have the layout evolving
towards the direction proposed by this PR. Of course, we didn't remove any of the features that were
originally present, we just arranged things differently to this respect.

We can't wait for your feedback :)

We already have more UI and code changes in mind, that we will talk about next week probably.

The PR can be merged ; we are aware of some bugs, but nothing too serious normally you can test
it on your side without problems.


meguiraun:
Hej guys, thanks for all of this work!!! Clearly the code improves a lot!

We are on a slightly different opinion regarding how the manual samples interface work, but I think is better if we discuss face to face next week. So I am happy to accept this great work as it is now.

Just one thing, we noticed that we can not add tasks to samples in the datacollection view (it comes from a previous PR), @bolmsten has a fix for that (hopefully later today a new PR will appear)

PR #442, opened 2017-01-13
fixes

bolmsten:
Fix for making it possible to show data collections from context menu and change button in taskitem.

I can not delete a task but this seems to be a server bug


mxcube3/ui/components/SampleView/ContextMenu.js
```diff
@@ -40,14 +40,14 @@ export default class ContextMenu extends React.Component {
 
   showModal(modalName) {
     const { sampleId, defaultParameters, shape, sampleData } = this.props;
-    const node = sampleData[sampleId];
+    console.log(sampleData);
```
mguijarr: would be great to remove the `console.log`

PR #447, opened 2017-01-23
Refactored queue, now only storing sample ids in queue

marcus-oscarsson:
Hey !

I have refactored the queue so that its only an array of sampleIDs, as we discussed during the meeting.

Please take a look, Ive done the basic testing here and it seems to work, but I might have forgotten something.

Cheers,
Marcus


meguiraun:
thanks for this, I have not tested but looks fine.

Issue #392, opened 2016-11-25
Forbid removal of past data collections


mguijarr:
Already collected tasks are part of history, and cannot be changed or removed.

mguijarr:
We also said:
- [ ] remove Change button, a click on Parameters link will show the parameters instead
- [ ] ensure past data collections cannot be changed
- [ ] Delete button to be moved as a cross at the top right corner
- [ ] link to Results as a text label instead of a button (more compact)

marcus-oscarsson:
Hey, 

So I did something for the delete buttons, Ill leave the redesign to you guys. Or just let me know if you would like me to do it ...

Marcus

amilan:
Hi, 
I'm cleaning the board and this issue appears as done, shall we close the issue? or is there anything missing?

marcus-oscarsson:
Hey !

I only did a part of it since some other parts involved some redesign of the queue UI. We discussed this during the meeting and It seemed to me that it was more Fredrik or Mikel that were going to take care of the queue UI part. So remaining to-do is:

- [ ] remove Change button, a click on Parameters link will show the parameters instead
- [ ] Delete button to be moved as a cross at the top right corner
- [ ] link to Results as a text label instead of a button (more compact)

But I might of course have misunderstood, just let me know if you think I should take of it

Marcus

meguiraun:
hej, you are right, it is on our desk.

meguiraun:
see https://github.com/mxcube/mxcube3/pull/462

marcus-oscarsson:
So I consider this done, right Mikel ? If not please re open it

PR #448, opened 2017-01-23
Fix for #392 (Forbid removal of past data collections)

marcus-oscarsson:
Hey !

Fix for #392 (Forbid removal of past data collections)

 - Removed the delete controls when a task is in a collected state ( state != TASK_UNCOLLECTED)
 - Preventing delete in action if state is other then TASK_UNCOLLECTED, (not sure that's actually needed)

I did not work anything on the layout of the TaskItem since we discussed some changes to it when you were here. I'm only hiding the delete button for the moment, Ill let you guys do the redesign as you like. Or just let me know if you prefer that I do it ...

Cheers,
Marcus


meguiraun:
great @marcus-oscarsson, thanks for this

PR #453, opened 2017-01-30
Grid improvements

marcus-oscarsson:
Hey,

- Replaced isotope with stone-cutter and removed isotope from project related files

- Re-factored SampleGrid, SampleGridItem and added an additional component TaskItem. Renamed SampleGridContainer to SampleGridViewContainer and created a new container from the component SampleGrid and renamed it SampleGridContainer.  ....  :)

- Worked on the usability as discussed during the MxCuBE meeting, fixed context menu, selection, and other smaller things that was broken due to previous PRs.

Cheers,
Marcus


meguiraun:
Hej @marcus-oscarsson. Thanks for this.

I am having some trouble running it. From a fresh clone:
```
ERROR in ./mxcube3/ui/containers/SampleGridContainer.jsx
Module build failed: SyntaxError: Only one default export allowed per module. (674:0)

  672 | }
  673 |
> 674 | export default connect(
      | ^
  675 |   mapStateToProps,
  676 |   mapDispatchToProps
  677 | )(SampleGrid);
```

Looks like it wors if I remove `export default` from the `class SampleGrid extends React.Component .....`

meguiraun:
Also, all the samples are coloured form the beginning, is this intentional? Also, I notice that picking samples is quite slow, it could be npm development issue.

image

meguiraun:
ahem... not to be very picky but...
```  queue actions
    ✓ should create an action to clear queue (3ms)
  queue reducer
    ✕ should return the initial state (8ms)
    ✓ should handle CLEAR_QUEUE

Test Suites: 1 failed, 1 total
Tests:       1 failed, 2 passed, 3 total
Snapshots:   0 total
Time:        3.114s
Ran all test suites.
npm ERR! Test failed.  See above for more details.
```

marcus-oscarsson:
Hey,

Thanks for your feedback I fixed the default export. I think we have different versions of Babel and that's why its working for me and not for you. What version of Babel are you using ?, I'm using the one that we get with npm install, but its perhaps not updated properly. In anyways that should be fixed now.

The samples should not be green, that seems a bit strange, maybe it works better when export is properly defined.

For the test, the test fails because of a change in a previous PR #447 I guess. So I fixed the test so it should now pass. A comment on the test, it actually tests two reducers the queueGUI and the queue but that's perhaps the intention ?

meguiraun:
Thanks for the fixes.
I am using babel from `package.json`... 
I still see the green cards from the beginning, I pull again, restart browser... I do not see any error in the console at all.

marcus-oscarsson:
Hm, did you restart the backend, because there is a update to the code that retrieves the samples.

See [this line] (https://github.com/marcus-oscarsson/mxcube3/blob/d40ad17eb4d5467cbf6a10814715549a1f65ecb2/mxcube3/routes/SampleChanger.py#L32)

meguiraun:
oppps... my bad. Everything seems to be fine now!!

PR #454, opened 2017-01-31
Filter

marcus-oscarsson:
Hi 

Based on #453, so merge it first ! The two first commits should then disappear !

Improved filter function, extended reducer with filterOptions object that replaces the old filterText string.

Marcus


meguiraun:
can you fix the git history, I still see the ones from the previous pr...

marcus-oscarsson:
Okey, its actually quite strange that its not done automatically in anyways its done now !

meguiraun:
great!! thanks

Issue #445, opened 2017-01-20
Top bar redrawing


meguiraun:
Redesign the top bar according to attached image

meguiraun:
![20170120_100427](https://cloud.githubusercontent.com/assets/12758327/22145051/14726902-df00-11e6-944f-636875d1b1f4.jpg)

marcus-oscarsson:
Hey !

It seems like a never ending topic but I thought about something that could be nice for the actuators.
Instead of the button below to do trigger the opposite action, one could display an overlay with radio buttons. One for each action (obviously), its maybe more intuitive and also quite easy to implement. One click extra for the overlay though, but we gain some space ...

Just an idea :)

mguijarr:
We (Antonia and I) don't like an extra click.

meguiraun:
+1 for minimising extra clicks

marcus-oscarsson:
As you guys like, its just that I'm not very happy with the layout if we add a button below like that. I'm not sure that its going to look very nice. Id like to have a clean look and feel for the top bar, so its easy to get the right information. There is going to be quite alot of controls there so there is a small risk that it might look crowded.

Its good to minimize the clicks needed, but we also need to asks our selfs if its something that is accessed frequently enough to be quickly accessible. Its perhaps acceptable to have an extra click to access the safety shutter, as long as its status is visible all the time.

beteva:
Hello,
good move to start the cryo component. Only it  looks a bit strange to me. The annealing is better to be as a part of the beamline actions (as it requires a combination of different equipment, but not at all the cryo temperature). Our scientists asked not only the reading, but also setting the temperature, so the cryo component might need the space you allocate for the annealing to be used for that.

PR #455, opened 2017-01-31
[WIP] Top bar

meguiraun:
Issue #445 
* Redesign of the beamline row
* Basic cryo component (allocating space)
* Dynamically create actuator component based on the objects present in `state.beamline.actuators`

WIP because:
* you may not like the design at all: I keep both open and close buttons, although I disable one at a time instead of hiding, so no big visual changes
* my git history can (and will) be improved

The following images are generated only removing the beamstop and safety_shutter xml files, no client code was changed, just refresh!

image

image


meguiraun:
#456 is the good one, cleaner git history, closing this one

PR #456, opened 2017-02-01
Top bar

meguiraun:
(Closing previous #455 , my git history is nicer here argghhh)

Issue #445

Redesign of the beamline row
* Basic cryo component (allocating space)
* Dynamically create actuator component based on the objects present in state.beamline.actuators

you may not like the design at all: I keep both open and close buttons, although I disable one at a time instead of hiding, so no big visual changes. But let me know!

The following images are generated only removing the beamstop and safety_shutter xml files, no client code was changed, just refresh!

image

image

### DIFF
--- a/mxcube3/ho_mediators/beamline_setup.py
+++ b/mxcube3/ho_mediators/beamline_setup.py
@@ -52,9 +52,9 @@ class _BeamlineSetupMediator(object):
         elif name == "fast_shutter":
             return self._ho_dict.setdefault(name, InOutHOMediator(ho, "fast_shutter"))
         elif name == "safety_shutter":
-            return self._ho_dict.setdefault(name, TangoShutterHOMediator(ho, "safety_shutter"))
+            return self._ho_dict.setdefault(name, InOutHOMediator(ho, "safety_shutter"))
         elif name == "beamstop":
-            return self._ho_dict.setdefault(name, BeamstopHOMediator(ho, "beamstop"))
+            return self._ho_dict.setdefault(name, InOutHOMediator(ho, "beamstop"))
         elif name == "capillary":
             return self._ho_dict.setdefault(name, InOutHOMediator(ho, "capillary"))
         elif name == "dtox":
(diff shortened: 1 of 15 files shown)


beteva:
Hello,
good move to start the cryo component. Only it looks a bit strange to me. The annealing is better to be as a part of the beamline actions (as it requires a combination of different equipment, but not at all the cryo temperature). Our scientists asked not only the reading, but also setting the temperature, so the cryo component might need the space you allocate for the annealing to be used for that.

meguiraun:
It was not intended to be a complete cryo component. 
I can move to beamline actions and leave visible only the temperature (set and get field)

marcus-oscarsson:
Hi

Sorry for taking some time to get back to you on this. I know that its a bit tedious to make this configurable, its great that we have the functionality. We have been speaking a bit about the design and we made some GIMP art :). 

What we are looking for is a "clean" design where its easy to get an overview of the beamline status and were its possible to perform the corresponding actions. The important thing is that its easy to see the current state, like the dashboard of a car (my favorite example), and then have the possibility to change it in an intuitive way.

So we have imagined a solid bar the color can of course be changed we just picked a light blue color (almost randomly :) ). A light gradient or a background with some texture would perhaps be nice too. Another possibility would be to choose a dark color for the top bar and lighter colors for the controls. We have to see how this looks like with entire application in thought, for instance the menu bar. Its perhaps not important to decide the exact color scheme right away.

While speaking about colors and doing these screenshot Daniele also tried a bit with the color blind application, Ill attach a screen shoot of that at the end. The green and the red we are using for signaling the states are very similar to people with this eye problem, its maybe something worth to consider. Although not 

Clicking on the state of a actuator would bring up an overlay with the control for setting it in the opposite state, in the case of beamstop that would for example be a OUT button. Maybe with a text, Beamstop is IN set OUT.

topbar-blue

Screen shot with the color blind app:
topbar-deutranopia

Marcus

meguiraun:
hej, the configurability will require some more work in the server side but at least we have something to start with. Glad you like it! Maybe a flag instead of the existence of the hwobj...

I am no that glad that you did not like the design... ;) The present PR is based on #445, which contains what we discussed during our last visit. I see that you want to remove the boxes, which is ok for our people as well. Regarding the behaviour of the actuators, we had a long discussion last time and we still do not see the overlay for the action buttons (or whatever is required for the actuator).

What about if I remove the boxes (where we all agree) and we (er... you) accept the PR and we hold a redesign discussion next time we have a meeting? Would that be ok?

JieNanMAXIV:
@marcus-oscarsson It looks nicer to remove the borders of each components. The intentions of improve looking and minimizing the clicking numbers to improve ergonomics are all very much appreciated. However, I'm not convinced that if that means a sacrifice of functionality. 

In the case of shutters and beamstops, to merge the status with buttons can be quite confusing. As these operations should not always be available, e.x. in the middle of data collection users should not move the beamstop, while in the centering phase users should not open the fast shutter. However, it's not clear for users if the buttons are clickable or not with the proposed design. 

The PR did fulfill the design that have been agreed upon during last meeting #445. Can we move on and accept the PR if it's only a design issue?  Thanks!

meguiraun:
Now it looks like this

image

ddesanctis:
@meguiraun please take the redesign as a suggestion for final convergence, good if you like it, but it was intended as alternative take for discussion.

@JieNanMAXIV Regarding the status buttons, I'm not sure, I'd give it a try, if we see that is confusing we go back. IMO it can save space and be intuitive to use. We had discussed about this in mxcube2 but then never done it. I do not see the link between the status button and the fact that the movement of some components should be blocked while doing other actions, but this is the case also when the status is displayed separately.

marcus-oscarsson:
Thanks for all your work Mikel !, lets continue the discussion on our next meeting (if we still feel that there is need for it).

meguiraun:
Thanks!

PR #459, opened 2017-02-07
SC Integration

marcus-oscarsson:
Hi,

I have refined the queue execution so that it works better with the features that we have foreseen.
There are still plenty of rough corners to fix, I tried to focus on getting a consistent and reasonable behavior based on the numerous discussions we had on this topic. There are still some smaller work to be done
until we can be completely satisfied with the behavior/functionality.

I'm patching the queue_entry.mount_sample method, this is something that we would like to solve in a nicer way. We have some ideas on how to do that, but that will come in a later PR.

Expected behavior of actions the user can perform (the name often being the same as the name of the
button that triggers it):

**AUTO MOUNT SAMPLE = False**

- collect => runs first sample in queue to the end, wait for user to perform action
- run     => re-collects the current sample
- next    => mounts and runs the next sample in the queue to the end, waits for user to perform action
- finish   => (for the moment only for last sample in queue), un-mounts sample and empties queue
- stop    => stops the queue
- pause  => pauses the queue

**AUTO MOUNT SAMPLE = True**

- collect => runs the queue to the end, mounting the next sample if nothing pauses the queue
- run     => run same as collect, but does not run already collected samples
- finish   => (for the moment only for last sample in queue), un-mounts sample and empties queue
- stop    => stops the queue
- pause  => pauses the queue

Marcus

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -118,6 +118,8 @@ if not app.debug or os.environ.get("WERKZEUG_RUN_MAIN") == "true":
 
         # SampleID of currently mounted sample
         app.CURRENTLY_MOUNTED_SAMPLE = ''
+        app.AUTO_MOUNT_SAMPLE = False
+        app.AUTO_LOOP_CENTER = False
 
         try:
             SampleCentring.init_signals()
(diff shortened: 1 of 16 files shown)


meguiraun:
thanks for this! 

Just one comment, if a sample has not tasks and automount is True, such sample will be mounted and immediately after unmounted. I am not sure if this is what we want, we can discuss it tomorrow. 

Accepting this PR independently of that issue.

marcus-oscarsson:
Hi, 

Ah, yes I think I mentioned that above, its true that its possible to combine the two in the interface, its perhaps something that should be prevented. I don't really see why you would add no tasks to a sample and use "auto mount next", since that will just cycle through the samples.

Marcus

PR #460, opened 2017-02-07
Small style update to topbar

marcus-oscarsson:
Hey,

We are having the user meeting here at ESRF and they are showing MxCuBE to future users. So 
we made a little update to the topbar to make it more "flashy". Its not at all something set in stone
just a quick fix to make it more presentable (for showing off).

I also removed some linting errors, that's why there are more than just changes to the style :)

Marcus


meguiraun:
thanks for this, I like it ;)

PR #462, opened 2017-02-09
Dc view queue refinement

meguiraun:
trying to finish https://github.com/mxcube/mxcube3/issues/392 (one Run button, removing buttons...)

Now it looks as:

image


mguijarr:
Hi,
There are some conflicts that prevent the branch to be merged. Please rebase it so we can merge :)

meguiraun:
let's see now

marcus-oscarsson:
Great !

A suggestion is to remove the delete collection button completely when the task is already collected, or to make it visible that its not possible to use.

Marcus

PR #463, opened 2017-02-13
Layout update, with menu on top

marcus-oscarsson:
Hi,

We have been discussing the general layout here a bit, and we though that it might look nicer to have the menu on top. We find that its difficult to achieve visual harmony with the vertical menu and the horizontal top bar, we tried a couple of things and we came up with this.

Another solution could be to have the top bar or the menu on the bottom. In anyways this is suggestion of how it could look like. 

It also gives us some more space for the queue which we think can benefit from this.

Please let us know what you think

Cheers,
Marcus


meguiraun:
I will miss the buttons... thanks for this.

mguijarr:
I really like the horizontal menu bar with text instead of the (bad) icons. It is less funky, though.

But we may still change it a bit... After a day working on MXCuBE 3 I find the menu/nav bar to be too dark,
and I am not very fond of the 'skeuomorphic' design at the end.

PR #467, opened 2017-02-15
Faster updates of sample grid

marcus-oscarsson:
Hey,
 
Ive "optimized" the sample grid a little bit, so it now reacts as fast as one would expect. Ive tried with up
to 1000 samples on the same page. It starts to become noticeably slower, but still usable, after 500 samples, using the production build speeds things up a bit. However I'm not sure that we will or should have 1000 samples on the same page, we should perhaps provide some default filtering, paging, dynamic loading or other mechanism to handle that.

Ive also added some visual effects to the grid, don't pay to much attention to that. However I do think its nice to have some kind of animation when samples are moved or filtered.

Marcus

### DIFF
--- a/mxcube3/ui/actions/queue.js
+++ b/mxcube3/ui/actions/queue.js
@@ -24,6 +24,11 @@ export function addSamplesToQueueAction(samplesData) {
 }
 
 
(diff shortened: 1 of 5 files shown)


meguiraun:
awesome! thanks for this!!

Issue #468, opened 2017-02-16
a centered point can be saved without a mounted sample


JieNanMAXIV:
can be quite confusing, as users will fail to add tasks to the point, but don't get any error msg.

meguiraun:
it only affects the 2d centring (aka double click)

meguiraun:
closing, solved by https://github.com/mxcube/mxcube3/pull/476

PR #469, opened 2017-02-16
[WIP] fixes

meguiraun:
a collection of small fixes here and there while we are testing in the beamline.

WIP because I guess more will come.

* in the queue state retrieval `sample_order` was not properly pop-ed
* missing `get_dc` if rest lims not available (this is an old friend)
* the prefix changed to {'d-', 'c-', 'l-'}name-acr based on collection type
* the root path supplied to the task was lost
* In the current tree clicking on 'next sample' was not mounting the sample, only running it
   * linked to that: it is a bit confusing that if you click 'next sample' it will start running that one no matter if you have run the queue first. This pr looks that it fixes the issue for us but I am not sure if it breaks something else (gonna work on this more, perhaps a flag that tells if the queue is already running?)
* Solving #468

### DIFF
--- a/mxcube3/ui/components/SampleQueue/CurrentTree.js
+++ b/mxcube3/ui/components/SampleQueue/CurrentTree.js
@@ -36,7 +36,8 @@ export default class CurrentTree extends React.Component {
 
   nextSample() {
     if (this.props.todoList[0]) {
-      this.props.runSample(this.props.todoList[0]);
+        this.props.mount(this.props.sampleList[this.props.todoList[0]]);
+        this.props.runSample(this.props.todoList[0]);
     }
   }
 
--- a/mxcube3/ui/components/SampleQueue/CurrentTree.js
+++ b/mxcube3/ui/components/SampleQueue/CurrentTree.js
@@ -36,8 +36,9 @@ export default class CurrentTree extends React.Component {
 
   nextSample() {
     if (this.props.todoList[0]) {
-        this.props.mount(this.props.sampleList[this.props.todoList[0]]);
-        this.props.runSample(this.props.todoList[0]);
+      this.props.mount(this.props.sampleList[this.props.todoList[0]]);
+      // TODO: discuss behaviour and logic a bit more
+      // this.props.runSample(this.props.todoList[0]);
     }
   }
 
--- a/mxcube3/ui/components/SampleView/ContextMenu.js
+++ b/mxcube3/ui/components/SampleView/ContextMenu.js
@@ -40,13 +40,22 @@ export default class ContextMenu extends React.Component {
 
   showModal(modalName) {
     const { sampleID, defaultParameters, shape, sampleData } = this.props;
+    let typePrefix = '';
+    if (modalName === 'Characterisation') {
+      typePrefix = 'c-';
+    } else if (modalName === 'DataCollection') {
+      typePrefix = 'd-';
+    } else if (modalName === 'Helical') {
+      typePrefix = 'l-';
+    }
+
     this.props.showForm(
       modalName,
       [sampleID],
       { parameters:
         { path: sampleData.sampleName,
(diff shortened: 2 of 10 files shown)


mxcube3/ui/components/SampleQueue/CurrentTree.js
```diff
@@ -36,7 +36,8 @@ export default class CurrentTree extends React.Component {
 
   nextSample() {
     if (this.props.todoList[0]) {
-      this.props.runSample(this.props.todoList[0]);
+        this.props.mount(this.props.sampleList[this.props.todoList[0]]);
```
marcus-oscarsson: You should not need to mount the sample before running a sample, its taken care of by the sample
queue entry. This already works here for us without the extra mount step so it should be working for you
to.

mxcube3/ui/components/SampleView/ContextMenu.js
```diff
@@ -40,13 +40,22 @@ export default class ContextMenu extends React.Component {
...
+      typePrefix = 'l-';
+    }
+
     this.props.showForm(
       modalName,
       [sampleID],
       { parameters:
         { path: sampleData.sampleName,
           ...defaultParameters[modalName.toLowerCase()],
-          prefix: sampleData.defaultPrefix
```
marcus-oscarsson: A small observation here, we should be care full when we are dealing with prefixes, run numbers and image numbers to make sure that they are as configurable as possible. Its also good if we can avoid logic related to those on the client as far as possible.

So my suggestion would be to implement this on the server logic.

marcus-oscarsson:
Hey Mikel,

Looks great, I just wanted to give you a heads up on some things I spotted, see the code comments. Let me know what you think about it. Prefixes and run numbers can be quite cumbersome to implement in a good way, so let me know if you'd like to discuss it a bit.

mxcube3/ui/components/SampleView/ContextMenu.js
meguiraun: I was expecting this comment ;)
I agree that it needs to be more configurable, at this moment I put it here just to move forward.

meguiraun:
Thanks for looking at this!! It is WIP and I know that I am messing with a few things that should be handle perhaps in a different way. Most of this comes from the deployment we are doing at the beamline, and there we only work in manual mode (so I did little hacks...).

I will refactor/clean up this PR before removing WIP tag. Mostly prefix-es and so on and the manual mode and queue based operation (now clicking next runs the next sample as well). (and I think the most ugly changes are not in this branch yet...)

marcus-oscarsson:
:), Okey

marcus-oscarsson:
Hey

A quick comment on one of the points you made above

- ...linked to that: it is a bit confusing that if you click 'next sample' it will start running that one no matter if you have run the queue first. 

Next sample, means just that; mount and run next sample, if you just want to mount it just click mount. If the sample does no have any tasks, the sample is only mounted and nothing else happens. We could perhaps add "run" to the text -> "Run next sample" (to clarify things). I'd like to avoid to have several buttons that just mounts a sample.

Marcus

meguiraun:
I see your point, I do not like many buttons either. What I see is that the 'next sample' should not run the next sample automatically if you have not hit the 'Run queue' before.

ddesanctis:
Hey,

Just to earn my two pence, I like the behavior of Next sample, because it
delegates to the sample changer what should be done next, and I think is
fine that it just does a mount if no DC is present and starts with the DC
if already there. Maybe we can fine tune with some confirmation
windows/pause?

D

meguiraun:
I understand and agree with the behavior both of your describe (and with the actual implementation). However, I think it does not fit very well when operating in a manual mode. Here there is a point for the next week's meeting!

meguiraun:
I am closing this one since most of the issues have been merger in several PRs.

Issue #403, opened 2016-11-28
Draw a grid for a mesh scan


amilan:
**User story**
As a user, I want to be able to draw a grid, so that will allow me to select an area to be scan.

**Acceptance criteria**
Given my application running,  
When I want to run a mesh scan I can draw a grid on the camera view,
Then the grid area is selected to be scanned.

beteva:
Please consider - need to be able to change numbering of the images in the grid drawing, depending on the type of mesh you will do (consecutive lines or zig-zag)

PR #470, opened 2017-02-16
Grid UI

marcus-oscarsson:
Hey,

Preliminary version of the grid UI, implements #403 

Its now possible to: 
 - Delete and save grids
 - Several grids supported
 - Define cell spacing or overlap (negative spacing)

The cell counting is by default done in "zig-zag" mode but is configurable and other methods can be added, methods currently available is "left-to-right" or "zig-zag"

A grid is: 
- Drawn either from the context menu or the video toolbar (the later will probably be removed).
- Saved or Deleted from the context menu
- Cell spacing is defined in the form that appears above the grid, (which probably will be moved into the
  context menu)

Marcus

### DIFF
--- a/mxcube3/ui/actions/sampleview.js
+++ b/mxcube3/ui/actions/sampleview.js
@@ -79,7 +79,6 @@ export function addCentringPoint(x, y) {
   };
 }
 
-
 export function addLine(p1, p2) {
   return {
     type: 'ADD_LINE', p1, p2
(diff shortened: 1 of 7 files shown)


meguiraun:
wow wow! I have shown this here and the response has been unanimous, everyone like it! Thanks for this hard work!

The only minor thing is that it is a bit hard to see the numbers with the mockup image background.

PR #471, opened 2017-02-17
Updates, refactoring and fixes

mguijarr:
Hi :)

This PR prepares for the next one I will do, as I am currently working on Beamline Actions and plotting.

For the moment here are mainly fixes and cleaning in general.

I also propose a small UI improvement regarding InOutSwitch2.

Have a nice week-end.

PS: I tried yarn instead of npm... It's 100 times better!!! I recommend it.


meguiraun:
thanks for this!

PR #472, opened 2017-02-17
Beamline actions

mguijarr:
Beamline actions list is built from the corresponding hardware object (called '/beamcmds' here), the exact same as what is already in MXCuBE 2. I made a `BeamlineActionsMockup` to be able to test.

People can run/abort/check output of each action via the Beamline Actions dropdown.
Only one action can be executed at a time.
When the output window is shown, it appears as a modal window. Log messages go to it.
Some actions (like annealing) may require parameters ; arguments can be entered via input fields in the output window.

Note: hardware objects need to be updated to be able to use the new mockup object (`git submodule update`).

Next (big) PR will be: support for plot output (in case of beamline realignment, for example, it is good to see what is happening directly in MXCuBE).


meguiraun:
hi! I am testing and the actuator operation seems ~broken (testing this pr, clean npm install, latest 2.2 hwobj)
```
2017-02-24 10:55:41,551 |exceptions|ERROR  | Uncaught exception while calling /mxcube/api/v0.1/beamline/fast_shutter
  File "/usr/lib/python2.7/site-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/lib/python2.7/site-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/mxcube/mxcube3/mxcube3/routes/Beamline.py", line 170, in beamline_set_attribute
    return jsonify(result)
  File "/usr/lib/python2.7/site-packages/flask/json.py", line 237, in jsonify
    return current_app.response_class(dumps(dict(*args, **kwargs),
ValueError: dictionary update sequence element #0 has length 1; 2 is required
```

mguijarr:
Gonna have a look later this afternoon...

mguijarr:
So, it should be fixed now... Indeed I introduced a bug before when I tried to 'homogenize' responses in our Beamline routes.
So, now I always try to use `jsonify` (which returns a Flask `Response` object with code = 200 and type set to `application/json`), except that sometimes we need to change the code.
And also when the client is not expecting a json answer, I reply with `application/text` instead of returning a json object, which is much cleaner.

meguiraun:
Thanks for the quick fix!
One note, the commit notifications from this PR I receive comes from Antonia, while it looks to me it's you...
image

PR #475, opened 2017-02-24
Attempt to fix #429

### LINKED ISSUE #429: Motor Values are not applied without pressing "return"
On the Data Collection view,  the new value of a motor position is not applied until after pressing "return". It should also be applied when the mouse leaves the input field, or otherwise put back the previous value, which is the real current motor position.  

As it is now, if a user presses somewhere else after changing the motor value (but without pressing return), the new value displayed on the UI is not what the motor has.

mguijarr:
... added 'edited' state for motor inputs.

The 'edited' state is _not_ part of the redux state, but I think it's not a problem.


meguiraun:
thanks for this!

PR #476, opened 2017-02-24
Fixes

meguiraun:
some the fixes from the wip pr (I removed the ones that need more work and some agreement)

### DIFF
--- a/mxcube3/routes/SampleCentring.py
+++ b/mxcube3/routes/SampleCentring.py
@@ -11,6 +11,7 @@ import json
 import signals
 import PIL
 import cStringIO
+from . import scutils
 
 SAMPLE_IMAGE = None
 CLICK_COUNT = 0
@@ -575,6 +576,11 @@ def wait_for_centring_finishes(*args, **kwargs):
     Executed when a centring is finished. It updates the temporary
     centred point.
     """
+    # we do not send/save any centring sata if there is no sample
+    # to avoid the 2d centring when no sample is mounted
+    if scutils.get_current_sample() == '':
+        return
+
     motor_positions = mxcube.diffractometer.centringStatus["motors"]
     motor_positions.pop('zoom', None)
     motor_positions.pop('beam_y', None)
--- a/mxcube3/routes/SampleCentring.py
+++ b/mxcube3/routes/SampleCentring.py
@@ -11,7 +11,7 @@ import json
 import signals
 import PIL
 import cStringIO
-from . import scutils
+import scutils
 
 SAMPLE_IMAGE = None
 CLICK_COUNT = 0
--- a/mxcube3/routes/qutils.py
+++ b/mxcube3/routes/qutils.py
@@ -227,9 +227,11 @@ def _handle_dc(sample_id, node):
     queueID = node._node_id
     enabled, state = get_node_state(queueID)
     parameters['subdir'] = parameters['path'].split(mxcube.session.get_base_image_directory())[1][1:]
-    try:
+
+    if mxcube.rest_lims:
         limsres = mxcube.rest_lims.get_dc(node.id)
-    except:
+    else:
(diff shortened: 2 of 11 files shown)


mxcube3/routes/SampleCentring.py
```diff
@@ -11,6 +11,7 @@
 import signals
 import PIL
 import cStringIO
+from . import scutils
```
mguijarr: `import scutils` should work exactly the same, no? Since Python first tries to load from current directory.

mxcube3/routes/qutils.py
```diff
@@ -537,11 +542,12 @@ def set_dc_params(model, entry, task_data):
     acq.path_template.base_prefix = params['prefix']
 
     full_path = os.path.join(mxcube.session.get_base_image_directory(),
-                             params.get('path', ''))
+                             params.get('subdir', ''))
```
mguijarr: Just for me to understand, what is the difference between `params.get('path')` and `params.get('subdir')` ?
Could you please show me with an example ?
It seems to me `full_path` won't be full path anymore with this change but I am not sure ?

mxcube3/routes/qutils.py
```diff
@@ -226,6 +226,11 @@ def _handle_dc(sample_id, node):
     parameters.pop('centred_position')
     queueID = node._node_id
     enabled, state = get_node_state(queueID)
+    parameters['subdir'] = parameters['path'].split(mxcube.session.get_base_image_directory())[1][1:]
+    try:
+        limsres = mxcube.rest_lims.get_dc(node.id)
+    except:
```
mguijarr: Would be really great to display the traceback or to raise an error if a real error happens, at the moment every error is eaten up by `except:`. Maybe it is enough to test for `if mxcube.rest_lims:` ?

mxcube3/routes/signals.py
```diff
@@ -218,11 +218,15 @@ def collect_oscillation_started(*args):
 
 
 def collect_oscillation_failed(owner, status, state, lims_id, osc_id, params=None):
+    try:
+        limsres = mxcube.rest_lims.get_dc(lims_id)
+    except:
```
mguijarr: Same comment as above, regarding `except:`

mxcube3/ui/components/Tasks/DataCollection.js
```diff
@@ -95,10 +95,13 @@ class DataCollection extends React.Component {
         prefix,
         run_number,
         beam_size,
-        path
+        subdir,
       },
       rootPath
     } = this.props;
+    console.log('******************')
```
mguijarr: to be removed

mxcube3/ui/reducers/queue.js
```diff
@@ -78,7 +78,7 @@ export default (state = initialState, action) => {
       {
         return {
           ...state,
-          rootPath: action.data.rootPath,
+          rootPath: action.data.beamlineSetup.path,
```
mguijarr: is `action.data.rootPath` removed from the action data ? (since we do not use it anymore)

mxcube3/routes/qutils.py
meguiraun: `full_path` is always _full_.  I created `subdir` because it is the one the user can populate (by default is the sample_name) and it was not consistent in general. Before this change, there was the `path` which was meaning the full_path but also the sample_name. I noticed this when I reloaded the interface and it was joining the path with itself.
I renamed the property to subdir.

mxcube3/ui/reducers/queue.js
meguiraun: yes, I think it has been lost...

mxcube3/routes/qutils.py
mguijarr: Ok thanks for the explanation. I was confused by all the paths and `.split()` and indexes (`[-1]`, `[1]`...)

PR #480, opened 2017-03-02
Fix task state

meguiraun:
**This already includes previos PRs commits! Accept that one first!**

I noticed that there was a mismatch regarding the state of the tasks, easy to miss but clear on refresh. Task's states (hopefully everywhere):
* 1: running
* 2: failed (3: warning)
* 4 done


mguijarr:
Hey Mikel,

Thanks for this.

I would propose the following: could we use constants instead of plain numbers ? That would make things much better, like READY=4 (or 2 ? Now I am puzzled), etc

mguijarr:
Btw, are you sure I should accept this one first ? It would add the 2 commits that are part of the previous PR. Maybe I should do the opposite ? Or you remove the other PR ?

beteva:
How about a state class with immutable enum, so we can do state.READY, state.FAILED ...

meguiraun:
ok I will add constants, a class for this? This pr was intended as a fix.
I mean accept the other one first.

mguijarr:
Not a class, maybe just module-global variables.

meguiraun:
sorry for the delay! let's see now

mxcube3/routes/qutils.py
```diff
@@ -29,6 +29,12 @@
 UNCOLLECTED = 0x0
 READY = 0
 
+TASK_COLLECTED = 4
+TASK_COLLECT_FAILED = 2
+TASK_COLLECT_WARNING = 3
+TASK_RUNNING = 1
+TASK_UNCOLLECTED = 0
```
mguijarr: @meguiraun correct me if I am wrong, but aren't TASK_* constants the same as the ones above? Like, `TASK_COLLECTED == COLLECTED`. So we don't need both...

Also, `WARNING` = 3 but this doesn't compose well, it would be better as 0x10 (16) I think - like, we can have something both `COLLECTED` and `WARNING` maybe. With 3 we cannot distinguish.

By the way I noticed we maybe made a mistake with READY/UNCOLLECTED = 0... We can't know if we have SAMPLE_MOUNTED and UNCOLLECTED. Well, maybe it's implicit if nothing is set it's READY/UNCOLLECTED...

mguijarr:
Hej @meguiraun :) Thanks a lot. A small question: is it ok for you if I merge it or are you still working on it? The warning constant should also be in js code or maybe we are not using it in javascript code ?

meguiraun:
It is ok to merge, and I think we are not using the warning in the client.

PR #492, opened 2017-03-14
Preliminary Workflow integration

marcus-oscarsson:
Hey !

Preliminary integration of workflows.

Added the possibility to:
  - get a list of available workflows, (displayed where appropriate: in sample grid and sample view context menus)
  - add workflow task to queue
  - start a workflow, (like any task in the queue)
  - stop a workflow, (like any task in the queue)
  - display a configurable dialog which is displayed on the signal parametersNeeded which can be sent
     by any HO.

The workflow methods are displayed according to the requires tag in the ednaparams.xml file. The name of the file is not really important but i kept the old name for now (its just an example file)

You need to update the HardwareObjects to get a small change to the EDNAWorkflow HO in order to get this to work. "git submodule update" should be enough

Still to be done is integration of workflow results, however I prefer to do that in another PR since many workflows does not use the result part.

Marcus

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -96,7 +96,7 @@ if not app.debug or os.environ.get("WERKZEUG_RUN_MAIN") == "true":
     ### Importing all REST-routes
     from routes import (Main, Login, Beamline, Collection, Mockups, Utils,
                         SampleCentring, SampleChanger, Diffractometer, Queue,
-                        lims, qutils)
+                        lims, qutils, workflow)
 
     ### Install server-side UI state storage
     from mxcube3 import state_storage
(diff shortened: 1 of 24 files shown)


meguiraun:
hej @marcus-oscarsson, thanks for this. Just one issue I noticed, I cannot add manual samples:
`Uncaught TypeError: toISOString is not a function`
I cannot find that reference in the code, do you have any idea about this? (clean npm install)

marcus-oscarsson:
Hey,

Sorry I was not in yesterday, Yes its redux form that "stringyfies" the default parameters, which were undefined. Fixed now !

Marcus

meguiraun:
great! thanks for this!

PR #494, opened 2017-03-14
[WIP] Char result

meguiraun:
Very wip... but i do not have ispyb to test the call `mxcube.rest_lims.get_dc(node.id)` in the case of a characterisation. I do not know what it returns ...

I will push here with any advance, and feel free to do the same


marcus-oscarsson:
Thanks Mikel,

I will try this when I get a good opportunity and maybe add the missing bits if we need to.

Marcus

meguiraun:
I close and will do another pr later, this has been stalled for too loooooong...

PR #495, opened 2017-03-14
Reset session

marcus-oscarsson:
Hey,

- Removed the reset session from navbar

- Clear sample grid now only clears the sample grid content, the queue (including current sample), also   clearing shapes since we are clearing current sample. Its at the moment not clear if its good idea to clear
the current sample (and shapes), alltough it seems intuitive to the user.

- Clearing all on logout so we can always achieve the old session reset by logging out.

- Two other completely unrelated fixes,  updated sub modules and fixed linting

Marcus


meguiraun:
tack!!

PR #497, opened 2017-03-15
Loading-Gif sample grid view

bamartos:
Hi,

I was assigned to this task by MaxIV mxcube developers. This task solves the issue with loading-gif during and Add sample/task to queue action and add the loading-gif in the SampleGrid during queue modifications.

Cheers,
Vasilis

### DIFF
--- a/mxcube3/ui/actions/queue.js
+++ b/mxcube3/ui/actions/queue.js
@@ -72,6 +72,8 @@ export function sendMountSample(sampleData) {
 
 export function addSamplesToQueue(sampleDataList) {
   return function (dispatch) {
+    dispatch(queueLoading(true));
+
     dispatch(addSamplesToQueueAction(sampleDataList));
 
     sendAddQueueItem(sampleDataList).then((response) => {
@@ -80,6 +82,7 @@ export function addSamplesToQueue(sampleDataList) {
         const sampleIDList = sampleDataList.map((sampleData) => (sampleData.sampleID));
         dispatch(removeSamplesFromQueueAction(sampleIDList));
       }
+      dispatch(queueLoading(false));
     });
   };
 }
@@ -424,10 +427,9 @@ export function addTask(sampleIDs, parameters, runNow) {
             dispatch(sendRunSample(sampleIDs[0], taskIndex));
           }
         }
+        dispatch(queueLoading(false));
       });
     }
-
-    dispatch(queueLoading(false));
   };
 }
 
--- a/mxcube3/ui/containers/SampleGridViewContainer.jsx
+++ b/mxcube3/ui/containers/SampleGridViewContainer.jsx
@@ -476,7 +476,7 @@ class SampleGridViewContainer extends React.Component {
           />
           {this.props.loading ?
             <div className="center-in-box" style={{ zIndex: 1200 }} >
-                <img src={loader} className="img-responsive" alt="" />
+                <img src={loader} className="img-responsive" alt="" width="430" />
               </div>
             : null
           }
@@ -636,4 +636,3 @@ export default connect(
   mapStateToProps,
   mapDispatchToProps
 )(SampleGridViewContainer);
-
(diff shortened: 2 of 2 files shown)


marcus-oscarsson:
Hi Vasilis,

Its nice to have you working on the project !, thanks for this first contribution. I guess its not that easy to get into the project right away, so let us know if there is anything we can do to help. It would have been nice to speak with you in person before you started the project, I guess we will perhaps see each other on the next Skype meeting

Its indeed important to have the busy indicator when we add samples great work :), we also need it when samples are moved around in the grid. 

I'm not entirely sure that a overlay like this is what we really want in the end but its maybe good enough for the moment. Its nice if we can come up with something more discrete but that still conveys the message, something to discuss ...

I guess you didn't notice but it seems like that the busy indicator is not always displayed. Its perhaps possible to preload the gif so that its display right away ?. The animation and the quality of the gif could also be improved, its not really animated when its shown and it looks a bit noisy around the edges like its 
scaled to much. At least here, using chrome on windows.

A slightly related matter is that tasks are added to the grid even if they are not added to the backend. This also needs to be handled somehow in the ADD_TASK reducer. So that we are sure that we are displaying whats really in the queue. That's really the reason for why we have this busy indicator to begin with.

Cheers,
Marcus

bamartos:
Hi Marcus,

Thanks for your fast feedback on this pull request! I fully agree with your suggestions but do you think that the current pull request depends on them? 

I suggest to post the suggestions above as requests/enhancement in the Issues section and implement them in the near future. I think like that we can create a lightweight agile workflow which can motivate more new users to contribute in the project. 

Cheers,

Vasilis

marcus-oscarsson:
Good morning !

The busy indicator is not only for one operation its for all queue operations. The feature we have been discussing, and thus are expecting, in our meetings is a busy indicator for all queue operations. It does not really matter to us how you divide this task within your team. Although I do think its better with one PR for this task so that we can get an appreciation of the intended solution as a whole.

So if your intention with this PR is to only provide the busy indicator for ADD and REMOVE tasks, we at least need to fix the "scaling" issue of the loading gif and also make sure that its animated when shown (and always shown not only sometimes). This is maybe something that I'm experiencing here with my setup and not you ?

You wrote in your message that this message was for "queue operations" so for me MOVE is also missing, but then its up to you guys if you'd like to do this in more PRs than one. 

Marcus

meguiraun:
god morgon!

The busy will appear in all queue operations, tested in add/delete and update for tasks and samples. In the case of the move, when you move samples in the grid no call is made to the server, (`queue.moveTask` action is never call, but it actually contains the `queueLoading(true/false)`). So, I would say that the busy indicator is added to all queue operations, although the move operation is not fully implemented in the grid.

We can fix the scaling limiting the size of the gif, I would say that it will make it ~nicer. For the other issue, I do not really know what you mean, it could be that what you noticed is due to very fast actions? Could you try to force a delay in the sever for those actions and see if the issue remains?

mikel

marcus-oscarsson:
Look !, I'm not going into a discussion of the definition of the task itself. The problem we need to solve is that we need to keep the queue synchronized between server and client, that's what was decided. 

We decided to do this by limiting the number of requests to one and thus displaying a busy indicator while this operation is in progress (effectively preventing other actions from being performed).

Move is undoubtedly a queue operation, the call to server might be missing if that is the case it needs to be added. If you feel that its for some reason is not relevant to add that call right now you can perhaps just pretend like its there ?. I can only see queueLoading for sendMountSample, addSamplesToQueue and addTask, did I miss something because it seems to me like you say that you've added it ?

In anyways it needs to be done sooner or later. What we need from this is is in the end predictable queue behavior, the rest is less important.

For the animation I noticed it when I add something (anything) to the grid. I sometimes see the busy spinner and sometimes not (when I see it its not animated). Its not that bad as long as the quality of the spinner is alright, otherwise just display a text.

Marcus

bamartos:
Hi Marcus,

I just improved the loading spinner quality by specifying a smaller width. About the issue with the appearance(or not) of the loading spinner which you mentioned above is due to the speed of loading. (sometimes it just spins for a few milliseconds so is not  visible by the eye, to debug you can add a delay of 2'' to the queue_add_item(),  queue_update_item() ..etc in the Queue.py to be able to see and check the spinner quality) 

About the queue operation tasks Mikel will take care as we discussed in the meeting last week. 

Cheers,

Vasilis

marcus-oscarsson:
Hi,

Thanks Vasilis, Ill merge and we will take a closer look at the display when we have the complete functionality.

Marcus

PR #498, opened 2017-03-15
Sample changer UI updates

marcus-oscarsson:
Hey,

Some smaller updates to the SC UI, feel free to add more functionality if you need as long as it can be considered generic. (Since this is the generic sample changer UI). 

Marcus


meguiraun:
thanks for this!

PR #499, opened 2017-03-20
Proposal selection

meguiraun:
Hej,

This pr shows a modal with a list of proposal for the user, right after the login is successful.
WIP because:
* The exact flow of the proposal selection needs a bit more of thinking by our scientist
* Methods' names are a bit confusing...
* I do not handle (properly) failures

Note: I decided to do some coding on my own to understand better mxcube3 _kernel_, and I took something that is not needed at this moment but eventually would be useful.


meguiraun:
hej guys,

A small update:
* After a successful login all available proposals are retrieved from ispyb if `loginType==User`
* User will see a modal and he/she has to select a proposal to work with
* All the mxcube internals related to proposal are updated

Issues:
* Here at Maxiv we use a different approach for the users (why not!), one consequence is that I need to prepare some directories once I know the proposal (that's the reason of this line:
https://github.com/meguiraun/mxcube3/blob/proposal_selection/mxcube3/routes/limsutils.py#L83
   * I not sure about that being the best place for such method, if you have any suggestion on where to put let me know
* The selection modal is displayed always, probably not needed if you login by proposal but otherwise you cannot test. I can filter later.


I keep this as WIP for now but I have tested successfully with our Ispyb and with mockups. But, please let me know your thoughts.

Cheers,

meguiraun:
hej guys, I have just made some small changes. Could you test in your environment?

It is working for mockups and for our ipysb (login as user), but I cannot test with proposal login mechanism in real life.

marcus-oscarsson:
Hi Mikel, 

it looks like there are some conflicts that needs to be solved

Sorry

meguiraun:
gonna fix conflicts

PR #500, opened 2017-03-21
Lint

marcus-oscarsson:
Hey !

Missed a lint commit for previous PR #492 

Marcus


meguiraun:
mmm, now you have both the biggest and the smallest PR in the mxcube v3 history! hehehe

PR #503, opened 2017-04-12
MPEG1 Streaming

marcus-oscarsson:
MPEG1 encoding and streaming displayed using JSMpeg.

Images are written as RGB24 to a v4l2 (Video 4 linux 2) loopback device (/dev/video0). The v4l2loopback kernel module needs to be loaded which will create the loopback device.

The video stream from /dev/video0 is encoded to MPEG1 by ffmpeg and written to a websocket that is read by JSMpeg.

There is a preliminary UI for selecting the video stream size but its not yet connected to the backend. It currently only does client side scaling, like before. There is a few second delay before the stream starts, we will look into if we can reduce this delay.

I've tried to get some performance data, see the graph with comments below

mpeg1-14fps-and-mjpeg-10fps

The red plot shows the original MJPEG stream and the green plot shows the MPEG1 stream. The camera was running in 14 FPS with a video size of 659x493 pixels and the same image (an empty loop was recorded for two minutes). Both plots are showing the TCP packages sent to the decoding client.

**MJPEG**

- Only 10 FPS was achieved with the MJPEG stream which most likely means that our encoding process takes a little bit longer than the exposure period of the camera so we are dropping 4 frames per second.

- Roughly 150000 bytes per second with MJPEG stream, the small dips in the plot corresponds to non full (last segment of the frame)  TCP packages.

**MPEG1**

We can clearly see the key frames (the peaks) in the graph

- The full 14 FPS received, no signs of bottlenecks within the encoding and streaming chain, We have not yet done any CPU performance measurements. It could perhaps be interesting. (The streaming and the encoding runs as separate processes)

- Roughly 31500 Bytes per second are sent, the compression of the JPEG key frame is likely much harder in the MPEG1 stream and that would explain why the key frame size ( 40Kb) is so much smaller than the MJPEG frames (150 Kb). The differential encoding is clearly visible in the plot as the points between the peaks.

**Comments**
We still need to perform some measurements on delay, the initial tests on the beamline shows no noticeable delay. The JSMpeg website http://jsmpeg.com/ claim that delays can be as low as 50ms,
a claim which we might need to verify :)

The MPEG1 encoding process is clearly much more efficient both in terms of data rate and most likely CPU (that's at least what the dropped frames indicate). The MJPEG stream could probably benefit from faster and much harder JPEG compression.

We finally need to see to this behaves on the internet, especially in regards to delay. It otherwise seems like the MPEG1 stream would be the one to use.

Please let us know if you have any comments or questions !?

Marcus and Matias


meguiraun:
huge thanks for this! Gonna test during today, I was having some trouble testing it in a docker container so I will move down to the beamline.

One thing I realised is that with this pr anyone trying to test mxcube3 will need to go through all the v4l2, ffmpeg installation. Would not be better to keep both (the current jpeg sender and this PR's steamer)? At launch time one will decide which one to use, someone that only wants to see the latest interface does not need the streamer.

Also, in our case, we have the python server serving the compiled js bundle. So, we do not have any js/node file in our server machine (we just scp those). With this update, we need to do an npm install in the server to make it work even if we only need a few packages (ws, ejs, websocket-relay). Is it possible to specify a kind of streaming tag in the webpack configuration?

Hopefully I will have time to test this today... but ... I am oncall...

meguiraun:
I have had quite a few problems running the streamer, but all points out to permission issues in our machines... at this moment mxcube launcher cannot run ffmpeg so I am running it independently.

However in our current setup (centos 7 in a vm for the server, client in my macbook, non lima camera), there is a clear loss of quality and perfomance, the image is too pixelated. I hope this is related to the v4l2 module, which I am having some issues to setup, specially for the video size. ffmpeg claims to be at ~30fps but I do not believe at all.

Also, I noticed some weird behavior with the websocket. If I move e.g. 200 deg the omega motor, I see the position updates in the motor input box and only when the final position is achieved the image is being updated. I do not see that issue if I move the motor continously, I guess the issue is there but hidden.

Also, in `mxcube3/ui/components/SampleView/SampleImage.jsx`, `${document.location.hostname}` returns always "localhost" in development mode.

I have not have much time to test this pr but I will continue, I think is best if I post here my comments so you know that I am on this.

mguijarr:
Hi @meguiraun,
Thanks for your tests. Some remarks and questions:
- what is the format of the images submitted to ffmpeg? if it is a JPEG, I think the quality loss comes from here because then it's encoded as a JPEG twice... Try with RGB data. For your information, Maatel and EMBL guys here have implemented their microdiff sample video with Lima 3 weeks ago, so now there is nothing preventing you to use Lima too to get frames (if it helps...)
- Regarding performance: I suspect the VM... Just run it on a real computer
- We never had any latency problem, and in fact with the streaming it's even better than before - so I cannot really see what is wrong on your side, I am also dubious about the VM ; it's clear a VM is pretty bad for I/O operations
- regarding v4l2loopback, could you please tell us more about the video size issues you had ? Video size is set in the Python code that does the streaming, it's not something to set on the v4l2 device externally all the code is there in the PR (maybe there is a bug if input data is not RGB 24?)
- you're right for `${document.location.hostname}`, I guess something has to be configured in webpack-dev-server to relay http requests to the right host (like we do for API calls)

meguiraun:
mmm... Lima... guess I will need to try that. Thanks for the info.

Yea, I know that running in a vm is not good for this stuff but a vm is where we run mxcube and we are not going to change that, at least for now...

my `v4l2-ctl --all` shows
```
****
Format Video Output:
	Width/Height  : 1360/1024
	Pixel Format  : 'RGB3'
	Field         : None
	Bytes per Line: 4080
	Size Image    : 4177920
	Colorspace    : SRGB
***
`
I have tried to change that size but no luck. I will try to test on a desktop computer, hopefully it will be better and I can give you proper feedback.

What do you think about keeping both old and new streaming in master? I am thinking on the troubles I am getting due to our environment, but also it would be a pain for anyone that just wants to have a look at mxcube3.

mguijarr:
@meguiraun You're right we should do something for people who just want to test MXCuBE 3. For this particular need, I think a static image is enough.
I am not so convinced by keeping both ways of sending frames to the client because development on this feature is not finished yet, and I would prefer to avoid having fingers in many pies... What we need to do next:
- selecting video size, to efficiently transmitting less bytes per frame
- stopping/restarting video from the client without affecting MXCuBE 3 server itself
- tweak settings to avoid latency and network congestion (we are just using defaults now, which seems fine, but there are a lot of settings to try)
If we keep the good ol' MJPEG streaming (not the new MPEG-1 one) it will soon look as a poor cousin compared to the other one if it is not at the same level in term of features. Do we really want this in the project ?

Second thing: you said you want to change video size. Why ? I don't see the point in your current tests,
I am asking myself if you're chasing the wrong problem here but maybe it's just me who didn't understand.

Changing video size is a feature we would like to implement. We have to tell ffmpeg to output another format for the frames in the MPEG-1 stream. The source (/dev/video0) video size is the same size as the source image, and this cannot be changed unless the source images themselves are changed.

meguiraun:
The problem I see is that this solution is not working for us now, I know that should not be any kind of impediment for the development, but it would make us more trouble to upgrade to master. That's why I say about keeping old a new ways (also for curious testers) for a while. I understand that the current stream will not be improved and all the efforts would be on this pr.

For lima, I tried a while ago what you mention (Lima device), I did not have luck and more urgent things came so I put back to my board...

For the video size, I was not expecting such size, so that's why I am confused. Still need to test better.

marcus-oscarsson:
Hey !

I have added fall back to MJPEG if there is no loop back device and improved the stream size selection.
Its also possible to enable or disable the client auto scale.

Marcus

mguijarr:
Thanks for the great work @marcus-oscarsson :)
With the fallback to MJPEG as it is now in mxcube3, I guess this can be merged safely :)
@meguiraun tell us if you find any problem with this PR.

meguiraun:
Hej, thanks for this. I just want to quickly test to avoid any unexpected issues, I will do as soon as I have some time

PR #504, opened 2017-04-13
Shapes (Mesh Scan Part 1)

marcus-oscarsson:
Hi,

Another big PR, sorry for that, (Please note that this PR is Dependent on PR #220 of HO)

Working on the mesh scans I noticed that the centered positions was not stored in a way that was compatible with how some parts of MXCuBE works. 

The Points  was simply attached to a dictionary on the diffractometer, this will of course works fine for just storing the positions.  However, it wont work in general, since other parts of MXCuBE needs to access that data. An API for accessing Shapes is already specified by the HO called ShapeHistory. This HO is dependent on Qt so it wont work for us, so I've created a new HO called Shapes that is completely independent of UI library.

So this PR:

- Adds Shapes HO which manages Shapes (Points, Lines and Grids) in a way which is compatible with
  MXCuBE

- Generalizes the REST-API for Points (previously called centrings in the REST-API) to handle Shapes in 
   general

- Refactors reducers and components to handle shapes instead of points.

All this is necessary to get the Mesh scans and workflows to work properly, so I will rename this 
Shapes (Mesh Scan Part 1), part 2 will put the final bits in place.

Marcus


mxcube3/routes/SampleCentring.py
```diff
@@ -13,6 +10,11 @@
 import cStringIO
 import scutils
 
+from mxcube3 import app as mxcube
+from mxcube3.routes import Utils
+from mxcube3.routes.transportutils import to_camel, from_camel
```
meguiraun: why not in Utils directly?

mxcube3/routes/SampleCentring.py
```diff
@@ -146,7 +142,7 @@ def subscribe_to_camera():
                     mimetype='multipart/x-mixed-replace; boundary="!>"')
 
 
-@mxcube.route("/mxcube/api/v0.1/sampleview/camera/unsubscribe", methods=['PUT'])
+@mxcube.route("/mxcube/ap i/v0.1/sampleview/camera/unsubscribe", methods=['PUT'])
```
meguiraun: extra space sneaked in

mxcube3/routes/SampleCentring.py
```diff
@@ -192,156 +188,92 @@ def get_image_data():
...
-    for pos in mxcube.diffractometer.savedCentredPos:
-        aux.update({pos['posId']: pos})
-        pos_x, pos_y = mxcube.diffractometer.motor_positions_to_screen(
-            pos['motor_positions'])
-        aux[pos['posId']].update({'x': pos_x, 'y': pos_y})
-    resp = jsonify(aux)
-    resp.status_code = 200
-    return resp
+    mxcube.shapes.delete_shape(sid)
+    return Response(status=200)
```
meguiraun: I am a bit confused, are we losing the ability of acting on individual shapes/points?

meguiraun:
Thanks a lot for this work!

mxcube3/routes/SampleCentring.py
marcus-oscarsson: Yes, we can act on individual shapes if we like to, but we don't need to in general. We want to remove all shapes when we mount or unmount a sample.

mxcube3/routes/SampleCentring.py
marcus-oscarsson: Because I believe that Utils is a name that could mean anything and its better to be specific in order to avoid the risk of creating big utility files.

marcus-oscarsson: We can move those functions but I would discourage or not advice it, as you guys like

meguiraun:
Now that the PR in hwobj has been accepted, could you update the submodule commit?

PR #505, opened 2017-04-19
Bug fixes

meguiraun:
a bunch of fixes that should have been submitted loooong ago.


meguiraun:
I noticed this morning that the copy to clipboard copies always the first box one no matter how many tasks you have in the queue... gonna fix that soon.

meguiraun:
removing WIP.

mxcube3/ui/components/SampleQueue/TaskItem.js
```diff
@@ -196,20 +196,21 @@ export default class TaskItem extends Component {
             <form>
               <div className="form-group row">
                 <label className="col-sm-12">File path:</label>
-                <div className="col-sm-12" style={{ display: 'flex' }}>
+                <div className="col-sm-12" style={{ display: 'flex' }} >
```
mguijarr: A few comments, at least on this code - I did not have time to review the whole PR yet:
- as we rely on react-bootstrap we should replace all old-style `form`+`input` etc to proper react-bootstrap equivalents, as it was done when we did the upgrade of libraries
- in React we should avoid as much as possible accessing the DOM tree directly with `document.getElementById` for example - there are references (callback refs) for doing this kind of thing
- we should avoid too much CSS styling (what is `display: flex` doing here ? Maybe it can be replaced by some options in react-bootstrap components ?)

mxcube-launcher
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/zsh
```
mguijarr: IMO this is a site-specific shell script that should not be included in the code repository.

mxcube3/ho_mediators/beamline_setup.py
```diff
@@ -638,16 +717,16 @@ def limits(self):
...
 
 
     def stop(self):
-        self._ho.stop()
+        self._ho.dtox.stop()
 
 
     def state(self):
-        return MOTOR_STATE.VALUE_TO_STR.get(self._ho.getState(), 0)
+	print self._ho.dtox.getState()
```
mguijarr: print to be removed

package.json
```diff
@@ -46,6 +49,9 @@
     "shortid": "^2.2.6",
     "socket.io-client": "1.4.5",
     "underscore": "^1.8.3"
+    "style-loader": "0.12.3",
+    "underscore": "^1.8.3",
+    "x-editable": "1.5.1"
```
mguijarr: We do not rely on x-editable: it has to be removed from deps

package.json
```diff
@@ -29,6 +30,8 @@
     "react": "15.4.2",
     "react-bootstrap": "^0.30.7",
     "react-contextmenu": "^2.3.0",
+    "react-bootstrap-switch": "^3.4.5",
```
mguijarr: Are we using react-bootstrap-switch ?

package.json
```diff
@@ -46,6 +49,9 @@
     "shortid": "^2.2.6",
     "socket.io-client": "1.4.5",
     "underscore": "^1.8.3"
+    "style-loader": "0.12.3",
```
mguijarr: style-loader and css-loader are devDependencies

mxcube-launcher
meguiraun: ouch!

mxcube3/ui/components/SampleQueue/TaskItem.js
meguiraun: I have not changed form-input and style, I will try to update anyway.
I use the react-clipboard library just for simplicity, I will try to do the same with standard html5

meguiraun:
Thanks for the review! I will try to fix your comments and update the pr

mxcube3/ui/components/SampleQueue/TaskItem.js
mguijarr: Ok - thanks Mikel ^^
Do not hesitate if you need help on anything. You can look at many examples of react-bootstrap form components on the webpage:
https://react-bootstrap.github.io/components.html#forms
I have nothing against react-clipboard I don't even know about the project, I am just asking myself if we are introducing dependencies that really bring something or not - maybe it's giving us some nice features and then it's good to keep it ? You can tell us.

mxcube-launcher
mguijarr: I am not talking about zsh I am talking about the whole mxcube-launcher file.

mguijarr:
Thanks Mikel.

mxcube-launcher
meguiraun: yes yes, I understood as the whole file.

meguiraun:
let's see now! we can discuss tomorrow!

PR #519, opened 2017-05-03
Fixes

meguiraun:
a bunch of fixes, see commits for description.


mxcube3/ui/actions/queue.js
```diff
@@ -74,13 +74,11 @@ export function addSamplesToQueue(sampleDataList) {
   return function (dispatch) {
     dispatch(queueLoading(true));
 
-    dispatch(addSamplesToQueueAction(sampleDataList));
-
     sendAddQueueItem(sampleDataList).then((response) => {
       if (response.status >= 400) {
         dispatch(showErrorPanel(true, 'Server refused to add sample'));
-        const sampleIDList = sampleDataList.map((sampleData) => (sampleData.sampleID));
```
mguijarr: @meguiraun 
Small question: before samples were added to the queue (`addSamplesToQueueAction`) - now with the new code, I don't see where they are added ; could you please explain to me how it works, just to keep me in sync.

mxcube3/ui/actions/queue.js
```diff
@@ -402,7 +400,7 @@ export function addTask(sampleIDs, parameters, runNow) {
                      checked: true };
 
       if (!state.queue.queue.includes(sampleID)) {
-        const sample = state.sampleGrid.sampleList[sampleID];
+        const sample = Object.assign({}, state.sampleGrid.sampleList[sampleID]);
```
mguijarr: @meguiraun: why do we need a copy here? (just a small question - if there is a copy here it may indicate our corresponding reducer is wrong, ie. not returning a real new state)

mxcube3/ui/components/Tasks/Characterisation.js
```diff
@@ -193,7 +191,7 @@ Characterisation = connect(state => {
   return {
     path: `${state.queue.rootPath}/${subdir}`,
     filename: `ref-${prefix}_${runNumber}.???`,
-    motorLimits: state.beamline.motorsLimits,
+    motorLimits: state.beamline.movables,
```
mguijarr: @meguiraun is it possible to rename motorLimits to motors? Or something like that. Because in fact it's now the list of movables, not limits. The same goes for all of them. What do you think?

mxcube3/ui/actions/queue.js
meguiraun: we need to check if the sample(s) are already in the queue, and also _push_ for the new task. If we do not copy the object we are actually pushing to the state before anything else happens, and thus, the task will appear in the sample grid before any call to the server is made.

mxcube3/ui/components/Tasks/Characterisation.js
meguiraun: sure! It totally makes sense, gonna change it.

mxcube3/ui/actions/queue.js
mguijarr: You're right - I misread it at first.

PR #523, opened 2017-05-12
Bl movables fixes

meguiraun:
small fixes since beamline set attribute was not working, and also solving https://github.com/mxcube/mxcube3/issues/506

### DIFF
--- a/mxcube3/ho_mediators/beamline_setup.py
+++ b/mxcube3/ho_mediators/beamline_setup.py
@@ -268,7 +268,7 @@ class EnergyHOMediator(HOMediatorBase):
         :rtype: float
         """
         try:
-            self._ho.startMoveEnergy(float(value))
+            self._ho.start_move_energy(float(value))
             res = self.get()
         except:
             raise
--- a/mxcube3/ui/components/PopInput/DefaultInput.jsx
+++ b/mxcube3/ui/components/PopInput/DefaultInput.jsx
@@ -8,15 +8,15 @@ export default class DefaultInput extends React.Component {
 
   constructor(props) {
     super(props);
-    this.state = {value: this.props.value};
+    this.state = { value: this.props.value };
     this.save = this.save.bind(this);
     this.cancel = this.cancel.bind(this);
     this.submit = this.submit.bind(this);
     this.onChange = this.onChange.bind(this);
   }
-  
+
   onChange(event) {
-    this.setState({value: event.target.value});
+    this.setState({ value: event.target.value });
   }
 
   getValue() {
--- a/mxcube3/ui/components/PopInput/DefaultInput.jsx
+++ b/mxcube3/ui/components/PopInput/DefaultInput.jsx
@@ -8,14 +8,19 @@ export default class DefaultInput extends React.Component {
 
   constructor(props) {
     super(props);
+    this.state = {value: this.props.value};
     this.save = this.save.bind(this);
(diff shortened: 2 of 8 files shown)


mxcube3/ho_mediators/beamline_setup.py
```diff
@@ -268,7 +268,7 @@ def set(self, value):
         :rtype: float
         """
         try:
-            self._ho.startMoveEnergy(float(value))
+            self._ho.start_move_energy(float(value))
```
marcus-oscarsson: I've updated this on my side to, alltough I have not made the PR yet.  In any ways not very important 
You should perhaps also pass wait=false to not block the UI

mxcube3/ui/components/PopInput/DefaultInput.jsx
```diff
@@ -8,14 +8,19 @@ export default class DefaultInput extends React.Component {
...
     super(props);
+    this.state = {value: this.props.value};
     this.save = this.save.bind(this);
     this.cancel = this.cancel.bind(this);
     this.submit = this.submit.bind(this);
+    this.onChange = this.onChange.bind(this);
+  }
+  
+  onChange(event) {
+    this.setState({value: event.target.value});
```
marcus-oscarsson: I have  already made a fix for this there is no PR yet. My fix does not involve setState which I think we should avoid.
Is there a reason for the setState other than the obvious one to retrieve the value ?

mxcube3/ui/components/PopInput/DefaultInput.jsx
```diff
@@ -8,14 +8,19 @@ export default class DefaultInput extends React.Component {
...
     super(props);
+    this.state = { value: this.props.value };
     this.save = this.save.bind(this);
     this.cancel = this.cancel.bind(this);
     this.submit = this.submit.bind(this);
+    this.onChange = this.onChange.bind(this);
   }
 
+  onChange(event) {
+    this.setState({ value: event.target.value });
```
marcus-oscarsson: I've already made a fix for this as well I have not made the PR yet though. My fix does not involve setState, I think we should avoid setState when we can. Is there any reason for using it here except for the obvious one to retrieve the value ?

marcus-oscarsson:
It looks like we are overlapping a bit I have also done some of the fixes that you have done, but slightly differently. I'm working on all the actuators and movables so please let me know if you intend to do more changes here. I've been wondering why we are separating between actuator and movable, it seems somehow unnecessary or ? 

Ill publish my PR next week, we can accept this PR, just for your information my PR will contain similar changes.

mxcube3/ui/components/PopInput/DefaultInput.jsx
meguiraun: I do not have any remarkable reason for the `setState`...

meguiraun:
Hej @marcus-oscarsson, you can keep this pr on hold, maybe your fixes are more elegant than  these ones. We can choose next week which fixes are better, there is not point on adding a similar fix twice.

For the actuators and movables, I remember making the split so that the interface populates automatically iterating over the actuators list (~shutters, etc). I would also say that the split is a way of having the list of beamline elements organized in the state. Of course, any improvement is always welcome.

I do not plan to work on this anymore.

In any case, this pr contains some small fixes apart form the set beamline attribute that would need to be merged at some point.

marcus-oscarsson:
Okey, thanks Mikel.

PR #524, opened 2017-05-17
Beamline attributes, movables and actuators

marcus-oscarsson:
Hello !

This is various fixes related to the actuators and movables. They were all working a while back a go when tested on a beamline, various refactoring and updates have then broken the functionality slightly. I could basically not use the controls to control the "movables" (energy, resolution ...). For those I could set the intermediate values (no value changes) were not handled.

So I have merged actuators and movables into one concept called attributes (as in beamline attributes), the name is maybe poorly chosen. It can definitely be changed if someone have a better suggestion. The reason for doing this was simply that the actuators and movables where all handled the same by the backend logic so there is no need to add extra logic to separate them in the UI or in the backend for that matter.

Ive added the machine info in the same way (as an attribute), and added the necessary logic to handle it.
The cryo attribute used a custom made CryoInput which I changed to a PopInput, I dont know what the original intent with that was but it did not seem to provide more functionality, please correct me if Im wrong.

So I finally got all the beamline attributes to work here (on ID29 and mockup ...), set, get and change events. It seems that we need to work a little bit more on abort,  it is not as immediate as one would like (at least in my opinion). A small detail that we need to look at is how to couple the resolution and detector distance in a convenient way.

Thanks,
Marcus

### DIFF
--- a/mxcube3/ho_mediators/beamline_setup.py
+++ b/mxcube3/ho_mediators/beamline_setup.py
@@ -157,6 +157,15 @@ class HOMediatorBase(object):
         """
         self._ho = ho
         self._name = name
+        self._precision = 1
+
+
+    def precision(self):
+        return self._precision
+
+
+    def step_size(self):
+        return math.pow(10, -self._precision)
 
 
     def __getattr__(self, attr):
@@ -248,7 +257,9 @@ class HOMediatorBase(object):
                 "limits": self.limits(),
                 "state": self.state(),
                 "msg": self.msg(),
-                "type": "FLOAT"
+                "type": "FLOAT",
+                "precision": self.precision(),
+                "step": self.step_size()
                 }
 
         return data
@@ -270,6 +281,7 @@ class EnergyHOMediator(HOMediatorBase):
     def __init__(self, ho, name=''):
         super(EnergyHOMediator, self).__init__(ho, name)
         ho.connect("energyChanged", self.value_change)
+        self._precision = 4
 
 
     def set(self, value):
@@ -300,7 +312,8 @@ class EnergyHOMediator(HOMediatorBase):
         """
         try:
             energy = self._ho.getCurrentEnergy()
-            energy = "{:3.4f}".format(round(float(energy), 4))
+            energy = round(float(energy), self._precision)
+            energy = ("{:3.%sf}" % self._precision).format(energy)
         except (AttributeError, TypeError):
             raise ValueError("Could not get value")
 
@@ -343,6 +356,7 @@ class WavelengthHOMediator(HOMediatorBase):
     def __init__(self, ho, name=''):
         super(WavelengthHOMediator, self).__init__(ho, name)
         ho.connect("energyChanged", self.value_change)
+        self._precision = 4
 
 
     def set(self, value):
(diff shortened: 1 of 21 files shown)


meguiraun:
hej @marcus-oscarsson, thanks for this, I am looking at this and one thing I notice is that `ui/components/Tasks/DataCollection.js` (and rest of the tasks) need to be adapted.

I'll keep looking and testing

mxcube3/ho_mediators/beamline_setup.py
```diff
@@ -728,7 +728,7 @@ def dict_repr(self):
 class DetectorDistanceHOMediator(HOMediatorBase):
     def __init__(self, ho, name=''):
         super(DetectorDistanceHOMediator, self).__init__(ho, name)
-        ho.connect("positionChanged", self.value_change)
+        #ho.connect("positionChanged", self.value_change)
```
meguiraun: then, how do we notify the client about changes in dtox?

marcus-oscarsson: Well at the moment we don't, its related to what I wrote at the end of my message. We need to find a good way to aggregate the resolution and the detector distance. I added the comment since it was causing me some problems when I was testing, its possible that you are not experiencing the same problem. But we will of course have a way to get those updates

marcus-oscarsson: Ill make a proposal for the resolution and detector distance, I think that will be in another PR if you dont mind.

meguiraun: I have not seen anything on our side, probably because for us dtox is a SardanaMotor. I totally agree on having separate PRs, better that way for everyone ;)

mxcube3/ho_mediators/beamline_setup.py
```diff
@@ -243,6 +269,7 @@ def value_change(self, *args):
         Signal handler to be used for sending values to the client via socketIO,
         data should normally be sent in the "hwr" namespace.
         """
+        print("Emitting beamline value change")
```
meguiraun: I missed this one... could you remove the print? (sorry for being picky)

mxcube3/ho_mediators/beamline_setup.py
```diff
@@ -383,10 +413,8 @@ def limits(self):
         :returns: The limits.
         """
         try:
-            energy_limits = self._ho.getLimits()
+            energy_limits = self._ho.getEnergyLimits()
```
meguiraun: don't we need to do the unit conversion?

marcus-oscarsson: We dont have a method called getLimits, simply getEnergyLimits

marcus-oscarsson: You seem to suggest that you do some additional unit conversion ?

marcus-oscarsson:
Okey so made the necessary changes to "aggregate" the resolution and detector distance. They are actually already aggregated by the resolution object so its just a matter of retrieving the detector distance through the resolution. So Im doing that now, which is at least better than before. Sorry for not waiting with that for another PR as we said, it was such a small change so it was simply better to do it right away.

mxcube3/ho_mediators/beamline_setup.py
meguiraun: I would say return the wavelength limits in A instead of eV. How is it done now?

meguiraun:
Thanks @marcus-oscarsson, just one more thing... `ui/components/Tasks/validate.js` needs also fixes. A refactoring is never easy...

No problem with the resolution/detdist, I though you were thinking in a bigger change.

mxcube3/ho_mediators/beamline_setup.py
marcus-oscarsson: Ahh, yes for the wavelength object sorry missed that, yes its getWavelengthLimits

PR #526, opened 2017-05-19
resolution related hwobj

meguiraun:
small fixes

### DIFF
--- a/mxcube3/ho_mediators/beamline_setup.py
+++ b/mxcube3/ho_mediators/beamline_setup.py
@@ -679,7 +679,8 @@ class ResolutionHOMediator(HOMediatorBase):
         super(ResolutionHOMediator, self).__init__(ho, name)
         ho.connect("valueChanged", self.value_change)
         self._precision = 3
-
+        self._ho.energy_hwobj = self.getObjectByRole("energy")
+        self._ho.dtox = self.getObjectByRole("dtox")
 
     def set(self, value):
         self._ho.move(round(float(value), 3))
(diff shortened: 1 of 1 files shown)


mxcube3/ho_mediators/beamline_setup.py
```diff
@@ -679,7 +679,8 @@ def __init__(self, ho, name=''):
         super(ResolutionHOMediator, self).__init__(ho, name)
         ho.connect("valueChanged", self.value_change)
         self._precision = 3
-
+        self._ho.energy_hwobj = self.getObjectByRole("energy")
+        self._ho.dtox = self.getObjectByRole("dtox")
```
marcus-oscarsson: Hm, this should be done by the actual hardware object in this case Resolution, its at least the case in
HardwareObjects/Resolution.py. So there should not be need add this extra assignment.

marcus-oscarsson: You do need to have the correct configuration in the corresponding resolution.xml for that to work though

meguiraun: that's what I though! But then, the mockup does not do the job properly.  I just using master files...

Gonna check more carefully...

marcus-oscarsson: Ahh, yes Ive made some local changes to mockups that I did not commit yet. Ill commit those later

marcus-oscarsson: There are some a PR for the HardwareObjects that improve the ResolutionMockup object, give it a try.

meguiraun:
with you hwobj pr this is no longer neeeded. Thanks

PR #528, opened 2017-05-19
Sample Image - Grid

marcus-oscarsson:
Hey !

Small fix for an issue that occurred when showing the context menu for grids.
Also not showing extra divider when there are no workflows.

Marcus


meguiraun:
thanks again!

PR #529, opened 2017-05-19
Sample Image - Incorrect number of cols

marcus-oscarsson:
Hi,

Just something I noticed when I was skimming through the code, the number of columns did
not add up to 12.

Marcus


meguiraun:
mxcube 3 is on fire!!

PR #531, opened 2017-05-19
Session - Fixing small issue with login

marcus-oscarsson:
Hello !

Fixing small issue with login, session = proplist[0]['Proposal']['number']['Sessions'][0] generates an exception effectively resetting proposal_code and proposal_number to None.

Cheers,
Marcus


meguiraun:
thanks!

Issue #481, opened 2017-03-10
Add proper detector filename extension (.cbf, .h5, ...) to full path


mguijarr:
The filename extension depends on the detector, and should come from the server.

meguiraun:
The `queue_model_object.PathTemplate.intersection` method should be adapted as well.

Issue #517, opened 2017-04-28
Sample name issues


meguiraun:
Cannot select the sample name string displayed in the current queue
Cannot rename a sample.

meguiraun:
In addition:
* Naming convention: <sample_name>_<position_name>_<run_number>_master.<file_type>
  * <position_name>: P2, L1,...
  * <run_number>: set automatically by mxcube, user **cannot** change it
  * <file_type>: h5, cbf, ...  (see also #481)

PR #545, opened 2017-05-29
Issue 481 detector file extension

meguiraun:
solving #481, The filename extension depends on the detector, and should come from the server.


meguiraun:
I took this one because it is related/required to #517

Issue #521, opened 2017-05-11
Selecting task path in text box


marcus-oscarsson:
It used to be possible to select the path or parts of the path to manually to for instance copy it.
Something that is no longer possible, It should be possible to select the path or parts of the path
for copying.

Trying to select the text now simply drags the entire task

meguiraun:
mmm, the same happens with the Sample Name in the top of the current queue. possibly related? #517

marcus-oscarsson:
I think its the drag event that is the problem so its likely that also effects the selection of other elements
within the TaskItem.

mguijarr:
What about removing drag and drop ?

meguiraun:
I vote for removing the drag&drop, fix this issue and if our scientist request it bring back the task drag&drop later

meguiraun:
it is back again, possibly due to #574

marcus-oscarsson:
Done by #582

PR #547, opened 2017-06-02
Issue 517 sample naming and 521

meguiraun:
* solving #517 
  * Cannot select the sample name string displayed in the current queue
  * Naming convention: <sample_name><position_name><run_number>_xxxx.<file_type>
* solving #521: without removing drag&drop
* also a but where `defaultPrefix` was lost after a page refresh
----
* Not done: rename a sample in # 517


marcus-oscarsson:
Thanks

Issue #511, opened 2017-04-19
Adding a task to multiple points


meguiraun:
As we have discussed already, here is the issue!

JieNanMAXIV:
Hi, when selecting 3 points or more, one can still set up a helical scan with the first two points, which doesn't make much sense. . Probably should disable the helical scan in this case.

PR #573, opened 2017-06-20
Issue 511 add tasks to multiple points

meguiraun:
solving #511 

Looks like it is working but since I am messing up a bit with tasks creation I still want to test a bit more. But you know what I am doing :)

If you select several points you can add a task to al of them at once (if number of points !=2, helical is not available)
Workflows do not work properly (although you can add the very first time), something related to the run_number update... looking at it.


meguiraun:
free to test (and merge) now.

I found a bug where the task item was lost wf parameters on page refresh. I added all the parameters at the queue node level (qmo.workflow) and fix the retrieval.

PR #574, opened 2017-06-20
Layout updates for interleaved

marcus-oscarsson:
Dear all,

I will mark this PR as WIP for the moment and I hope it can be some basis for discussion, regarding the interleaved. We are in general happy with it and it can be accepted as it is.

This PR introduces some of the layout changes that we spoke about when we discussed the interleaved
feature. It adds a more table like structure for the tasks which will make it easier to create and get a good
overview of the interleave parameters.

I also took care of a few other things that are of lesser importance but that still needed to addressed in some way. 

- Flux was removed from the machine current display and added in its own display component
   (LabeledValue)
     
- Also add a component called LabeledPopInput which displays the beamline attributes vertically,
   which maybe looks more consistent.

 - The video display now takes 7 instead of 8 columns which maybe does not look very nice at the
    moment but it removes the vertical page scroll. Its a solution until we come up with a better way
    of setting the video height. Also we might need more vertical space for the queue so it gives us
    some idea of how it will look like.

Cheers,
Marcus


meguiraun:
thanks for this! Looks really nice!

Just some small comments:
* The resolution unit is wrong on the top bar
* It is difficult to differentiate several tasks specially if they are not run. Could we compress/indent a bit the table and the path line
* The progress bar also appears in run tasks

On another level, I do not see beamline actions (list empty) neither any actuator.

marcus-oscarsson:
Hey !

Feel free to merge when you like, see my comments below

- I have all beamline actions and actuators, are you using mockups, and are you using the same mockups as me, for instance ShutterMockup and MicrodiffInOutMockup ?. 

- I don't have any progress bar in run tasks, do you mean the text ?

meguiraun:
I mean that I see the progress bar on top of an already executed task (green).

PR #575, opened 2017-06-21
small bug fix

meguiraun:
remove unused method call, missed from previous pr

### DIFF
--- a/mxcube3/routes/Login.py
+++ b/mxcube3/routes/Login.py
@@ -125,7 +125,6 @@ def loginInfo():
 
     login_info = login_info["loginRes"] if login_info is not None else {}
     login_info = limsutils.convert_to_dict(login_info)
-    limsutils.update_mxcube_session(login_id, login_info)
 
     res = {"synchrotron_name": mxcube.session.synchrotron_name,
            "beamline_name": mxcube.session.beamline_name,
--- a/test/HardwareObjectsMockup.xml/lims.xml
+++ b/test/HardwareObjectsMockup.xml/lims.xml
@@ -4,7 +4,7 @@
   </ws_root>
   <ws_username></ws_username>
   <ws_password></ws_password>
-  <loginType>user</loginType>
+  <loginType>proposal</loginType>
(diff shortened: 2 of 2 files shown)


mxcube3/routes/Login.py
```diff
@@ -125,7 +125,6 @@ def loginInfo():
 
     login_info = login_info["loginRes"] if login_info is not None else {}
     login_info = limsutils.convert_to_dict(login_info)
-    limsutils.update_mxcube_session(login_id, login_info)
```
marcus-oscarsson: Hm, how is this unused ?

meguiraun: because it is done in the select_proposal

marcus-oscarsson: It looks to me like some logic that once were in limsutils is now missing. The method update_mxcube_session made sure that the mxcube Session object was kept in sync when for instance the server is restarted.

Did you put it somewhere else ?, the data paths will not be correct if this information is not kept.

meguiraun: see https://github.com/mxcube/mxcube3/blob/master/mxcube3/routes/limsutils.py#L80

Of course... there could be a bug

marcus-oscarsson: Hm, well okey and what calls select_proposal ?, what hapends for instance when the server is restarted and we reconnect the client ? How is the mxcube Session object updated ?

Well in anyways, I guess Ill accept this and we have to fix it later ...

meguiraun: select_proposal is called in loginInfo, I think this is the same solution as the other method, just taking into account user/proposal login. Did you try running?

No need to accept per se if you do not agree. I can modify on this pr.

marcus-oscarsson: Hm, okey Ill merge now and I will try on the beamline later.

Thanks Mikel

meguiraun: thanks, let me know if you have issues, this is a tricky part and different in our sites.

PR #577, opened 2017-06-26
Mesh results

meguiraun:
Setting the basic functionality. A new overlay control for setting the opacity of the grid and cell results. At this moment I fill the cells with random values, so we can test the heat map function and so on (I will remove it later).

image


meguiraun:
You can merge as it is if you do not mind to have the random cell filling.

meguiraun:
I have been having some fun with the pr... just a note: small changes in the hwobj are needed and not yet accepted in the pending PRs (I still need to address the comments)

meguiraun:
so far...

I added a new route (`/sampleview/shape_mock_result/<sid>`), so if you do a GET on that a mockup-random grid result is linked to the given shape (grid) and send to the client through socketio channel, and stored in the app state.

The heatmap is then calculated and displayed if you ask for that. 

Still WIP because:
- I would like to have your review, specially for the code in `DrawGridPlugin`
- Small additions are needed in the HR repo

image

mxcube3/__init__.py
```diff
@@ -158,6 +161,7 @@ def complete_initialization(app):
         except Exception:
             sys.excepthook(*sys.exc_info())
 
+        logging.getLogger("HWR").info("Mxcube3 server initialized, it took %.1f seconds" % (time.time() - t0))
```
marcus-oscarsson: I guess this some kind of debug information, is it something that you are still using ? Does it mean that we have some un-deterministic delay when starting the server ?

If its meant to be permanent could we fix the casing of Mxcube3

mxcube3/routes/Queue.py
```diff
@@ -532,6 +532,40 @@ def get_default_char_params():
...
+            'resolution': acq_parameters.resolution,
+            'energy': acq_parameters.energy,
+            'transmission': acq_parameters.transmission,
+            'shutterless': acq_parameters.shutterless,
+            'detector_mode': acq_parameters.detector_mode,
+            'inverse_beam': False,
+            'take_dark_current': True,
+            'skip_existing_images': False,
+            'take_snapshots': True,
+            'cell_counting': mxcube.beamline['default_mesh_values'].getProperty('cell_counting', 'None'),
```
marcus-oscarsson: Maybe replace None with a method that we know exists, like zig-zag so that we by accident won't get None as method ?

mxcube3/routes/SampleCentring.py
```diff
@@ -242,6 +242,36 @@ def get_shapes():
     return resp
 
 
+@mxcube.route("/mxcube/api/v0.1/sampleview/shapes/<sid>", methods=['GET'])
+def get_shape_with_sid(sid):
+    """
+    Retrieve all the stored centred positions.
```
marcus-oscarsson: Check this comment, status 409 is never set if there is an error and it refers to centered position where I should be shape

mxcube3/routes/SampleCentring.py
```diff
@@ -264,8 +294,16 @@ def update_shapes():
         # Shape does not have any refs, create a new Centered position
         if not refs:
             x, y = shape_data["screen_coord"]
-            mpos = mxcube.diffractometer.get_centred_point_from_coord(x, y)
-            shape = mxcube.shapes.add_shape_from_mpos([mpos], (x, y), t)
+            mpos = mxcube.diffractometer.get_centred_point_from_coord(x, y, return_by_names=True)
+            pos.append(mpos)
+            # MD3 needs the center of the grid instead of the top left corner
```
marcus-oscarsson: The position of the Shape object should always be top left corner, we can add a method or similar that calculates the center position if needed. This could also and should perhaps be handled in the diffractometer specific code. I guess it wont work for routines that expect position to be top left corner like it is now ?

mxcube3/routes/qutils.py
```diff
@@ -681,6 +681,17 @@ def set_dc_params(model, entry, task_data):
         acq.acquisition_parameters.centred_position = cpos1
         acq2.acquisition_parameters.centred_position = cpos2
 
+    elif params["mesh"]:
+        grid = mxcube.shapes.get_shape(params["shape"])
+
+        acq.acquisition_parameters.mesh_range = grid.get_grid_range()
```
marcus-oscarsson: What is the range and how is it used, is it not the same as width or height ?

mxcube3/ui/components/SampleView/ContextMenu.js
```diff
@@ -99,11 +100,27 @@ export default class ContextMenu extends React.Component {
   showModal(modalName, wf = {}, _shape = null) {
     const { sampleID, defaultParameters, shape, sampleData } = this.props;
     const sid = _shape ? _shape.id : shape.id;
+    const cellCount = shape.gridData ? defaultParameters[modalName.toLowerCase()].cell_counting
+      : 'None';
+
+    let aux;
```
marcus-oscarsson: So aux is actually num_images ?

mxcube3/ui/components/SampleView/DrawGridPlugin.js
```diff
@@ -2,7 +2,7 @@ import 'fabric';
 const fabric = window.fabric;
 
 /**
- * Fabric Shape for drawing grid (defined by GridData)
+ * Fabric Shape for d§wing grid (defined by GridData)
```
marcus-oscarsson: Please fix this comment

mxcube3/ui/components/SampleView/SampleImage.jsx
```diff
@@ -417,13 +515,28 @@ export default class SampleImage extends React.Component {
...
     if (!this.props.drawGrid) {
       this.hideGridForm();
       this.drawGridPlugin.reset();
     }
   }
 
+  updateGridResults() {
+    const gd = this.selectedGrid();
+
+    console.log('updating grid result... ', this.props);
```
marcus-oscarsson: Please grep for console.log and remove them if you dont need them

marcus-oscarsson:
Hey Mikel,

As you know Im vacation but im taking a look here in anyways, I noticed that you asked for a review so I   tried to give you some feedback. Thanks for all the nice work looks like we are almost there. I can't test with the environment I have here so I did not actually run the code.

I started to look at the semantics of the results and I noticed a comment here:
   https://github.com/mxcube/mxcube3/pull/577/files#diff-a310a5ec003226fd77f95daea98c208cR272

Does it mean that you would like to add something to make it complete ?. Im not sure we need a specific method called showHeatmap, it was atleast not my idea when I did the grid. Each cell have a result value and its either drawn or not drawn depending on various conditions. Like if there is a result at all or if the user does want to display it. We just need to update the grid data and draw it on screen ... Maybe thats what you already are doing and I misunderstood the thing, maybe the method name confuses me ?

Regards,
Marcus

meguiraun:
hej @marcus-oscarsson! Hope you had good time! I will be off until Thursday and I will address your comments then. Perhaps we can have a skype talk and discuss this pr: I am adding all the result to the griddata and let the user chose if she/he wants to see the heatmap, hence the checkbox.

For the comment you noticed, as the current status of this code, there is something wrong with the redux state, I need this line: `this.props.sampleActions.updateShape(gridData);` in order to trigger the rendering (and I do not know what I am missing, since it should not be needed).

marcus-oscarsson:
Hey !

Allright then I understand a bit more, okey sounds like a good idea with a Skype meeting, I wanted to give you some feedback since you asked for a review. I won't be back until  Wednesday next week so the skype meeting have to wait a few days. Enjoy your time off see you in roughly a week then !

Marcus

mxcube3/__init__.py
meguiraun: I see much longer time now in the beamline, ~35 s. But, we also have more hwobj in place so I cannot tell if there is something around mxcube3 or just Maxiv hwobjs. Also, you never know which is the last step of the process, so a msg at the end seems useful.

mxcube3/routes/SampleCentring.py
meguiraun: I do not remove the top-left corner, just adding an extra field for the center of the grid. Which is then applied in qutils.py L688 `acq.acquisition_parameters.centred_position = grid.get_centred_position()`. I do not know about other diffs but ours needs the center position, and the top-left pos is needed for the client.

mxcube3/routes/qutils.py
meguiraun: basically yes, I added the method for convenience

mxcube3/ui/components/SampleView/ContextMenu.js
meguiraun: Is more that num_images will be aux... the num_images is determined by the num_{cols, rows} and the cell-counting method and then fixed after so the user cannot change it.

mxcube3/routes/SampleCentring.py
meguiraun: I will try to move out into specific hwobj

mxcube3/ui/components/SampleView/ContextMenu.js
meguiraun: we keep on hold this issue since we have different opinions

mxcube3/routes/SampleCentring.py
meguiraun: mmm.. I having trouble for moving down in the hierarchy, neither collect_hwobj or diff_hwobj know anything about the shape selected or shape's details/params. I do not want to start creating custom hwobj for the shapes...

What about each grid has two positions (top-left and center) as I did here before, but when creating the acquisition parameters I select which cpos to use for the mesh:

```
mesh_center = mxcube.beamline['default_mesh_values'].getProperty('mesh_center', 'top-left')

if mesh_center == 'top-left':
    acq.acquisition_parameters.centred_position = grid.get_centred_positions()[0]
 else:
    acq.acquisition_parameters.centred_position = grid.get_centred_positions()[1]
```
Unless you have a better idea...

meguiraun:
To be discussed during next meeting with scientist:
* How do we treat image numbers on the modal? here it is num_images per line
* Horizontal/vertical spacing: here only one can be displayed at a time (configurable in xml: vert or hor or none)

meguiraun:
let's see now.

* Both horizontal and vertical spacing appear in the xml contains both.
* Mesh modal shows the total number of images (= cell count)

mxcube3/routes/SampleCentring.py
marcus-oscarsson: Okey we can keep the center, just make sure that we always have the top left corner and center for all grids independent of diffractometer. We can also remove the MD3 specific comment.

marcus-oscarsson:
Thanks, I will test this on the beamline I hope that we can run some grids scans at the beginning of next week.

meguiraun:
great, I guess that there are still differences between us... let's hope they are small

marcus-oscarsson:
Could you remove the MD3 specific comment on line 299 of SampleCentring.py ?, we just need to say that we also add the center for grids.

Marcus

marcus-oscarsson:
Ive done shallow tests on one of our beamlines here and it seems to work fine, Ive not yet started the integration, I hope to start that on Monday. Ive noticed something that is a bit disturbing but I guess not critical for the moment; I did some tests with the latest chrome on linux. The entire browser crashes when using this PR. It looks like it have something to do with WEB GL and it was only with the latest stable Chrome on Linux.

I don't think this is a huge problem because I suspect that it has something to do with the graphics drivers on that computer, but please keep an eye on that kind of behavior, is it something you have seen ?

Marcus

mxcube3/ui/components/SampleView/ContextMenu.js
marcus-oscarsson: Do we still need to pass numCols or numRows as number of images ?

meguiraun:
I have not seen what you describe, I saw a few times (months ago with our older branch in the beamline) that firefox occasionally went unresponsive. But I have not seen that recently

mxcube3/ui/components/SampleView/ContextMenu.js
meguiraun: yes, because num_images is the value that will go to the acquisition parameters. It is removed from the modal display (substituted by total_num_images) but still needed

marcus-oscarsson: I think that num_images should always be num_images, if we need to pass num cols and num rows we should pass them as num_cols and num_rows and use them where needed.

num_images should always be total number of images

marcus-oscarsson:
Okey, thanks for the fixes. I'm ready to merge when we all agree on how to pass num_cols, num_rows, and num_images.

meguiraun:
I have changed a few things and now I pass num_images, num_cols and num_rows. And then, rigth before adding the task to queue I decide which will be the applied value for the num_images parameter (based on the zig-zag method and num_{cols, rows})

marcus-oscarsson:
Ive tried to integrate all this and it seems to roughly work here. There seem to be some issues with the results part that we need to look at but maybe in another PR.

So we can now run and display the results of the mesh scan, I needed to change the result format to make it work. We can discuss the format later and Ill make a PR with my changes. It also seems like we only can have one result at a time, that is the same result is applied to all grids.

In anyways, thanks for the work done :) Id suggest that we resolve the remaining issues in another PR.

Here is how it currently looks like when you run mesh scans
mx3-3-mesh-2
mx3-3-mesh

meguiraun:
Nice pictures!! I would be very happy if we finally can close this pr!

For the pending things as the result data format, we need to discuss about it because I think we have ~opposing ideas.

PR #578, opened 2017-06-26
Fixing minor problem with code related to login

marcus-oscarsson:
Hey !

Solving a minor problem that prevents us from logging in properly. The proposal was sometimes passed as only the proposal number but it the entire proposal  (code and number) is needed.

Also removed code that handled differences between mockups and concrete implementations. The mockup objects should return the data in the same format as the real objects. Any differences between formats should either be in site specific code (if really needed) or in "extreme/unusual" cases agreed upon beforehand.

Marcus


meguiraun:
thanks @marcus-oscarsson

PR #579, opened 2017-06-26
Adding tasks even if no shape defined

marcus-oscarsson:
Hey !

The "fast" task creation broke the task creation from the sample grid, so heres a fix.

Also run_number should be handled on the sever and there should be a site specific option
for those who does not want to have the a shape identifier, the indexOf can give the wrong
number even if sorted, we need to use the actual ID of the shape.

Marcus


meguiraun:
thanks!

PR #581, opened 2017-06-28
update and lock versions

meguiraun:
a bit of cleaning and version locking since master was broken due to a mismatch in version numbers.


marcus-oscarsson:
Hi

Sorry for not commenting on this earlier, which were the package mismatches ? And do we really need to lock the versions at this stage ?

Marcus

meguiraun:
react version was not the correct one, I would lock since that way we do not risk that much breaking master, it happen from time to time and we only see when we do a fresh clone&install. The version numbers that I put are the ones that I have in my machine, we can put some other versions as well if preferable.

mguijarr:
Hello :)

We would prefer if versions were updated, but not locked - because we would like to have the opportunity to update all packages from time to time quite easily, to follow the evolution of libraries. If something starts to break, then it needs to be fixed (like we did for redux-form and react-bootstrap in the past).

So, **for development** I would like to always be able to do `npm update` and get newer versions.

For **production**, of course we don't want that - normally we should create production bundles with `webpack -p` (we have `npm run production` if I remember well), and then we should only serve the production files.

Instead of locking I would then recommend to have versions like this:

    "dependencies": {
        "dep1": "^0.4.0"
        ...
    }

with `0.4.0` being the one we put in production last, for example. Then if we do `npm update` we can get `0.5` for example.

Some people may think it is a dangerous approach but we didn't experience so much trouble in the past (did we ?) and so I would advocate to continue like this for now.

meguiraun:
I do not mean not to update versions if needed. I want to avoid a master that can not be tested, at this moment there is a small issue in the package.json, and we have had that quite a few times before. As soon as we need a newer package, we can absolutely update and lock for such new version.

meguiraun:
since there is no consensus I will close this, we will need to discuss version locking with/after the fist release.

Issue #584, opened 2017-06-30
porposal_list not defined 


marcus-oscarsson:
Hi,

The login routine is giving me some errors here it seems like proposal_list is not defined
the first time loginInfo is called so we get an attribute error.

It seems like the proposal_list is also attached on the fly to the mxcube session, which is part of the problem.
It should either be defined on the Session hardware object directly or perhaps in the flask session.

Marcus

meguiraun:
ok, gona have a look

meguiraun:
https://github.com/mxcube/mxcube3/pull/587 solves the issue (without moving the dict to somewhere else)

marcus-oscarsson:
Sorry for being persistent but is still get this AttributeError, not urgent to fix, but if you have the time please take a look.

You get it when you for instance kill the server, but stay logged in, then restart the server and refresh the page.

Cheers,
Marcus

meguiraun:
see #608

mguijarr:
Seems like it is fixed

PR #587, opened 2017-07-04
fix for the first call to login_info

meguiraun:
solving #584

### DIFF
--- a/mxcube3/routes/Login.py
+++ b/mxcube3/routes/Login.py
@@ -135,7 +135,7 @@ def loginInfo():
            "observerName": remote_access.observer_name()
            }
 
-    if res["loginType"].lower() != 'user':
+    if res["loginType"].lower() != 'user' and login_info:
         # autoselect proposal
         limsutils.select_proposal(LOGGED_IN_USER)
         res["selectedProposal"] = LOGGED_IN_USER


marcus-oscarsson:
Hey Mikel !

This probably solves the problem of proposal_list not being defined but it suggest that select_proposal
must be called in a certain order dependent of some other logic. Id prefer if proposal_list was properly
defined so that we always can call select_proposal with a deterministic result.

Whats your opinion on that, am I somehow mistaken ?

Marcus

meguiraun:
The select proposal is only called after a successful login. The first call to `login_info` does nothing, then `login` is successful and then, the next call to `login_info` does the select_proposal. It assumes that a user/proposal has logged in successfully, if not, it should not be executed.

I do not fully grasp what you mean, perhaps the current order is a bit messy and we just need to put some better order? Or you prefer that we always call it `set_proposal` but it checks the login state itself?

I would suggest to accept this pr, since it essentially solves the issue. We can discuss tomorrow any improvement on the login flow. Would that be ok? I am not against your idea, but I think it needs a bit of discussion, and the refactoring can be a bit tricky.

marcus-oscarsson:
What I don't find very appealing is that we get an attribute error (that is we have a variable that is not defined) if we don't execute things in a certain order or context. So we will probably need to change that at the time when or if we move the call to select_proposal. 

The solution would simply be to define the selected_porposals list in the right place

meguiraun:
with this fix the attribute error disappears, you will not access that variable if the conditions are not meet.

PR #593, opened 2017-07-12
Task progress

marcus-oscarsson:
Hello,

Improved the task progress, the progress is sent as a value between 0 and 1 and displayed in the
per task progress bar on the client.

The implementation of the progress for the interleaved is not very elegant and I hope to fix that a later moment. It requires some alterations to the routine that performs the interleaved, so Ill do that for the
next update of HardwareRepository.

Cheers,
Marcus

### DIFF
--- a/mxcube3/routes/SampleCentring.py
+++ b/mxcube3/routes/SampleCentring.py
@@ -30,12 +30,8 @@ def init_signals():
     callback method.
     """
     for signal in signals.microdiffSignals:
-        if signal in signals.task_signals:
-            mxcube.diffractometer.connect(mxcube.diffractometer,
-                                          signal, signals.task_event_callback)
-        else:
-            mxcube.diffractometer.connect(mxcube.diffractometer, signal,
-                                          signals.motor_event_callback)
+        mxcube.diffractometer.connect(mxcube.diffractometer, signal,
+                                      signals.motor_event_callback)
     for motor in mxcube.diffractometer.centring_motors_list:
         @Utils.RateLimited(3)
         def pos_cb(pos, motor=motor.lower(), **kw):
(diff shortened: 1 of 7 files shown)


mxcube3/routes/qutils.py
```diff
@@ -1081,15 +1079,18 @@ def init_signals(queue):
     """
     import signals
 
-    for signal in signals.collect_signals:
-        mxcube.collect.connect(mxcube.collect, signal,
-                               signals.task_event_callback)
```
JieNanMAXIV: With this change, task_signals including "collect_finished" and "warning" are not connected to any event and will not be handled.

mxcube3/routes/SampleCentring.py
```diff
@@ -30,12 +30,8 @@ def init_signals():
     callback method.
     """
     for signal in signals.microdiffSignals:
-        if signal in signals.task_signals:
-            mxcube.diffractometer.connect(mxcube.diffractometer,
-                                          signal, signals.task_event_callback)
```
JieNanMAXIV: how about the warning signal? it will be connected to motor_event_callback

mxcube3/routes/qutils.py
marcus-oscarsson: Yes it will, because its explicitly connected at for [instance](https://github.com/mxcube/mxcube3/pull/593/files/6e8794a3cebaa651b08a8df8dd34c99c36cc646e#diff-aaa7ddef1045c72f72d678cdfc7b7c84R1091)

They way it was done before was to ambiguous and could potentially lead to clashes depending on the signal name. So Id prefer if we could use a more explicit  way, either like its now or some other way if you have another idea.

mxcube3/routes/SampleCentring.py
marcus-oscarsson: The list of signals might need to be cleaned up a bit and I might have made a mistake here, but I cant really find were the signal warning is emitted, also it seems to be present in other signals lists as well which makes me wonder a little bit

marcus-oscarsson:
What do you think, do you have some further comments or can we merge ?

Issue #48, opened 2015-11-17
isotope-layout problems with NPM V.3


bolmsten:
It seems that Isotope-layout has a problem when run with webpack and NPM V.3,  see 

https://github.com/metafizzy/isotope/issues/979

This will be resolved in the future and all that is needed is to download the latest version of Isotype.

mguijarr:
Are you using npm v3 ? Here I am still running node 0.12.7 with npm 2.11.3

bolmsten:
I was using v3 but I have now changed to 2.14.7 and it works. I was just adding this as a reminder for the future to update Isotope.

mguijarr:
ok thanks !

mguijarr:
I think this is fixed by PR #95

PR #597, opened 2017-08-04
Adapt support for SampleChanger / add SampleChangerMaintenance

vrey01:
A wiki page explains all the changes proposed by this PR. 

The PR is mostly related with SampleChanger in mxcube3 but contains also another two small commits.
(1)- fix for mach_info_updated()  that had been lost 
(2)- change method names in "scutils" to use those of the GenericDiffractometer rather than old method names.

(3)- Bigger commit with all the changes related with SampleChanger support. [see wiki for this]

@meguiraun

### DIFF
--- a/mxcube3/routes/SampleChanger.py
+++ b/mxcube3/routes/SampleChanger.py
@@ -7,13 +7,18 @@ from mxcube3 import app as mxcube
 from . import limsutils
 from . import scutils
 
-from .qutils import UNCOLLECTED
+from .qutils import UNCOLLECTED, SAMPLE_MOUNTED
 from .scutils import set_current_sample
 
 
 def init_signals():
     """Initialize hwobj signals."""
     mxcube.sample_changer.connect('stateChanged', signals.sc_state_changed)
+    mxcube.sample_changer.connect('loadedSampleChanged', signals.loaded_sample_changed)
+    mxcube.sample_changer.connect('contentsUpdated', signals.sc_contents_update)
+
+    if mxcube.sc_maintenance is not None:
+        mxcube.sc_maintenance.connect('globalStateChanged', signals.sc_maintenance_update)
 
 
 @mxcube.route("/mxcube/api/v0.1/sample_changer/samples_list", methods=['GET'])
@@ -24,6 +29,12 @@ def get_samples_list():
     order = []
 
     for s in samples_list:
+        if not s.isPresent():
+            continue
+        if s.isLoaded():
+            state = SAMPLE_MOUNTED
+        else:
+            state = UNCOLLECTED
         sample_dm = s.getID() or ""
         coords = s.getCoords()
         sample_data = {"sampleID": s.getAddress(),
@@ -31,7 +42,7 @@ def get_samples_list():
                        "sampleName": "Sample-%s" % s.getAddress().replace(':', ''),
                        "code": sample_dm,
                        "loadable": True,
-                       "state": UNCOLLECTED,
+                       "state": state,
                        "tasks": [],
                        "type": "Sample"}
         order.append(coords)
@@ -45,6 +56,25 @@ def get_samples_list():
     order.sort()
 
     return jsonify({ 'sampleList': samples, 'sampleOrder': [samplesByCoords[coords] for coords in order] })
+
+@mxcube.route("/mxcube/api/v0.1/sample_changer/state", methods=['GET'])
+def get_sc_state():
+    state = mxcube.sample_changer.getStatus().upper()
+    
+    return jsonify({'state': state})
+
+@mxcube.route("/mxcube/api/v0.1/sample_changer/loaded_sample", methods=['GET'])
+def get_loaded_sample():
+    sample = mxcube.sample_changer.getLoadedSample()
+
+    if sample is not None:
+       address = sample.getAddress()
+       barcode = sample.getID()
+    else:
+       address = ''
+       barcode = ''
+    
+    return jsonify({'address': address, 'barcode': barcode})
 
 @mxcube.route("/mxcube/api/v0.1/sample_changer/contents", methods=['GET'])
 def get_sc_contents():
@@ -68,7 +98,7 @@ def get_sc_contents():
         return ""
 
     def _addElement(parent, element):
-        new_element = { "name": element.getAddress(),
+        new_element = { "name": element.getAddress(), 
                         "status": _getElementStatus(element),
                         "id":_getElementID(element),
                         "selected": element.isSelected() }
@@ -84,7 +114,8 @@ def get_sc_contents():
     contents = { "name": root_name }
 
     for element in mxcube.sample_changer.getComponents():
-        _addElement(contents, element)
+        if element.isPresent():
+           _addElement(contents, element)
 
     return jsonify(contents)
 
@@ -111,6 +142,12 @@ def unmount_sample(loc):
     set_current_sample('')
     return get_sc_contents()
 
+@mxcube.route("/mxcube/api/v0.1/sample_changer/unmount_current/", methods=['GET'])
+def unmount_current():
+    mxcube.sample_changer.unload()
+    set_current_sample('')
+    return get_sc_contents()
+
 @mxcube.route("/mxcube/api/v0.1/sample_changer/mount", methods=["POST"])
 def mount_sample_clean_up():
     try:
@@ -128,3 +165,46 @@ def unmount_sample_clean_up():
         return Response(status=409)
     else:
         return Response(status=200)
+
+@mxcube.route("/mxcube/api/v0.1/sample_changer/get_maintenance_cmds", methods=['GET'])
+def get_maintenance_cmds():
+    try:
+        if mxcube.sc_maintenance is not None:
+            ret = mxcube.sc_maintenance.get_cmd_info()
+        else:
+            ret = "SC maintenance controller not defined"
+    except Exception, exc:
+        print("There was an exception %s" % str(exc))
+        return Response(status=409)
+    else:
+        return jsonify(cmds=ret)
+
+@mxcube.route("/mxcube/api/v0.1/sample_changer/get_global_state", methods=['GET'])
+def get_global_state():
+    try:
+        if mxcube.sc_maintenance is not None:
+            ret = mxcube.sc_maintenance.get_global_state()
+            state, cmdstate, msg = ret
+            print("Command returns %s" % str(ret))
+        else:
+            print("No SC maintenance controller")
+            return jsonify(ret) 
+    except Exception, exc:
+        print("There was an exception %s" % str(exc))
+        return Response(status=409)
+    else:
+        return jsonify(state=state, commands_state=cmdstate, message=msg)
+
+@mxcube.route("/mxcube/api/v0.1/sample_changer/send_command/<cmdparts>", methods=['GET'])
+def send_command(cmdparts):
+    print("Sending command to sample changer: %s" % (cmdparts, ))
+    try:
+        ret = mxcube.sc_maintenance.send_command(cmdparts)
+    except Exception, exc:
+        msg = str(exc);
+        msg = msg.replace("\n", " - ")
+        return 'Cannot execute command', 406, {'Content-Type': 'application/json',
+                                                'message': msg,
+                                                }
+    else:
+        return jsonify(response=ret)
--- a/mxcube3/ui/actions/general.js
+++ b/mxcube3/ui/actions/general.js
@@ -165,6 +165,38 @@ export function getInitialState() {
         'Content-type': 'application/json'
       }
     });
+    const loadedSample = fetch('mxcube/api/v0.1/sample_changer/loaded_sample', {
+      method: 'GET',
+      credentials: 'include',
+      headers: {
+        Accept: 'application/json',
+        'Content-type': 'application/json'
+      }
+    });
+    const sampleChangerState = fetch('mxcube/api/v0.1/sample_changer/state', {
+      method: 'GET',
+      credentials: 'include',
+      headers: {
+        Accept: 'application/json',
+        'Content-type': 'application/json'
+      }
+    });
+    const sampleChangerCommands = fetch('mxcube/api/v0.1/sample_changer/get_maintenance_cmds', {
+      method: 'GET',
+      credentials: 'include',
+      headers: {
+        Accept: 'application/json',
+        'Content-type': 'application/json'
+      }
+    });
+    const sampleChangerGlobalState = fetch('mxcube/api/v0.1/sample_changer/get_global_state', {
+      method: 'GET',
+      credentials: 'include',
+      headers: {
+        Accept: 'application/json',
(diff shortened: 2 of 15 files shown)


mxcube3/ui/actions/sampleChanger.js
```diff
@@ -19,18 +36,34 @@ export function refresh() {
       credentials: 'include'
     }).then(response => {
       if (response.status >= 400) {
-        dispatch(setState('ERROR'));
+        // dispatch(setSCState('ERROR'));
```
marcus-oscarsson: Small observation this line is commented did you leave it by purpose ?

mxcube3/routes/SampleChanger.py
```diff
@@ -24,14 +29,20 @@ def get_samples_list():
     order = []
 
     for s in samples_list:
+        if not s.isPresent():
+            continue
+        if s.isLoaded():
+            state = SAMPLE_MOUNTED
+        else:
+            state = UNCOLLECTED
```
marcus-oscarsson: This logic seems to indicate that SAMPLE_MOUNTED and UNCOLLECTED are mutually exclusive. Which is of course not the case since a sample can be SAMPLE_MOUNTED and UNCOLLECTED. Did you consider this case and are we just confused by the names of the states ?

mxcube3/ui/actions/general.js
```diff
@@ -165,6 +165,38 @@ export function getInitialState() {
         'Content-type': 'application/json'
       }
     });
+    const loadedSample = fetch('mxcube/api/v0.1/sample_changer/loaded_sample', {
```
marcus-oscarsson: Question about the sample changer related routes; it seems like most of these are used only to initialize the sample changer, after for instance refresh. The routes could perhaps be grouped into one route with an appropriate name, in order to minimize requests and to simplify the code somewhat, we could for instance group:

   'mxcube/api/v0.1/sample_changer/contents'
   'mxcube/api/v0.1/sample_changer/loaded_sample'
   'mxcube/api/v0.1/sample_changer/state'

Finally this route have, if we have understood things correctly, a somewhat confusing name. We understand it to be the state for maintenance ?, if its the case why not include maintenance in the route name ?

   'mxcube/api/v0.1/sample_changer/get_global_state'

mxcube3/ui/actions/general.js
```diff
@@ -201,10 +233,24 @@ export function getInitialState() {
...
+        state.sampleChangerCommands = json;
+      }).catch(notify),
+      sampleChangerGlobalState.then(parse).then(json => {
+        state.sampleChangerGlobalState = json;
+      }).catch(notify),
       observers.then(parse).then(json => { state.remoteAccess = json.data; }).catch(notify),
       workflow.then(parse).then(json => { state.workflow = json; }).catch(notify)
     ];
 
+    console.log('init-state', state.sampleChangerState);
```
marcus-oscarsson: Please remove console.log if you are finished debugging

mxcube3/ui/serverIO.js
```diff
@@ -154,6 +159,11 @@ class ServerIO {
     });
 
     this.hwrSocket.on('sc', (record) => {
+      this.dispatch(setLoading((record.signal === 'operatingSampleChanger'),
+                               'Sample changer in operation',
+                               record.message, true, () => (this.dispatch(sendStopQueue()))));
```
marcus-oscarsson: What does this signal signify ?, and are we certain that we like stopQueue as a abort action ?

mxcube3/ui/actions/sampleChanger.js
```diff
@@ -65,20 +98,20 @@ export function scan(address) {
...
 
-      dispatch(setState('READY'));
+      // dispatch(setSCState('READY'));
     });
   };
 }
 
 export function loadSample(address) {
   return function (dispatch) {
-    dispatch(setState('MOVING'));
```
marcus-oscarsson: Why is 'MOVING' state dispatching removed ? We think it is better to immediately reflect that we are doing something with the sample changer, instead of waiting for the next state update. Indeed, it would make it more responsive for users and it has the added benefit of preventing crazy 'clickers' to send many commands while one has already been sent.
We can ask ourselves the same question for the READY state at the end, although it is not so clear because theoretically the last updated state is the right one, and in this case the new code seems more correct.

meguiraun: better instead to use the busy indicator in this case

mxcube3/ui/actions/general.js
meguiraun: Only one route for getting the initial state in `actions.general.js`,  we can keep the rest of the routes as they are

mxcube3/ui/actions/sampleChanger.js
meguiraun: Right after the loading press button the loading modal appears, so there is no need of additional busy indicator

meguiraun:
a small fix for the sample changer mockup is needed, already submitted in a pr for 2.2 (master already was solved)

meguiraun:
please check if the comments have been solved and you can test this pr

marcus-oscarsson:
Hi,

Thanks for the changes Ive done some tests here, have you tested with the mock sample changer ?.
The interface already looks a bit better than before thats great, I'm not sure about the status bar spanning the entire page and the contents being in a separate box. Thats maybe just a detail.

So to the tests, I tired to list the contents but there does not seem to be any (with the mock SC3), I tried refresh, scan all containers and Unload actions but they don't seem to do anything. There are no messages in the server log.

I tried context menu -> mount and then I get the following error on the server:
```
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/oscarsso/projects/mxcube/mxcube3/mxcube3/routes/SampleChanger.py", line 152, in unmount_current
    mxcube.sample_changer.unload()
TypeError: unload() takes exactly 3 arguments (1 given)
```

meguiraun:
I think I know... this PR depends on https://github.com/mxcube/HardwareRepository/pull/48 and on the last one already accepted by matias.

I can list the available samples and mount, could you try with the code not yet merged?

meguiraun:
I noticed something weird with the notification of the currently mounted sample, looking at it. It looks as a mismatch between what you try to mount and what gets displayed in the ui.

meguiraun:
Let's see now, I have improved all the unload/load signalling, and tested with mockups (I did not tested against mockups earlier with care enough). I also have updated SampleChangerMockup, see PR.

marcus-oscarsson:
Hi,

Ive tested again, with the updates made, I also made sure that the HR is up to date.
So, I now get the contents of the SC


I can still not use the unload button it gives me:
```
   2017-08-21 16:40:54,351 |exceptions|ERROR  | Uncaught exception while calling /mxcube/api/v0.1/sample_changer/unmount_current/
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/oscarsso/projects/mxcube/mxcube3/mxcube3/routes/SampleChanger.py", line 154, in unmount_current
    mxcube.sample_changer.unload(None, wait=True)
  File "/home/oscarsso/projects/mxcube/mxcube3/mxcube3/HardwareRepository/HardwareObjects/SampleChangerMockup.py", line 54, in unload
    self._triggerLoadedSampleChangedEvent(self.getLoadedSample())
  File "/home/oscarsso/projects/mxcube/mxcube3/mxcube3/HardwareRepository/HardwareObjects/SampleChangerMockup.py", line 64, in getLoadedSample
    return self.getComponentByAddress(Pin.getSampleAddress(self._selected_basket, self._selected_sample))
  File "/home/oscarsso/projects/mxcube/mxcube3/mxcube3/HardwareRepository/HardwareObjects/sample_changer/Container.py", line 213, in getSampleAddress
    return str(basket_number) + ":" + "%02d" % (sample_number)
TypeError: %d format: a number is required, not NoneType

The scan all containers button gives me an error on the client:
```
main.js:87594 Uncaught (in promise) Error: Error while scanning sample changer @ 
    at main.js:87594
    at <anonymous>
```

un mount tho this position still gives me:
```
2017-08-21 16:46:45,460 |exceptions|ERROR  | Uncaught exception while calling /mxcube/api/v0.1/sample_changer/unmount/1:02
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/oscarsso/projects/mxcube/mxcube3/mxcube3/routes/SampleChanger.py", line 148, in unmount_sample
    mxcube.sample_changer.unload(loc, wait=True)
  File "/home/oscarsso/projects/mxcube/mxcube3/mxcube3/HardwareRepository/HardwareObjects/SampleChangerMockup.py", line 54, in unload
    self._triggerLoadedSampleChangedEvent(self.getLoadedSample())
  File "/home/oscarsso/projects/mxcube/mxcube3/mxcube3/HardwareRepository/HardwareObjects/SampleChangerMockup.py", line 64, in getLoadedSample
    return self.getComponentByAddress(Pin.getSampleAddress(self._selected_basket, self._selected_sample))
  File "/home/oscarsso/projects/mxcube/mxcube3/mxcube3/HardwareRepository/HardwareObjects/sample_changer/Container.py", line 213, in getSampleAddress
    return str(basket_number) + ":" + "%02d" % (sample_number)
TypeError: %d format: a number is required, not NoneType

```
Is it maybe because #48 is missing ?

meguiraun:
hej marcus, thanks for the tests! you do need the hardwarerepository code

marcus-oscarsson:
Hej !

Okey, Ill take a look when that code is merged then. 

Its a bit strange though because loading, unloading and scanning have been working before so it makes me wonder why its suddenly not working without changes to the HR. Maybe you have some idea ?.

In anyways its good if it works now, waiting for the PR to be merged.

Marcus

meguiraun:
I think this PR changes few things on loading, signals etc... Not much, but enough to annoy...

marcus-oscarsson:
Okey, it makes me a bit worried. We probably need to test this on real hardware then, ill try to get some time on a beamline. Ill try to do it asap, in the meantime we need to get #48 on HR merged

meguiraun:
We did the test on our real hardware and it was working, but glad if you also want to test on your side.

marcus-oscarsson:
I understand that you did some tests on your side, but I guess since the Mockup implementation used to work and there was some changes in GenericSampleChanger that some how changes the signaling of mounting and unmounting. We have to make sure that that still works on our hardware here, since its different from yours.

marcus-oscarsson:
Hi again,

So I had some time to perform some tests on the beamline with the FlexHCD. I understand that its impossible for you to know exactly how these things effect the FlexHCD since you don't have one, so I'm happy to do the final touches if needed.

The FlexHCD used to work fine, list contents, mount and umount. There are several problems with it now that makes it impossible to use. We cant list the contents so its difficult to say if we can load or unload. Heres a summary, as I said id be happy to help you out with these. Matias will take a deeper look at the HR PR to see if there is something there that might cause these issues.

- No contents
- The terminal is spamed with: "loaded sample changed now is:" even if nothing was done with the sample changer.
- Refresh does nothing
- Scan all baskets does nothing ( I think thats normal, not sure it worked with the Flex, to be confirmed with Matias)

Marcus

meguiraun:
thanks for your tests, I will try to review and see if I can see anything. However, the changes on the GenericSampleChanger.py are small...

Do you have any idea if the problem is on mxcube3 side or hardware object side?

marcus-oscarsson:
I guess the main issue is how the contents of the sample changer is exchanged, maybe this changed in some way ?. Maybe you could start at looking how the initial contents are retrieved and see if there is a difference on how its done now and how it was done before.

Also, whats the intended behavior of refresh, to update the contents ?

vrey01:
Ok guys.  

I have revisited the HardwareRepository original PR into smaller ones.  Concerning changes in HardwareObjects other than those related with Cats the only change is to provide a new signal that sampleChanger objects could emit to notify of content changes. If you want we can wait for those PR to be merged to continue on this, but I dont see the need.  
Submitted changes in HWOs should, really, not affect the GUI.

On the gui/web side there are more significant changes.  In my understanding, before my changes, one had to click in "Update Contents" to get the contents updated. The new signal in the HWO (not connected yet) would allow to automatically respond to contents changes, if not connected, or not emitted, still the "Update Contents" should work as before.  
The scan button was not doing anything at all.   
There were a number of issues to be addressed (state at initialization reflecting the real state of the sample changer, abort actions and other... ) in a general way.

My changes (in the gui part) concerned more the logic of the initialization of sample state (to handle also a proper refresh on initialization. Before the changes, "loaded sample" information would be correct only on change,  not if already loaded on application opening for example.

New buttons for abort (absolutely necessary) / unload were added

There are changes to avoid the sample-changer to become "READY" unpromptly that was disturbing the operation as it did not take into account the real state of the hardware (setting MOVING state on sending the command is maybe acceptable. Changing to  READY on actions happening only in the GUI are in my opinion unacceptable). 
Still setting MOVING may affect logic on, for example, pop-ups that check for the state to be LOADING (I solved some issues with this too, so that the code would be more resistant to intermediate states from the hardware during loading operation).  My opinion is that the state should reflect as much as possible the real state of the hardware, shortcuts or changes of state in the GUI itself should be done with the aim of helping to provide a rapid update for a expected future update from the SC, and still... with care on the logic of those state changes (regarding the latest known SC state)

The rest of the changes are related with SampleChangerMaintenance.

I dont know the current state of tests.  I am glad to re-test here the Mockup if there is any issue pending with this code.   I can unfortunately not test the FlexHCD but I will be happy to discuss with you any remaining issue.

meguiraun:
Hej guys, can we merge this code?

marcus-oscarsson:
HI Mikael,

I guess you did not notice but there are merge conflicts so we cant :) After that we need to re test since nothing was done on this PR since my last comment. So I imagine that the problems remain 

Marcus

meguiraun:
as far as I understood, @vrey01  already address your concerns?

marcus-oscarsson:
Well, there were no commits added after my last comment so have not taken the time to re test it. Maybe the issues was more related to the PRs on Hardware Repository, so its maybe a good idea to test again. 
We might have some time to re test this on Friday.

Marcus

meguiraun:
Okey thanks, I will try to solve the conflicts before that date so it is easier to test.

mguijarr:
If you want I recommend you to merge plotting first, then solve the conflicts on this one - otherwise I will have to solve conflicts again in my PR ;)

Issue #265, opened 2016-06-29
Installation


mguijarr:
Installation procedure has to be improved.

TO DO:
- remove dependencies/devDependencies => dependencies

bolmsten:
It this issue solved by the docker that Antonio created?

mguijarr:
I do not think so - I do not think we will deploy mxcube with docker? Since it needs to access to the beamline equipments, I am afraid it would make things more difficult and less efficient in term of performance within a docker container. I have no experience with docker so please tell me if you disagree.

In anyways, I created this issue during last MXCuBE meeting when people were struggling with MXCuBE 3 development environment installation. I am not sure we are compatible with most recent versions of npm? I noticed it may help if we do not make the distinguish between dependencies and devDependencies...

bolmsten:
Okay I will have to look into the difference between dependencies and devDependencies, at the moment I think we mix without knowing why. 

Regarding docker I will leave this for @amilan.

mguijarr:
Ok. Thanks Frederik.

mguijarr:
I think our installation problems can all be solved by switching from `npm` to [yarn](https://yarnpkg.com/). AFAIK it is compatible with all features supported by `npm` and uses the same  `package.json`.
This project is quite new but it seems to be very well appreciated (17.000+ stars on github !) I will give it a try soon and see if it makes installing the development environment for MXCuBE 3 faster and more predictable.
It can also be an exercise to upgrade some of our dependencies...

bolmsten:
Bixente was here on visit and for him we used docker to set up the development enviroment. It worked so good that I have now started to run mxcube only through docker as well.

mguijarr:
I think we are talking about 2 different things.

The docker image is perfect for someone to be able to get MXCuBE 3 running very quickly, and it is even possible to do some development as you are doing it now. So far so good.

That said, someone has to set up the development environment (at least when creating the docker image...), and the packages have to be installed. For this task, I think we would be better served by this new tool -yarn- compared to npm.

I do not know Docker very much, I know everybody is very fond of it so it's probably very good. But it seems a bit overkill to have to run MXCuBE 3 within a lightweight virtual machine thing just because dependencies cannot be properly installed using the normal way. Or maybe I am missing something ?

mguijarr:
In fact we should have a better look at devDependencies & dependencies, to make sure we do not include unnecessary files in our bundles

meguiraun:
[...]

PR #599, opened 2017-08-21
 Fix for issue #265 - Added simple install script to be used with curl or wget and bash

marcus-oscarsson:
Hey

Simple install script to be used with curl or wget and bash, currently only for Debian packages, that means Debian, and Ubuntu like systems. The system packages also needs to be fairly recent. Can easily be adapted to rpm based systems.

Usage:

```
curl -o- https://raw.githubusercontent.com/mxcube/mxcube3/master/debian_install.sh | bash
wget -qO- https://raw.githubusercontent.com/mxcube/mxcube3/master/debian_install.sh | bash

```

Related to issue #265 

Cheers,
Marcus

### DIFF
--- a/debian-install.sh
+++ b/debian-install.sh
@@ -1,0 +1,71 @@
+#!/usr/bin/env bash
+
+{ # this ensures the entire script is downloaded #
+
+system_has() {
+  type "$1" > /dev/null 2>&1
+}
+
+mxcube_source() {
+  echo "https://github.com/mxcube/mxcube3.git"
+}
+
+mxcube_install_dir() {
+  echo "mxcube3"
+}
+
+mxcube_download() {
+  if ! system_has git; then
+    echo >&2 'You need git, curl, or wget to install MXCuBE'
+    exit 1
+  fi
+    
+  command git clone "$(mxcube_source)" "$(mxcube_install_dir)" || {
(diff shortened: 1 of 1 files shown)


meguiraun:
could you update https://github.com/mxcube/mxcube3/wiki/How-to:-Installation so both wiki and script are consistent in terms of packages?

debian-install.sh
```diff
@@ -0,0 +1,71 @@
...
+  pip2.7 install -r requirements.txt
+}
+
+install_node() {
+  # nvm
+  command curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.2/install.sh | bash
+  export NVM_DIR="$HOME/.nvm"
+  source "$NVM_DIR/nvm.sh"
+  source "$NVM_DIR/bash_completion"
+  nvm install --lts
```
meguiraun: why nvm?

marcus-oscarsson: I found Node Version Manager (nvm) the easiest way to install node, there are maybe others but this is fairly platform independent and makes the script easier to port.

marcus-oscarsson:
No problem ill update the wiki page when the script exists, id like to do some additional test from the repository before i add the instructions to the install page.

meguiraun:
creating an issue so we do not forget to update the wiki

Issue #598, opened 2017-08-17
run-now in task wrong order


meguiraun:
If you run a task only (by run now method) it tries to execute the next task in that sample, which might not even exist

PR #600, opened 2017-08-21
Fix for issue #598 (run-now in task wrong order)

marcus-oscarsson:
Hi,

  Fix for issue #598 (run-now in task wrong order)

Marcus


meguiraun:
thanks!

PR #603, opened 2017-08-22
Issue 485 - Sample Grid: replace 'mount and collect' icon with right-click

marcus-oscarsson:
Hi !

Fix for issue #485 Sample Grid: replace 'mount and collect' icon with right-click

Right clicking and selecting mount and collect adds the selected samples to the queue
and mounts the first one (if several samples where selected).

Marcus


meguiraun:
thanks

Issue #541, opened 2017-05-24
Disable motor while moving


meguiraun:
The motor should be disabled during its movement. Specific case of the zoom motor to be fixed first

marcus-oscarsson:
Fixed by #605

Issue #540, opened 2017-05-24
Disable motors during data collection


meguiraun:
* Most motors and beamline controls should be disabled during data collection
* Question: which motors? which beamline controls?

marcus-oscarsson:
Fixed by #605

PR #605, opened 2017-08-23
Issue #541 and Issue #540

marcus-oscarsson:
Fixes for  Issue #541 and Issue #540

 - Disabling motors while collecting, zoom and light not disabled.
 - Disabling light motor and zoom motor input when the corresponding motor is moving

I discovered that there is a "free" zoom attribute directly under the root of the beamline store, what is this for ? maybe it can be removed. Because we also have the zoom represented as a motor but it was not initialized correctly (corrected with this PR as well). Please let me know.

Also these changes rely on that the state of motors are updated correctly which is not the case for the zoom and light mockups. The rest should work fine in mockup mode.

Marcus

### DIFF
--- a/mxcube3/ui/components/SampleView/MotorControl.js
+++ b/mxcube3/ui/components/SampleView/MotorControl.js
@@ -34,6 +34,7 @@ export default class MotorControl extends React.Component {
                 decimalPoints="2"
                 state={phi.Status}
                 stop={stop}
+                disabled={this.props.motorsDisabled}
               />
             </div>
 
@@ -49,6 +50,7 @@ export default class MotorControl extends React.Component {
                 decimalPoints="2"
                 state={kappa.Status}
                 stop={stop}
+                disabled={this.props.motorsDisabled}
               />
             </div>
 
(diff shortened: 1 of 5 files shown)


meguiraun:
I do not know what is doing zoom so free, but I agree it should not be there. I would leave out of this PR, I do not the consequences of removing it, it has been there quite a long time (I just checked with an old state representation that I printed a while ago).

When you say that zoom and light are not updated correctly, do they remain disabled? Or it has no effect at all?

marcus-oscarsson:
They remain disabled since the state of the mockup does not change from moving to ready. Or that state change is not emitted by the HO it seems. A fix to be done on the mockup HO. Refreshing the browser re initializes the state to ready.

meguiraun:
Thanks for these improvements, I am doing a quick test and I noticed:
* the abort button in the motor input is disabled while moving
* Motor inputs are not re-enabled after dc finishes

marcus-oscarsson:
Hey,

Thanks for your feedback, I fixed the issue with the abort button. For the re-enable after queue stop.
We look at the queue state and the controls are enabled when the state is not "QueueRunning". So is the state for some reason not updated ?. It works fine for the tests that Ive done here, is there a scenario in particular that does not work ?

Marcus

meguiraun:
ok, thanks for the quick fix.
I was (apparently) receiving the queue stop signals... Now, I cannot reproduce the issue @#@#@?? ....

PR #610, opened 2017-09-01
Issue 555: Characterisation - Diff plan

meguiraun:
this: https://github.com/mxcube/mxcube3/issues/555

~~Essentially it is kind of working, but WIP because:~~
* ~~still missing the checkbox in queue control for automatically adding diff plan (now everything manual)~~
* ~~task enable/disable not working properly?~~
* ~~a bit more of testing it is never a bad idea~~ (the code is a bit messy right now)
* so you know that I am on this

### DIFF
--- a/mxcube3/routes/Queue.py
+++ b/mxcube3/routes/Queue.py
@@ -335,6 +335,11 @@ def update_sample(sample_id):
         return Response(status=409)
 
 
+@mxcube.route("/mxcube/api/v0.1/queue/<sid>/<node_id>/accept", methods=['POST'])
+def accept_diff_plan(sid, node_id):
+    mxcube.queue.diffraction_plan[int(node_id)]['diffractionPlanAccepted'] = True
+    return Response(status=200)
+
 @mxcube.route("/mxcube/api/v0.1/queue/<node_id>/toggle", methods=['PUT'])
 def toggle_node(node_id):
     '''
@@ -707,5 +712,6 @@ def create_diff_plan(sid):
     group_entry.set_enabled(False)
     sample_entry.enqueue(group_entry)
     group_entry.enqueue(dc_entry)
+
 
     return Response(status=200)
--- a/mxcube3/routes/Queue.py
+++ b/mxcube3/routes/Queue.py
@@ -325,7 +325,7 @@ def update_sample(sample_id):
         # TODO: update here the model with the new 'params'
         # missing lines...
         sample_entry.set_data_model(sample_node)
-        #qutils.save_queue(session)
+        # qutils.save_queue(session)
         logging.getLogger('HWR').info('[QUEUE] sample updated')
         resp = jsonify({'QueueId': node_id})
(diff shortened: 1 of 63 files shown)


marcus-oscarsson:
Hi Mikel,

I noticed that you add the diffraction plans on a structure that is attached directly on `mxcube.queue`.
It would in my opinion be much better if we could extend the Characterisation TaskNode to hold the results, for instance as a list of data collections.

Like this we can just update the Characterisation entry in the queue.

Marcus

mxcube3/routes/Queue.py
```diff
@@ -640,5 +645,74 @@ def set_autmount():
...
     return resp
+
+@mxcube.route("/mxcube/api/v0.1/queue/mock/diff_plan/<sid>", methods=["GET"])
+def create_diff_plan(sid):
+    '''Juts for creating a diff plan as if it were created by edna and so on.
+    '''
+    from mock import Mock
+
+
+    acq_parameters = mxcube.beamline.get_default_acquisition_parameters()
```
marcus-oscarsson: Its not very important since its just mockup code, the file suffix and precision should be handled by _set_dc_params.
Which you are calling below, so these few lines should not be necessary.

marcus-oscarsson: Lines 659 - 662

mxcube3/routes/qutils.py
```diff
@@ -301,18 +301,39 @@ def _handle_char(sample_id, node):
...
            "type": "Characterisation",
            "parameters": parameters,
            "checked": node.is_enabled(),
            "sampleID": sample_id,
            "taskIndex": node_index(node)['idx'],
            "queueID": node._node_id,
-           "state": state
+           "state": state,
+           "diffractionPlan": task,
+           "diffractionPlanID": originID
```
marcus-oscarsson: I made a general comment about it this earlier. Its perhaps better if the counteraction contains  a list of all the collections in the diffraction plan ie diffractionPlan: [task1, task2 ... taskn]

We can like this display them in the chracterisation entry and the user can select
which one they like to use if not all.

mxcube3/routes/qutils.py
```diff
@@ -1024,22 +1047,31 @@ def queue_model_child_added(parent, child):
...
-            if isinstance(origin_model, qmo.Characterisation):
-                # Handle characterization specific logic here
-                print("CHARACTERIZATION")
+        if child_model.get_origin():
+            origin_model, origin_entry = get_entry(child_model.get_origin())
+
+        # we do not want to process when TaskGroup (get_name contains 'Diffraction Plan')
+        # is added, but when a dc_entry is added to such TaskGroup
+        # the origin of the task group is the char node, but not for the dc_entry
+        if isinstance(child, qmo.DataCollection) and isinstance(origin_model, qmo.Characterisation):
```
marcus-oscarsson: See my general comment on how to possibly avoid this case handling

PR #612, opened 2017-09-05
Grid integration

marcus-oscarsson:
Hi,

This PR contain the necessary changes to make the Grid scans work with the worklflow meshes

Cheers,
Marcus

### DIFF
--- a/mxcube3/routes/SampleCentring.py
+++ b/mxcube3/routes/SampleCentring.py
@@ -264,14 +264,17 @@ def get_shape_with_sid(sid):
 @mxcube.route("/mxcube/api/v0.1/sampleview/shape_mock_result/<sid>", methods=['GET'])
 def shape_mock_result(sid):
     shape = mxcube.shapes.get_shape(sid)
+    res = {}
 
-    from random import random
+    if shape:
+        from random import random
 
-    res = map(lambda x: map(lambda y: random(), range(shape.num_rows)), range(shape.num_cols))
+        for i in range(1, shape.num_rows*shape.num_cols + 1):
+            res[i] = [i, [int(random() * 255), int(random() * 255),
+                          int(random() * 255), int(random())]]
 
-    mxcube.shapes.set_grid_data(sid, res)
-
-    signals.grid_result_available(to_camel(shape.as_dict()))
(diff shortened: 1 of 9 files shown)


mxcube3/routes/SampleCentring.py
```diff
@@ -594,7 +601,7 @@ def accept_centring():
     """
     Accept the centring position.
     """
-    mxcube.diffractometer.acceptCentring()
+    mxcube.diffractometer.saveCurrentPos() #acceptCentring()
```
meguiraun: this does not exists in the generic difractometer

mxcube3/ui/components/SampleView/DrawGridPlugin.js
```diff
@@ -208,10 +214,9 @@ export default class DrawGridPlugin {
     canvas.renderAll();
   }
 
-  heatMapColorforValue(gd, value) {
-    const h = (1.0 - value) * 240;
-    const dataFill = `hsla(${h}, 100%, 50%, ${this.overlayLevel})`;
-
+  heatMapColorForValue(gd, value) {
+    let dataFill = `rgba(${parseInt(value[0], 10)}, ${parseInt(value[1], 10)},`;
+    dataFill += `${parseInt(value[2], 10)}, ${this.overlayLevel})`;
```
meguiraun: but then the color coding comes from the data sent by the server, you do not expect a single score value

mxcube3/routes/SampleCentring.py
marcus-oscarsson: Aha, okey I will revert back to acceptCentring then.

Thanks

mxcube3/ui/components/SampleView/DrawGridPlugin.js
marcus-oscarsson: I think we already discussed this, and we had some sort of agreement. So the grid results have always been given as R,G,B values and thats what we need to adapt to today. We have added a fourth field in the results so that its now R,G,B,score to fulfill your needs. Where score is given by the analysis software, handling of the score value needs to to be added if needed. We can have a discussion on migrating to only basing the the results on sore for version 3.1.

PR #614, opened 2017-09-07
[issue #558] - Signal handling, zoom and light motor fix

### LINKED ISSUE #558: Beam size not updated after zooming

marcus-oscarsson:
Hi,

I started to look at issue #558 and issues related to it that we have observed. I discovered some
issues related to the signal handling while doing this. So I took the opportunity to solve those and
clean up the code a bit.

In short I tried to make the signal handling a bit more effective, when it comes to the number of
signals sent and calculations made in signal handlers. It speeds things up quite alot for us, so I hope it will do the same for you.

This PR fixes issue #558 

Marcus

### DIFF
--- a/mxcube3/ui/components/SampleView/SampleImage.jsx
+++ b/mxcube3/ui/components/SampleView/SampleImage.jsx
@@ -308,7 +308,7 @@ export default class SampleImage extends React.Component {
     const { sendGoToBeam } = sampleActions;
 
     // Only move to beam if the click was done directly on the canvas.
-    if (e.target.tagName === 'CANVAS') {
+    if (e.target.tagName === 'CANVAS' && e.shiftKey) {
       sendGoToBeam(e.layerX * imageRatio, e.layerY * imageRatio);
     }
   }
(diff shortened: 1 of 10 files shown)


mxcube3/ui/components/SampleView/SampleImage.jsx
```diff
@@ -308,7 +308,7 @@ export default class SampleImage extends React.Component {
     const { sendGoToBeam } = sampleActions;
 
     // Only move to beam if the click was done directly on the canvas.
-    if (e.target.tagName === 'CANVAS') {
+    if (e.target.tagName === 'CANVAS' && e.shiftKey) {
```
meguiraun: why adding the shift for this? I would rather keep without...

meguiraun:
thanks for this, I cannot merge since it has conflicts, could you fix it please?

mxcube3/ui/components/SampleView/SampleImage.jsx
marcus-oscarsson: Because it happens that users accidentally double click on the canvas and then the sample is moved. It also happens that one during 3-click centering clicks a bit fast and then accidentally also send move to beam. Thats why we though it would be better with a modifier for move to beam

marcus-oscarsson:
Okey, conflict solved :)

PR #616, opened 2017-09-07
Fixing issue with reference image file name

marcus-oscarsson:
Hi,

This fixes an issue with the characterization reference image filename. The file suffix was
not retrieved correctly so the name ends up with ..? instead of .suffix

Marcus


meguiraun:
could you change the xml file as well?

PR #617, opened 2017-09-08
General improvements

mguijarr:
Most notably, this increases Python libraries versions - so, when testing this do not forget to do a pip install upgrade beforehand otherwise it won't work as some code is now specific to Flask-Socketio 2.9.

Otherwise look at commit messages for details.

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import
-from flask import Flask, request
-from flask.ext.socketio import SocketIO
-from flask.ext.session import Session
+from flask import Flask, request, session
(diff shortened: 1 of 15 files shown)


meguiraun:
which is the reason for upgrading python libs?

mguijarr:
The same as when we updated javascript libs. First, we can hope that some bugs get fixed (socketio...) and second it makes it easier to upgrade later since the differences won't be so much. When updating flask-socketio I had to change something else in the code, now it's done we can reasonably think next time we upgrade flask-socketio it will be easier (because the library is not changing so much between releases). Well, that's just an example...

Do you have a problem with updates? Let me know.

Issue #508, opened 2017-04-19
3Click centring


meguiraun:
It would be nice to display the instructions dialog only in the beginning of the experiment/session.

meguiraun:
In addition:
* Remove the information button? (text copied from the todo list, 20170510)
* On canvas text: 'you are now in 3 click centring', 'n clicks left'
* Users miss the click often: move the button to the left of the controls-bar and see if improves user experience

meguiraun:
Update:
* keep the md in centring mode until users saves the point or cancels the procedure (so you can continuously keep doing centrings...)

PR #637, opened 2017-09-20
[issue: #508] Show sample centring message on canvas

marcus-oscarsson:
Hi

  Show sample centring message on canvas, closing issue #508 

Marcus


meguiraun:
thanks!

PR #642, opened 2017-09-21
Xrf results issue 22

meguiraun:
Finally is here!

I extended a bit the plot component and reorganise where XRF plots are stored (on finish I move it to the xrf task result, any finished plot comes from the server, they are stored in the queue model, so refresh works). During execution you can click on 'plot available' and you will see the plot being populated with data.

Note: see HardwareRepository additions for a new mockup and small changes on the queue entry.

image

### DIFF
--- a/mxcube3/routes/Beamline.py
+++ b/mxcube3/routes/Beamline.py
@@ -52,6 +52,13 @@ def init_signals():
     except Exception, ex:
         logging.getLogger('HWR').error("error loading plotting hwo: %s" % str(ex))
 
+    try:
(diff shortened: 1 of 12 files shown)


beteva:
Well done @meguiraun!  Some cosmetic changes, though. As we said from the beginning, Scan should not be used, as there is nothing scanned. Spectra should be used instead. I am a bit puzzled by the numbers on both axes, but this may be only because of the dummy values.

meguiraun:
thanks!! Ok with your comments (I add to the list), I know the numbers have no real sense, it will be more meaningful with real data

Issue #554, opened 2017-06-07
Reset to default parameters button


meguiraun:
* New button to reapply the default parameters for the task modal
* Display it in the modals

PR #649, opened 2017-09-26
Issue 554

meguiraun:
#554 reset to defaults on modals. Dc, char, mesh and helical

I started on the plane, so better to finish before I forget what is about...

image


mguijarr:
Thanks Mikel !

It looks ok to me, let's see if Marcus has comments.

marcus-oscarsson:
Great !

PR #651, opened 2017-09-30
Fix for production build, and package updates

mguijarr:
This PR also removes __test__ directory and test packages, since we do not use them. I recommend to refer to the tests PR from Fredrik when we will want to have some tests, but for the moment we can just get rid of unused packages and unread file(s).

The problem with dygraphs was: code was written in ES6 in the package files in `node_modules` ; the
minification could not happen properly because UglifyJS still has problems with ES6 (I tried latest beta,
could not make it work...). So I include dygraphs to our Babel chain, to convert ES6 to ES5 ; then
everything works :)

I made all package updates I could do, because it narrows the gap next time we will have to do upgrades... And we get bug fixes etc. (and maybe new bugs - but we never know !).


meguiraun:
thanks!

Issue #661, opened 2017-10-05
Sample snapshots


marcus-oscarsson:
Number of snapshots needs to be weird
  - No snapshots for interleaved sub-wedges

PR #665, opened 2017-10-19
[issue 661] Number of snapshots

marcus-oscarsson:
Hi,

   Wired number of snapshots, solves issue #661  

Marcus


meguiraun:
thanks!

PR #670, opened 2017-10-24
Fixes, pause/resume and take snapshots

marcus-oscarsson:
Hi,

Minor fixes for propagating pause and resume events. Also fixing small issue with take snapshots

Marcus


meguiraun:
thanks!

PR #671, opened 2017-10-26
Improved top bar layout

marcus-oscarsson:
Hi,

   Improved top bar layout

Cheers,
Marcus


meguiraun:
hej, could you paste here an screenshoot? so I can see how it looks

marcus-oscarsson:
Ohh, sure well you can always check it out "live" by running it. But here is a screenshot just in case
tb

meguiraun:
thanks, I was in a messy branching situation...

Issue #664, opened 2017-10-05
Characterisation - auto add diffraction plan


marcus-oscarsson:
- The auto add option is not always taken into account
 - Refresh is needed for diffraction to be added to task, when auto add is false
 - Cant add resulting diffraction plan, something with associated shape ID

meguiraun:
see PR #675

marcus-oscarsson:
Thanks, all fixed !

meguiraun:
great!

marcus-oscarsson:
Alltough could you test it try with the built version to see if you have some issues. I got some issues when i used the build

meguiraun:
ok, I will test

PR #675, opened 2017-10-30
Issue 664 char bugs

meguiraun:
Trying to solve #664. Wip so you know I am on this, also I might need your input debugging.

### DIFF
--- a/mxcube3/routes/qutils.py
+++ b/mxcube3/routes/qutils.py
@@ -1466,7 +1466,7 @@ def queue_model_diff_plan_available(char, index, collection_list):
             socketio.emit('add_diff_plan', {"tasks": [task]}, namespace='/hwr')
 
 
-def se_auto_add_diffplan(autoadd, current_sample=None):
+def set_auto_add_diffplan(autoadd, current_sample=None):
     """
     Sets auto add diffraction plan flag, automatically add to the queue
     (True) or wait for user (False)
--- a/mxcube3/routes/qutils.py
+++ b/mxcube3/routes/qutils.py
@@ -1406,7 +1406,7 @@ def queue_model_child_added(parent, child):
     """
     Listen to the addition of elements to the queue model ('child_added').
     Add the corresponding entries to the queue if they are not already
-    added. Handels for instance the addition of referance collections for
+    added. Handels for instance the addition of reference collections for
     characterisations and workflows.
     """
     parent_model, parent_entry = get_entry(parent._node_id)
@@ -1431,25 +1431,28 @@ def queue_model_child_added(parent, child):
             parent_entry.enqueue(dcg_entry)
 
 
-def queue_model_diff_plan_available(char, index, collection):
-    if isinstance(collection, qmo.DataCollection):
-        if collection.get_origin():
-            origin_model, origin_entry = get_entry(collection.get_origin())
-        collection.set_enabled(False)
-        dcg_model = char.get_parent()
-        sample = dcg_model.get_parent()
-        sampleID = sample._node_id
-        queue = queue_to_dict()
-        tasks = queue[str(sampleID)]['tasks']
+def queue_model_diff_plan_available(char, index, collection_list):
+    # TODO: The client is not prepared for handling several collections for now @#@#!
+    for collection in collection_list:
+        if isinstance(collection, qmo.DataCollection):
+            if collection.get_origin():
+                origin_model, origin_entry = get_entry(collection.get_origin())
+            else:
+                origin_model, origin_entry = get_entry(char._node_id)
+            collection.set_enabled(False)
+            dcg_model = char.get_parent()
+            sample = dcg_model.get_parent()
+            sampleID = sample._node_id
+            queue = queue_to_dict()
+            tasks = queue[str(sampleID)]['tasks']
 
-        for t in tasks:
-            if t['queueID'] == char._node_id:
-                shape = t['parameters']['shape']
-                setattr(collection, 'shape', shape)
-
-        task = _handle_dc(sample._node_id, collection)
-        task.update({'isDiffractionPlan': True, 'originID': origin_model._node_id})
-        socketio.emit('add_diff_plan', {"tasks": [task]}, namespace='/hwr')
+            for t in tasks:
+                if t['queueID'] == char._node_id:
+                    shape = t['parameters']['shape']
+                    setattr(collection, 'shape', shape)
+            task = _handle_dc(sample._node_id, collection)
+            task.update({'isDiffractionPlan': True, 'originID': origin_model._node_id})
+            socketio.emit('add_diff_plan', {"tasks": [task]}, namespace='/hwr')
 
 def set_auto_add_diffplan(autoadd, current_sample=None):
     """
--- a/mxcube3/routes/qutils.py
+++ b/mxcube3/routes/qutils.py
@@ -1438,16 +1438,25 @@ def queue_model_diff_plan_available(char, index, collection):
         collection.set_enabled(False)
         dcg_model = char.get_parent()
         sample = dcg_model.get_parent()
+        sampleID = sample._node_id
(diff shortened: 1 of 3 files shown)


mxcube3/routes/qutils.py
```diff
@@ -1431,23 +1442,36 @@ def queue_model_child_added(parent, child):
...
+            for t in tasks:
+                if t['queueID'] == char._node_id:
+                    shape = t['parameters']['shape']
+                    setattr(collection, 'shape', shape)
+            task = _handle_dc(sample._node_id, collection)
+            task.update({'isDiffractionPlan': True, 'originID': origin_model._node_id})
+            socketio.emit('add_diff_plan', {"tasks": [task]}, namespace='/hwr')
+
+
+def se_auto_add_diffplan(autoadd, current_sample=None):
```
mguijarr: se_ = set_

meguiraun: aarrgggggggggg!!!!!!! I hate git!!!!!!
thanks!

meguiraun:
summarising:

- _Refresh is needed for diffraction to be added to task, when auto add is false_: fixed by commit 6424380. But, if `collection_list` has more than one collection the client will not work properly (known limitation)
- _Cant add resulting diffraction plan, something with associated shape ID_: I fixed by copying the shape id of the char that generated the dc
- _The auto add option is not always taken into account_: not sure about this one

I tested with mockups (but hacking the queue_entry so I have dummy edna_results). I would suggest to test on your side (I cannot fully test in the beamline here) before removing WIP (and eventually merging).

mxcube3/routes/qutils.py
```diff
@@ -1422,6 +1422,17 @@ def queue_model_child_added(parent, child):
             parent_entry.enqueue(dc_entry)
             sample = parent.get_parent()
 
+            sampleID = sample._node_id
```
marcus-oscarsson: Cant we attach the shape in the corresponding queue entry ?, I think that would be a better solution.

mxcube3/routes/qutils.py
```diff
@@ -1431,23 +1442,36 @@ def queue_model_child_added(parent, child):
...
-        if collection.get_origin():
-            origin_model, origin_entry = get_entry(collection.get_origin())
-        collection.set_enabled(False)
-        dcg_model = char.get_parent()
-        sample = dcg_model.get_parent()
-        task = _handle_dc(sample._node_id, collection)
-        task.update({'isDiffractionPlan': True, 'originID': origin_model._node_id})
-        socketio.emit('add_diff_plan', {"tasks": [task]}, namespace='/hwr')
+def queue_model_diff_plan_available(char, index, collection_list):
+    # TODO: The client is not prepared for handling several collections for now @#@#!
```
marcus-oscarsson: I don't really understand what you mean with that the queue is not prepared for handling several collections ? What's missing ?

marcus-oscarsson:
Hey Mikel,

Please see my comments above, we can have a quick skype similair if you like. Mainly two questions: 
  -Why cant we handle more than one collection ?
  -Can we move the shape assignment to the actual queue entry logic or is that for some reason not possible?

Cheers,
Marcus

mxcube3/routes/qutils.py
meguiraun: but the queue_entry does not have any info about the shapes. I though about it, but it would meant o modify the queue_entry only for this specific case.

mxcube3/routes/qutils.py
meguiraun: For the `case auto_add_diff_plan = false`, we allow the user to review the diff plan by opening a standard data collection modal. If the diff plan contains more than one collection, we still display the same modal and I think show only the first collection (I am not fully sure about which one though). We discussed this on my last visit, we said that for now it is ok like this.

What is missing is to design a new modal that could display several collections from the diff plan in a fancy way.

mxcube3/routes/qutils.py
marcus-oscarsson: Hm the DataCollectionModel should have information about its CenterdPosition, I think you can get the shape ID from there. Or just copy the entire CenteredPosition object over to the new collection.

It should be something like model.acquisitions[0].acquisition_parameters.centred_position

Maybe it helps ?

Cheers,
Marcus

mxcube3/routes/qutils.py
marcus-oscarsson: Aha, I see. 

I agree that its nice to have a dialog where we can change the parameters of multiple collections at once. However I believe, unless you have any idea on how to do that right now, that it something that requires some thought and time to implement. Its not a very important feature so we should probably wait for this until maybe 3.1 or an intermediate release 3.0.x.

Id rather that we use what we have today in-order to achieve what we want. So Id suggest, the user clicks add diffraction plan, and gets no dialog, all collections are added to the queue. The user can then remove or change the parameters of any of those like she would with any other task. 

There are of course alternatives to this, like the user could select which collections to add and so on. Or we only display the dialog when there is one collection on no dialog when there are multiple.

My interpretation was that it was okey for the moment as development in progress bu not to give to users. In anyways sorry for any misunderstanding.

Characterisations is something that are used alot so the behavior needs to be consistent and convenient. So I see two solutions for 3.0 we only allow diffraction plans with on collection in them, thus what you've made today is basically fine. The second solution, which I prefer, is something along with what I described above (we only need to remove the dialog).

Marcus

meguiraun: I kind of agree with you. In fact, I remember that we said that if more than one collection, add all collections without user intervention.

But, surprise surprise... I had completely forgoten but we already have that implemented!
https://github.com/mxcube/mxcube3/blob/master/mxcube3/ui/components/SampleQueue/CharacterisationTaskItem.jsx#L71

I do not know if it is working...

mxcube3/routes/qutils.py
meguiraun: I will have a look... no promises...

mxcube3/routes/qutils.py
marcus-oscarsson: Okey, so it remains to be tested :)

mxcube3/routes/qutils.py
meguiraun: the issue is that we need the shape_id, I can copy the raw cpos object from the char entry, but it does not contain the shape_id, only motor positions (in fact it is already created). I could also add the shapes_hwobj and retrieve the id from there (comparing dicts?), but in this case the cpos from the diff plan entry would contain motor positions + shape_id, while the originated char entry does not.

Maybe the best thing would be to add the shape_id whenever a cpos is created, thus before the diff plan is created. but for the purpose of this PR I would suggest to accept as it is. What do you think?

mxcube3/routes/qutils.py
meguiraun: I will try to test with mockups

meguiraun: working with the last fix

marcus-oscarsson:
I guess you are happy with this PR as you have removed the WIP ? We have no beam today so
I cant do a real test here until tomorrow.

Ill try to do some basic testing today and the rest tomorrow.

mxcube3/ui/reducers/shapes.js
```diff
@@ -24,6 +24,16 @@ export default (state = initialState, action) => {
 
         return { ...state, shapes };
       }
+    case 'UPDATE_SHAPES':
```
marcus-oscarsson: Hey I already have this code in my master, did you do a rebase or merge with master ?

marcus-oscarsson:
I will merge this for now but not close the issue, I will continue the tests with beam tomorrow.

Thanks !

mxcube3/ui/reducers/shapes.js
meguiraun: I do not what I did! You know how much I like git...

meguiraun:
great, thanks!

PR #681, opened 2017-11-07
[ISSUE 656] Taking account of files already on disk when mxcube is started

marcus-oscarsson:
Hi,

Taking account of files already on disk when mxcube is started

Marcus


meguiraun:
thanks!

PR #682, opened 2017-11-07
Fixing smaller issue introduced by previous fixes

marcus-oscarsson:
Hi 

Fixing smaller issue introduced by previous fixes,
  - diffractometer_hwobj does not exist
  - we should check if we can set_phase before doing anything, this code is kind of difractometer
    specific so we need to be a bit careful with it. Please consider a better check.

Also please consider defining a variable for "Centring" i.e. CENTRING_STATE or something
like that

Marcus


meguiraun:
thanks!

Issue #663, opened 2017-10-05
Login, Longing-out and then logging-in again does not work


marcus-oscarsson:
Longing-out and then logging-in again does not work, cant login

meguiraun:
Looking at this.

I cannot reproduce:
* when login as username (our case) and real Ispyb hwobj.
* when login as proposal and mockups

Using ispyb mockups and user login, it happens due to inconsistent dict between the dummy `__test_proposal` and the session obtained after calling `get_todays_session`. Probably something similar is happening with real hwobj and proposal login, how do you see this issue?

I will try to improve those dict structures, they have been giving problems in the past as well...

(testing with #653 and corresponding pr in HR, it may alleviate a bit the problem, but it definitively does not solve)

marcus-oscarsson:
I have an issue when for instance doing this: sign out -> sign in -> goto sample grid -> go to datacollection (fails)

 - It seems like current sample is not properly cleared on the server

Whats otherwise the status of this ?

meguiraun:
is this solved now with the last PR? (I think so)

PR #684, opened 2017-11-09
Save mounted sample on logout

meguiraun:
On #663 

I reproduce the behaviour you describe only if I mount a sample before signing out.
this line `mxcube.CURRENTLY_MOUNTED_SAMPLE = ''` in the signout api would do the fix.

I am a bit confused with how this affects the sample changer operation. I understand that currently only a unload command clears that attribute. 

So, for the cases where `unload` is not used and a sample was mounted before logout, the user logouts, If anyone logs in again, he/she does not see anything mounted but on the sample changer page. So, no queue at all but a sample mounted.

We could send and unload together with the signout, but it will probably make any test problematic. An alternative could be to create a queue with that single sample (no tasks)

### DIFF
--- a/mxcube3/routes/Login.py
+++ b/mxcube3/routes/Login.py
@@ -101,6 +101,10 @@ def signout():
     mxcube.queue = qutils.new_queue()
     mxcube.shapes.clear_all()
 
+    if mxcube.CURRENTLY_MOUNTED_SAMPLE_IS_MANUAL:
+        mxcube.CURRENTLY_MOUNTED_SAMPLE = ''
+        mxcube.CURRENTLY_MOUNTED_SAMPLE_IS_MANUAL = False
+
     LOGGED_IN_USER = None
     if remote_access.is_master(session.sid):
         state_storage.flush()
--- a/mxcube3/routes/Login.py
+++ b/mxcube3/routes/Login.py
@@ -102,7 +102,7 @@ def signout():
     mxcube.shapes.clear_all()
 
     if mxcube.CURRENTLY_MOUNTED_SAMPLE:
-        if mxcube.CURRENTLY_MOUNTED_SAMPLE_DATA['location'] == 'Manual':
+        if mxcube.CURRENTLY_MOUNTED_SAMPLE.get('location', '') == 'Manual':
             mxcube.CURRENTLY_MOUNTED_SAMPLE = ''
 
     LOGGED_IN_USER = None
--- a/mxcube3/routes/Login.py
+++ b/mxcube3/routes/Login.py
@@ -101,9 +101,9 @@ def signout():
     mxcube.queue = qutils.new_queue()
     mxcube.shapes.clear_all()
 
-    if mxcube.CURRENTLY_MOUNTED_SAMPLE_IS_MANUAL:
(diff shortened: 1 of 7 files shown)


marcus-oscarsson:
Hey !

It does not necessarily effect sample changer operation, since we can have manually mounted samples.
This means that there is not any sample mounted by the sample changer. We should only clear this if it was a manually mounted sample, if its not the case we should get the actually mounted sample from the
SC. 

So we should include the current sample when we set the INITIAL STATE of the client and I guess clear it if it was a manually mounted (maybe not even necessary)

Cheers,
Marcus

meguiraun:
Edited title to something more meaningful.

So, I modified so that we keep the sample and its main params after sign out. On the next login, we send back a queue with that sample to the client for the initial state, so the user will see a sample not only in the sample change page but also in the queue as mounted sample, and in the sample grid.

Let's see now

marcus-oscarsson:
Great !

mxcube3/routes/Login.py
```diff
@@ -101,6 +101,10 @@ def signout():
     mxcube.queue = qutils.new_queue()
     mxcube.shapes.clear_all()
 
+    if mxcube.CURRENTLY_MOUNTED_SAMPLE_IS_MANUAL:
```
marcus-oscarsson: Do we really need this check cant we just check id the sample changer is available and have a sample mounted
if its not the case its manual ?

meguiraun:
I cleaned a bit the code, instead of  a ~MANUAL flag I check in the sample data object ('location' key)

mxcube3/routes/Login.py
```diff
@@ -101,6 +101,10 @@ def signout():
     mxcube.queue = qutils.new_queue()
     mxcube.shapes.clear_all()
 
+    if mxcube.CURRENTLY_MOUNTED_SAMPLE:
+        if mxcube.CURRENTLY_MOUNTED_SAMPLE_DATA['location'] == 'Manual':
```
marcus-oscarsson: Where is  CURRENTLY_MOUNTED_SAMPLE_DATA defined the first time, and could it not be merged with CURRENTLY_MOUNTED_SAMPLE ?

meguiraun: scutils Line 108, once a sample is mounted.. and yeah, I guess it can be merged... I will do that

mxcube3/routes/scutils.py
```diff
@@ -101,10 +107,10 @@ def mount_sample(beamline_setup_hwobj,
 
 def mount_sample_clean_up(sample):
     try:
-        msg = '[SC] mounting %s (%r)', sample['location'], sample['sampleID']
+        msg = '[SC] mounting %s (%r)' % (sample['location'], sample['sampleID'])
         logging.getLogger('HWR').info(msg)
         set_sample_to_be_mounted(sample['sampleID'])
-        set_current_sample(sample['sampleID'])
+        set_current_sample(sample)
         mxcube.CURRENTLY_MOUNTED_SAMPLE_DATA = sample
```
marcus-oscarsson: Seems like you forgot this one then .... or ?

meguiraun: thanks!

PR #689, opened 2017-11-15
Zoom motor fix

marcus-oscarsson:
Hi,

  Yet another fix for the zoom motor, removed the error handling which creates some issues when using the 
  z+scroll. Its perhaps possible to come up with a more robust error handling.

Marcus

### DIFF
--- a/mxcube3/ui/components/SampleView/SampleImage.jsx
+++ b/mxcube3/ui/components/SampleView/SampleImage.jsx
@@ -461,14 +461,14 @@ export default class SampleImage extends React.Component {
         // Focus out
         sendMotorPosition('Focus', motors.focus.position - parseFloat(motorSteps.focusStep, 10));
       }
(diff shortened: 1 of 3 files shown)


mxcube3/ui/components/SampleView/SampleImage.jsx
```diff
@@ -461,14 +461,14 @@ export default class SampleImage extends React.Component {
         // Focus out
         sendMotorPosition('Focus', motors.focus.position - parseFloat(motorSteps.focusStep, 10));
       }
-    } else if (keyPressed === 'z' && motors.zoom.state === 2) {
+    } else if (keyPressed === 'z' && motors.focus.state === 2) {
```
meguiraun: why focus here?

marcus-oscarsson: Sorry my misstake

PR #693, opened 2017-11-15
Improved handling of LIMS data and LIMS synchronization on backend

marcus-oscarsson:
Hi,

  Improved handling of LIMS data and added LIMS synchronization on backend

Marcus


meguiraun:
oops, could you fix the conflict?

PR #702, opened 2017-11-20
Linting

marcus-oscarsson:
Hey

 :)

Marcus


meguiraun:
you are on fire! thanks

Issue #512, opened 2017-04-19
Disable unavailable params in modals


meguiraun:
The task configuration modal allows to write into all parameters even if they are not available in the beamline.

The modal should check in the state for available actuators/movables and disable the input fields for the ones missing (or a better way)

marcus-oscarsson:
Is this one completely done ?

meguiraun:
not really. I was thinking on something like this https://github.com/mxcube/mxcube3/pull/704

I will extend and submit a full pr soon

marcus-oscarsson:
What was the progress on this ?

marcus-oscarsson:
Done

PR #704, opened 2017-11-20
params disabled

meguiraun:
for #512

### DIFF
--- a/mxcube3/ho_mediators/beamline_setup.py
+++ b/mxcube3/ho_mediators/beamline_setup.py
@@ -247,6 +247,10 @@ class HOMediatorBase(object):
         """
         return ""
 
+    def isDisabled(self):
+        state = self.state()
+        return state == "NOTINITIALIZED" or state == "UNUSABLE"
+
     def dict_repr(self):
         """
         :returns: The dictionary representation of the hardware object.
@@ -259,7 +263,8 @@ class HOMediatorBase(object):
                 "msg": self.msg(),
                 "type": "FLOAT",
                 "precision": self.precision(),
-                "step": self.step_size()
+                "step": self.step_size(),
+                "disabled": self.isDisabled()
                 }
 
         return data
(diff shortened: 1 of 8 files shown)


meguiraun:
Removed WIP.

Trying to solve issue #512, preventing typing an attribute that is in fault. It is not exactly what the issue was about, for example if there is no transmission, the beamline hwobj will not even initialize so there is not much we can do at client level.

also, small fix for the form validation

marcus-oscarsson:
Okey so there seems to be things related to two issues in one PR here, so first for issue #512:

All beamlines does not have the same capabilities for instance some have movable wavelength others don't. So we need to be able to configure to account for this, and there is already a configuration in beamline-setup.xml for energy:

   beamline-setup.xml:  <tunable_wavelength>True</tunable_wavelength>

So whats missing today is to add configuration for missing attributes and account for it in the UI and if necessary in other logic. 

This kind of includes handling faulty equipment because such equipment can be added to the list of attributes to ignore.

I don't think its a good idea to automatically handle faulty equipment in the UI 

The validation fix is of course always good, but could it be included in a separate PR ?

Marcus

meguiraun:
hej!

I did not understand the issue exactly that way. Initially it was there because in our case many beamline attributes were not available at all, but the modal did (do) not reflect that. So we were able to set the transmission and have no effect at all in the beamline. That's  why disabling a parameter seems useful if the attribute is not available in the beamline.

And well, I did not  automatically handle faulty equipment in the UI. I just prevent from acting on those in case of a fault, perhaps I am misunderstanding your sentence.

What I can do is to modify the `ho_mediators` and ensure that we always will have a dict for all the actuators, for the case when they are not available I add a keyword in the dict for disabling the inputs.

marcus-oscarsson:
Hi again, 

Okey so thats exactly what I interpret the issue to be about. For instance if you don't have transmission 
on your beamline, you need to be able to configure mxcube so that that transmission is not taken into account. This is done via beamline-setup.xml where you could say for instance:

```<has_transmission>True/False</has_transmission>```

Just an example there is perhaps a better attribute name. Just like we have beamlines with tunable and fixed wavelength. So we need to add such configuration option for all beamline properties that are
"optional".

Marcus

meguiraun:
I would avoid adding those configuration options at hwobj level, it is hard to keep track of what we currently can configure, so I adding more configuration options I think is a mess.

And I do not think is strictly necessary. What we have now, is that if transmission hwobj does not exits an empty dict is generated. So on this pr I added a flag. Like you suggest but not in the beamline xml. The "optional property" is already defined by how your beamline or how you build you hwobj-s.

PR #708, opened 2017-11-23
Disable sample handling

marcus-oscarsson:
Hello,

Disabling instead of removing samples when "deleting" a sample from queue

   - Handle enable/disable event from server to client
   - Handle enable/disable event form client to server

Marcus

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -158,6 +158,7 @@ if not app.debug or os.environ.get("WERKZEUG_RUN_MAIN") == "true":
         app.INITIAL_FILE_LIST = []
         app.SC_CONTENTS = {"FROM_CODE": {}, "FROM_LOCATION": {}}
         app.SAMPLE_LIST = {"sampleList": {}, 'sampleOrder': []}
+        app.TEMP_DISABLED = []
 
         # set up streaming
         from mxcube3.video import streaming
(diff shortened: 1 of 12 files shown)


meguiraun:
I do not follow. Then how do you remove a sample from the queue?

marcus-oscarsson:
When a user deselects a sample is simply disabled not removed from the queue, that is how the queue works today in for instance MXCuBE2.  I found it better to do like this since we dont need to re add a sample if it needs to be re-collected. This also removes the need to synchronize uncollected tasks with the sample list when a sample is removed. Overall the logic becomes a bit simpler. The delete call still exists but we dont use it like we did before.

meguiraun:
To me this looks like the enabling/disabling checkbox we once had (in mxcube3). I think removing a sample from the queue should be simply remove. Otherwise we will have a queue which is not what would be executed. 
With this approach all the samples are in the queue all the time?

I do not see this PR, to me looks like a big change in the queue ~operation. Perhaps I am missing something, we could have an skype talk during the day and discuss it?

marcus-oscarsson:
We are simply maintaining the function of the queue as it is today. We can have a skype meeting about if you like. Id like this to get done ASAP since its important for our operation.

meguiraun:
some issues while testing:
* on sample grid: cannot unpick a sample that has a task
* on `limsutils.sample_list_get()` (from `/mxcube/api/v0.1/queue_state`) call: I get the samplelits but there is no `"checked": false` in the samples (only if have added and removed)

PR #709, opened 2017-11-23
Sample loading

marcus-oscarsson:
HI

Hopefully better handling of sample loading 
  - Snakecase to camelcase in mount_sample
  - Removed Pin check from loaded_sample_changed
  - Added bw argument to take_snapshot needed by automatic loop centring

Marcus

### DIFF
--- a/mxcube3/routes/scutils.py
+++ b/mxcube3/routes/scutils.py
@@ -59,6 +59,7 @@ def get_sample_to_be_mounted():
 def mount_sample(beamline_setup_hwobj,
                  view, data_model,
                  centring_done_cb, async_result):
+    from signals import loaded_sample_changed
 
     logging.getLogger('user_level_log').info("Loading sample ...")
 
@@ -77,7 +78,8 @@ def mount_sample(beamline_setup_hwobj,
     else:
         sample_mount_device = beamline_setup_hwobj.sample_changer_hwobj
 
-    if sample_mount_device.getLoadedSample().getAddress() == data_model.loc_str:
+    if sample_mount_device.getLoadedSample() and \
+            sample_mount_device.getLoadedSample().getAddress() == data_model.loc_str:
         return
 
     if hasattr(sample_mount_device, '__TYPE__'):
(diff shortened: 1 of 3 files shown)


mxcube3/routes/scutils.py
```diff
@@ -109,17 +113,17 @@ def mount_sample(beamline_setup_hwobj,
                     msg = "Manual centring used, waiting for" +\
                           " user to center sample"
                     log.warning(msg)
-                    dm.start_centring_method(dm.CENTRING_METHOD_MANUAL)
+                    dm.startCentringMethod(dm.MANUAL3CLICK_MODE)
```
meguiraun: this does not exist in the `GenericDiffractometer` (snakecase yes )

mxcube3/routes/signals.py
```diff
@@ -137,14 +137,13 @@ def loaded_sample_changed(sample):
         barcode = ''
 
     logging.getLogger("HWR").info('loaded sample changed now is: ' + address)
-
-    if isinstance(sample, Pin):
+ 
+    try:
```
meguiraun: why did you remove this? I think it could also be ~Container

marcus-oscarsson: The code as it were before generated errors when executed, since sample_data was undefined if
sample was not a Pin

And I dont understand why it would be container, the event is sample changed, so sample should 
not be a container. For me it indicates the the event is called when its not supposed to be called

If its the case it should just return at the top of the function, because all the logic in this function corresponds to a sample beeing loaded and is not valid for a container.

mxcube3/routes/scutils.py
marcus-oscarsson: Yes I noticed, its unfortunately what we have on our beamlines so there is not much we can do about it for the moment. Both the Minidfiff and Microdiff classes uses the camel case version, which one are you using ?

meguiraun: we are using the `snake_case`. I am fine with a `hasattr`, like we have done before in similar cases.

marcus-oscarsson: This could get quite problematic, since for instance camel case is used in the queue_entry and perhaps in other places. The interface of GenericDiffractomter is not really applied, which is 
a problem in itself. This is I believe is due to that the work with it is not complete yet.

I noticed that you already have some camel case aliases in your diffractometer so Id suggest that you add one for startCentringMethod aswell so that its backward compatible with the rest of the queue.

We will continue the discussion with the new interface for hardware objects 3.0

Marcus

mxcube3/routes/signals.py
meguiraun: Maybe I am mixing containers/baskets/pins etc...

This code was working with mockups and the Pin was needed (for some error I cannot remember), but I see and agree with your point, so we can put everything on the same `if`

mxcube3/routes/scutils.py
meguiraun: I guess I can do that

mxcube3/routes/signals.py
meguiraun: this is a merge problem, the `try` should be inside the `if`

PR #717, opened 2017-11-28
Several fixes

marcus-oscarsson:
Hey,

So this PR contains several smaller fixes, sorry in advance for not making several PRs but I thought it would be simpler like this.

Summary:
   - Some layout updates to task nodes in the sample queue, darker background color on body to make some elements "pop"
   - Pausing instead of stopping queue on connection loss 
   - Fix for characterization results
   - Added  delete hot key for shapes and escape hot key for canceling centring
 
Cheers,
Marcus


meguiraun:
thanks for this!

PR #719, opened 2017-12-11
[WIP] various fixes

meguiraun:
Hej,

A temporary PR for fixes. Some are only relevant to us (like the prepare_beamline on new sample, in manual mode), and the dc link for non exi.

I found a few bugs with the centring mechanism when several points are on the canvas, point selection has a bug that prevents working with helical under some circumstances (like removing points before), added Cryo, transmission and Mach Info, and default number of snapshot in  the mxcollect.xml.

I am doing this PR so you can see and perhaps cherry-pick fixes if you cannot wait, I still need to apply these on master to be mergeable.

### DIFF
--- a/mxcube3/routes/Beamline.py
+++ b/mxcube3/routes/Beamline.py
@@ -254,3 +254,17 @@ def beamline_get_data_path():
     """
     data = mxcube.session.get_base_image_directory()
     return jsonify(data)
+
+@mxcube.route("/mxcube/api/v0.1/beamline/prepare_beamline", methods=['PUT'])
+def prepare_beamline_for_sample():
+    """
+    Prepare the beamline for a new sample.
+    """
+    
+    try:
+        if hasattr(mxcube.collect, prepare_for_new_sample):
+            mxcube.collect.prepare_for_new_sample()
+    except Exception:
+        logging.getLogger('HWR').error('Cannot prepare the Beamline for a new sample')
+        return Response(status=200)
+    return Response(status=200)
--- a/mxcube3/routes/Beamline.py
+++ b/mxcube3/routes/Beamline.py
@@ -262,7 +262,7 @@ def prepare_beamline_for_sample():
     """
     
     try:
-        if hasattr(mxcube.collect, prepare_for_new_sample):
+        if hasattr(mxcube.collect, 'prepare_for_new_sample'):
             mxcube.collect.prepare_for_new_sample()
     except Exception:
         logging.getLogger('HWR').error('Cannot prepare the Beamline for a new sample')
--- a/mxcube3/routes/SampleCentring.py
+++ b/mxcube3/routes/SampleCentring.py
@@ -55,7 +55,6 @@ def centring_remove_current_point():
 def centring_add_current_point():
     global CENTRING_POINT_ID
     shape = mxcube.shapes.get_shape(CENTRING_POINT_ID)
-
     if shape:
         shape.state = "SAVED"
         signals.send_shapes(update_positions = False)
@@ -64,7 +63,6 @@ def centring_add_current_point():
 
 def centring_update_current_point(motor_positions, x, y):
     global CENTRING_POINT_ID
-
     if CENTRING_POINT_ID:
         point = mxcube.shapes.get_shape(CENTRING_POINT_ID)
         point.move_to_mpos([motor_positions], [x, y])
@@ -657,20 +655,21 @@ def wait_for_centring_finishes(*args, **kwargs):
     centred point.
     """
     centring_status = args[1]
-
     # we do not send/save any centring data if there is no sample
     # to avoid the 2d centring when no sample is mounted
     if scutils.get_current_sample().get('sampleID', '') == '':
         return
-    try:
-        motor_positions = mxcube.diffractometer.centringStatus["motors"]
-    except KeyError:
-        msg = "[SAMPLEVIEW] Centring error, cannot retrieve motor positions."
-        logging.getLogger('HWR').exception(msg)
-        return
+    #try:
+    #    motor_positions = mxcube.diffractometer.centringStatus["motors"]
+    #except KeyError:
+    #	motor_positions = centring_status["motors"]
+    #    msg = "[SAMPLEVIEW] Centring error, cannot retrieve motor positions."
+    #    logging.getLogger('HWR').exception(msg)
+    #    return
 
     # If centering is valid add the point, otherwise remove it
     if centring_status['valid']:
+    	motor_positions = centring_status["motors"]
         motor_positions.pop('zoom', None)
         motor_positions.pop('beam_y', None)
         motor_positions.pop('beam_x', None)
@@ -688,7 +687,7 @@ def accept_centring():
     """
     mxcube.diffractometer.acceptCentring()
     centring_add_current_point()
-
+    global CENTRING_POINT_ID
     return Response(status=200)
 
 
@@ -709,7 +708,7 @@ def move_to_beam():
     params = request.data
     params = json.loads(params)
     click_position = params['clickPos']
-    logging.getLogger('HWR').info("A point submitted, x: %s, y: %s"
+    logging.getLogger('HWR').info("Moving to beam, A point submitted, x: %s, y: %s"
                                   % (click_position['x'],
                                      click_position['y']))
     if getattr(mxcube.diffractometer, 'moveToBeam') is None:
--- a/mxcube3/routes/qutils.py
+++ b/mxcube3/routes/qutils.py
@@ -1728,7 +1728,7 @@ def set_auto_mount_sample(automount, current_sample=None):
         enable_sample_entries(current_queue["sample_order"], automount)
 
     # No automount, enable the current entry if any
-    if not automount and sample:
+    if not automount and sample in current_queue["sample_order"]:
         enable_sample_entries([sample], True)
 
 
--- a/mxcube3/routes/qutils.py
+++ b/mxcube3/routes/qutils.py
@@ -21,7 +21,7 @@ from . import limsutils
 
 # Important: same constants as in constants.js
 QUEUE_PAUSED = 'QueuePaused'
-QUEUE_RUNNING = 'QueueStarted'
+QUEUE_RUNNING = 'QueueRunning'
 QUEUE_STOPPED = 'QueueStopped'
 QUEUE_FAILED = 'QueueFailed'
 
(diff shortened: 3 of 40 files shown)


marcus-oscarsson:
Hey

I see that you are doing quite alot of work there :), do you think it would be possible to divide this PR into several ones and to get rid of the merge commits ?

You state for instance that there are things that are BioMax specific or only relevant to you guys. Could you put those things in a separate PR, and perhaps try organize the rest so that its easier to follow ?

Thanks
Marcus

mxcube3/routes/qutils.py
```diff
@@ -304,7 +304,15 @@ def _handle_dc(sample_id, node, include_lims_data=True):
         limsres = mxcube.rest_lims.get_dc(lims_id)
 
     # Always add link to data, (no request made)
-    limsres["limsTaskLink"] = mxcube.rest_lims.dc_link(lims_id)
```
marcus-oscarsson: Id prefer if this was done slightly differently, either you implement your ISPyBRestClient HO, with the correct method, which perhaps is the preferred way (using exi or not your choice)

The other way would be to add a method in limsutils, for instance called get_dc_link that performs the right logic. There is also no need for the .use_exi attribute since you can simply check for mxcube.rest_lims if it does not exist
use db_connection. This would at lest the duplication of this logic ...

Id recommend you to use the ISPyBRestClient though, I think the long term strategy is to phase out the SOAP interface, so we are in a transition phase right now. Another question is why you would not use EXI ?

Marcus

mxcube3/routes/qutils.py
```diff
@@ -1631,8 +1652,8 @@ def execute_entry_with_id(sid, tindex=None):
         node_id = current_queue[sid]["tasks"][int(tindex)]["queueID"]
 
         node, entry = get_entry(node_id)
-        # in order to fill lims data, we execute first the parent (group_id missing)
-        parent_id = node.get_parent()._node_id
+	# in order to fill lims data, we execute first the parent (group_id missing)
```
marcus-oscarsson: Watch out this indent seems to off by one

mxcube3/routes/qutils.py
```diff
@@ -822,18 +844,18 @@ def _queue_add_item_rec(item_list, current_queue):
...
     for item in item_list:
         item_t = item["type"]
         # If the item a sample, then add it and its tasks. If its not, get the
         # node id for the sample of the new task and append it to the sample
         sample_id = str(item["sampleID"])
 
         # Do not add samples that are already in the queue
-        if item_t == "Sample" and item["sampleID"] not in current_queue:
-            sample_node_id = add_sample(sample_id, item)
+	# TODO: fix this!!!
```
marcus-oscarsson: In this case I would sugest:

 if item["sampleID"] not in current_queue and item_t == "Sample": 

What is the TODO: fis this about ?

mxcube3/ui/components/SampleView/ContextMenu.js
```diff
@@ -105,10 +105,10 @@ export default class ContextMenu extends React.Component {
     const { sampleID, defaultParameters, shape, sampleData } = this.props;
     const sid = _shape ? _shape.id : shape.id;
 
-    if (this.props.clickCentring) {
-      this.props.sampleActions.stopClickCentring();
```
marcus-oscarsson: This seems strange to me, why do we need to stop a click centering which is not started ?

The 2D centring routine should either start a click centring, or if its not really a click centering but translation of sampx sampy, just do sendAddShape or maybe sendAcceptCentring.

I dont know exactly how your 2D centring is implemented, is it simply a call to move to beam ?

mxcube3/routes/Beamline.py
```diff
@@ -254,3 +254,17 @@ def beamline_get_data_path():
     """
     data = mxcube.session.get_base_image_directory()
     return jsonify(data)
+
+@mxcube.route("/mxcube/api/v0.1/beamline/prepare_beamline", methods=['PUT'])
```
marcus-oscarsson: What does prepare beamline do, could it be generic to all beamines ?

We maybe don't need to send this explicitly from the client but we can deduce when its needed in the server, like
we call it when a sample is added, like this we don't need any client side implementation of this at all ?

mxcube3/routes/SampleCentring.py
```diff
@@ -688,7 +687,7 @@ def accept_centring():
     """
     mxcube.diffractometer.acceptCentring()
     centring_add_current_point()
-
+    global CENTRING_POINT_ID
```
marcus-oscarsson: It seems like something is missing here ?

mxcube3/routes/SampleCentring.py
```diff
@@ -657,20 +655,21 @@ def wait_for_centring_finishes(*args, **kwargs):
     centred point.
     """
     centring_status = args[1]
-
     # we do not send/save any centring data if there is no sample
     # to avoid the 2d centring when no sample is mounted
     if scutils.get_current_sample().get('sampleID', '') == '':
         return
-    try:
```
marcus-oscarsson: Yes indeed, cant really tell how it was before but I imagine that 

    motor_positions = mxcube.diffractometer.centringStatus["motors"]

used to retrieve the centering status and the if statement maybe checked motor_positions[valid], but then someone passed centering status as an argument and this block became redundant.

So the commented block can be removed

mxcube3/routes/qutils.py
meguiraun: We are using both soap and rest hwobj. But, the dc_link in the rest hwobj only works with exi like urls. Thats why I cannot check in in the rest hwobj.

I can add a method in limsutils but I still need a selector somewhere.

We do not have exi in place here, it is not straightforward deployment although we have a test run. We have seen things here and there and we have enough with one ispyb flavour for now...

marcus-oscarsson:
Hey 

I've checked up to the commit: "dc link based on exi usage (or not)", The first 5 commits are completely fine and can be merged then there are some on which I have some questions and remarks. let me know if you prefer to discuss this over Hangouts or similar.

mxcube3/routes/qutils.py
meguiraun: A temporary fix for the addition of samples, I had troubles with this before. I know you commit some fixes and perhaps this is no longer needed, but I need to check here.

mxcube3/ui/components/SampleView/ContextMenu.js
meguiraun: Our 2d is a move_to_beam. 

This allows to add a task and save the point in the middle of the centring procedure, on a temporary point.

mxcube3/routes/qutils.py
marcus-oscarsson: Ah, so maybe we can add some configuration attributes to the ISPyBRestClient, now there is EXI root:

  <exi_root>
   https://your.exi.org
  </exi_root>

  <restRoot>
   https://your.lims.org
  </restRoot>

in lims-rest.xml

Maybe this can be changed to something more generic for the dc_link method or you can just use that to point at another url for the link ?

like:

  <result_link_root>
   ...
  </result_link_root>

It would be more generic I think

mxcube3/routes/Beamline.py
meguiraun: In our case, in our Collect hwobj
- Move detector back
- Close safety shutter
- Diff to transfer

We trigger from the client in order to do those steps before the sample is actually added to the queue. We saw (in the previous user run, but not commited this before :( ) that the users were more aware of that this way.

mxcube3/routes/SampleCentring.py
meguiraun: yes, my bad

mxcube3/routes/SampleCentring.py
meguiraun: okey, then I was not crazy...

mxcube3/routes/Beamline.py
marcus-oscarsson: Hm, this should in my opinion definitely be handled by the control or hardware object layer. Maybe even in the collect hwobj or in the sample changer hw object.

Adding such logic in the client makes the behavior of the beamline control dependent on the UI.

meguiraun:
Thanks for the review!

I will split in PRs as soon as I can, we are having users today and there are small glitches... so I am having fun ... to say in a way...

mxcube3/ui/components/SampleView/ContextMenu.js
meguiraun: no no, we do a 2d centring, get a temp point, and on  that try to add a task. So it should also save the point automatically. No matter if you are in a centring procedure

mxcube3/routes/Beamline.py
meguiraun: We can discuss about this, but for me the beamline acts on a user request, so I do not see an issue. Is later on the hwobj level where it is triggered.

mxcube3/ui/components/SampleView/ContextMenu.js
marcus-oscarsson: So then its maybe enough to just save the position, id like that we avoid calling acceptCentring if its not necessary, especially since the route handling acceptCentring performs no checking before the call is made.

So I see two options, guard the acceptCentring route or just make a call to updateShape

mxcube3/routes/Beamline.py
marcus-oscarsson: In a way everything acts on a users request, this needs to be pushed back to the backend or better to the hardware object layer, of there are no really good reasons for not doing that.

mxcube3/routes/qutils.py
meguiraun: what I do not like on that approach is that in our case we do not use the rest interface for getting the link, so I feel like it should be on the soap hwobj, which is anyway something missing now.

meguiraun:
see https://github.com/mxcube/mxcube3/pull/724
I leave this one open while I am splitting, so I see the comments.

mxcube3/routes/qutils.py
marcus-oscarsson: Ok, so then something like this

- Add 

<result_link_root>
...
</result_link_root>

To lims.xml (the SOAP HO) and maybe also to lims-rest.xml (Rest HO)

- Add a method in limsutils called get_result_link(lims_id),
That first tries to get the one from the Rest HO and if nothing is returned the one from the SOAP 

HO. Something like that.

meguiraun: The dc_link method in the rest_hwobj always returns a url, not None. Which points to a non existing page in our case. Thats I only see a selector flag as a solution (well, move everything to rest, but thats the future ;))

meguiraun:
Closing this one, the only thing that remains is the prepare_beamline action, I will submit a RP next year ;)

Issue #408, opened 2016-11-28
Input window for energy scan


amilan:
**User story**
As a user I want to input data through a periodic table.

beteva:
Please consider - L1, L2 or L3 edge, when L edges involved and only relevant elements (static configuration)

amilan:
Thanks! 
We don't plan to work on this feature before our next Skype meeting, so we can consider this when we will define the acceptance criteria.

Issue #407, opened 2016-11-28
Energy scan results into data collection


amilan:
**User story**
As a user, I want to handle scan results and feed them into data collection.

**Acceptance criteria**
Given a data collection,
When results are ready,
Then they are added to the data collection.

PR #720, opened 2017-12-12
Sample Sequence, Energy Scan UI, Bug fixes

marcus-oscarsson:
Hey,

Here is a big PR that contain six major categories of features/fixes, I should maybe have made several PRs instead. I thought it would be easier to handle like this, please let me know if you prefer that I divide
it. Each commit could almost be considered a PR on its own so you can treat as such.

The four major categories are
  - Improved sample by sample sequence
  - Bug fixes for bugs introduced when adding enable/disable sample
  - Optimized server side add_task_item
  - UI for Energy scan, including result handling on back-end
  - Various other minor bug fixes, like EDNA defaults
  - Some style fixes, like miss aligned elements

Solves issues #408, #407

Regards,
Marcus

### DIFF
--- a/config.gui.prod.js
+++ b/config.gui.prod.js
@@ -1,4 +1,4 @@
 module.exports = {
-	phaseControl: true,
+	phaseControl: false,
 	apertureControl: true
 }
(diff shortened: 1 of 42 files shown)


config.gui.prod.js
```diff
@@ -1,4 +1,4 @@
 module.exports = {
-	phaseControl: true,
+	phaseControl: false,
```
meguiraun: why did you remove this?

marcus-oscarsson: Yes I understand, but its false by default, you can enable it if you need it. I disabled it because most beamlines do not have phase control, so its like an additional option.

PR #724, opened 2017-12-15
Fixes pr 1

meguiraun:
only fixes.

Leaving out: dc link, save 2d centring, prepare beamline, will come later


mxcube3/routes/qutils.py
```diff
@@ -5,7 +5,7 @@
 import redis
 import itertools
 import xml.etree.ElementTree as et
-
+import logging
```
marcus-oscarsson: logging is never used in this file

mxcube3/__init__.py
```diff
@@ -153,7 +153,11 @@ def complete_initialization(app):
         app.AUTO_MOUNT_SAMPLE = False
         app.AUTO_ADD_DIFFPLAN = False
         app.CENTRING_METHOD = CENTRING_METHOD.LOOP
-        app.NUM_SNAPSHOTS = 4
+        num_snapshots = app.collect.getProperty('num_snapshots')
+        if num_snapshots is not None:
```
marcus-oscarsson: I think you can simply do

num_snapshots = app.collect.getProperty('num_snapshots', 4) 

to avoid the if ...

marcus-oscarsson:
Hey 

Thank for all the good fixes, looks good in general.

Please have a look at my comments, and is it possible to do something about the last merge commit ?

Marcus

mxcube3/routes/qutils.py
meguiraun: it was before I merge with master, removed

meguiraun:
About the merge commit, I already did a rebase, cherry-pick, split ... we have many of those already in master... unless you tell me an easy way to do

mxcube3/__init__.py
meguiraun: thanks

marcus-oscarsson:
What is upstream for you ?

meguiraun:
thanks

meguiraun:
Upstream is the master repo in github

PR #725, opened 2017-12-15
Save 2d centred point

meguiraun:


mxcube3/ui/components/SampleView/ContextMenu.js
```diff
@@ -244,9 +244,14 @@ export default class ContextMenu extends React.Component {
...
 
   savePoint() {
+    const sid = this.props.shape.id;
+    const shape = this.props.shapes[sid];
+
     if (this.props.clickCentring) {
       this.props.sampleActions.stopClickCentring();
       this.props.sampleActions.sendAcceptCentring();
+    } else if (!this.props.clickCentring && shape.state === 'TMP') {
+      this.props.sampleActions.sendAcceptCentring();
```
marcus-oscarsson: I'm not sure that sendAcceptCentring is that what we would like to do here, there is no centering started ?
I think its enough with just saveShape or similar ?

I think its okey for us, but we don't really know what, mxcube.diffractometer.acceptCentring(), does if there 
is no centring started.

Marcus

meguiraun: it just sets a centring point as valid and accepted (created already by move_to_beam), and then it sends centringAccepted signal

marcus-oscarsson: Aha, so it works for you guys because your diffractomter.move_to_beam starts a centring ?, so for you when you do diffractometer.acceptCentring() it saves the current centring.

But its not evident by that move_to_beam should start or setup for centring, that will probably only work for you. Following the semantics of the application we should do start a centring and then either accept or abort it, anything else would be kind of work-arounds.

marcus-oscarsson: I personally see two alternatives, either you explicitly start a centering when you do move to beam then its okey to accept it, or you just perform a saveShape or similar as I said before.

meguiraun: Our `move_to_beam` starts but also finishes a centring. The result is a temporary point on the canvas. Then the user will save or keep centring or whatever. This a quick way of getting a centring points, and not a workaround, the user can chose the precise 3 click mechanism, or the quick `move_to_beam` and get a point.

The moment you have a temporary point on the canvas, I do not see any problem on how that point has been generated. `sendAcceptCentring` is totally fine, a centring has been performed with a temporary point as output.

If your move_to_beam does not do a centring, nothing changes, there will be no temporary point, so in your case nothing needs to be accepted, and that line is never triggered.

marcus-oscarsson: Okey, that is maybe convenient for you but Id suggest that we keep the semantics as clear as we can in order to make 2D centring easily available on other sites. Building it into move_to_beam in the diffractometer makes it difficult to understand whats really happening. 

Id like that we at least try to handle the centering methods homogeneously, automatic loop centering, 3-click, n-click any other fancy centering.

Actually the logic in savePoint is a bit strange, stopClickCentring should perhaps even be on the server side. And save on all TMP points should call acceptCentring, the only call that should be in savePoint is basically acceptCentring. 

Its perhpas me who added the stopClickCentring there once, without really thinking to hard about it.

It would in my opinion be great if we could harmonize this a bit more so that it becomes more consistent.

beteva: The more generic the GUI, the better. This looks to me like: start/save whatever centring method you choose on the GUI part and the HO implements the actions to be taken when centring started and when saved. Move to beam is a completely different procedure, which does not belong to the centring methods on the GUI side, but may be the action performed on the HO side if the HO implements it such. On the other hand on the GUI there should be a clear distinction between a temporary and a saved point and there should be a possibility to save the current position as a collection point, even if it is not marked as temporary. It seems to me that the procedure, described by  @meguiraun is a centring method, which should have its own start/save, independent from move to beam.

meguiraun: Yes, our move_to_beam is (also) a 2d centring. The start of the centring is included within the move_to_beam code, the user only needs to double click on the canvas (or right clilck move_to_beam). And latter save the point (= acceptCentring). It is only our internal hwobj the one that eventually defines this operation as a centring, which in your case it is not. For me, the only difference is that there is a point as a result. And you may or may not use it

Let see if I understand your concerns correctly, if there is any kind of centring, you want the user to click somewhere that triggers the 'start_xxx_centring'? You do not accept that a centring is happening under the hood?

marcus-oscarsson: There are basically two things, in my opinion

 - all centrings should be handled in the same way whithin the mxcube core
 - move to beam, means exactly move to beam and nothing else, if we intend to do a 2d centring
   instead we should then trigger the 2d centring procedure and not move to beam.

PR #726, opened 2017-12-18
dc link based on exi usage (or not)

meguiraun:
I still do not see any cleaner alternative since we are using both interfaces here right now, but no exi.

We can discuss here or during tomorrows meeting (I will finally attend it)

### DIFF
--- a/mxcube3/routes/qutils.py
+++ b/mxcube3/routes/qutils.py
@@ -309,7 +309,15 @@ def _handle_dc(sample_node, node, include_lims_data=True):
         limsres = mxcube.rest_lims.get_dc(lims_id)
 
     # Always add link to data, (no request made)
-    limsres["limsTaskLink"] = mxcube.rest_lims.dc_link(lims_id)
+    try:
+        if mxcube.db_connection.use_exi:
+            limsres["limsTaskLink"] = mxcube.rest_lims.dc_link(lims_id)
+        else:
+            limsres["limsTaskLink"] = mxcube.db_connection.dc_link(lims_id)
+    except Exception:
+        limsres["limsTaskLink"] = "#"
+        msg = "Could not get lims link for collection with id: %s" % lims_id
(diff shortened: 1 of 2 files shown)


marcus-oscarsson:
Okey, but I think the meeting is on Wednesday right ?

Just to clarify a couple of things, the ISPyBRestClient does not necessarily have to be coupled with EXI.
Infact it can be used like you are using it just to get access to the REST services. So the ISPyBRestClient needs to be updated with a link that points to the user interface used. Right now its been called <exi_root> (but I dont think its used), we can imagine a better name for it like <lims_ui_root> or if its only for the result link <result_link_root> like I stated before.

Since you are using the SOAP services that is ISPyBClient we also need to add the same thing for the
ISPyBClient.

So my suggestion is that we; add a routine to limsutils.py that decides which ones to use, like I suggested before, we try get_result_link from ISPyBRestClient and then ISPyBClient.

I hope its clear, let me know if you have any questions ?

Marcus

meguiraun:
yes, this Wednesday 10:00, I sent a email I think.

I see what you mean, but currently the dc_link on the rest interface is only valid for an exi like url. And it always returns a link, valid or not it does not matter.

I can add a method in limsutils, but I still need a flag somewhere to accept or not the exi link that comes from the rest interface.

marcus-oscarsson:
Yes you sent a mail but in your first message you said tomorrow so I just wanted to be sure that nothing changed :).

Then we need to update the dc_link method so that if there is no result or ui link configured it returns None so that we can implement appropriate logic. Another way would be to check if the link url is set or not and then use the right method depending on that.

Marcus

meguiraun:
okey, sorry for the confusion.
ok, I will modify the dc_link on the rest interface (return None if there is no exi_url of someting set) and the add a method on the limsutils.

I will try it later today, I am now with a mockup you will like ;)

PR #727, opened 2017-12-18
Sample changer maintenance 

meguiraun:
small fix and mockup hwobj, HR pr coming soon


meguiraun:
this is what you get:

image

PR #729, opened 2017-12-18
Remote access slave read only

marcus-oscarsson:
Hi,

Remove access slave read only and small badge on-top of RA icon indicating number of observers

Marcus

### DIFF
--- a/mxcube3/routes/Login.py
+++ b/mxcube3/routes/Login.py
@@ -186,16 +186,16 @@ def request_control():
     """
     data = request.get_json()
 
+    remote_addr = remote_access.remote_addr()
+
     # Is someone already asking for control
     for observer in remote_access.OBSERVERS.values():
-        if observer["requestsControl"]:
+        if observer["requestsControl"] and observer["host"] != remote_addr:
             msg = "Another user is already asking for control"
             return make_response(msg, 409)
 
-    remote_addr = remote_access.remote_addr()
-
     remote_access.OBSERVERS[remote_addr]["name"] = data["name"]
-    remote_access.OBSERVERS[remote_addr]["requestsControl"] = True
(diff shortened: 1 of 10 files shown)


meguiraun:
I have an issue, when the second guy logins the grayed out overlay is above the 'Observer Mode', so I cannot type my name or click ok

marcus-oscarsson:
Oh, yes i forgot to fix that, ill upload a fix for that ASAP

meguiraun:
thanks for the fix. Small issues still:
* When asking for control, there is a blank cover from rigth under the 'Ask for control' until the end of the page
* When giving control the second users receives the 'obersever mode' dialog, and then the 'You were given control'
* From that point onwards, none of the users can operate the gui

marcus-oscarsson:
Hey,

Im aware of those issues although they are separate from this PR, this PR is only for the read only div and the number badge not for the entire RA feature.

The take and give control feature have been broken by previous PRs and needs to be fixed, thats what I intend to do next. The blank cover is also something else, which we need to take a look at.

meguiraun:
yes sure, I do not intend to fix the whole remote operation, I know it required more work. But with this PR is in a worse situation. We cannot apply the read only div to both master and slave at the same time

marcus-oscarsson:
That's, strange were are normally not doing that, it works fine here, given that the give take control is a bit broken/glitchy but giving control usually works one time.

I can imagine that there is a bug in the give take control feature that prevents us from correctly passing from salve to master, so that we sometimes have no master. Effectively the read only div stays. So I already started to take a look at it and there will be a PR for that as well.

marcus-oscarsson:
So I took the opportunity to add a couple of fixes to the give/take control mechanism its really not a part of this PR but I imagine it makes things easier for you to test.

meguiraun:
thanks

PR #732, opened 2017-12-20
Submodule update

marcus-oscarsson:
Hey,

:),

Marcus


marcus-oscarsson:
I will merge this myself since everybody else are on vacation and its a submodule update

PR #735, opened 2018-01-11
Login fix

marcus-oscarsson:
Hi,

Fixes to login introduced by select proposal PR, two main issues,

 - We cant categorically add mx like it was done in select_proposal.
 - The session ID is accessed directly on the session and not through session.session (which normally should not exist, do you have this structure or was it a mistake ?)

Marcus

### DIFF
--- a/mxcube3/routes/limsutils.py
+++ b/mxcube3/routes/limsutils.py
@@ -166,7 +166,7 @@ def lims_login(loginID, password):
         for prop in session['proposal_list']:
             # if len(prop['Session']) == 0:
             todays_session = mxcube.db_connection.get_todays_session(prop)
-            prop['Session'] = [todays_session]
+            prop['Session'] = [todays_session['session']]
             # elif not prop['Session'][0]['scheduled']:
             #     todays_session = mxcube.db_connection.get_todays_session(prop)
             #     prop['Session'] = [todays_session]
--- a/mxcube3/routes/limsutils.py
+++ b/mxcube3/routes/limsutils.py
@@ -214,12 +214,8 @@ def get_proposal_info(proposal):
 
 
 def select_proposal(proposal):
-    if not proposal.lower().startswith('mx'):
-	aux_prop = "{}{}".format('mx', proposal)
-    else:
-	aux_prop = proposal
-    
-    proposal_info = get_proposal_info(aux_prop)
+    proposal_info = get_proposal_info(proposal)
+
     logging.getLogger('HWR').info("[LIMS] Selecting proposal: %s" % proposal)
     logging.getLogger('HWR').info("[LIMS] Proposal info: %s" % proposal_info)
     if mxcube.db_connection.loginType.lower() == 'user' and 'Commissioning' in proposal_info['Proposal']['title']:
@@ -230,7 +226,7 @@ def select_proposal(proposal):
     if proposal_info:
         mxcube.session.proposal_code = proposal_info.get('Proposal').get('code', '')
         mxcube.session.proposal_number = proposal_info.get('Proposal').get('number', '')
-        mxcube.session.session_id = proposal_info.get('Session')[0].get('session').get('sessionId')
+        mxcube.session.session_id = proposal_info.get('Session')[0].get('sessionId')
 
         if hasattr(mxcube.session, 'prepare_directories'):
             try:
--- a/mxcube3/routes/Login.py
+++ b/mxcube3/routes/Login.py
@@ -166,15 +166,18 @@ def loginInfo():
 
     if res["loginType"].lower() != 'user' and login_info:
         # autoselect proposal
+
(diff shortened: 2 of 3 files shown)


mxcube3/routes/limsutils.py
```diff
@@ -214,12 +214,8 @@ def get_proposal_info(proposal):
 
 
 def select_proposal(proposal):
-    if not proposal.lower().startswith('mx'):
-	aux_prop = "{}{}".format('mx', proposal)
-    else:
-	aux_prop = proposal
-    
-    proposal_info = get_proposal_info(aux_prop)
+    proposal_info = get_proposal_info(proposal)
```
meguiraun: this was not a bug (although it could be improved). It happens that the proposal name can be 'MX12323' or 'mx12323' depending on ispyb (we might have slightly different versions). Please, leave this like it was..

mxcube3/routes/limsutils.py
```diff
@@ -230,7 +226,7 @@ def select_proposal(proposal):
     if proposal_info:
         mxcube.session.proposal_code = proposal_info.get('Proposal').get('code', '')
         mxcube.session.proposal_number = proposal_info.get('Proposal').get('number', '')
-        mxcube.session.session_id = proposal_info.get('Session')[0].get('session').get('sessionId')
+        mxcube.session.session_id = proposal_info.get('Session')[0].get('sessionId')
```
meguiraun: we have this working. So there is (again) a mismatch in ispyb data between us. arggghhh!!!

meguiraun:
I made a few comments, what about a skype meeting tomorrow and we discuss them? I think we have somehow different answer from ispyb.
The code that is in master now is working for us, for you?

mxcube3/routes/limsutils.py
marcus-oscarsson: Yest but the test is 

`if not proposal.lower().startswith('mx'):`

So it will match all other codes, we are for instance not using mx for all our accounts so we will get
mx prepended to our proposals, we have opid, ix, ifx and so on. So with this code we get mxopid... mxix...

So the prepending of mx needs to be more specific or handled diffrently

mxcube3/routes/limsutils.py
marcus-oscarsson: That seems really strange to me, because that data should be returned canonically by the IPSyB web-services. Ill check with Alejandro (our ISPyB developer) to see if there is the possibility of differences in the internal database structure that might cause this kind of problem.

marcus-oscarsson:
Hi Mikel,

The code that was on master was working for us before you made this change and does not work any longer, after the latest select_proposal PR that you made. We can certainly discuss things but there is not much to discuss.

 - The code that are appending or handling of mx to the proposal needs to be more specific
 -  For the session, It seems to me that this is more on the ISPyB side and such differences should
    not be in MXCuBE. It however seems strange that you have to access session twice ?

Marcus

mxcube3/routes/limsutils.py
meguiraun: aahh ok, I see

meguiraun:
That's the beauty of ispyb! I will check all those dictionaries, we have had these kind of issues before so I am not surprised at all.

marcus-oscarsson:
Have you reported those issues to the ISPyB community and do you know why you had those issues. I have spoken with our ISPyB developers and they don't really understand why you have this problem. Which version of ISPyB are you using ?

In the meantime, when you get the time could you please accept the PR, since this should have nothing todo with the MXCuBE code itself. If you have no other comments of course ?

meguiraun:
I think the issue is that in our case we made calls to `get_proposals_by_user` and in your case you get the proposal info from the `login` method (based on `user_type`). Something is kicking my head about that, I will look in detail, but the issue could be that the ispyb hwobj returns different dictionaries depending on the methods you use for getting the proposal info (probably not in ispyb, we are using a ~recent version, will check).

For the acceptance, well ... I could also say the opposite, currently is working for us. We should make this pr working for both (I mean that I can also commit to this and I will start looking at the issue later today, or at latest tomorrow)

marcus-oscarsson:
Okey,

Don't really know what to say Mikel, either you come up with a working solution or you simply accept this PR and we work from there.

Marcus

meguiraun:
@marcus-oscarsson , please could you check that the commit I did works for you? I did a quick test here and this PR with my commit works for us.
I saw that `get_todays_session` was adding an extra `session` field

mxcube3/routes/limsutils.py
meguiraun: Sorry, it was not exactly for what I said. It is for ensuring that the proposal code is also supplied together with the name. However, I am not sure why that code is there :)
It is not needed from our side, it was earlier probably, so you removal is ok

marcus-oscarsson:
Thanks, It looks fine

PR #737, opened 2018-01-11
Fixes

meguiraun:
small fixes

### DIFF
--- a/mxcube3/routes/Diffractometer.py
+++ b/mxcube3/routes/Diffractometer.py
@@ -11,7 +11,7 @@ def init_signals():
     callback method.
     """
     diffractometer = mxcube.diffractometer
-    diffractometer.connect("minidiffPhaseChanged", signals.diffractometer_phase_changed)
+    diffractometer.connect("phaseChanged", signals.diffractometer_phase_changed)
 
 
 @mxcube.route("/mxcube/api/v0.1/diffractometer/phase", methods=['GET'])
--- a/mxcube3/routes/Diffractometer.py
+++ b/mxcube3/routes/Diffractometer.py
@@ -11,7 +11,7 @@ def init_signals():
     callback method.
     """
     diffractometer = mxcube.diffractometer
-    diffractometer.connect("minidiffPhaseChanged", signals.minidiff_phase_changed)
+    diffractometer.connect("minidiffPhaseChanged", signals.diffractometer_phase_changed)
 
 
 @mxcube.route("/mxcube/api/v0.1/diffractometer/phase", methods=['GET'])
--- a/mxcube3/routes/Diffractometer.py
+++ b/mxcube3/routes/Diffractometer.py
@@ -1,9 +1,17 @@
 from flask import Response, jsonify, request, session
 from mxcube3 import app as mxcube
-from mxcube3.routes import Utils
+from mxcube3.routes import Utils, signals
 
 import json
 import logging
+
+def init_signals():
+    """
+    Connect all the relevant hwobj signals with the corresponding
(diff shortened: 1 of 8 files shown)


mxcube3/ui/components/SampleQueue/QueueControl.js
```diff
@@ -110,27 +110,34 @@ export default class QueueControl extends React.Component {
...
-    const idx = this.props.queue.indexOf(this.props.mounted);
     let nextSample = [];
-
-    if (this.props.queue[idx + 1]) {
-      const sampleData = this.props.sampleList[this.props.queue[idx + 1]];
-      const sampleName = sampleData.sampleName ? sampleData.sampleName : '';
-      const proteinAcronym = sampleData.proteinAcronym ? `${sampleData.proteinAcronym} - ` : '';
-
-      nextSample = [{ text: `Next Sample (${proteinAcronym}${sampleName})`,
+    let idx = -1;
```
marcus-oscarsson: This definition of idx seems unnecessary, cant it be moved to the line just under where idx is really set ?, idx is never
reassigned so it can be a const ...

mxcube3/ui/components/SampleQueue/QueueControl.js
```diff
@@ -110,27 +110,34 @@ export default class QueueControl extends React.Component {
...
+        const sampleData = this.props.sampleList[this.props.queue[idx + 1]];
+        const sampleName = sampleData.sampleName ? sampleData.sampleName : '';
+        const proteinAcronym = sampleData.proteinAcronym ? `${sampleData.proteinAcronym} - ` : '';
+
+        nextSample = [{ text: `Next Sample (${proteinAcronym}${sampleName})`,
+                        class: 'btn-default',
+                        action: this.nextSample,
+                        key: 2 }];
+      } else {
+        nextSample = [{ text: 'There is not next sample',
```
marcus-oscarsson: I'm not entirely sure about this, actually its maybe better to not display the button at all if we cant use it ?, thats what we
thought at least ...

mxcube3/routes/Diffractometer.py
```diff
@@ -1,10 +1,18 @@
...
 import json
 import logging
 
+def init_signals():
+    """
+    Connect all the relevant hwobj signals with the corresponding
+    callback method.
+    """
+    diffractometer = mxcube.diffractometer
+    diffractometer.connect("minidiffPhaseChanged", signals.minidiff_phase_changed)
```
marcus-oscarsson: We should probably avoid to use the name minidiff to be more generic, maybe just phaseChanged ?

meguiraun: the signal minidiffPhaseChanged comes like that from the hwobj.
I will change the method name in signals

marcus-oscarsson: I see, but you are the one that crated the signal so imagine that you can change that as well. So that someone that does have phaseChanged but not a minidiff still can use the code without having to add minidiff to their signal name.

meguiraun: not exactly, it comes from the GenericDiffractometer.
Do you have/use this signal?

mxcube3/ui/components/SampleQueue/QueueControl.js
marcus-oscarsson: This is still an issue

mxcube3/routes/Diffractometer.py
meguiraun: no big deal. Changing signal name to listen

PR #739, opened 2018-02-08
Bug fixes

marcus-oscarsson:
Hi,

Here comes what seems to be a huge PR but which is in reality a series of smaller bug fixes. Its bug fixes from this weeks testing, I think you will realize when you browse through it that there are no major changes. 

It might be interesting to know that we have been running energy scans and XRF scans all day and it works fine :)

Cheers,
Marcus

### DIFF
--- a/mxcube3/routes/Queue.py
+++ b/mxcube3/routes/Queue.py
@@ -527,6 +527,7 @@ def get_default_char_params():
     resp.status_code = 200
     return resp
 
+
 @mxcube.route("/mxcube/api/v0.1/queue/mesh", methods=['GET'])
 def get_default_mesh_params():
     """
@@ -559,6 +560,25 @@ def get_default_mesh_params():
             'subDirTemplate': '{ACRONYM}/{ACRONYM}-{NAME}',
         },
         })    
+    resp.status_code = 200
+    return resp
+
+
+@mxcube.route("/mxcube/api/v0.1/queue/xrf", methods=['GET'])
+def get_default_xrf_parameters():
+    """
+    returns the default values for a xrf scan
+    """
+    int_time =  mxcube.beamline.getObjectByRole('xrf_spectrum').\
+                getProperty('default_integration_time', '5').strip()
+
+    try:
+        int(int_time)
+    except ValueError:
+        int_time = 5
+
+    resp = jsonify({"countTime": int_time})
+
     resp.status_code = 200
     return resp
 
--- a/mxcube3/routes/Queue.py
+++ b/mxcube3/routes/Queue.py
@@ -159,7 +159,7 @@ def queue_get():
               409: On error, could not retrieve queue
     """
     logging.getLogger('HWR').info('[QUEUE] queue_get called')
-    resp = qutils.queue_to_json_response()
+    resp = qutils.queue_to_json_response(include_lims_data = True)
     resp.status_code = 200
     return resp
 
--- a/mxcube3/routes/Queue.py
+++ b/mxcube3/routes/Queue.py
@@ -569,13 +569,20 @@ def get_default_xrf_parameters():
     """
     returns the default values for a xrf scan
     """
-    int_time =  mxcube.beamline.getObjectByRole('xrf_spectrum').\
-                getProperty('default_integration_time', '5').strip()
+    int_time = 5
 
     try:
-        int(int_time)
-    except ValueError:
-        int_time = 5
+        int_time =  mxcube.beamline.getObjectByRole('xrf_spectrum').\
+                    getProperty('default_integration_time', '5').strip()
+        try:
+            int(int_time)
+        except ValueError:
+            pass
+
+    except Exception:
+        msg = "Failed to get object with role: %s" % name
+        msg += "cannot get default values for XRF"
+        logging.getLogger("HWR").exception(msg)
 
     resp = jsonify({"countTime": int_time})
 
--- a/mxcube3/routes/Queue.py
+++ b/mxcube3/routes/Queue.py
@@ -65,7 +65,7 @@ def queue_stop():
         try:
             qe.stop()
         except Exception as ex:
-            print 'exception...', ex
+            print str(ex)
 
         logging.getLogger('user_level_log').info('Queue execution was aborted, ' + str(qe.get_data_model()))
 
@@ -462,7 +462,7 @@ def get_default_dc_params():
             'helical': False,
             'mesh': False,
             'prefixTemplate': '{PREFIX}_{POSITION}',
-            'subDirTemplate': '{ACRONYM}/{NAME}-{ACRONYM}',
+            'subDirTemplate': '{ACRONYM}/{ACRONYM}-{NAME}',
         },
         'limits': mxcube.beamline.get_acquisition_limit_values()
     })
@@ -503,7 +503,7 @@ def get_default_char_acq_params():
             'skip_existing_images': False,
             'take_snapshots': True,
             'prefixTemplate': '{PREFIX}_{POSITION}',
-            'subDirTemplate': '{ACRONYM}/{NAME}-{ACRONYM}',
+            'subDirTemplate': '{ACRONYM}/{ACRONYM}-{NAME}',
             'strategy_complexity': 'SINGLE',
             'account_rad_damage': True,
             'opt_sad': False,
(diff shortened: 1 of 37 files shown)


marcus-oscarsson:
Added a small addition to this PR it now also contains a help page for sending feedback and getting help in general.

meguiraun:
thanks @marcus-oscarsson, I have just come back to work today, I will review this as soon as I have some time, hopefully tomorrow...

marcus-oscarsson:
Okey, sounds good !

mxcube3/routes/Utils.py
```diff
@@ -227,4 +232,44 @@ def take_snapshots(self, snapshots=None, _do_take_snapshot=_do_take_snapshot):
...
+def send_mail(_from, to, subject, content):
+    smtp = smtplib.SMTP('smtp', smtplib.SMTP_PORT)
+    date = email.Utils.formatdate(localtime=True)
+
+    email_msg = "From: %s\r\nTo: %s\r\nSubject: %s\r\nDate: %s\r\n\r\n%s" % (
+        _from, to, subject, date, content)
+
+    try:
+        error_dict = smtp.sendmail(_from, to.split(','), email_msg)
+    except smtplib.SMTPException, e:
```
meguiraun: can we add the content dat to the log output in case of exception? so it does not get lost

mxcube3/ui/actions/login.js
```diff
@@ -30,6 +30,19 @@ export function selectProposal(prop) {
   };
 }
 
+export function sendMail(sender, content) {
```
meguiraun: `sender` is not used later in the server

mxcube3/routes/scutils.py
```diff
@@ -152,11 +152,16 @@ def mount_sample_clean_up(sample):
...
+                mxcube.sample_changer.load(sample['sampleID'], wait=True)
             elif mxcube.sample_changer.getLoadedSample().getAddress() != sample['location']:
-              mxcube.sample_changer.load(sample['sampleID'], wait=False)
+              mxcube.sample_changer.load(sample['sampleID'], wait=True)
+
         mxcube.shapes.clear_all()
 
+        if sample['location'] != 'Manual':
+            C3D_MODE = mxcube.diffractometer.C3D_MODE
+            mxcube.diffractometer.startCentringMethod(C3D_MODE)
```
meguiraun: this should be dependent on the `auto_loop_centring`, not by default

mxcube3/routes/Queue.py
```diff
@@ -556,13 +557,33 @@ def get_default_mesh_params():
...
     resp.status_code = 200
     return resp
 
 
+@mxcube.route("/mxcube/api/v0.1/queue/xrf", methods=['GET'])
+def get_default_xrf_parameters():
+    """
+    returns the default values for a xrf scan
+    """
+    int_time =  mxcube.beamline.getObjectByRole('xrf_spectrum').\
```
meguiraun: if there is not `xrf_spectrum` this will fail, move to the following try

mxcube3/ui/components/SampleGrid/TaskItem.jsx
```diff
@@ -158,7 +161,7 @@ export class TaskItem extends React.Component {
       taskStatus = 'Collected';
     }
 
-    return `${task.label}${point} (${taskStatus})`;
+    return `${task.label} (${taskStatus})`;
```
meguiraun: a bit picky I know, but why removing the point here?

mxcube3/ui/components/SampleQueue/CharacterisationTaskItem.jsx
```diff
@@ -172,7 +172,7 @@ export default class TaskItem extends Component {
       <tr>
         <td><a>{parameters.osc_start.toFixed(2)}</a></td>
         <td><a>{parameters.osc_range.toFixed(2)}</a></td>
-        <td><a>{parameters.exp_time.toFixed(2)}</a></td>
+        <td><a>{parameters.exp_time.toFixed(3)}</a></td>
```
meguiraun: I smiled at this line, they made me put 4 here. No problem, 3 is fine

mxcube3/ui/components/SampleView/ContextMenu.js
```diff
@@ -246,8 +244,9 @@ export default class ContextMenu extends React.Component {
   savePoint() {
     if (this.props.clickCentring) {
       this.props.sampleActions.stopClickCentring();
-      this.props.sampleActions.sendAcceptCentring();
     }
+
+    this.props.sampleActions.sendAcceptCentring();
```
meguiraun: I do not understand this change

mxcube3/ui/components/Tasks/Characterisation.js
```diff
@@ -115,19 +115,14 @@ class Characterisation extends React.Component {
               </FieldsRow>
               <FieldsRow>
                 <InputField propName="exp_time" type="number" label="Exposure time (s)" />
-                <SelectField
-                  propName="beam_size"
-                  label="Beam size"
-                  list={this.props.apertureList}
-                />
+                <InputField propName="resolution" type="number" label="Resolution (Å)" />
```
meguiraun: should we disabled by default beam_size instead

mxcube3/ui/components/Tasks/DataCollection.js
```diff
@@ -87,11 +87,6 @@ class DataCollection extends React.Component {
            >
              Default Parameters
            </Button>
-           <Button bsSize="xsmall" bsStyle="default"
```
meguiraun: why? Is it confusing to have both?

meguiraun:
Thanks for all of this!

However I have some small questions, but more importantly I really did not like at all such a big pr with not only bugs (design related, code cleanup, new feedback component). It is very hard to review properly, although the code may seem clear to you. So please, I would appreciate if you split the pr next time.

I am very happy that you are running energy and XRF scans routinely!

marcus-oscarsson:
Hi Mikel, 

Thanks for the review I will take a closer look at your comments tomorrow. In general I know that its a quite large PR and I hesitated a bit before I added the help component. I did so because I though it was quite clear and that it would be easier for both me and you. I'm sorry if it made things more complicated for you to review.

The rest, which I guess could be broken up in smaller parts, is in my opinion bug fixes even if not pure logical code bugs, usability bugs. So that's why there are some design changes. In anyways Ill try to categorize even the bug fixes in the future so that it will be easier to review.

Cheers,
Marcus

mxcube3/routes/scutils.py
marcus-oscarsson: Its not the decision that we took here for the moment, it can definitely be argued but they prefer that its performed by default for the moment. The idea was the the auto_loop_centring options only applies to the queue, and that we have a consistent behavior for the rest. 

Does not really matter to me I can understand both opinions, in this case Ive let our users decide for the moment and we will see if it something that makes sense in general later on.

mxcube3/routes/Utils.py
marcus-oscarsson: I've, added some more information to the logs, you cant see it here though because your comment is on the row above the affected lines

mxcube3/ui/actions/login.js
marcus-oscarsson: I'm now attaching the sender information the user provides to the subject of the mail

mxcube3/ui/components/SampleGrid/TaskItem.jsx
marcus-oscarsson: The point was removed because some users said that they thought it might be confusing to display the point number for a already collected sample. That is confusion between points on the mounted sample and already collected ones. Also there were some cases when it was undefined which also looked strange.

So it was removed until we have come up with a better way to display it

mxcube3/ui/components/SampleQueue/CharacterisationTaskItem.jsx
marcus-oscarsson: Okey, this could perhaps become configurable, although we see little need for that at the moment. I think there is more or less an agreement on the precision needed for these values. So good if 3 is fine for you

mxcube3/ui/components/SampleView/ContextMenu.js
marcus-oscarsson: Its to accept a temporary point provided by the auto-loop centering

mxcube3/ui/components/Tasks/Characterisation.js
marcus-oscarsson: We thought it was better to just remove it since its not at all wired at the backend. It's a feature we would like but there is no way of using it at the moment. The only thing it does right now is to take space in the interface

mxcube3/ui/components/Tasks/DataCollection.js
marcus-oscarsson: Yes, the scientist here found it a bit confusing and questioned the need for doing reset and it was difficult to argue for keeping it.

marcus-oscarsson:
Hey !

I've tried to do something about your comments, I hope you find this a bit better.

Cheers,
Marcus

mxcube3/routes/scutils.py
meguiraun: okey, we have our sample changer operational yet so cannot say much against. Perhaps we could add later the option at init. Fine as it is now.

meguiraun:
thanks for the comments, and for the commits!

PR #741, opened 2018-02-14
Fixes

meguiraun:
a bunch of small fixes.

HR commit need, I will do it asap...


marcus-oscarsson:
Thanks, looks good. Could you add the sub-module update ?

PR #743, opened 2018-02-15
Exporting video files and minor fixes

marcus-oscarsson:
Hi,

Exporting video files for static builds through file-loader, fixing the sub directory name for samples that does not have sample acronym and some other fixes.

Cheers,
Marcus


meguiraun:
thanks!

PR #136, opened 2016-03-10
Result of Workshop, new queue design, new version of SampleView

bolmsten:
This PR is from the work done in the workshop and the result of the new design. 

The changes mostly affects the queue and some changes has also been made to the SampleView, the changes to the SampleView was done in order to test the new queue. I have also tried to clean up a lot of code so the warnings goes away.

Please check and give feedback.

PR #744, opened 2018-02-16
UX Fixes

marcus-oscarsson:
Hi,

  Three (together with the HR PR #136) smaller UX fixes
     -  The mounted sample cannot be removed from the queue once en-queued (unmounted needed)
     -  Dark grey border instead of blue for selected samples
     -  Points are named Point-n instead of Pn (could be confused with space groups)

Cheers,
Marcus


meguiraun:
thanks!

PR #746, opened 2018-02-22
Bug fixes and UX improvements from testing on ID29

marcus-oscarsson:
Hey,

Here's a set of bug fixes and UX improvements from last weeks operation on ID29

Bug fixes:

- Browser feels slow after a while, user reported after around 10 samples, video stream fills up browser memory, added a explicit video reset, also automatically resetting video at the start of collect 
- Preventing clicks on video while collecting
- Fixed missing session in proposal login
- Not displaying "empty" result link for characterization result

UX:

- Higher quality loader
- Retrieving LIMS information when task details is displayed
- Added possibility to filter on limsresults
- One click to sync and get samples
- Using Point, Line and Grid labels for items in the queue

Marcus


mxcube3/routes/limsutils.py
```diff
@@ -167,26 +170,24 @@ def lims_login(loginID, password):
...
-                                    login_data['Proposal']['number'])
+                       get_proposal(login_res['Proposal']['code'],
+                                    login_res['Proposal']['number'])
+
         except:
             logging.getLogger('HWR').error('[LIMS] Could not login to LIMS')
             return dict({'status': {'code': '0'}})
 
-        login_res['proposalList'] = [proposal]
-        login_res['status'] = status
```
meguiraun: is this removed on purpose? (status)

marcus-oscarsson:
Yes because login_res already contains status so we don't need it

meguiraun:
thanks!

PR #748, opened 2018-03-12
Added remote access chat feature

marcus-oscarsson:
Hey

Added a preliminary version of the remote access chat

Marcus


meguiraun:
thanks!

PR #751, opened 2018-03-14
Minor fixes

marcus-oscarsson:
Hi,

Some minor fixes

Cheers,
Marcus

### DIFF
--- a/mxcube3/video/streaming.py
+++ b/mxcube3/video/streaming.py
@@ -114,7 +114,8 @@ def new_frame_received(img, width, height, *args, **kwargs):
         # start the streaming process if not started or restart if terminated
         if not VIDEO_STREAM_PROCESS or VIDEO_STREAM_PROCESS.poll() is not None:
             sfpath = os.path.join(os.path.dirname(__file__), "streaming_processes.py")
-            VIDEO_STREAM_PROCESS = subprocess.Popen(["python", sfpath, VIDEO_DEVICE.name, VIDEO_SIZE])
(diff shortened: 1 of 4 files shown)


mxcube3/video/streaming.py
```diff
@@ -114,7 +114,7 @@ def new_frame_received(img, width, height, *args, **kwargs):
         # start the streaming process if not started or restart if terminated
         if not VIDEO_STREAM_PROCESS or VIDEO_STREAM_PROCESS.poll() is not None:
             sfpath = os.path.join(os.path.dirname(__file__), "streaming_processes.py")
-            VIDEO_STREAM_PROCESS = subprocess.Popen([sys.executable, sfpath, VIDEO_DEVICE.name, VIDEO_SIZE])
+            VIDEO_STREAM_PROCESS = subprocess.Popen(["python", sfpath, VIDEO_DEVICE.name, VIDEO_SIZE])
```
mguijarr: It seems like the more portable and reliable way is to use a trick like this:

```import os
python_executable = os.sep.join(os.path.dirname(os.__file__).split(os.sep)[:-2]+["bin", "python"])
```
I am afraid using just "python" could lead to surprising results when virtual envs are used, but maybe not...

marcus-oscarsson: I dont mind using this fix, but if the problem is that "python" does not point to the correct version of python within a virtual environment I would rather consider it as a problem with that virtual environment

meguiraun:
thanks to both! @mguijarr  good to see you around ;)

mguijarr:
@meguiraun thanks ;) I am turning into a jedi master "force ghost" for MXCuBE ;)

PR #753, opened 2018-03-16
Remote access, access restriction

marcus-oscarsson:
Hey !

Preliminary work to make remote access a bit more secure:

Implemented the following convention in the login routine:

A user is granted access IFF the following criteria is met:
 - User on local host and valid credentials
 - User on remote host, session already exists and valid credentials

All the routes except login is decorated with the restrict decorator which gives 
a 404 to anyone without a valid session.

This will be followed up by some more enhancements to the RA feature, such as giving local operators certain privileges to retake control, enable and disable remote access completely.

Cheers,
Marcus


meguiraun:
quick comments (no proper review, will come...)
- Why decorator instead of app.before_request? 
- I did implement here an app.before_request (not PR-ed since I was not fully sure about the feature), but allowing all the following (api path of method name)
    session_paths = ['login', 'login_info', 'signout', 'clear', '/']
    session_endpoints = ['login', 'loginInfo', 'signout', 'queue_clear', 'serve_static_file', 'static']
Which I believe should be allowed all time.

marcus-oscarsson:
Well I chose to use a decorator instead of  app.before_request we get a better control 
of what we restrict. Since app.before_request is called before all request we the need to implement the exclusion logic there. I considered to patch the route decorator and do only explicitly stating when we allow access, but people here that was deviating to much from the standard behavior of flask. Other suggestions are certainly welcome

I think we should only allow the login route as it is today, but there might be some other routes that we would like to allow in the future.

There is no point of allowing logininfo or signout for users that are not logged in, I see no specific reason for allowing them. 

queue_clear seems even dangerous to allow since anyone could clear the queue from basically anywhere.

Cheers !

meguiraun:
okey for the decorators.
The client asks first for the login_info, and with the returned info it goes to login. Or it was like that before.
I agree with your comments about signout and clear routes, but they are very useful for debugging and for cleaning the server, at least should be available locally.

marcus-oscarsson:
Well I don't think we are asking for login_info before login anymore, unless I missed it, it would be potentially dangerous to do that before the user is actually authenticated. So we should definitely avoid that if its the case. My tests here have worked fine so far but I might have missed something.

For the testing, I can see two options: 

- You inject routes from the test environment which you can use for testing. I don't think that is a particular good way of doing it.

- You make sure that the test scripts authenticate correctly, which I think would be the preferred way of doing it.

For cleaning the server, Id rather ask why you would need to clean the server from outside of the application ?. Maybe we have some issues that we need to take a look at if its the case. Also one could simply make script that logs in as a local user and flushes the server if that's really what you'd like to do.

Marcus

marcus-oscarsson:
So I took a closer look, we do indeed call login_info before the user is authenticated to display the proposal selection dialog. Its not an issue for us since we are not using it, but it will be an issue for you. 

Do you think you could move that functionality to something that only asks for the type of login, that we don't have to restrict ?

Marcus

meguiraun:
yes, I can do a `getUserType` instead of a `getLoginInfo` in the `signIn` action (This is what you meant right?). No problem.

Regarding the cleaning, if the application/server goes to an undesired state, one could clean the queue and signout any user instead of restarting the server. It is useful here if the user does not logout when going home, then the next one has problems.

As you mentioned in the description, the local operators are missing certain privileges, so is it ok to leave those api-s un-restricted for now? And put it back once the local operator privileges are granted?

marcus-oscarsson:
Yes, exactly a getUserType would be great !

I understand that its not always so easy to get the time to investigate why the server is in a undesired state, but I think we should do our best to make sure that it does not happen. We have seen very few such cases here, so I have to admit that its a bit strange that you need a routine for doing that ?

But if the server in anyways goes to an undesired state, would it not be easier to just restart it. To make sure that its in a clean state ?

I would say that even if the user is a local user ("in house" as we call it in MXCuBE) it needs to be logged in properly to be able to perform those actions, even if its done from local maintenance scripts.

Marcus

meguiraun:
I will do the `getUserType` after this pr.

This is probably due to the user login differences, the fact that a user does not logout does not affect you (that much) because your login as proposal, and then the proposal is internally auto-selected and overrides whatever was before. While in our case we expect the user manually select a proposal from a list.

I prefer to make a single API call that solves the issue rather than a full restart, also, it is not me the one  who does this but the beamline staff. It is a bit ugly that they need to restart the server.

In any case, I accept this PR as it is, you will use this code earlier than us anyway. But we do need a privileged way of kicking out users and clearing the queue

PR #756, opened 2018-04-13
Fixes

meguiraun:
small fixes here and there


mxcube3/ui/components/SampleGrid/SampleGrid.css
```diff
@@ -8,7 +8,9 @@
   display: inline;
 }
 
-.samples-grid{ 
+.samples-grid{
```
marcus-oscarsson: Not so certain about this, I think we prefer when the page scrolls at least until now. Why did you limit the height ?

meguiraun: I did this way so the top bar and filter row are always visible. If not, it is quite annoying when many samples are loaded.

mxcube3/routes/Login.py
```diff
@@ -115,7 +115,8 @@ def signout():
     mxcube.queue = qutils.new_queue()
     mxcube.shapes.clear_all()
     qutils.reset_queue_settings()
-    mxcube.session.clear_session()
+    if hasattr(mxcube.session, 'clear_session'):
```
marcus-oscarsson: Why do we need this check for session in particular and not for all HO's ?, it seems to be like there is a configuration problem if you need to do this check. If that's the case we better check for shapes as well ...

mxcube3/ui/components/SampleGrid/SampleGrid.css
marcus-oscarsson: Top bar and filter should always be visible, with the floating bar feature

mxcube3/routes/qutils.py
```diff
@@ -1752,3 +1752,9 @@ def get_task_progress(node, pdata):
 def is_interleaved(node):
     return hasattr(node, "interleave_num_images") and \
         node.interleave_num_images > 0
+
+
+def reset_queue_settings():
```
marcus-oscarsson: This could be called init, like this you can use it in both __init__ and Login

mxcube3/ui/components/Tasks/warning.js
```diff
@@ -0,0 +1,31 @@
+const warn = (values, props) => {
```
marcus-oscarsson: Why are you doing this, is it because you want to tell the users that the value differs or is it because the validation somehow is to strict ?

mxcube3/ui/components/SampleGrid/SampleGrid.css
meguiraun: then it stopped working, at least for me. Could you confirm that it works for you?

mxcube3/routes/Login.py
meguiraun: I have not submit those lines for the hardware repo, so I added this here in the meantime

mxcube3/routes/qutils.py
meguiraun: gonna rename

mxcube3/ui/components/Tasks/warning.js
meguiraun: simply to warn the users. Some mistakes were made here by users because it was not clear. This is only a warning, visual thing.
The strict validation is still needed.

mxcube3/ui/components/SampleGrid/SampleGrid.css
marcus-oscarsson: Yes, it seems like that feature is broken we have to take a look at it. Its maybe some CSS that changed

mxcube3/ui/components/Tasks/warning.js
marcus-oscarsson: Okey, I need to discuss this one with Daniele, Ill get back to you later

mxcube3/__init__.py
```diff
@@ -117,6 +117,8 @@ def init_app_state(app):
     # SampleID of currently mounted sample
     app.CURRENTLY_MOUNTED_SAMPLE = None
     app.SAMPLE_TO_BE_MOUNTED = ''
+    app.AUTO_MOUNT_SAMPLE = app.collect.getProperty('auto_mount_sample', False)
```
marcus-oscarsson: But I noticed that you are not using init_queue_settings here ?

mxcube3/routes/SampleCentring.py
```diff
@@ -128,6 +128,16 @@ def state_cb(state, motor=motor, **kw):
         except Exception as ex:
             logging.getLogger('HWR').exception(str(ex))
 
+    def zoom_state_cb(state, motor='zoom', **kw):
+        movable = Utils.get_movable_state_and_position('zoom')
+        signals.motor_state_callback(movable['zoom'], **kw)
+
+    try:
```
marcus-oscarsson: Why are you adding these back again, I think we removed them in a previous PR ?

marcus-oscarsson: There is a comment in Utils that says this:
   ```
   def get_centring_motors_info():
       # the centring motors are: ["phi", "focus", "phiz", "phiy",
       # "zoom", "sampx", "sampy", "kappa", "kappa_phi"]
   ```

So zoom is already handled in the code above

meguiraun: I will double check tomorrow in the beamline

meguiraun: You are right.

PR #757, opened 2018-04-16
Remote access: Give and take control by force

marcus-oscarsson:
Hi,

Implemented give and take control by force as well as a functionality to disable and enable remote access completely.

Moved utility functions defined in remote_access.py to loginutils and implemented all remote access functionality in ra.py

Marcus


meguiraun:
regarding the ALLOW_REMOTE, why do not we simply enable/disable during initialisation? Instead of making someone to login and enable

mxcube3/routes/ra.py
```diff
@@ -0,0 +1,228 @@
...
+    new_op["message"] = "You were given control"
+
+    socketio.emit("observersChanged", observers, namespace='/hwr')
+    socketio.emit("setMaster", new_op, room=new_op["socketio_sid"], namespace='/hwr')
+    socketio.emit("setObserver", current_op, room=current_op["socketio_sid"], namespace='/hwr')
+
+    return make_response("", 200)
+
+
+@mxcube.route("/mxcube/api/v0.1/ra/remoteaccess", methods=["GET"])
```
meguiraun: could we simplify the route to '/ra'?

meguiraun:
A small note, I feel a bit confused by the naming of the `mxcube3/routes/ra.py` file and the internal routes `xxx/ra/xxx`. I would rename to `RemoteAccess.py` and `xxx/remote_access/xxx` (or something easy to understand)

marcus-oscarsson:
Hey !

Thanks for taking a look;

For ALLOW_REMOTE we do it like that because we might want to enable/disable remote operation while operating MXCuBE. We don't want to restart MXCuBE to run remote, we simply allow it when we need it and then disable it when we dont.

For the route names I just took a name and I found ra (Remote Access) to be good. No particular reason for it.

Marcus

meguiraun:
okey, I do not have strong opinion for the enable/disable ra feature, so it is fine

mxcube3/routes/beamlineutils.py
```diff
@@ -46,13 +46,14 @@ def get_aperture():
 
 
 def get_beam_definer():
-    bd = mxcube.beamline.getObjectByRole("beam_info").beam_definer_hwobj or \
-         mxcube.beamline.getObjectByRole("beam_info").aperture_hwobj
+    if hasattr(mxcube.beamline.getObjectByRole("beam_info"), "beam_definer_hwobj"):
+        bd = mxcube.beamline.getObjectByRole("beam_info").beam_definer_hwobj
+    else:
+        bd = mxcube.beamline.getObjectByRole("beam_info").aperture_hwobj
```
meguiraun: you read my mind! I found this one just 5 min ago!

meguiraun:
thanks

PR #758, opened 2018-04-17
Fixed import and improved login routine

marcus-oscarsson:
Hi,

While testing on the beamline I noticed that I forgot to commit a changed import. We also refined the remote access login routine. So now we are using the following protocol:

A user (non in-house accounts) can login remotely:

- IFF no other account is already using the beamline, there is a valid session and remote access is enabled

In-house users can only login from the local control computer
-   They can give and take control by force as well as enable/disable remote access

Both in-house and normal users can login from the local control computer to for instance create a session

Marcus

### DIFF
--- a/mxcube3/routes/Login.py
+++ b/mxcube3/routes/Login.py
@@ -62,18 +62,28 @@ def login():
     password = params.get("password", "")
 
     try:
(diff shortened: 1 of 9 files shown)


meguiraun:
tack!

PR #759, opened 2018-04-17
VIDEO_STREAM_URL Config files

marcus-oscarsson:
Hi,

   Adding optional config files for VIDEO_STREAM_URL, so that we can keep these values out of git

Marcus


meguiraun:
thanks!

PR #762, opened 2018-04-24
New grid form

marcus-oscarsson:
Hi,

Designed a new form that gives better control of the grids. We have also foreseen to add more results to the grid, so this form is needed to be able to handle these new results in a convenient way.

Also did some clean-up of the grid related code

Marcus


mxcube3/routes/SampleCentring.py
```diff
@@ -428,6 +428,36 @@ def delete_shape(sid):
                "Zoom 6", "Zoom 7", "Zoom 8", "Zoom 9", "Zoom 10"]
 
 
+@mxcube.route("/mxcube/api/v0.1/shapes/rotate_to", methods=['POST'])
```
meguiraun: Juts to understand, why is this needed? Is it not enough with move to point?

marcus-oscarsson: Its to be able to rotate back to the angle where a grid was defined but not to move to a certain
position

PR #766, opened 2018-05-21
Fixs

meguiraun:
small things...

### DIFF
--- a/mxcube3/routes/loginutils.py
+++ b/mxcube3/routes/loginutils.py
@@ -93,7 +93,8 @@ def set_operator(sid):
     user = get_user_by_sid(sid)
     user["operator"] = True
 
-    select_proposal(user["loginID"])
+    if mxcube.db_connection.loginType.lower() != 'user':
+        select_proposal(user["loginID"])
 
 def users():
     return mxcube.USERS
--- a/mxcube3/routes/loginutils.py
+++ b/mxcube3/routes/loginutils.py
@@ -73,7 +73,10 @@ def logged_in_users(exclude_inhouse=False):
     users = [user["loginID"] for user in mxcube.USERS.itervalues()]
 
     if exclude_inhouse:
-        ih_users =["%s%s"% (p, c) for (p, c) in mxcube.session.in_house_users]
+        if type(mxcube.session.in_house_users[0]) == tuple:
+            ih_users =["%s%s"% (p, c) for (p, c) in mxcube.session.in_house_users]
+        else:
+            ih_users = mxcube.session.in_house_users
         users = [user for user in users if user not in ih_users]
 
     return users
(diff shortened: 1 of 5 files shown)


mxcube3/routes/loginutils.py
```diff
@@ -73,7 +73,10 @@ def logged_in_users(exclude_inhouse=False):
     users = [user["loginID"] for user in mxcube.USERS.itervalues()]
 
     if exclude_inhouse:
-        ih_users =["%s%s"% (p, c) for (p, c) in mxcube.session.in_house_users]
+        if type(mxcube.session.in_house_users[0]) == tuple:
```
marcus-oscarsson: Pay attention to that the session that we have today have been written to handle proposal code and number it would probably not scale to handle the kind of user accounts that you currently are using.

There is no issue with this snippet but keep it in mind, because it might cause you a problem later, for instance when generating paths ...

mxcube3/routes/loginutils.py
```diff
@@ -93,7 +93,8 @@ def set_operator(sid):
     user = get_user_by_sid(sid)
     user["operator"] = True
 
-    select_proposal(user["loginID"])
+    if mxcube.db_connection.loginType.lower() != 'user':
```
marcus-oscarsson: This is important for remote usage; to change the data path when operator changes. You would probably need to do something here to handle that.

mxcube3/routes/loginutils.py
meguiraun: I have not seen anything on this, probably because we inherited the session hwobj. But thanks for the hint anyway! You never know

mxcube3/routes/loginutils.py
meguiraun: Okey I see, in our case we rely on the select proposal modal in the client, so I would need to implement that... later ok?

meguiraun:
thanks for the comments.

WIP because I saw that the chat is a bit messed up with the z-index and is not available for the slave, gonna fix

marcus-oscarsson:
Okey, yes please fix that ! For the rest its alright with me, just a heads up !

meguiraun:
A few more things:
* Help page and chat always on top
* Fix for deny control when requesting the control

mxcube3/routes/ra.py
```diff
@@ -162,19 +162,19 @@ def request_control_response():
     if not data['giveControl']:
         data["sid"] = session.sid
 
-        # Reset request of observer, since it was denied
-        observers = loginutils.get_observers()
+        socketio.emit("setObserver", new_op, room=new_op["socketio_sid"], namespace='/hwr')
```
marcus-oscarsson: Actually I have some updates on all the code in ra.py, that I will push later. I will accept this for now, but there are some more issues that Ive solved ...

PR #767, opened 2018-05-22
Grid updates, Remote access bug fixes and improvements

marcus-oscarsson:
Hi

Grid is now stored in microns internally which makes it easier to handle zooming, consequently zooming now updates the size of the grid. There is also some more information displayed in the grid dialog, number of cells and grid dimensions.

I updated fabric to the latest version 2.2.4 since there was some issues with the version we were using. I also tried to optimize the grid drawing, we had some performance issues with large grids.

Several fixes for the remote access, most importantly giving control and file paths. Also added a sample changer status component.

Cheers,
Marcus

### DIFF
--- a/mxcube3/routes/Login.py
+++ b/mxcube3/routes/Login.py
@@ -108,7 +108,7 @@ def login():
         # Create a new queue just in case any previous queue was not cleared
         # properly
         mxcube.queue = qutils.new_queue()
-        qutils.add_defualt_sample()
+        qutils.add_default_sample()
 
         # For the moment not loading queue from persistent storage (redis),
         # uncomment to enable loading.
--- a/mxcube3/routes/Login.py
+++ b/mxcube3/routes/Login.py
@@ -108,6 +108,7 @@ def login():
         # Create a new queue just in case any previous queue was not cleared
         # properly
         mxcube.queue = qutils.new_queue()
+        qutils.add_defualt_sample()
 
         # For the moment not loading queue from persistent storage (redis),
         # uncomment to enable loading.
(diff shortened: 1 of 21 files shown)


mxcube3/ui/components/RemoteAccess/ObserverDialog.jsx
```diff
@@ -66,7 +72,7 @@ export class ObserverDialog extends React.Component {
           <FormControl
             inputRef={(ref) => { this.name = ref; }}
             type="text"
-            placeholder="Your name"
+            default={this.props.loginInfo.selectedProposal}
```
meguiraun: do we have anywhere the loginID? So that it displays username or proposal based on logintype.

mxcube3/routes/qutils.py
```diff
@@ -1751,3 +1751,17 @@ def init_queue_settings():
     mxcube.NUM_SNAPSHOTS = mxcube.collect.getProperty('num_snapshots', 4)
     mxcube.AUTO_MOUNT_SAMPLE = mxcube.collect.getProperty('auto_mount_sample', False)
     mxcube.AUTO_ADD_DIFFPLAN = mxcube.collect.getProperty('auto_add_diff_plan', False)
+
+
+def add_defualt_sample():
```
meguiraun: typo

mxcube3/routes/Login.py
```diff
@@ -108,6 +108,7 @@ def login():
         # Create a new queue just in case any previous queue was not cleared
         # properly
         mxcube.queue = qutils.new_queue()
+        qutils.add_defualt_sample()
```
meguiraun: typo

mxcube3/ui/components/RemoteAccess/ObserverDialog.jsx
marcus-oscarsson: You probably know that better than me, there is maybe something in loginInfo. We are still not using user based accounts so did not really look into that

meguiraun: okey, not big deal. I can fix that later

meguiraun:
thanks!

PR #768, opened 2018-05-23
[WIP] added a basic beamline webcam viewer

meguiraun:
It displays a button next to the chat, it is open on-demand by the user and displays the webcams defined in the config_gui (youtube video for mockups). Based on iframe, it works for us but let me know if this is not working for your cams, or it gives problems.

I know there is more work to do in this topic (like open automatically when sample loading, switch canvas, etc.), this is a first approach to have a webcam in mxcube3.

image

### DIFF
--- a/config.gui.prod.js
+++ b/config.gui.prod.js
@@ -1,4 +1,16 @@
 module.exports = {
-	phaseControl: false,
-	apertureControl: true
+	phaseControl: true,
+	apertureControl: true,
+	beamlineCameras: [{
+		url: "http://www.youtube.com/embed/xDMP3i36naA",
+		width:"320px",
+		height:"180px",
+		location:"Experimental Hutch"
+		},
+		{
+		url: "http://www.youtube.com/embed/8CnFo7qinng",
+		width:"320px",
+		height:"180px",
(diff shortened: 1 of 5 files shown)


marcus-oscarsson:
Thanks for this, I will take a look to see what we can do with this. As we have said no new 
features before 3.0, so we will have to take a careful look at this.

meguiraun:
Sure, have a look and let me know your thoughts.

marcus-oscarsson:
Hey !,

So I looked at this a while back ago, I have two technical questions/remarks: 

    - The video icon is very small and not centered, it would be nice if it could be done in a better way,
       the placement is also a bit strange since its not really in the corner of the screen 

    - The source for this video is in config.gui.prod.js and set to these demo videos by default, could we
comment those by default and also set phaseControl to false by default.

I also have one  more conceptual question and that is about the placement and choice of icon. I'm not sure the current placement is the best, especially not in remote access were it can be mistaken for some kind of video chat.

Then we of course need to evaluate this in remote access where its really needed, did you already perform some tests ?

Marcus

meguiraun:
I will close this one, it is still in my todo list but...

PR #770, opened 2018-05-25
Shape selection

marcus-oscarsson:
Good morning !

- Fixes to selection code broken by fabric.js update
- Handling the case when there is no centered position in a more consistent way

Cheers,
Marcus


meguiraun:
Thanks!!!

PR #772, opened 2018-05-28
Remote access and other minor fixes 

marcus-oscarsson:
Hi,

 - Made sure that lims and session information is treated when remote access operator / observer changes.
 - Handling centringAccepted signal in a uniform way, between UI and workflow

I would also like to improve the login routine a bit it seems unnecessarily complicated at the moment. Let me know if you have any idea or thoughts on that !?

I did some minor fixes to the error icons, mostly placement, let me know what you think.

Cheers,
Marcus


meguiraun:
thanks!

For the login, I agree is complex now, but I do not know how to simplify since we have different login mechanism plus remote access... Although I am of course open to discuss, what do you have in mind?

PR #778, opened 2018-06-14
Logging out user if no session on server and FF crash bug

marcus-oscarsson:
Hi,

 - Making sure user is logged out if the session is lost 
 - Firefox sometimes crashes on this.player.destroy()

Marcus


meguiraun:
thanks!

PR #779, opened 2018-06-14
prop_id to state and index fix

meguiraun:
small bug fix

### DIFF
--- a/mxcube3/ui/containers/SampleGridViewContainer.jsx
+++ b/mxcube3/ui/containers/SampleGridViewContainer.jsx
@@ -233,7 +233,7 @@ class SampleGridViewContainer extends React.Component {
     let proposalId;
 
     try {
-      proposalId = this.props.loginData.proposalList[0].Proposal.proposalId;
+      proposalId = this.props.loginData.selectedProposalID;
     } catch (e) {
       return;
     }
@@ -703,7 +703,7 @@ class SampleGridViewContainer extends React.Component {
  */
 function mapStateToProps(state) {
   return {
-    loginData: state.login.data,
+    loginData: state.login,
     queue: state.queue,
     loading: state.queueGUI.loading,
     selected: state.sampleGrid.selected,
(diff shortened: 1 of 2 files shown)


mxcube3/ui/containers/SampleGridViewContainer.jsx
```diff
@@ -233,7 +233,7 @@ class SampleGridViewContainer extends React.Component {
     let proposalId;
 
     try {
-      proposalId = this.props.loginData.proposalList[0].Proposal.proposalId;
+      proposalId = this.props.loginData.selectedProposalID;
```
marcus-oscarsson: Just a quick question, why are we not using selectedProposal here, it seems like
there is both selectedProposal and selectedProposalID. Does not selectedProposal also contain selectedProposalID ?

meguiraun: selectedProposal is simply the string `'MX122385'`,  selectedProposalID is a number

meguiraun: unless there is some mismatch between your login type and ours?

marcus-oscarsson: I see, okey I guess we could have stored the entire proposal but that's fine I guess. We can refine that at some other moment maybe, what do you think ?

meguiraun: I totally agree we need to review this part of the state. It is very confusing at this moment and some info is duplicated, I did not dare to refactor now since it might have unpleasant consequences... Shall we discuss in our next meeting?

marcus-oscarsson: Okey, I agree on that, lets make that one of the tasks for 3.1 since its working like this right now.

marcus-oscarsson: Thanks

PR #780, opened 2018-06-18
Chat message count and limits values

marcus-oscarsson:
Hi

 - Chat message count in chat badge
 - Formatted limit values

Marcus


meguiraun:
Thanks!

PR #786, opened 2018-07-10
Various fixes

marcus-oscarsson:
Hi 

These are some minor fixes that we have applied during the last two-three weeks of operation. Its mostly related to how samples are handled in the sample grid when collected, or when lims is synchronized.

Apart from that some fixes to the group path, and added path validation.

The commits are otherwise rather atomic, and I think the commit messages says whats done.

Cheers,
Marcus

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -219,7 +219,8 @@ def exception_handler(e):
 
 t0 = time.time()
 
-app = Flask(__name__,  static_url_path='')
+template_dir = os.path.join(os.path.dirname(__file__), "templates")
+app = Flask(__name__,  static_url_path='', template_folder=template_dir)
 app.debug = False
 
 app.config['SESSION_TYPE'] = "redis"
--- a/mxcube3/routes/Login.py
+++ b/mxcube3/routes/Login.py
@@ -94,7 +94,6 @@ def login():
         # sample to the queue
         if sample:
             scutils.get_sample_list()
-            qutils.queue_add_item([mxcube.CURRENTLY_MOUNTED_SAMPLE])
 
         # For the moment not loading queue from persistent storage (redis),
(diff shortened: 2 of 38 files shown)


meguiraun:
tack for this, I will try to review during the day

meguiraun:
I understand the point, but this modal is the only way of preventing the user to click in wrong places (send another command to the sc, move any motor, etc.).

marcus-oscarsson:
Ok, the users and scientist have found it quite useful here so I hope its the same for you ...

meguiraun:
Hi @marcus-oscarsson, our scientist did not like the idea at all. They do not want any mistake by a user by hitting anything on the sample changer, queue and so on.

Could you remove this commit from this PR so I accept it? And then we can discuss further during next meeting how to deal with the blocking in a way that fits both sites? What do you think?

mikel

marcus-oscarsson:
I would rather like to keep this as it is until we have decided what to do.

meguiraun:
Ok, as you prefer. I will try to deploy your branch and give to our scientist so they have a better feeling on the implications

marcus-oscarsson:
Added a few other fixes:

 - Replacing all colons (:) with dashes (-) in path to avoid need for escaping the path and for making it possible for certain software to even read the data.

 - Opening the sample link in a new tab

 - Prepared for server side generation of the ISPyB results, to make it possible to integrate already existing result views from lims directly in MXCuBE and to make the the result part less dependent on lims flavor.

marcus-oscarsson:
Hi,

I've made a few more commits as you can see, all related to not allowing "special" characters in the paths and improving the lims result display.

We had a short discussion here and I've made the SC dialog blocking again since it is the most common solution, that require least changes on both sides.

Cheers,
Marcus

PR #787, opened 2018-07-12
Fixes improvements

meguiraun:
hi!

- Some minor fixes for improving notification messages
- remove focus control from the on canvas controls (configurable and by default visible)
- add an open hutch button (configurable and by default hidden)
- A fix for the top bar not being static
- Enclose SC sample list within scrollable are in the SC view

### DIFF
--- a/config.gui.prod.js
+++ b/config.gui.prod.js
@@ -1,4 +1,6 @@
 module.exports = {
 	phaseControl: false,
 	apertureControl: true
+	focusControlOnCanvas: true,
+	apertureControl: true
 }
--- a/config.gui.prod.js
+++ b/config.gui.prod.js
@@ -1,6 +1,6 @@
 module.exports = {
 	phaseControl: false,
-	apertureControl: true
+	apertureControl: true,
 	focusControlOnCanvas: true,
 	apertureControl: true
 }
--- a/mxcube3/routes/SampleChanger.py
+++ b/mxcube3/routes/SampleChanger.py
@@ -122,6 +122,13 @@ def mount_sample(loc):
 def unmount_sample(loc):
     mxcube.sample_changer.unload(loc, wait=True)
     set_current_sample(None)
+    try:
+        mxcube.sample_changer.unload(loc, wait=True)
+        set_current_sample(None)
+    except Exception as ex:
+	    return 'Cannot load sample', 409, {'Content-Type': 'application/json',
+					   'message': str(ex)	
+					}
     return jsonify(get_sc_contents())
 
 @mxcube.route("/mxcube/api/v0.1/sample_changer/unmount_current/", methods=['GET'])
@@ -129,6 +136,13 @@ def unmount_sample(loc):
 def unmount_current():
     mxcube.sample_changer.unload(None, wait=True)
     set_current_sample(None)
+    try:
+        mxcube.sample_changer.unload(None, wait=True)
+        set_current_sample(None)
+    except Exception as ex:
+        return 'Cannot unload sample', 409, {'Content-Type': 'application/json',
+					   'message': str(ex)	
+					}
     return jsonify(get_sc_contents())
 
 @mxcube.route("/mxcube/api/v0.1/sample_changer/mount", methods=["POST"])
@@ -136,8 +150,10 @@ def unmount_current():
 def mount_sample_clean_up():
     try:
         scutils.mount_sample_clean_up(request.get_json())
-    except Exception:
-        return Response(status=409)
+    except Exception as ex:
+        return 'Cannot load sample', 409, {'Content-Type': 'application/json',
+					   'message': str(ex)	
+					}
     else:
         return Response(status=200)
 
@@ -146,8 +162,10 @@ def mount_sample_clean_up():
 def unmount_sample_clean_up():
     try:
         scutils.unmount_sample_clean_up(request.get_json())
-    except Exception:
-        return Response(status=409)
+    except Exception as ex:
+        return 'Cannot unload sample', 409, {'Content-Type': 'application/json',
+					   'message': str(ex)	
+					}
     else:
         return Response(status=200)
 
--- a/mxcube3/ui/components/MXNavbar/MXNavbar.css
+++ b/mxcube3/ui/components/MXNavbar/MXNavbar.css
@@ -1,3 +1,8 @@
+.navbar-fixed-top {
+  position:sticky !important;
+  top:0 !important;
+}
+
 .navbar li {
   font-weight: bold;
   letter-spacing: 0.5px;
--- a/mxcube3/ui/components/SampleQueue/QueueControl.js
+++ b/mxcube3/ui/components/SampleQueue/QueueControl.js
@@ -46,7 +46,7 @@ export default class QueueControl extends React.Component {
           { text: 'New Sample', class: 'btn-primary', action: this.showForm, key: 1 },
         ],
         LastSample: [
-          { text: 'Unmount and terminate the queue', class: 'btn-primary',
+          { text: 'Unmount', class: 'btn-primary',
             action: this.nextSample, key: 1 },
         ]
       }
--- a/mxcube3/ui/components/SampleQueue/QueueControl.js
+++ b/mxcube3/ui/components/SampleQueue/QueueControl.js
@@ -46,7 +46,8 @@ export default class QueueControl extends React.Component {
           { text: 'New Sample', class: 'btn-primary', action: this.showForm, key: 1 },
         ],
         LastSample: [
-          { text: 'Finish', class: 'btn-primary', action: this.nextSample, key: 1 },
+          { text: 'Unmount and terminate the queue', class: 'btn-primary',
+            action: this.nextSample, key: 1 },
         ]
       }
     };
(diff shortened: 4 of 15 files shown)


mxcube3/ui/components/SampleQueue/QueueControl.js
```diff
@@ -46,7 +46,8 @@ export default class QueueControl extends React.Component {
           { text: 'New Sample', class: 'btn-primary', action: this.showForm, key: 1 },
         ],
         LastSample: [
-          { text: 'Finish', class: 'btn-primary', action: this.nextSample, key: 1 },
+          { text: 'Unmount and terminate the queue', class: 'btn-primary',
```
marcus-oscarsson: Oh, this is great, un-mount is indeed more descriptive. I'm not so certain about "terminate queue" though, I think Unmount or "Unmount and Finish" or "Finish and Unmount" even better.

mxcube3/routes/Beamline.py
```diff
@@ -265,11 +265,27 @@ def prepare_beamline_for_sample():
...
+
     try:
         if hasattr(mxcube.collect, 'prepare_for_new_sample'):
             mxcube.collect.prepare_for_new_sample()
     except Exception:
         logging.getLogger('HWR').error('Cannot prepare the Beamline for a new sample')
         return Response(status=200)
     return Response(status=200)
+
+@mxcube.route("/mxcube/api/v0.1/beamline/open_hutch", methods=['PUT'])
```
marcus-oscarsson: We already have a hutchtrigger hardware object that can be configured through the beamline actions, what exactly do you want to achieve with this ?.

mxcube3/ui/components/SampleQueue/QueueControl.js
meguiraun: Our people was confuse for the 'finish', they understood finish as 'finish the rest of the queue' instead of leaving where it is

marcus-oscarsson:
Thanks for these fixes !

I have one more general concern though and it is that we are adding quite a few things to the config.gui.prod.js. I think its absolutely necessary that we have room for each site to add site specific functionality. I do think however it becomes a problem when we use it as a way to represent or alter the way shared functionality is presented to the users. This would be the case for focus, maybe even phase (depending on how its used), and now also hutch trigger (if I understood the intent correctly ?)

mxcube3/routes/Beamline.py
meguiraun: that's rigth, but I have not way at this moment to link the hutch trigger with signals from the door opening PLC system. Moreover, our scientist request this button so that the user is more 'conscious' on this. I did it configurable so it only appears if you want to

mxcube3/ui/components/SampleQueue/QueueControl.js
meguiraun: Terminate is a bit socking, but we could not figure out a better text

marcus-oscarsson: Yes finish is not a good word.

What the finish button really does, is to unmount the last sample in a queue that was just executing. The queue is not running when the user presses this button. The button is there to prevent users to forget samples on the gonio.

There are normally not any other samples in the queue when this button is presented so there should be no scenario for 'finish the rest of the queue'.

I agree that finish is not a good term, but I don't think adding terminate in anyway would make it any better ?

meguiraun:
I see your point. We do not have a way for site specific variations but the config.gui.prod.js file, and we need to use that for these small deviations on the interface we present to users. 
For the focus, I could simply remove completely but I think you use it. It seems an unnecessary button (at least here) so why giving the user the choice (thats the reason behind the change). For the open hutch button, it is a request from here. I know is better to use the Hutch Trigger hwobj, but availability of the signals made me follow this implementation.

mxcube3/routes/Beamline.py
marcus-oscarsson: I see, this is exactly what the HutchTrigger hardware object is for and its possible to enable and disable this feature from the beamline actions menu. I would rather see that we use the beamline actions way unless you don't see any reasons for doing it differently or somehow altering beamline actions to suite your needs.

mxcube3/ui/components/SampleQueue/QueueControl.js
meguiraun: The problem is that if there are some un-executed tasks in the current sample, finish could be understood as 'finish the remaining tasks'.

So, maybe simply 'Unmount sample'?

mxcube3/routes/Beamline.py
meguiraun: I thought about the beamline actions thing but not much...
I can implement this as a bl action, but the bl actions are somehow hidden, and this is a action closely related to the queue operation / sample management and so on, that why we though on putting the button next to the queue.

marcus-oscarsson:
I think we should try to avoid deviations however small in the interface unless they are necessary for certain functionality that is specific to the beamline or in other ways not general. 

Focus is more a property of the microscope than it is a diffractometer motor (even if focus often is the same as phi-x). So we for this reason display both, if one should be hidden or removed its more the one in the motors list.

I'm not sure I follow the problem with the hutch trigger ?, how are the signals available now ?

mxcube3/ui/components/SampleQueue/QueueControl.js
beteva: Why not "Simply unmount" or "Unmount only". It might be more explicit if you do it the other way arround: "Unmount and countinue" and "Unmount".

mxcube3/routes/Beamline.py
marcus-oscarsson: Hm, I maybe misunderstood what you want to do with your hutch trigger.

How is the hutch trigger closely related to the queue operation and sample management, its only triggered when you enter or exit the hutch ?, there should be neither of those going on when you are in the hutch ?

beteva: With mxcube2 we have the hutchtrigger on/off as well, with the possibility to configured it as automatic or manual. The experience shows that it has never been used as manual, except for us doing tests of the software. For me hutchtrigger is more hoe to protect the equipment when somebody enters the hutch. It has lost its original idea of saving time when closing the hutch, as the actions are more related to mount/unmount sample. I do not think there should be a button next to the queue - this will only make the design crowded. Action, not related to a signal from PLC, are not really "hutchtrigger" related and are maybe better implemented in specific procedures, hidden from the users, as they will very likely forget to hit the button.

meguiraun:
we are at minimum here and things pop continuously... I will remove the hutch thing from here into bl actions. Will try later today...
Thanks for the feedback

meguiraun:
removed the open hutch button and changed the queue control to 'Unmount'

marcus-oscarsson:
Thanks, I will give this a try tomorrow !

config.gui.prod.js
```diff
@@ -1,4 +1,6 @@
 module.exports = {
 	phaseControl: false,
 	apertureControl: true
```
marcus-oscarsson: There is a comma missing at the end of this line, you get a syntax error when you build the project ...

mxcube3/ui/components/MXNavbar/MXNavbar.css
```diff
@@ -1,3 +1,8 @@
+.navbar-fixed-top {
```
marcus-oscarsson: Hm, this does not seem to be the fix that I had in mind, this makes the entire "nav" bar floating, which I agree might be useful . But the original problem is that the sample grid "control" bar used to be floating, which I think is even more important. Either both or only the sample grid "control" bar should be floating.

The problem seems to be that the "nav" bar hides the sample grid "control" bar when its floating, applying an offset when floating might work ...

mxcube3/ui/index.html
```diff
@@ -4,7 +4,7 @@
   <meta charset="UTF-8">
   <title>MXCuBE 3</title>
 </head>
-<body style="padding-top: 55px; background-color: #F5F5F5; height:100%;">
+<body style="/*padding-top: 55px;*/ background-color: #F5F5F5; height:100%; ">
```
marcus-oscarsson: Its nice if we can remove commented code completly

mxcube3/routes/SampleChanger.py
```diff
@@ -122,22 +122,38 @@ def mount_sample(loc):
 def unmount_sample(loc):
     mxcube.sample_changer.unload(loc, wait=True)
     set_current_sample(None)
+    try:
+        mxcube.sample_changer.unload(loc, wait=True)
+        set_current_sample(None)
+    except Exception as ex:
+	    return 'Cannot load sample', 409, {'Content-Type': 'application/json',
```
marcus-oscarsson: The formatting here seems a bit funny, or is it just me ?

meguiraun: I do not understand your comment

mxcube3/ui/components/MXNavbar/MXNavbar.css
meguiraun: I would say both should be always visible, I do not know about the 'control' bar, but the nav bar was floating before so thats what I tried to fix. I know it is not the "best" fix, but it is working as intended

mxcube3/routes/SampleChanger.py
marcus-oscarsson: The {} are missaligned for one and the dictionary keys as well ..., at least that's what I see here ?

PR #789, opened 2018-07-17
Passing missing prop

marcus-oscarsson:
Hi,

Discovered a missing prop in warnings.js

Marcus


meguiraun:
thanks and sorry for the delay...

PR #790, opened 2018-07-25
 Navigation cross and 2D Centring

marcus-oscarsson:
Hi,

We have been working with and trying out  "2D-centerd points" and a "navigation cross" the last few days. We hope, except for being convenient, that it will be very useful for plate and jet experiments. The conditions under which one can perform the different actions might need to be refined a little bit.

Please give it a try and let us know what you think ?

This PR depends on #206 on HR so please merge that one first.

Introduces:
-  "2D-centered points", points that are normally only valid in the plane seen from the microscope.
- A "navigation cross" / Joystick for aligning the sample in the plane, using horizontal and vertical pseudo motors for moving sampx and sampy.

The micro diffractometer already have a native vertical pseudo-motor and the horizontal motor is the same as phiy or phyz (depending on orientation of spindle)

Cheers,
Marcus

### DIFF
--- a/config.gui.prod.js
+++ b/config.gui.prod.js
@@ -1,6 +1,5 @@
 module.exports = {
-	phaseControl: false,
-	apertureControl: true,
-	focusControlOnCanvas: true,
-	apertureControl: true
+  phaseControl: false,
+  apertureControl: true,
+  focusControlOnCanvas: true
 }
--- a/mxcube3/routes/Diffractometer.py
+++ b/mxcube3/routes/Diffractometer.py
@@ -85,7 +85,6 @@ def md_in_plate_mode():
 @mxcube.restrict
 def get_movables_state():
     ret = Utils.get_centring_motors_info()
-
     ret.update(Utils.get_light_state_and_intensity())
 
     resp = jsonify(ret)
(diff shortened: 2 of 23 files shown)


meguiraun:
hej! thanks for this!

I only have one little concern. So far, we always need a point before adding a new task. with these changes one could add a task without being aware of any point. What do you think about having one single entry in the context menu for 'create a 2D point'?

For the joystick ... looks great!

marcus-oscarsson:
Hey Mikel,

Just getting back from vacation, so reading through all the mails and issues. Sorry for not
responding sooner !

We wanted to avoid the extra step of creating a position, the position is created "on the fly". I don't believe that adding this extra step will actually help the users, I think most users will actually see it 
as an unnecessary step. But there might of course be cases we did not think about, do you have anything special in mind ?

Cheers,
Marcus

meguiraun:
I assumed that, hope you had great time! We had a great weather here in Sweden!

Ok, I do not have our scientist around so fine as it is. We can fine tune later if needed.

thanks for this!

meguiraun:
By the way, sprint meeting next week? at least you and me?

PR #792, opened 2018-09-03
Bug fixes

meguiraun:
Hej, various fixes here and there


marcus-oscarsson:
Sorry, I missed that you removed the WIP, Ill take a close look tomorrow when I have sometime to test.
Its a good initiative to add testing, I wanted to speak about it on our last meeting but you were not available. 

My intention would be to test at the level of the UI-API which means that both MXCuBE3 and 2 can use the same test suite. Both testing the UI API itself and the basic functionality of the core part of MXCuBE. I guess we could still have some tests specifically for the REST API but it might also mean that we are doing the same effort twice. Lets see if we can conclude something regarding this on Wednesday

Regards,
Marcus

marcus-oscarsson:
There is a minor issue with the dequeue sample(s), its now possible to dequeue the currently mounted sample, which we have said should not be possible.

meguiraun:
good catch! your functional testing is better than mine!

Fixed and merged with master.

We used to have a UI testing some time ago that was lost in the process.

The UI-API is somehow out of mxcube3 (at least for now), if we follow that route, then yes, better focus the testing in the common part. But, in the meantime having some tests in mxcube3 is a nice thing, at the end is a matter of priorities,  we have lived without them in mxcube3 so long already, and I would say that an incomplete set of test (as it is now) is more or less useless...

marcus-oscarsson:
Thanks, for the fix. Works great now !

PR #793, opened 2018-09-03
LIMS Results, Grid improvement, SC

marcus-oscarsson:
Hi,

- More generic LIMS result handling, and added characterization results 

- Added omega angle to grid form, and hiding grid when out of a certain, configurable, angle 

- Improved the handling of certain sample changer mount and unmount cases, like when there is
  no sample on a certain or when a incorrect gripper is used. (Basically making sure that the exception 
  triggered on load reaches the user)

- And some minor fixes to validation

Marcus


meguiraun:
hej! Thanks for this, could you post some images if how the results page looks?

marcus-oscarsson:
The view is exactly the same as in ISPyB, but shown in MXCuBE. If you really need it ill try to get some real data but I need to wait until the beamline is free, or I can give you sample from our ISPyB if you like.

marcus-oscarsson:
The results would look like this:

ex

meguiraun:
Thanks for the plot!

it is hard to test with mockups, I have a DC collected but it says: 'No results yet, processing ...'
Can we populate with something for the mockup result? so that there is always something to render the template. I do not mind if there is not data at all, just render the page with 0-s everywhere. Or if it is too much to ask, at least do not remove the current task summary if there are no results.

I also needed to install package babel-plugin-transform-class-properties, otherwise I could not launch the dev server, is it only me?

marcus-oscarsson:
Oh, I see what I do is that I hook up our mockup version to our test version of ISPyB so that I can 
get something that resembles reality. I guess you could implement a method in the ISPyBRestClient that always returns some test data.

marcus-oscarsson:
Sorry i missed the thing about "babel-plugin-transform-class-properties", We are using class properties since before as far as I understand class properties ...

marcus-oscarsson:
Not showing results but task overview if there are no results to show ...

This is how it looks for a normal data collection when there are results:
id30a3-dc-results

meguiraun:
Great! thanks for the fixes!

For the babel-plugin, I know, I saw this when I tried with a fresh clone... something to keep an eye on...

rhfogh:
Hi,

Just FYI, Global Phasing has an image simulator that will simulate images for a test crystal depending on orientation, strategy etc. It is what I use for those demos, so we can analyse the characterisation images and have the workflow program take decisions based on the result. If people are  interested we could try to set it up so that this image simulation facility was generally available in mock mode. 

Rasmus

marcus-oscarsson:
Hi Rasmus,

That would probably really useful in some cases, this PR more concerned retrieving and displaying information from LIMS (ISPyB), that is information that already exists ... some one that probably be even more interesting in simulating collects is Olof Svensson here at ESRF ...

PR #796, opened 2018-09-27
Refactoring moved routes/*utils to /core

marcus-oscarsson:
Hi all,

So I took some time to refactor the code a bit, its WIP'ed (Work in progress) until
we are happy with it. 

This divides the application into three more clearly separated "layers" which are:
web server, MXCuBE3 code and finally access to HardwareObjects (blcontrol)

Except for technical advantages, this also makes it easier for new comers to 
understand the code and know where to look to find certain parts of the application.

 - Introduce an object (module) through which we access all the Hardware Objects in
   Hardware Repository. I've called this object "blcotnrol", I'm open for other 
   suggestions on the name

 - Renamed the flask "app" object to "server"

 - Introduce an app object (module) through which we access all the MXCubE3 specific
   logic. I have for the moment imported this as mxcube. Again let me know if you
   prefer another name

 - Moved all MXCuBE3 specific logic that was previously in routes/*utils to a new module
   called core. So that routes only contain client server communication and "relays" the
   rest to core or blcontrol

Please let me know what you think, especially on naming but also on general organisation.
I would like to move on with this as soon as possible. There are still some smaller changes
to be made, but those can be done gradually after we have discussed those in more detail.

For a place to start looking please see, [blcontrol](https://github.com/mxcube/mxcube3/blob/refactor/mxcube3/blcontrol.py), [app](https://github.com/mxcube/mxcube3/blob/refactor/mxcube3/app.py), [core](https://github.com/mxcube/mxcube3/blob/refactor/mxcube3/core)

Thanks,
Marcus


meguiraun:
I still need to change my mindset... But, I still see confusion between `app` and `blcontrol`. How about  merging both into one file? `beamlineinterface.py`? I know is not exactly the same purpose but for the sake of simplicity

meguiraun:
also, the tests fail! I am thinking if we should populate the test cases (**first**) to be safe for this refactoring

marcus-oscarsson:
Hi Mikel,

Thanks for taking a look at this and for already testing it. I have as I mentioned earlier not yet gone through all routes and tested them. I wanted to share this to get input on the general idea and perhaps implement any changes before doing more extensive tests.

So the contents of this PR should not be seen as a finished solution. I'm happy to discuss it and fix the issues that we have with it. That being said I think we should get this done as soon as we can, I've already done some preliminary tests and fixed some of the issues you mention.

Most of the changes are just moving files around so its possible that there are still some imports and
references that are wrong, but it should be quite easy to sort out.

I think its a good thing to improve the tests, I have not looked at the test cases that you added but we can make sure that they run before this is accepted. I would prefer if we move on with this before adding further tests, and do the testing as we have done up till now. However, if you have some more test cases to add while we are doing this, please do so.

Ill think about somehow mixing "blcontrol" and "app" and what one in that case would call such a thing.

The purpose of "blcontrol" is to access the beamline control layer, and the idea with app is that it encapsulates all application wide variables. So their roles are clearly different but they can perhaps
somehow be combined.

Marcus

marcus-oscarsson:
So, I've gone through the basically all routes and corrected the errors that I've found. The problem with next sample and the tests that you mentioned @meguiraun should now be fixed, along with the other things that I've found.

Thanks,
Marcus

meguiraun:
thanks a lot @marcus-oscarsson for this big work!

For the testing, I can take care of half of the routes, but it it will not happen until next week I think

marcus-oscarsson:
Excellent, lets divide the routes between us !. I will also have difficulties to start this
 week, maybe on Friday. Would you like to continue with queue ?, it seems like you already
started with that ?

I can continue with sample changer, beamline and maybe lims ?

meguiraun:
Ok, I take the queue and then I pick another one

PR #799, opened 2018-10-09
Tests

marcus-oscarsson:
Hi,

So I fixed a small issue with the is_local_host function returning false when the test client is used. So routes with @ restrict could not be tested, which was basically all ! That's sorted now though, and all the tests run and pass !

I've added a test "suite" for the diffractomter routes. I might have found a problem with the mockups and set_phase, you might be interested in looking into that (since you are the ones using set_phase). 

Basically the test fails at the end when I move back to the original phase. I commented out that part of the test for the moment, please have a look if you have the time.

Cheers,
Marcus

### DIFF
--- a/mxcube3/core/loginutils.py
+++ b/mxcube3/core/loginutils.py
@@ -183,9 +183,14 @@ def is_local_host():
     localhost_list = socket.gethostbyname_ex(socket.gethostname())[2]
     localhost_list.append("127.0.0.1")
 
-    remote_addres = remote_addr()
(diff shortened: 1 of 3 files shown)


meguiraun:
thanks for this. I will check the set_phase just in case, it must be something related to mockups, I did no notice anything in real life.

By the way, @13bscsaamjad (Abdullah) has joined our team recently and I gave him the task for building the tests for the queue. He is having fun with the queue these days, hopefully he will do a PR soon.

PR #800, opened 2018-10-11
[WIP] Harmonized motor, attributes and movables to simply movables

marcus-oscarsson:
Hi,

Attempt to harmonize, motors, attributes and movables. They are now all simply referred to
as movables and they are manipulated through the same routes (beamline). Its possible to
take this a bit further to also create a clear definition of what each movable is, which I hope
to get time to do in the coming weeks. This is however a start that hopefully makes things
a bit clearer. 

There are a lot of files touched but its mostly name changes :)

The tests are updated and are all passing.

Marcus

### DIFF
--- a/mxcube3/core/beamline_setup.py
+++ b/mxcube3/core/beamline_setup.py
@@ -54,8 +54,10 @@ class _BeamlineSetupMediator(object):
         socketio.emit("workflowParametersDialog", params, namespace="/hwr")
 
     def getObjectByRole(self, name):
+        name = name.lower()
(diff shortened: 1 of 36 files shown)


meguiraun:
I am not fully convinced of accessing diffractometer motors via beamline api, it seems to me more confusing. How do I get now all the diff motors at once (without beamline movables)?

Also `{item_name: {"name": item_name, "state": None, "value": None}}`, if this is a movable I would call `Position` instead of `value`

PR #801, opened 2018-10-15
Handling sample "loading and loaded" state in better way

marcus-oscarsson:
Hi,

Improved the handling of the sample loading and unloading so that the corresponding dialog is displayed correctly.

Handling sample "loading and loaded" state mount and unmount instead of in sc_state_changed callback. The sample changers seem to handle loading and loaded events poorly and uses ready and busy instead, so its difficult to accurately know when a sample have been mounted or unmounted. This code relies on that the mount and unmount returns when the actual loading or unloading phase is done.

Regards,
Marcus

### DIFF
--- a/mxcube3/ui/actions/sampleChanger.js
+++ b/mxcube3/ui/actions/sampleChanger.js
@@ -1,5 +1,6 @@
 import fetch from 'isomorphic-fetch';
 import { showErrorPanel } from './general';
+import { setCurrentSample, clearCurrentSample } from './queue';
 
 export function setContents(contents) {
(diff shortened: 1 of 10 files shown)


mxcube3/ui/actions/sampleChanger.js
```diff
@@ -97,50 +98,64 @@ export function scan(address) {
...
+            successCb();
+          }
+        }
+      });
+    }
   };
 }
 
-export function unloadSample(address) {
+export function unloadSample(location) {
```
meguiraun: is `sampleData` in loadSample arg same datatype as here?
Should we name them the same?

PR #802, opened 2018-10-17
Added beamstop alignment and cleaned up MotorInput related components

marcus-oscarsson:
Hi,

Added dialog for beamstop alignment, currently only the beamstop distance, but the y and z positions  can easily added if needed. Adjusting the beamstop distance can be used to decrease the beamstop shadow.

The beamstop alignment dialog is only available if the beamstop_distance motor is present, and its displayed when the label of the beam stop in-out component is clicked. It's not certain for the moment if the beamstop alignment dialog will remain there or will be moved in the future. We chose this placement for the moment since it's discrete and the feature is not used very often.

Also cleaned up some motor related components.

Cheers,
Marcus


meguiraun:
@marcus-oscarsson, I discussed with our scientist, and at this moment we are not in a situation where we can allow the users manipulate the beam stop distance. Would it be possible to make i the component appear to be configurable? Or perhaps, the beamstop hwobj should be readonly in such scenario?

marcus-oscarsson:
Hi Mikel,

The input only appears of you have the beamstop distance motor configured, is that enough as configuration or do you also need an options somewhere ?

Marcus

meguiraun:
Ok, then that will do it.

PR #803, opened 2018-10-18
queue api testing

13bscsaamjad:
Hey, I am Abdullah. I recently started working at MAXIV and for the purpose of learning mxcube, I added the tests for one of the routes; queue with the help of Mikel. I also removed some of the unused API calls from the queue routes.

Cheers.


mxcube3/routes/queue.py
```diff
@@ -255,37 +248,6 @@ def update_sample(sample_id):
         return Response(status=409)
 
 
-@server.route("/mxcube/api/v0.1/queue/<node_id>/toggle", methods=['PUT'])
```
marcus-oscarsson: I'd rather that we keep this method, its true that its not used for the moment but it's a existing feature of the queue, and we might want to use it.

test/test_queue_routes.py
```diff
@@ -3,83 +3,360 @@
...
+    parameter_update_value = 10
+    task_to_update['parameters'][parameter_to_update] = parameter_update_value
+    rv = client.post(('/mxcube/api/v0.1/queue/{}/{}').format(queue_id, task_queue_id),
+                     data=json.dumps(task_to_update),
+                     content_type='application/json'
+                     )
+    assert (rv.status_code == 200 and json.loads(rv.data).get('parameters')[parameter_to_update] == parameter_update_value)
+
+def test_queue_start(client, add_sample, add_task):
+    """Test if we can start the queue. The queue requires a sample and a task to start which are added by fixtures."""
```
marcus-oscarsson: What do you think about also checking if the queue was annually started ?

test/test_queue_routes.py
```diff
@@ -3,83 +3,360 @@
...
+    rv = client.put('/mxcube/api/v0.1/queue/start',
+                    data=json.dumps({'sid':'1:05'}),
+                    content_type='application/json')
+    assert rv.status_code == 200
+
+def test_queue_stop(client, add_sample, add_task):
+    """Test if we can stop the queue. The queue is started and then stopped."""
```
marcus-oscarsson: Same here, we could also check if the queue was actually stopped ?

marcus-oscarsson:
Hi Abdullah,

Welcome to the project :), looks great in general. I spotted a few tests that does not test the actual outcome of the particular route tested. Its perhaps not necessary depending on what the test is
designed to do, but maybe worth looking into.

The code in general looks great, and I'm a bit picky I know ... We don't really have an official
code style, but I think its good if we try to aim for PEP8 (maybe good to know as a newcomer).

Also, I did not so far run this PR. Does all the tests pass at your side ?

Marcus

test/test_queue_routes.py
13bscsaamjad: We had problems writing tests for the routes that are using gevent; pytest doesn't seem to follow the gevent calls; maybe because the calls are spawned on another thread. That is the reason we skipped pause and unpause route calls for now. The same applies for these. We couldn't understand completely how to cater this for now so we will have a look at it later.

test/test_queue_routes.py
13bscsaamjad: Same problem as with the start.

test/test_queue_routes.py
marcus-oscarsson: I don't see how the gevent calls could be spawned in another thread and even if they would be how would it be a problem ?. Your doing a HTTP call through the test client so that should in principle be independent on whatever thread running on the server, or am I wrong ?

What do you mean when you say that pytest have problems following gevent code, how did you do the actual queue state verification ?

13bscsaamjad:
Hi Marcus,

Thanks :) Thats a good observation on your part for the tests but we are being limited because of gevent as I explained in reply to your comment. 

I will look into the PEP8 guide and try to add a plugin in my editor to conform to that.

And yes, all the tests are passing on my side :)

Abdullah

test/test_queue_routes.py
mguijarr: @13bscsaamjad are you talking about coverage report? You need to configure pytest with
concurrency=gevent. 

See https://stackoverflow.com/questions/22890779/determining-test-coverage-with-py-test-gevent

meguiraun: @mguijarr , good to see you around!
we try to get the state of the queue by calling `/queue_state` (after starting it, tried with some delay as well) and then looking at the queueStatus field, it is alwasy queueStopped.
We follow the call until `HardwareObjects.QueueManager.execute()`, 
the call `self._root_task = gevent.spawn(self.__execute_task)` seems that is never executed, or well, it has no effect, the greenlet is created but we cannot see it running... we do not see any error as well.

mguijarr: @meguiraun I am very busy with BLISS but sometimes I read messages here ;) I think I will see you at NOBUGS next week right ? I will talk about this error with Marcus tomorrow... For the call you mention I guess it returns (self._root_task == a gevent task) but if it is not running it means the gevent loop is not running...

marcus-oscarsson: I see, so its likely that the gevent loop is not running properly. It might be sufficient to move the monkey patching done in the mxcube3-server startup script to mxcube3/__init__.py.

As it is now the test code is running un-patched so that is likely a problem. I'm however not sure that it's the problem that we are seeing here (but probable). I think it's quite important that we find out the reason for this issue to make sure that we are not doing something wrong.

Also I did not find a test that tests queue execution, I guess that would be problematic for the same reason ? Once solved I guess it would be good to have such a test.

marcus-oscarsson:
Hi again,

Lets see if we can sort out this potential gevent issue. What do you think about moving the monkey patching as I suggest above, it might perhaps help ?

None of the methods that you removed are used so its not that important but I was more interested in keeping the toogle_node method because it can become useful in the future. The set_queue method would probably be less useful for the moment but I don't mind if we keep it.

Cheers,
Marcus

13bscsaamjad:
Hi,

We tried moving the monkey patching as you suggested and it seems to work. I will now be writing the remaining tests and commit those soon.

Yeah, I added back set_queue by mistake, it should have been toggle_node. I will add that back as well.

Cheers,
Abdullah

test/test_queue_routes.py
13bscsaamjad: Applied moving the monkey patching and it worked for start, stop and abort queue tests.

For queue execution, the call is blocking and it expects something to be returned. However, the server does not return anything. I could not figure it out until now. I might add those tests later.

test/test_queue_routes.py
```diff
@@ -96,43 +106,88 @@ def test_add_and_edit_task(client, add_sample, add_task):
...
     assert rv.status_code == 200
 
+    rv = client.put('/mxcube/api/v0.1/queue/unpause')
+    assert rv.status_code == 200
+
+    time.sleep(2)
+    rv = client.get('/mxcube/api/v0.1/queue_state')
+    assert (rv.status_code == 200 and json.loads(
+        rv.data).get('queueStatus') == 'QueueRunning')
+
```
marcus-oscarsson: What do you think about making sure that the queue also executes to the end, I.E waiting for the queue and then making sure the queue is stopped or even that all collections passed, or is it better to do that in a separate test ?

meguiraun: can we leave that for a later PR? I would say that such case is more about internal hwobj behavior, while these tests are more API related.

marcus-oscarsson: I think it would be a pity to wait and not add it while we are working on it, but I understand that you might be to busy to do it right now. So I'll merge this if you don't have anything else to add ?

meguiraun: I would say yes, please merge this as it is for now.

marcus-oscarsson:
Thanks for adding the extra tests and fixing the small issues :)

PR #804, opened 2018-10-23
[WIP]LIMS sample id and grid cell spacing

marcus-oscarsson:
Hi all,

- Updating the LIMS sample id of samples that are already in the queue when synch with LIMS is performed
- Adjusted grid cell spacing to mean spacing between beams instead of between cells
- Improved grid click handling and added rows and cols display to grid table

Regards,
Marcus


marcus-oscarsson:
Replacing this PR with to smaller ones

PR #807, opened 2018-11-05
Fixes

meguiraun:
various fixes, see commit messages

### DIFF
--- a/mxcube3/core/qutils.py
+++ b/mxcube3/core/qutils.py
@@ -1854,8 +1854,6 @@ def queue_stop():
             print str(ex)
         
         try:
-            logging.getLogger('user_level_log').info(
-                'Queue execution was aborted, ' + str(qe.get_data_model()))
             blcontrol.queue.queue_hwobj.set_pause(False)
             # the next two is to avoid repeating the task
             # TODO: if you now run the queue it will be enabled and run
--- a/mxcube3/core/qutils.py
+++ b/mxcube3/core/qutils.py
@@ -1014,7 +1014,7 @@ def set_dc_params(model, entry, task_data, sample_model):
 
     # Only get a run number for new tasks, keep the already existing
     # run number for existing items.
-    if not params.get("queueID", ""):
+    if not task_data.get("queueID", ""):
         acq.path_template.run_number = get_run_number(acq.path_template)
 
     model.set_enabled(task_data['checked'])
--- a/mxcube3/core/qutils.py
+++ b/mxcube3/core/qutils.py
@@ -1851,16 +1851,18 @@ def queue_stop():
             qe.stop()
         except Exception as ex:
             print str(ex)
-
-        logging.getLogger('user_level_log').info(
-            'Queue execution was aborted, ' + str(qe.get_data_model()))
-
-        blcontrol.queue.queue_hwobj.set_pause(False)
-        # the next two is to avoid repeating the task
-        # TODO: if you now run the queue it will be enabled and run
-        qe.get_data_model().set_executed(True)
-        qe.get_data_model().set_enabled(False)
-        qe._execution_failed = True
+        
+        try:
+            logging.getLogger('user_level_log').info(
+                'Queue execution was aborted, ' + str(qe.get_data_model()))
+            blcontrol.queue.queue_hwobj.set_pause(False)
+            # the next two is to avoid repeating the task
+            # TODO: if you now run the queue it will be enabled and run
+            qe.get_data_model().set_executed(True)
+            qe.get_data_model().set_enabled(False)
+            qe._execution_failed = True
+         except Exception as ex:
+            print str(ex)
 
         blcontrol.queue.queue_hwobj._is_stopped = True
         signals.queue_execution_stopped()
--- a/mxcube3/core/qutils.py
+++ b/mxcube3/core/qutils.py
@@ -1847,25 +1847,22 @@ def queue_stop():
         blcontrol.queue.queue_hwobj.stop()
     else:
         qe = blcontrol.queue.queue_hwobj.get_current_entry()
-        # check if a node/tas is executing and stop that one
-        try:
-            qe.stop()
-        except Exception as ex:
-            print str(ex)
-        
-        try:
+        # check if a node/task is executing and stop that one
+        if qe:
+            try:
+                qe.stop()
+            except Exception as ex:
+                print str(ex)
             blcontrol.queue.queue_hwobj.set_pause(False)
             # the next two is to avoid repeating the task
             # TODO: if you now run the queue it will be enabled and run
             qe.get_data_model().set_executed(True)
             qe.get_data_model().set_enabled(False)
             qe._execution_failed = True
-         except Exception as ex:
-            print str(ex)
 
-        blcontrol.queue.queue_hwobj._is_stopped = True
-        signals.queue_execution_stopped()
-        signals.collect_oscillation_failed()
+            blcontrol.queue.queue_hwobj._is_stopped = True
+            signals.queue_execution_stopped()
+            signals.collect_oscillation_failed()
 
 
 def queue_pause():
--- a/mxcube3/core/qutils.py
+++ b/mxcube3/core/qutils.py
@@ -276,6 +276,7 @@ def _handle_dc(sample_node, node, include_lims_data=False):
     parameters = node.as_dict()
     parameters["shape"] = getattr(node, 'shape', '')
     parameters["helical"] = node.experiment_type == qme.EXPERIMENT_TYPE.HELICAL
+    parameters["mesh"] = node.experiment_type == qme.EXPERIMENT_TYPE.MESH
 
     parameters.pop('sample')
     parameters.pop('acquisitions')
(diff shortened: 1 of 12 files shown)


mxcube3/core/qutils.py
```diff
@@ -1851,16 +1852,18 @@ def queue_stop():
...
-
-        blcontrol.queue.queue_hwobj.set_pause(False)
-        # the next two is to avoid repeating the task
-        # TODO: if you now run the queue it will be enabled and run
-        qe.get_data_model().set_executed(True)
-        qe.get_data_model().set_enabled(False)
-        qe._execution_failed = True
+        
+        try:
+            logging.getLogger('user_level_log').info(
```
marcus-oscarsson: What is this try handling ?, and would it not be better to write any exception to log ?

mxcube3/core/scutils.py
```diff
@@ -19,6 +19,7 @@ def init_signals():
 
     """Initialize hwobj signals."""
     blcontrol.sample_changer.connect('stateChanged', signals.sc_state_changed)
+    blcontrol.sample_changer.connect('pathSafeChanged', signals.path_safe_changed)
```
marcus-oscarsson: Do you think we can come up with a name that is more general name
than pathSafeChanged maybe isSafe or safe ?

mxcube3/routes/signals.py
```diff
@@ -121,6 +121,15 @@ def sc_unload(location):
 
     socketio.emit('sc', msg, namespace='/hwr')
 
+def path_safe_changed(*args):
+    # responds to the 'pathSafeChanged' signal in cats like SC
```
marcus-oscarsson: Maybe remove cats specific comment, since this applies basically all sample changers using an robotic arm ?

mxcube3/routes/signals.py
```diff
@@ -121,6 +121,15 @@ def sc_unload(location):
...
+    new_state = args[0]
+    #we are only interested when it becames true
+    if new_state:
```
marcus-oscarsson: Wont it always remain true then ?, or is it re-setted somewhere else ?

mxcube3/routes/signals.py
```diff
@@ -129,7 +138,7 @@ def loaded_sample_changed(sample):
     else:
         address = ''
         barcode = ''
-
+    mxcube.shapes.clear_all()
```
marcus-oscarsson: I've guess you've added this because you are mounting samples from outside of MXCuBE ?, the way this is handled today is in mount_sample_cleanup. It turns out that the different sample changers fire this event at different moments which sometimes causes some unexpected behavior.

Its fine to keep it like this, if you are mounting/un-mounting samples from other software often. Otherwise I would remove it and rely on the behavior in mount_sample_cleanup.

Another alternative is to move some of mount_sample_cleanup to this method, but this might become more complicated than we want because of what I mentioned above.

marcus-oscarsson:
Thanks for all these fixes, please take a look at my comments

mxcube3/routes/signals.py
meguiraun: we only use mxcube ... for everything!

But I still have seen situations where not all the samples were deleted after a sample mount, thats why I ensure the deletion of old shapes here. Have you seen anything like this?

mxcube3/routes/signals.py
meguiraun: it will switch between false/true. We only want to react when is safe, so that the loading panel disappears. The signal comes from the SC

mxcube3/core/qutils.py
meguiraun: removed the unnecessary log.
It can happen that one tries to stop an empty queue (current_entry none), then you will see the red error bar in the interface.

mxcube3/core/scutils.py
meguiraun: you mean for the signal? safeAreaChanged maybe?

mxcube3/core/qutils.py
marcus-oscarsson: Hm, I see that's really strange because we can only stop a queue that is running and the queue can only be running if contains something, right ?

In anyways, I guess its good to handle the scenario jut in case. I am however wondering if its not a effect of something else going wrong that you are seeing ?

mxcube3/core/scutils.py
marcus-oscarsson: Yeah, I mean it should probably even be part of generic sample changer, it would be great if both signal and handler had a more generic name. I guess its in reality implemented as a "safe volume", but we don't really care about that only that the sample changer "is safe".

mxcube3/routes/signals.py
marcus-oscarsson: I see, but what hides the dialog ?

mxcube3/routes/signals.py
marcus-oscarsson: Hm, I have to say that we are not seeing that. The most elegant solution would be if this was handled in the loaded_sample_changed but its for the moment handled in mount_sample_cleanup which is completely fine as long as one assumes that we are only loading samples within MXCuBE. 

So all the clean up associated with mounting a sample is is handled in mount_sample_cleanup, hence its name :). I guess something else is going wrong before clear_all is called if the shapes are not removed ? Its maybe worth looking into ...

mxcube3/core/qutils.py
meguiraun: When mounting a sample (no queue operation) the cancel button on the pop up modal is linked to `sendStopQueue`, this method will fail if the queue is empty, which is the case if you simply mount a sample.
I changed a little.

The alternative would be to send abort to the SC instead of stopping the queue, but I am not sure if that is a good solution...

mxcube3/routes/signals.py
meguiraun: I could not find the issue other than this line :(

mxcube3/core/qutils.py
marcus-oscarsson: I see, okey I think the sendStopQueue is there in case we are running the queue in pipeline mode.

I think this is fine, for now but its maybe worth looking into what the desired action of cancel would be.

marcus-oscarsson: what do you think ?

mxcube3/core/scutils.py
marcus-oscarsson: We do have the safe area, its for the moment linked to the busy state so that's how its handled today but we could definitely use a feature like this. I think its general to most sample changers using a robotic arm that they have a "safe zone". My comment is simply that we could perhaps find a more general name than pathSafeXXX, the intention is in the end to indicate that the sample changer or parts of it in some notion is "safe". Simply isSafe, isCollisionSafe or something in that direction would seem logical to me ?

mxcube3/routes/signals.py
marcus-oscarsson: Okey, but what if it for some reason goes false after that ?

marcus-oscarsson: To clarify a bit, I guess its not a big deal in the end. But imagine this case loading is false and "is safe" turns false, what would happen then ? Not sending the when state is false would prevent us from handling such cases. They are probably rare so its maybe not a big deal ?

mxcube3/core/scutils.py
beteva: The business about the safe area in our case is mostly used to speed up the loading - in the sample changed hardware object we start actions once the sample is on the gonio head. I would agree with @marcus-oscarsson that we need something more generic like isCollisionSafe.

mxcube3/core/qutils.py
meguiraun: I could see that if we are cancelling via SC in operation modal, one should send a abortSC and then stopQueue if the queue was running. But, letting the users abort the SC operation may cause extra trouble (e.g. stop in the middle of a loading and the they would need to recover from there)

mxcube3/core/scutils.py
meguiraun: changed to `isCollisionSafe `

mxcube3/routes/signals.py
marcus-oscarsson: Okey, well lets leave it like this for the moment then. I still think that it's a good idea to consider situations when we are not loading but are "not safe" but for now its fine then.

mxcube3/core/qutils.py
marcus-oscarsson: Yes, that's also the reason for why are using stopQueue today and that we a while back wanted to remove the cancel possibility entirely. Its good if its possible but it does often cause other maybe more serious problems. Although its finally an "issue" of the sample changer abort function it often causes us undesired side effects

marcus-oscarsson: Lets leave it like it is for the moment, Ill re-discuss it with the people here to see if any other ideas have surfaced.

mxcube3/routes/signals.py
marcus-oscarsson: Well something is clearly going wrong if the shapes are not removed at the end of mount_sample_cleanup, how are you mounting the sample ?

marcus-oscarsson: ahh, perfect !

PR #809, opened 2018-11-09
set canvas lib

meguiraun:
fixing versions so that download snapshot is back to normal. Otherwise image is cropped


marcus-oscarsson:
Thanks, for spotting this one. Do you know whats causing this and what version ?

meguiraun:
library canvas 1.6.3 definitely wrong, which was the one trying to compile in my mac. Since it was nothing specified in the package.json, maybe you are getting the correct file

marcus-oscarsson:
I guess I was using the latest version at the time, since there was not entry for canvas. 

However you are also pinning fabric ?. 

We should not use either fabric-js or fabricjs since they don't really exist (anymore), so they can be removed.

meguiraun:
removed fabricjs

marcus-oscarsson:
So do you know what the problem was, why we had the cropped snapshots ?

meguiraun:
honestly not, I saw the library was not doing was it should so I upgraded it

marcus-oscarsson:
Ok, I see. That's enough sometimes :). I would prefer if we keep the ^ and follow the minor version since its a good way to spot any problematic updates. The version is in anyways pinned when we release ....

PR #813, opened 2018-12-05
[BUG-FIX] ProposalID lost in remote, workflow dialog

marcus-oscarsson:
Hi,

Two minor fixes: 

  - The ProposalID was previously not sent properly from the backend
  - Removed possibility to hide the workflow parameters dialog

Regards,
Marcus

### DIFF
--- a/mxcube3/core/loginutils.py
+++ b/mxcube3/core/loginutils.py
@@ -332,17 +332,11 @@ def login_info(login_info):
     user = get_user_by_sid(session.sid)
 
     if user:
-        if res["loginType"].lower() != "user":
-            res["selectedProposal"] = user["loginID"]
-        elif proposal_info:
-            code = proposal_info.get("Proposal").get("code")
-            number = proposal_info.get("Proposal").get("number")
-            proposalId = proposal_info.get("Proposal").get("proposalId")
-            res["selectedProposal"] = code + number
-            res["selectedProposalID"] = proposalId
-
+        res["selectedProposal"] = user["loginID"]
+        res["selectedProposalID"] = res['loginRes']['Session']['session']['proposalId']
     else:
-        res["selectedProposal"] = {}
+        res["selectedProposal"] = ""
+        res["selectedProposalID"] = ""
 
     return user, res
 
--- a/mxcube3/core/loginutils.py
+++ b/mxcube3/core/loginutils.py
@@ -330,10 +330,11 @@ def login_info(login_info):
     }
 
     user = get_user_by_sid(session.sid)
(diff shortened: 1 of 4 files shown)


mxcube3/core/loginutils.py
```diff
@@ -332,17 +332,11 @@ def login_info(login_info):
     user = get_user_by_sid(session.sid)
 
     if user:
-        if res["loginType"].lower() != "user":
-            res["selectedProposal"] = user["loginID"]
-        elif proposal_info:
```
meguiraun: we need this for the username based login

marcus-oscarsson: But you still have it, like this is the same regardless, I was a bit puzzled by

```
elif proposal_info:
  ...
```
Because proposal_info is always undefined, its never assigned to

marcus-oscarsson: What did you intend proposal_info to be ?

meguiraun: I guess I added myself on https://github.com/mxcube/mxcube3/commit/c34f0a7e555df31e87af7fbb43a611b5356b85e0#diff-ac7ec31afc04b853184223e8a487bce4
But I see that it does not make sense since `proposal_info` is not defined,??
I think the line we need would be the first if clause, otherwise res["selectedProposal"] would not have the correct proposal but username.
Cannot test this now, but could you change your loginType in the mockups to 'user' and test it yourself?

meguiraun: The proposal will be selected with a call from the client right after

marcus-oscarsson: yes it must have been you who added that piece of code, well Im not overly enthusiastic about testing it but I will since you can't

marcus-oscarsson: Can I just ask, what you are using today ?

meguiraun: thanks.
using today? In what sense? We are not using this code in production, we are still behind. If thats what you asked

marcus-oscarsson: Whats the logic that you are using today to select proposal and send this information ?

meguiraun: Rigth after login, the client checks the response and looks for the `loginType === 'User'`, if so ProposalsForm form is displayed, there the user selects a proposal from the available ones for that user.  
That form calls then:
 actions.login.sendSelectProposal -> postProposal  -> API ../lims/proposal -> limsutils.select_proposal

marcus-oscarsson:
So, I changed the logic slightly fixing the broken user type login

meguiraun:
thanks!

PR #819, opened 2019-02-04
Removed unused imports and variables and added future imports

marcus-oscarsson:
Hi,

Removed unused imports and variables and added future imports

Marcus

### DIFF
--- a/mxcube3/core/qutils.py
+++ b/mxcube3/core/qutils.py
@@ -111,8 +111,6 @@ def node_index(node):
 
         sample = sample_model.loc_str
         task_groups = sample_model.get_children()
-        [group.get_children() for group in task_groups]
-
         tlist = []
 
         for group in task_groups:
@@ -1332,13 +1330,6 @@ def add_data_collection(node_id, task):
     dc_model, dc_entry = _create_dc(task)
     set_dc_params(dc_model, dc_entry, task, sample_model)
 
-    dc_model.acquisitions[0].path_template
-
-#    if blcontrol.queue.check_for_path_collisions(pt):
-#        msg = "[QUEUE] data collection could not be added to sample: "
-#        msg += "path collision"
-#        raise Exception(msg)
-
     group_model = qmo.TaskGroup()
     group_model.set_origin(ORIGIN_MX3)
     group_model.set_enabled(True)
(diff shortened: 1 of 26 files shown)


mxcube3/core/qutils.py
```diff
@@ -1330,7 +1332,7 @@ def add_data_collection(node_id, task):
     dc_model, dc_entry = _create_dc(task)
     set_dc_params(dc_model, dc_entry, task, sample_model)
 
-    pt = dc_model.acquisitions[0].path_template
+    dc_model.acquisitions[0].path_template
```
meguiraun: is this line needed at all?
same for the next ones

mxcube3/core/utils.py
```diff
@@ -103,7 +106,7 @@ def get_light_limits():
 
 
 def get_movable_state_and_position(item_name):
-    ret = dict()
+    pass
```
meguiraun: why this?

mxcube3/routes/mockups.py
```diff
@@ -29,12 +29,11 @@ def mockup_newres(newres):
 def create_diff_plan(sid):
     """Juts for creating a diff plan as if it were created by edna and so on.
     """
-    from mock import Mock
 
     acq_parameters = blcontrol.beamline.get_default_acquisition_parameters()
     ftype = blcontrol.beamline.detector_hwobj.getProperty("file_suffix")
     ftype = ftype if ftype else ".?"
-    n = int(blcontrol.session["file_info"].getProperty("precision", 4))
+    int(blcontrol.session["file_info"].getProperty("precision", 4))
```
meguiraun: should go out?

mxcube3/routes/signals.py
```diff
@@ -163,7 +167,7 @@ def loaded_sample_changed(sample):
 
     try:
         # recreate the dict with the sample info
-        q = queue_to_dict()
+        queue_to_dict()
```
meguiraun: out

mxcube3/routes/signals.py
```diff
@@ -384,9 +388,9 @@ def collect_oscillation_failed(
 
     if not qutils.is_interleaved(node["node"]):
         try:
-            limsres = blcontrol.rest_lims.get_dc(lims_id)
+            blcontrol.rest_lims.get_dc(lims_id)
         except BaseException:
-            limsres = {}
+            pass
```
meguiraun: this try bloc is not needed

meguiraun:
thanks for this! I only noticed small thins, basically lines no longer needed (I guess you removed the unused var)

mxcube3/routes/signals.py
marcus-oscarsson: The call to get_dc can still go wrong so the try block would still be needed, unless you see a scenario when this never will go wrong ?

marcus-oscarsson:
Hi Mikel, 

Exactly I removed a few unused variables. As you noticed  I missed to remove some of the calls without side effects. However that's taken care of now.

Marcus

meguiraun:
thanks!

mxcube3/routes/signals.py
meguiraun: sorry, I missed this comment. I meant the actual call is not needed in this method

PR #822, opened 2019-02-20
Submodule update and test configuration

marcus-oscarsson:
Hi,

I made a few fixes on HWR which fixes the remaining two broken test cases. So all tests should now run
correctly.

Marcus


meguiraun:
Thanks!

PR #824, opened 2019-03-14
Updated dev dependencies

marcus-oscarsson:
Hi all,

Updated dev dependencies and added necessary fixes to fix compatibility issues. Mostly indentation issues due to updated eslint. The new version of eslint also added a whole bunch of new rules that we currently are NOT following. I have disabled these for the moment, since they are new, but we can re-enable some of them if we think that they are worth having.

Cheers,
Marcus


meguiraun:
thanks

PR #825, opened 2019-04-08
Fixes

meguiraun:
hej, a list of fixes I had locally...

### DIFF
--- a/mxcube3/core/qutils.py
+++ b/mxcube3/core/qutils.py
@@ -261,8 +261,12 @@ def get_queue_state():
 
     queue = queue_to_dict(include_lims_data=True)
     sample_order = queue.get("sample_order", [])
+    try:
+        current =  scutils.get_current_sample().get('sampleID', '')
+    except:
+        current = ''
 
-    res = {"current": scutils.get_current_sample().get('sampleID', ''),
+    res = {"current": current,
            "centringMethod": mxcube.CENTRING_METHOD,
            "autoMountNext": get_auto_mount_sample(),
            "autoAddDiffPlan": mxcube.AUTO_ADD_DIFFPLAN,
--- a/mxcube3/routes/lims.py
+++ b/mxcube3/routes/lims.py
@@ -111,8 +111,11 @@ def get_results():
                 r = apply_template("data-collection-results.html", data)
 
         elif isinstance(model, qmo.Characterisation) or isinstance(model, qmo.Workflow):
-            if result_file_test("characterisation-results.js"):
-                url_list = data["limsResultData"]["workflow_result_url_list"]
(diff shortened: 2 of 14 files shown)


mxcube3/core/qutils.py
```diff
@@ -261,8 +261,12 @@ def get_queue_state():
 
     queue = queue_to_dict(include_lims_data=True)
     sample_order = queue.get("sample_order", [])
+    try:
+        current =  scutils.get_current_sample().get('sampleID', '')
```
marcus-oscarsson: Do you know why get_current_sample fails ?

meguiraun: I could not figure out ... :(
Happening with real hw, some mismatch during initialisation

marcus-oscarsson: I guess It would be good to find out, right ?

meguiraun: yes sure, but at least with this it does not fail on initialisation, solution for now?
But I will look in more detail as soon as I can run tests in the beamline

marcus-oscarsson:
Is this WIP ?

mxcube3/core/qutils.py
marcus-oscarsson: Hm, I'm not super happy to add temporary fixes on master like this It would be good if we could get a more definitive fix.

meguiraun:
I do not consider this WIP, unless you request changes

meguiraun:
just saw travis failing... fixing it asap

test/test_samplechanger_routes.py
```diff
@@ -0,0 +1,86 @@
...
+
+def test_get_sc_contents_view(client):
+    """
+    Checks retrieval of the sample changer contents
+    """
+    resp = client.get("/mxcube/api/v0.1/sample_changer/contents")
+    data = json.loads(resp.data)
+    print(data)
+    assert isinstance(data["children"], list)
+    assert len(data["children"]) == 17  # pucks
```
marcus-oscarsson: Be careful here, because this only tests this specific implementation of the sample changer in this case SampleChanger Mockup. Other sample changers might have other layouts. It would probably, if possible, be better to get the actual number of baskets and samples from the sample changer HardwareObject.

meguiraun: But we run all tests with mockups

marcus-oscarsson: Sure, but we might want to use a different layout one day or even test against a real sample changer. It would be good if we could avoid hard coding values like these. There is also a print left in the test, is that intentional ?

meguiraun: modify a bit the test and added a new url for getting the capacity

PR #826, opened 2019-04-08
Sc switch

meguiraun:
A small component to allow SC on/off from the data collection view.
It comes from the workshop training

### DIFF
--- a/mxcube3/ui/containers/BeamlineSetupContainer.jsx
+++ b/mxcube3/ui/containers/BeamlineSetupContainer.jsx
@@ -4,7 +4,8 @@ import { connect } from 'react-redux';
 import { Row, Col, Table, Popover } from 'react-bootstrap';
 import PopInput from '../components/PopInput/PopInput';
 import BeamlineActions from './BeamlineActionsContainer';
-import InOutSwitch from '../components/InOutSwitch/InOutSwitch';
+import InOutSwitch2 from '../components/InOutSwitch2/InOutSwitch2';
+import SampleChangerSwitch from '../components/SampleChangerSwitch/SampleChangerSwitch';
 import LabeledValue from '../components/LabeledValue/LabeledValue';
 import MachInfo from '../components/MachInfo/MachInfo';
 import OneAxisTranslationControl from '../components/MotorInput/OneAxisTranslationControl';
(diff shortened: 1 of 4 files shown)


mxcube3/ui/containers/BeamlineSetupContainer.jsx
```diff
@@ -4,7 +4,8 @@ import { connect } from 'react-redux';
 import { Row, Col, Table, Popover } from 'react-bootstrap';
 import PopInput from '../components/PopInput/PopInput';
 import BeamlineActions from './BeamlineActionsContainer';
-import InOutSwitch from '../components/InOutSwitch/InOutSwitch';
+import InOutSwitch2 from '../components/InOutSwitch2/InOutSwitch2';
```
marcus-oscarsson: InOutSwitch2 and InOutSwitch have been merged to InOutSwitch. The idea was to only use InOutSwitch

mxcube3/ui/containers/BeamlineSetupContainer.jsx
```diff
@@ -269,12 +271,10 @@ class BeamlineSetupContainer extends React.Component {
               </Col>
               {this.createActuatorComponent()}
               <Col sm={2} className="pull-right">
-                <LabeledValue
-                  suffix=""
-                  name="Sample changer"
-                  value={this.props.sampleChanger.state}
-                  level={this.props.sampleChanger.state === 'READY' ? 'info' : 'warning'}
-                  look={'vertical'}
+                <SampleChangerSwitch
```
marcus-oscarsson: Does this also show the state of the sample changer or only if its on or off ?

marcus-oscarsson:
Could you take a closer look at why Travis does not pass ?, there seems to be some issue with "npm run build"

meguiraun:
yes, I will. Probably because the InOutSwitch, strange I did not see locally... on it.

marcus-oscarsson:
Thanks Mikel !

meguiraun:
It has been a while since I last run master... more problems that expected :)

mxcube3/ui/containers/BeamlineSetupContainer.jsx
meguiraun: It shows state, same functionality as with the shutters
image

marcus-oscarsson: Ok, great :)

PR #829, opened 2019-10-30
adds 'tango-controls' channel to conda environment

elmirjagudin:
Include 'tango-controls' channel in the conda environment
description file, otherwise conda will not find the
'pytango' with default settings.

### DIFF
--- a/conda-environment.yml
+++ b/conda-environment.yml
@@ -1,4 +1,6 @@
 name: mxcube
+channels:
+  - tango-controls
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
Thanks, There seems to be a  conflict from the last merge, could you take a look at that  ?

elmirjagudin:
I have re-based  the change onto latest master, it can now be merged without conflicts.

PR #833, opened 2019-11-14
Minor updates to HWR and server side check for "in control" when remote 

marcus-oscarsson:
Hi All,

   A few smaller update to adapt to some of the latest changes on HWR
   Added possibility to require control for certain resources

Cheers,
Marcus


marcus-oscarsson:
So I finally removed the a few commits because we have not yet made the necessary PR to HardwareRepository. I will make another PR with the API updates when the PR to HWR is done.

@meguiraun so this is no WIP any more :)

marcus-oscarsson:
Alright, yes I think that's reasonable for the moment. We should make the API changes fairly soon though. There are a few more changes for HWR to be made and then we can do those ....

PR #834, opened 2019-12-06
Get v4l2 from fork to build with Python 3

laispc:
Hi,

Just a minor fix in `requirements.txt` to get v4l2 from the fork (following it is done in `conda-environment.yml`).

Thanks,
Laís


marcus-oscarsson:
Thanks :)

PR #840, opened 2020-01-10
Remote operation

meguiraun:
First things first... Many apologies for this long delay submitting this PR. Mostly due to the looong process it took for us to have the remote operation in place and busy with other things.

What this PR does:
- Reformatted user info box 
- Staff users defined (comes from session hwobj), staff users can kick out other users, and they can always login
- Login error msg propagated into the client (instead of generic error)
- If a user is already logged in somewhere else, now he/she can force logout himself
- Safer select proposal modal, there were some ugly corner cases
- Bugs, bugs, bugs

image

Have a look and let me know what you think. Once we are happy with these changes I probably should submit this to master... 

I have more PRs in my backlog: random bugfixes, sample changer related, raster related, characterization related, webcam viewer still pending... they will come ~soon

cheer,

mikel


marcus-oscarsson:
Thanks for all these nice fixes, I have not yet had the time to go trough them all. The new user info box looks really nice, well done !. Just a quick question before going into details whats the difference between staff and the already existing inhouse user and how do you configure the staff user, is there an example config somewhere ?

meguiraun:
no hurries :)

staff comes from the in_house user list in the session hwobj. Our session hwibj is as follows:

```xml
.
.
  <inhouse_users>jienan,mikegu, ......</inhouse_users>
.
.
```

mxcube3/routes/limsutils.py
```diff
@@ -178,6 +178,9 @@ def lims_existing_session(login_res):
 def lims_is_inhouse(login_res):
     return login_res.get("Session", {}).get("is_inhouse", False)
 
+def lims_is_staff(loginID):
```
marcus-oscarsson: Hm, we need to double check the usage of `lims_is_staff` and `lims_is_inhouse` and especially if its compatible with whats done in ISPyBClient2, the lims_is_staff check is basically already done in:

https://github.com/mxcube/HardwareRepository/blob/3caa18f0d853167d43e00bead253917c20df9a56/HardwareObjects/ISPyBClient2.py#L775

The problem is that the inhouse proposals definition in session is used for both users and proposals. Actually the only thing that can be inhouse according to the old definition is a proposal (as it has to do with data storage). With many proposals per user this should be changed to in_house_proposals, we could then imagine to have a in_house_users with usernames.

mxcube3/routes/Login.py
```diff
@@ -42,21 +44,36 @@ def login():
...
+        inhouse = limsutils.lims_is_inhouse(login_res)
         info = {"valid": limsutils.lims_valid_login(login_res),
                 "local": is_local_host(),
                 "existing_session": limsutils.lims_existing_session(login_res),
                 "inhouse": inhouse}
 
-        _users = logged_in_users(exclude_inhouse=True)
+        _users = logged_in_users(exclude_inhouse=False)
+
+        common_proposal = False
```
marcus-oscarsson: Whats this `common_proposal` ?

mxcube3/routes/Login.py
```diff
@@ -42,21 +44,36 @@ def login():
 
     try:
         login_res = limsutils.lims_login(loginID, password)
-        inhouse = limsutils.lims_is_inhouse(login_res)
 
+        if login_res['status']["code"] is not "ok":
```
marcus-oscarsson: I know very picky and not so important  but the usage of " and ' is a bit inconsistent

mxcube3/routes/Login.py
```diff
@@ -42,21 +44,36 @@ def login():
...
         info = {"valid": limsutils.lims_valid_login(login_res),
                 "local": is_local_host(),
                 "existing_session": limsutils.lims_existing_session(login_res),
                 "inhouse": inhouse}
 
-        _users = logged_in_users(exclude_inhouse=True)
+        _users = logged_in_users(exclude_inhouse=False)
+
+        common_proposal = False
+        if mxcube.SELECTED_PROPOSAL is not None:
```
marcus-oscarsson: So we dont have to have a selected proposal to login ?

mxcube3/routes/Login.py
```diff
@@ -113,22 +132,112 @@ def signout():
...
+        # we are here because the user force logut itself
+        mxcube.queue = qutils.new_queue()
+        mxcube.shapes.clear_all()
+        mxcube.session.clear_session()
+
+        if mxcube.CURRENTLY_MOUNTED_SAMPLE:
+            if mxcube.CURRENTLY_MOUNTED_SAMPLE.get('location', '') == 'Manual':
+                mxcube.CURRENTLY_MOUNTED_SAMPLE = ''
+
+        LOGGED_IN_USER = None
```
marcus-oscarsson: Is this somekind of gobal or is it just a local variable with caps ?

marcus-oscarsson:
I've left a few comments it looks like good fixes in general, the most important thing to workout is the comment about `in_house_users`. Its also a rather big and diverse PR so I have to admit that its a bit tricky to digest. Id suggest that we have a quick chat about these things tomorrow maybe, so we can get this merged ?

meguiraun:
Yeah I see your point with `in_house_users`. Sure lets talk tomorrow!

Thanks for the review

mxcube3/routes/Login.py
meguiraun: two different users can login only if they are in the same proposal only, defined by this `common_proposal`

mxcube3/routes/Login.py
meguiraun: you have good eyes

mxcube3/routes/Login.py
meguiraun: When a first user authenticates, then he will select a proposal (via SelectProposal form, call to /lims/proposal) or it will be autoselected if `loginType != user` (call to logintutils.set_operator -> select_proposal ).

At that point mxcube.SELECTED_PROPOSAL will have the proposal name.

When another user tries to login, mxcube checks first if the login is valid and then if there is a proposal already selected. This `if` clause is to determine if they share a proposal, for later in 
```
        if not is_staff and not common_proposal and _users and (loginID not in _users):
            return deny_access("Another user is already logged in")
``` 

Which is probably not the best error message though...

We are using this logic in prod here with remote, multiple users, staff etc. But I could not tested very deeply with your login method. Anyway, maintaining both login methods is not easy so let me know if you have ideas to make things clearer.

mxcube3/routes/Login.py
meguiraun: it's a global, but I think is dead code... need to check

mxcube3/routes/Login.py
rhfogh: While we are nitpicking:
'!= "ok"' would be better, 'is not' dep[enmds on how strings are stored internally, whether both cases point to the same string object,. or different but identical objects.

marcus-oscarsson: very true, and maybe more important ...

meguiraun:
I have addressed the comments... at least I hope ;)
- Proposal info displayed in the Users list (RA tab), except staff users shadowing common users
- Refactor a bit in_house, is_staff etc. To make it more compatible with the traditional in_hous_proposals approach (`session.xml` has `staff_users` list)

mxcube3/routes/Login.py
```diff
@@ -113,22 +139,107 @@ def signout():
...
+
+    mxcube.SELECTED_PROPOSAL = None
+    mxcube.SELECTED_PROPOSAL_ID = None
 
+    session.clear()
+    socketio.emit("signout", {}, namespace='/hwr')
     return make_response("", 200)
 
+@mxcube.route("/mxcube/api/v0.1/forceusersignout", methods=["POST"])
+def forceusersignout():
```
marcus-oscarsson: Do you really want all this logic directly in the route an cant we reuse some logic from the logout routine ?

marcus-oscarsson:
Thanks for your fixes looks good, please let me know what you think about my last comment :)

meguiraun:
I did remove some unused code, and I agree for the simplification of remaining signout methods, but I am facing some issues and would suggest that I do that clean up when I do the PR for master. The files are in different structure in master so I want to avoid too much duplicated job. Would that be ok? (I want to start with the PR to master once this PR is merged)

marcus-oscarsson:
I would be okey with that :), good idea !

PR #841, opened 2020-02-10
Remote operation master

meguiraun:
Same changes as the previous #840 but this time to master. Still need to test due to the changes on the HW, and the tricky issues while merging (the was not much difference, but enough for renaming and moving for a headache)

I still need to tidy a bit the code for the signout and forcesignout


meguiraun:
Hej @marcus-oscarsson, ready for your eyes! let me know your thoughts. I tested with mockups for both login approaches.

It is essentially the original #840 adapted to master.

marcus-oscarsson:
Hi @meguiraun thanks for all this work, I've been quite busy so I have not had the time to go through it yet.

We are waiting for a few PR's on master to get accepted and then we will make a update with all the HR refactoring work. So id like to merge that before we merge this one, then id like to look at this and test it on the beamline. So it might take a few more days, sorry for the delay :)

Just a quickly there i maybe a good explanation for it but why are there three commits from me in the beginning ?

meguiraun:
Sure, I understand. I cannot expect anything fast since it took a long time from my side...

Those commits from you come from the 3.0.1 branch

meguiraun:
hej! Travis fails for python38, not sure why since it works in my machine... looking at it. Anyway, you can review while I figure out...

meguiraun:
closing this. Its already big and difficult to manage, your comments are in essence partially achieved (is_house is almost same as staff). However, since we have already discussed about a new login module, better to focus on that one first.

I will create new MR with bug fixes only (no functionality addtion)

marcus-oscarsson:
Thanks, that sounds good @meguiraun

PR #844, opened 2020-03-18
Updated JS dependencies and a few HWR API updates 

marcus-oscarsson:
Dear all,

Updated some of the JS dependencies and added few more HWR API updates

Marcus


meguiraun:
tack!

PR #858, opened 2020-05-06
[WIP] Added flask-RESTX to add the possibility to define resources through models

marcus-oscarsson:
Hi,

flask-RESTX makes it possible to define Resources through models, and provides a mechanism for request and response marshaling. It also provides a swagger specification based on the models and the swagger test-ui.

There is a also a Blueprint like feature called Namespace that makes it possible to define the API without importing the server instance.


marcus-oscarsson:
Putting this at the side for the moment :)

PR #863, opened 2020-09-21
use nodejs version 12.x

elmirjagudin:
A couple of commits for running front-end with nodejs version 12.x:

* force travis-ci to use nodejs 12.x by explicitly running 'nvm' commands

This seems to be the officially recommended way to ensure minimal version of nodejs for project's with mixed languages. Previously the travis build where using nodejs 8.x, which are quite old now. 

* fix linter ARIA related errors, which I get when running nodejs 12.4.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,15 @@
 language: python
+
 python:
   - "3.7"
   - "3.8"
-node_js:
-  - "10"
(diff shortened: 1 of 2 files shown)


marcus-oscarsson:
Thanks alot !

PR #865, opened 2020-10-23
send video sizes as list of tuples

elmirjagudin:
Make sure that we send video sizes as a list of tuples,
as this is what is expected by UI.

Otherwise, for cases when Video Device is not used, the
UI drop-down for 'Video size' will be broken.


marcus-oscarsson:
Thanks @elmirjagudin

PR #867, opened 2020-12-01
send 'hidden' mesh grid state to backend

elmirjagudin:
When a mesh grid is hidden/unhidden in the UI, send the new state to the backend server.

This way, back-end will save grids state. This solves the issue of grids getting 'unhidden' when sample is moved,
or the UI page is reloaded.


marcus-oscarsson:
Thanks !

PR #870, opened 2020-12-18
Minor fixes regarding packages used

elmirjagudin:
* include redis package, so that we get 'redis-server' out-of-box when creating conda environment
* use node version 10.* both in conda environment  and when running travis CI


marcus-oscarsson:
Thanks !

PR #871, opened 2021-02-08
Minor fixes

marcus-oscarsson:
Hi,

Just a few fixes related to the work done the last few months, mainly:

-   Fix to the mesh PNG result display
-   Adapted to new EMBL Flex sample changer software

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -98,18 +98,22 @@ def exception_handler(e):
 
 
 def kill_processes():
-    pid_list = []
+    # Killing the processes causes pytest to fail because
+    # of non zero exit code, so we dont kill the processes
+    # when running the tests
+    if not server.testing:
+        pid_list = []
 
-    with open("/tmp/mxcube.pid", "r") as f:
-        pid_list = f.read().strip()
-        pid_list = pid_list.split(" ")
-        pid_list.reverse()
+        with open("/tmp/mxcube.pid", "r") as f:
+            pid_list = f.read().strip()
+            pid_list = pid_list.split(" ")
+            pid_list.reverse()
 
-    with open("/tmp/mxcube.pid", "w") as f:
-        f.write("")
+        with open("/tmp/mxcube.pid", "w") as f:
+            f.write("")
 
-    for pid in pid_list:
-        os.kill(int(pid), signal.SIGKILL)
+        for pid in pid_list:
+            os.kill(int(pid), signal.SIGKILL)
 
 
 t0 = time.time()
@@ -137,12 +141,6 @@ socketio.init_app(server)
 if not server.debug or os.environ.get("WERKZEUG_RUN_MAIN") == "true":
     logging.getLogger("MX3.HWR").info("Starting MXCuBE3...")
     atexit.register(kill_processes)
-
-    # Killing the processes causes pytest to fail because
-    # of non zero exit code, so we dont register kill_process
-    # when running the tests
-    if "COV_CORE_SOURCE" not in os.environ:
-        atexit.register(kill_processes)
 
     with open("/tmp/mxcube.pid", "w") as f:
         f.write(str(os.getpid()) + " ")
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -102,8 +102,6 @@ def kill_processes():
     # of non zero exit code, so we dont kill the processes
     # when running the tests
     if not server.testing:
-        pid_list = []
-
         with open("/tmp/mxcube.pid", "r") as f:
             pid_list = f.read().strip()
             pid_list = pid_list.split(" ")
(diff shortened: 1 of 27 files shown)


mxcube3/core/limsutils.py
```diff
@@ -335,11 +335,11 @@ def select_proposal(proposal):
         blcontrol.beamline.session.proposal_number = proposal_info.get("Proposal").get(
             "number", ""
         )
-        blcontrol.beamline.session.session_id = proposal_info.get("Session").get(
+        blcontrol.beamline.session.session_id = proposal_info.get("Session")[0].get(
```
elmjag, line 340: With this change logging with mocked beamline does not work. For mocked beamline the 'Session' is still a dictionary here.

I'm guessing the mocks needs to be updated somehow.

marcus-oscarsson: It should be fixed now :), there is a small update done to the mockup

mxcube3/__init__.py
```diff
@@ -98,18 +98,22 @@ def exception_handler(e):
 
 
 def kill_processes():
-    pid_list = []
+    # Killing the processes causes pytest to fail because
+    # of non zero exit code, so we dont kill the processes
+    # when running the tests
+    if not server.testing:
+        pid_list = []
```
elmjag, line 105: No need to declare pid_list here, the code below takes care of it already.

mxcube3/core/limsutils.py
```diff
@@ -448,6 +450,11 @@ def synch_with_lims():
         sample_info["defaultPrefix"] = get_default_prefix(sample_info, False)
         sample_info["defaultSubDir"] = get_default_subdir(sample_info)
 
+        regexp = re.compile("^[a-zA-Z0-9:+_-]+$")
+
+        if not regexp.match(sample_info["sampleName"]):
+            raise AttributeError("sample name contains an incorrect character")
```
elmjag, line 463: Suggestion, define a global variable out-side of this function, e.g:

VALID_SAMPLE_NAME_REGEXP = re.compile("^[a-zA-Z0-9:+_-]+$")

and then use it on line 455 as :

if not VALID_SAMPLE_NAME_REGEXP.match(sample_info["sampleName"]):

This way you compile the regexp only once, when the module is loaded. Otherwise  there is no point using compiled regexp here, it is compiled each time we enter the function.

mxcube3/routes/samplecentring.py
```diff
@@ -368,10 +368,19 @@ def centre_3_click():
...
-    data = sviewutils.start_manual_centring()
 
-    resp = jsonify(data)
-    resp.status_code = 200
+    try:
+        data = sviewutils.start_manual_centring()
+    except Exception as ex:
+        resp = (
+            "Could not move motor %s" % str(ex),
+            406,
```
elmjag, line 377: I think status code here should be 409.

The "409 Conflict" response.

mxcube3/routes/samplecentring.py
```diff
@@ -368,10 +368,19 @@ def centre_3_click():
...
         :statuscode: 409: error
     """
-    data = sviewutils.start_manual_centring()
 
-    resp = jsonify(data)
-    resp.status_code = 200
+    try:
+        data = sviewutils.start_manual_centring()
+    except Exception as ex:
+        resp = (
```
elmjag, line 375: I suggest 'early return' on error. I.e. the whole code will be:

    try:
        data = sviewutils.start_manual_centring()
    except Exception as ex:
        return (
            "Could not move motor %s" % str(ex),
            406,
            {"Content-Type": "application/json", "msg": str(ex)},
        )

    resp = jsonify(data)
    resp.status_code = 200

    return resp

IMHO, this style of returning on errors is easier to read.

mxcube3/__init__.py
marcus-oscarsson: Yes, indeed but I actually prefer to declare variables like this, I know that it's maybe not very pytonic but it least indicates what the intended type is. It's true that one waste one assignment. I guess the name already indicates that it should be a list so I will remove it this time :)

mxcube3/routes/samplecentring.py
marcus-oscarsson: Yes, thanks :)

mxcube3/routes/samplecentring.py
marcus-oscarsson: We can continue this discussion on our next meeting, but I don't really agree that the "early return" makes the code easier to read. I'm more a fan of only one return per function, except for in certain cases when it's clearly more readable with "early returns" I don't think this is such a case. I'm in anyways happy to discuss it and I happy to use whatever style we agree on :) 

What do you think about that ?

marcus-oscarsson:
Thanks, I had a look and corrected the issues you commented on

meguiraun (approved):

PR #872, opened 2021-02-12
fix js-lint errors

elmjag:
fixes for following errors

mxcube3/ui/components/SampleView/DrawGridPlugin.js:

  434:11  error  Unexpected if as the only statement in an else block  no-lonely-if
  441:37  error  Missing semicolon                                     semi


marcus-oscarsson:
Oh thanks, I thought I already did that but I must have missed that :)

PR #876, opened 2021-06-08
mxcubcore adaptation

marcus-oscarsson:
- Improved existing App object and renamed it to MXCUBEApplication
- Introduced MXCUBECore that replace the old blcontrol module
- Cleanup initialization and start up
- Attempt to cleanup route handling


mxcube3-server
```diff
@@ -25,6 +25,6 @@ if __name__ == "__main__":
         print("No Redis server is running, exiting")
         sys.exit(1)
 
-    from mxcube3 import server, socketio
+    from mxcube3 import main
 
-    socketio.run(server, host="0.0.0.0", port=8081)
+    main()
```
marcus-oscarsson, line 30: Simply moved the server start to a main function cleaning up the initialization

mxcube3/__init__.py
```diff
@@ -23,134 +23,82 @@
...
-sys.path.insert(0, fname)
-
-from mxcubecore import HardwareRepository as hwr
-
-hwr.add_hardware_objects_dirs([os.path.join(fname, "HardwareObjects")])
-
-import app as mxcube
-from config import Config
+from mxcube3.config import Config
+from mxcube3.app import MXCUBEApplication
```
marcus-oscarsson, line 27: Introduced a MXCUBEApplication object that encapsulates (or represents) "application" wide data and functions

mxcube3/__init__.py
```diff
@@ -23,134 +23,82 @@
...
+from mxcube3.server import Server
```
marcus-oscarsson, line 28: Introduced a Server object that encapsulates flask, flask_socketio and a few other things that are server specific

mxcube3/routes/beamline.py
```diff
@@ -1,149 +1,147 @@
...
 import logging
 
-from flask import Response, jsonify, request, make_response
+from flask import Blueprint, Response, jsonify, request, make_response
 
-from mxcube3 import server
-from mxcube3 import blcontrol
 from mxcube3.core import beamlineutils
 
+def init_routes(mxcube):
```
marcus-oscarsson, line 9: attempt to isolate the routes and make them less dependent on a global import, please let me know what you think about this ?

mxcube3/app.py
```diff
@@ -2,288 +2,400 @@
...
+    beamline = None
+    # XMLRPCServer
+    actions = None
+    # Plotting
+    plotting = None
 
-# Initial file list for user, initialized at login, for creating automatic
-# run numbers
-INITIAL_FILE_LIST = []
+    @staticmethod
```
meguiraun, line 39: why static method? (as opposed to adding the initialization in the __init__)

mxcube3/app.py
```diff
@@ -2,288 +2,400 @@
...
+
+        with open("stored-mxcube-session.json", "w") as fp:
+            json.dump(data, fp)
+
+    @staticmethod
+    def load_settings():
+        """
+        Loads application wide variables from "stored-mxcube-session.json"
+        """
+        with open("stored-mxcube-session.json", "r") as f:
```
meguiraun, line 381: should not this be better in redis?

meguiraun:
I cannot mount a sample:
```
  File "/Users/mikegu/MXCUBE/mxcubecore_upstream/mxcubecore/HardwareObjects/abstract/sample_changer/Container.py", line 221, in get_sample_address
    return str(basket_number) + ":" + "%02d" % (sample_number)
TypeError: %d format: a number is required, not str
```
casting `sample_number` to int does the job

mxcube3/__init__.py
meguiraun: its a neat what of separating domains. I like it

mxcube3/routes/beamline.py
meguiraun: are you suggesting to do the same for all? I mean completing following method?
``` 
    def register_routes(mxcube):
        from mxcube3.routes.beamline import init_routes as init_beamline_routes
        Server.FLASK.register_blueprint(init_beamline_routes(mxcube))
```

beteva:
> ```
>   File "/Users/mikegu/MXCUBE/mxcubecore_upstream/mxcubecore/HardwareObjects/abstract/sample_changer/Container.py", line 221, in get_sample_address
>     return str(basket_number) + ":" + "%02d" % (sample_number)
> ```
This has been the code from the very beginning, but indeed we should cast the sample_number to int.

mxcube3/routes/beamline.py
marcus-oscarsson: Yes, I just started with one route as we might need to discuss before we arrive to something final

mxcube3/app.py
marcus-oscarsson: I wanted singleton like behavior for these classes (But I'm not certain we actually need a singleton). So I created a static class for this an the variables that were previously global are this class variables and all methods static (as there are no instance)

marcus-oscarsson:
Thanks for having a look :)

The above should probably go on mxcubecore, and might also be a problem with the SampleChangerMockup because this code works with the or sample changers. We have to make sure that it works with both :)

mxcube3/routes/beamline.py
meguiraun: its a good start

meguiraun:
thanks for this huge work!

PR #877, opened 2021-06-18
Possible usage beamline adapter and pydantic

marcus-oscarsson:
- Provided access to BeamlineAdapter directly through MXCUBECore.beamline and access to the underlying hardware object (if needed) through MXCUBECore.beamline_ho

- Using Pydantic and Spectree to pass data and to validate the rest routes

The interesting changes are in reality in `core/models.py`, `core/beamlineutils.py` and `routes/beamline.py`. The idea would be to remove the core/beamlineutils.py and replace all logic in a similar way to whats done for get and set beamline_attribute. Basically all of the core/*utils modules could be handled in a similar way.

The idea would be to provide an adapter class for each hardware object that handles its signals and uses Pydantic models for the data that is exchanged between mxcubecore and the REST layer.


meguiraun:
thanks marcus

PR #881, opened 2021-09-17
WIP Adding GPhL workflow, with some refatoring,

rhfogh:
Goes with a similar PR for mxcubecore.

The GPhL part is still not mature for looking at, but I would greatly appreciate comments on the non-GPhL, refactoring parts.

Notably I have made a change so you can pass multiple paths for the xml configuration files path1:path2, as was always possible in mxcubecore. This significantly simplifies keeping a remote set of repositories that you might want to test with different beamline set-ups.


beteva (commented):
Wow, 152 files changed. Is this really the PR to be merged? It includes the @marcus-oscarsson changes as well.

rhfogh:
Thans for checking and OK'ing it. At least I now know I am not running in the wrong directions

Actually, I should probably have PRed it to the restruc branch - that is what it is built on.

It is going to be painful no matter how it is done, I suspect. Thre are a lot of changes needed just for integrating the GPhL workflow, and in the current state of the code there is still a certain amount of incompletely refactored stuff that presents itself for fixing while I am doing that. It is hard for me to see how I could split this into smaller chunks.

For now, should I just close this one and make a PR to the restruc branch instead?

rhfogh:
PR moved from master to restruct branch - where it should havde been all along.

PR #882, opened 2021-09-24
WIP Rhfogh restruct4 - Adding GPhL workflow, with some refactoring,

rhfogh:
This PR is moved from mxcube3/master, since it was originally forked from restruct.

Comment was:
It goes with a similar PR for mxcubecore.

The GPhL part are still not mature for looking at, but I would greatly appreciate comments on the non-GPhL, refactoring parts.

Notably I have made a change so you can pass multiple paths for the xml configuration files path1:path2, as was always possible in mxcubecore. This significantly simplifies keeping a remote set of repositories that you might want to test with different beamline set-ups.


mxcube3/__init__.py
```diff
@@ -2,26 +2,29 @@
 from __future__ import division
 from __future__ import print_function
 
-from gevent import monkey
+import sys
+import mock
+import os
 
+from gevent import monkey
+# NB HardwareRepository must be imported *before* the gevent monkeypatching
```
marcus-oscarsson, line 10: Ok

mxcube3/__init__.py
```diff
@@ -2,26 +2,29 @@
...
-import time
-import traceback
-import atexit
 
-import gevent
+# import signal
+# import logging
+# import time
+# import traceback
+# import atexit
```
marcus-oscarsson, line 21: thanks for the cleanup, we can remove these imports

mxcube3/__init__.py
```diff
@@ -88,18 +91,20 @@ def parse_args():
...
     cmdline_options, args = parse_args()
 
-    cfg = Config(os.path.abspath(os.path.join(
-        cmdline_options.hwr_directory, 
-        "mxcube-server-config.yml"
-    )))
+    # This refactoring (with other bits) allows you to pass a 'path1:path2' lookup path
+    # as the hwr_directory. I need it for sensible managing of a multi-beamline test set-up
+    # without continuously editing teh main config files.
+    # Note that the machinery was all there in the core alrady. rhfogh.
```
marcus-oscarsson, line 97: Okey, Nice :)

mxcube3/app.py
```diff
@@ -137,8 +136,13 @@ def init(app, hwdir):
 
     @staticmethod
     def _import_adapter_cls(adapter_cls_str):
+        # NBNB Cannot be imported at top level since that causes cirtular imports
+        # NBNB It is no good that a geeneric converter live in a file tat imports
```
marcus-oscarsson, line 140: good point, the logic that does the actual import should perhaps be moved to core.utils (or do you have another
idea). We can keep it as it is in this PR and move it a another PR.

mxcube3/core/qutils.py
```diff
@@ -343,6 +342,59 @@ def _handle_dc(sample_node, node, include_lims_data=False):
     return res
 
 
+def _handle_gphl_wf(sample_node, node, include_lims_data=False):
```
marcus-oscarsson, line 345: Well done !

mxcube3/ui/components/Tasks/GphlWorkflow.jsx
```diff
@@ -0,0 +1,158 @@
+import React from 'react';
```
marcus-oscarsson, line 1: Well done

mxcube3/core/qutils.py
```diff
@@ -104,10 +104,9 @@ def node_index(node):
         sample = node.loc_str
     # TaskGroup just return the sampleID
     elif node.get_parent():
-        if isinstance(node, qmo.TaskGroup):
-            sample_model = node.get_parent()
-        else:
-            sample_model = node.get_parent().get_parent()
+        # NB under GPhL workflow, nodes do not have predictable distance
```
marcus-oscarsson, line 107: Ok, yes the get_sample_node is a nicer way of doing this

marcus-oscarsson:
Very nice :) For the general part I can only see improvement. I'm also happy to see that you were able to get started with the JS parts !.

I'm happy to merge this if its in working condition or is it still WIP ?

mxcube3/app.py
rhfogh: It looks like I put htis comment in the wrong place and/or the code has changed over the lst two months.
Anyway, _import_adapter_cls does not seem to be used anywhere currently, so it can easily be left where it is until such a time as we can consider removing it )it does look like a bit of a hack?).

My beef was actually with putting str_to_snake (etc.) in utils as opposed to somewhere in mxcubecore. Which I still think should happen, but which can wait till 1) I need it from somewhere else, or 2) we feel like doing a cleaning day.

rhfogh:
Thanks for the kindly comments! If you think this is good enough, I would be very happy to have it merged - do you want me to remove the WIP? 
It is still 'in progress' to the extent that I am not by any means finished with getting the GPhL workflow integrated, either for interactive use (MXCuE3) or MASSIF1. But it would be great if we can get this in as a milestone, so as to make future PR's easier.

marcus-oscarsson:
That sounds good !

PR #894, opened 2021-11-10
Additional fix for restruc

rhfogh:
Bug fixes, propose to merge it into restruc before restruc is merged into master.

Also removed MXCUBECore.beamline_ho as it is no longer in use.
I suspect that MXCUBECore.actions and MXCUBECore.plotting may be unused as well, but since I am not sure I did not modify those.

11/11: Added two more fixes


marcus-oscarsson:
Thanks !

PR #899, opened 2022-01-14
Improved remote access feature

marcus-oscarsson:
Improved remote access feature:

 - Now using flask-security
 - Simplified logic not relying on socket-io callbacks
 - Added force logout user
 - Added disconnect on idle

### DIFF
--- a/mxcube3/core/components/user/usermanager.py
+++ b/mxcube3/core/components/user/usermanager.py
@@ -1,10 +1,13 @@
 import logging
 import json
 import uuid
+import datetime
+import time
 
 import flask
 import flask_security
 import flask_login
+import flask_socketio
 
 from mxcube3.core.components.component_base import ComponentBase
 from mxcube3.core.components.user.models import User
@@ -22,7 +25,7 @@ class BaseUserManager(ComponentBase):
         return [
             user
             for user in User.query.all()
-            if (not user.in_control and user.is_authenticated)
+            if ((not user.in_control) and user.is_authenticated and user.is_active)
         ]
 
     def get_operator(self):
@@ -52,6 +55,15 @@ class BaseUserManager(ComponentBase):
 
         return users
 
+    def get_user(self, username):
+        user = None
+
+        for _u in User.query.all():
+            if _u.username == username:
+                user = _u
+
+        return user
+
     def set_operator(self, username):
         user = None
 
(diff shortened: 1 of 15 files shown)


rhfogh (approved):
I am not really capable of evaluating this. My current set-up is not in a state where I can test, and this is java-and-web, where I am not at all familiar. But I am crrtainly not going to hold it up, so I approve - without taking responsibility..

mxcube3/ui/components/RemoteAccess/UserList.jsx
```diff
@@ -14,14 +14,24 @@ class UserList extends React.Component {
...
           </div>
           { this.props.login.user.inControl ?
-            (<div className="col-xs-4">
-              <Button className="btn-sm" onClick={() => this.props.sendGiveControl(observer.username)}>
+            (<div className="col-xs-5">
+               <Button className="btn-sm" onClick={() => this.props.sendGiveControl(observer.username)}>
                  Give control
                </Button>
+               { this.props.login.user.isstaff ?
+                (<span>
```
meguiraun, line 26:
```suggestion
                 (<span>
```

mxcube3/ui/components/RemoteAccess/UserList.jsx
```diff
@@ -14,14 +14,24 @@ class UserList extends React.Component {
...
+                  &nbsp;
+                  <Button className="btn-sm" onClick={() => this.props.sendLogoutUser(observer.username)}>
+                    Logout
+                  </Button>
+                 </span>)
+                :
```
meguiraun, line 32:
```suggestion
                 :
```

mxcube3/ui/components/RemoteAccess/UserList.jsx
```diff
@@ -14,14 +14,24 @@ class UserList extends React.Component {
...
+                null
```
meguiraun, line 33:
```suggestion
                 null
```

mxcube3/core/components/user/usermanager.py
```diff
@@ -22,7 +25,7 @@ def get_observers(self):
         return [
             user
             for user in User.query.all()
-            if (not user.in_control and user.is_authenticated)
+            if ((not user.in_control) and user.is_authenticated and user.is_active)
```
meguiraun, line 28: not entirely sure about this:

```suggestion
            if ((not user.in_control) and user.is_authenticated and user.active)
```

meguiraun:
I see this in ui state, is this username intended? Is it mixed with userID?
image

mxcube3/routes/ra.py
```diff
@@ -89,27 +82,47 @@ def give_control():
...
+        return make_response("", 200)
+
+    @bp.route("/logout_user", methods=["POST"])
+    @server.restrict
+    def logout_user():
+        """
+        """
+        username = request.get_json().get("username")
+        user = app.usermanager.get_user(username)
+        if not user.in_control:
```
meguiraun, line 104: I dont understand this `if`. If I am the one in control and I log out, I would like to kick out all the observers. But if I am _not_ in control, I would like to leave the others logged in.

This is what we do here (more or less), but of course we have users login instead of proposal.

Can the same proposal login be used at the same time?

mxcube3/ui/components/RemoteAccess/UserList.jsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { bindActionCreators } from 'redux';
 import { connect } from 'react-redux';
 import { Button, Panel } from 'react-bootstrap';
-import { sendGiveControl } from '../../actions/remoteAccess';
+import { sendGiveControl, sendLogoutUser } from '../../actions/remoteAccess';
 
 class UserList extends React.Component {
   getObservers() {
```
meguiraun, line 8: would be good to include the one in control too

meguiraun:
I get this too:

```
ERROR in
/Users/mikegu/MXCUBE/mxcube3_upstream/mxcube3/ui/actions/remoteAccess.js
  23:16  error  'getRaState' was used before it was defined  no-use-before-define
```

meguiraun:
The user type login seems broken now, SelectProposal component does not get all the required info or the like. We take care of that ourselves, unless you see a 15 min fix...

I only have small comments, I probably will have more once we fix&try our own user login. We added here a few additions:
- Display all users in the RA page
- If the user is staff, it can force kickout any non staff user
- If the user is staff, it can always login (no matter if it does not belong to the proposal)

If we agree on the comments I did we (MAX IV) should do another PR for the user type login after.

mxcube3/core/components/user/usermanager.py
marcus-oscarsson: Well you could use both, active is the attribute and is_active is the property pointing to the same. I prefer to use the attribute in this case.

mxcube3/routes/ra.py
marcus-oscarsson: - username has replaced userID in the question above

-  The method above is used to force logout a user, so the if statement is just a security check to make sure that if against all odds something goes wrong no-one can force logout a user that is already in control

mxcube3/ui/components/RemoteAccess/UserList.jsx
marcus-oscarsson: I don't think I understood this one ?, do you want to force logout the user in_control ?

marcus-oscarsson:
@meguiraun Thanks for the feedback we can discuss exactly how to integrate your local changes on our next meeting. It should be on Tuesday but we can meet before if you prefer. Then we can go through the changes and I can explain how things works in more detail. One of the ideas with this was,as you know, to make it easier to make local adaptations to the login procedure.

I fixed the comments above, this is the preliminary work to get things working as they were before with some minor additions. So can we merge this as it is now, provided that we will follow it up with other PRs ?

mxcube3/ui/components/RemoteAccess/UserList.jsx
meguiraun: aah no, just list the one in control in the User list

meguiraun:
Thanks for this. Yes, lets discuss the follow up actions on our next meeting

marcus-oscarsson:
:), thanks

PR #898, opened 2021-12-02
Reinstated lost fixes - Needed for GPhL workflow to run. See issue #895

rhfogh:
It looks like these fixes got lost in the merging of PRs. One is needed for Issue #895

Issue #895, opened 2021-11-30
Problems with signals / thread communication?


rhfogh:
Some problems with core/MXCuBE3 communication that I do not know what to do about - all arise when running MXCuBE3 in mock mode using the CollectEmulator image simulation.

-1- There is an intermittent problem, where the Acquisition data entries in the queue are shown twice in the on-screen tree display. This can happen both for characterisation (five scans, shown as ten), or for acquisition proper (two scans, shown as four). It is my impression that the duplication only happens in the UI, not in the actual objects to execute.

-2- The collectEmulator spawns off a process to generate images, and waits for the process to finish:

>             running_process = subprocess.Popen(
>                 command_list, stdout=fp1, stderr=fp2, env=envs
>             )
> [...]
>             return_code = running_process.wait()

While this thread is waiting, control is passed to somewhere else, causing the queue route
    
> @bp.route("/<sid>/<tindex>/execute", methods=["PUT"])

to be executed. This calls execute_entry_with_id, and puts the currently executing sample on the queue as the next queue_entry to be executed. Which obviously leads to downstream errors. I was unable to figure out where this command came from, and patched the problem by refusing to put a queue_entry on the queue if said object is in the queue already - see PR pointing to this issue. A more permanent fix would avoid this action rather than just blocking it.

Is there anyone who has a better understanding of the signal passing to the UI, who can figure out what might be going on?

Issue #660, opened 2017-10-05
Task dialogs, show more sometimes confussing


marcus-oscarsson:
Show more in task dialog are sometimes confusing, difficult to see what is about to be shown or hidden. Characterization is good example if this, perhaps possible to solve in another way.

PR #900, opened 2022-01-14
Using asynchronous version of execute entry see PR mxcubecore#660  

marcus-oscarsson:
Fix for issue #895 and replaces #898, 

@rhfogh please let me know if it does the job or if you think that both fixes are necessary


marcus-oscarsson:
I use the underscore because async is a reserved word and that using async (as it is a statement) will cause a syntax error
even if its used as a argument

rhfogh (approved):
Looks good, and thanks for fixing it. It must have been tricky to find. See comments to #898 and mxcubecore #660 for related feedback.

meguiraun (approved):
looks good

PR #904, opened 2022-01-21
Improved route registration

marcus-oscarsson:
Added automatics tagging of the view functions that are not already tagged so that the routes gets categorized automatically.

### DIFF
--- a/mxcube3/routes/log.py
+++ b/mxcube3/routes/log.py
@@ -8,7 +8,6 @@ def init_route(app, server, url_prefix):
 
     @bp.route("/", methods=["GET"])
     @server.restrict
-    @server.validate(tags=["log"])
     def log():
         """
         Retrive log messages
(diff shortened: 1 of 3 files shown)


mxcube3/routes/log.py
```diff
@@ -8,7 +8,6 @@ def init_route(app, server, url_prefix):
 
     @bp.route("/", methods=["GET"])
     @server.restrict
-    @server.validate(tags=["log"])
```
meghdadyazdi, line 11: Should the `tags=["beamline"]` also be removed from [here](https://github.com/mxcube/mxcube3/blob/8b37bc56c17abd2dca30a777baf268e60aaca834/mxcube3/routes/beamline.py#L31)?

marcus-oscarsson: Yes, indeed I missed that one :)

meghdadyazdi (approved):

PR #906, opened 2022-02-09
Fixed staff configuration

marcus-oscarsson:
This fixes the issue that the staff role was not taken from the mxcubecore session.xml configuration

- Added the possibility to use the users defined as inhouse in session.xml as staff

### DIFF
--- a/mxcube3/core/components/user/usermanager.py
+++ b/mxcube3/core/components/user/usermanager.py
@@ -275,6 +275,16 @@ class BaseUserManager(ComponentBase):
         self.app.server.user_datastore.put(user)
         self.app.server.user_datastore.commit()
 
+    def _get_configured_roles(self, user):
+        roles = []
+
+        _ihs = ["%s%s" % prop for prop in HWR.beamline.session.in_house_users]
+
+        if self.config.inhouse_is_staff and user in _ihs:
+            roles.append("staff")
+
+        return roles
+
     def db_create_user(self, user, password, lims_data):
         sid = flask.session["sid"]
         user_datastore = self.app.server.user_datastore
@@ -297,10 +307,11 @@ class BaseUserManager(ComponentBase):
                 session_id=sid,
                 selected_proposal=user,
                 limsdata=json.dumps(lims_data),
-                roles=["staff"],
+                roles=self._get_configured_roles(user)
             )
         else:
             _u.limsdata = json.dumps(lims_data)
+            user_datastore.append_roles(_u, self._get_configured_roles(user))
 
         self.app.server.user_datastore.commit()
 
--- a/mxcube3/core/components/user/usermanager.py
+++ b/mxcube3/core/components/user/usermanager.py
@@ -276,12 +276,17 @@ class BaseUserManager(ComponentBase):
         self.app.server.user_datastore.commit()
 
     def _get_configured_roles(self, user):
-        roles = []
+        roles = set()
 
         _ihs = ["%s%s" % prop for prop in HWR.beamline.session.in_house_users]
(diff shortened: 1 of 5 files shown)


mxcube3/core/models/configmodels.py
```diff
@@ -35,13 +35,31 @@ class UIPropertiesListModel(BaseModel):
     __root__: Dict[str, UIPropertiesModel]
 
 
+class UserManagerUserConfigModel(BaseModel):
+    username: str = Field("", description="username")
+    role: str = Field("staff", description="Role to give user")
+
+
+class UserManagerConfigModel(BaseModel):
```
marcus-oscarsson, line 43: Configuration model for the UserManager

mxcube3/core/models/configmodels.py
```diff
@@ -35,13 +35,31 @@ class UIPropertiesListModel(BaseModel):
     __root__: Dict[str, UIPropertiesModel]
 
 
+class UserManagerUserConfigModel(BaseModel):
```
marcus-oscarsson, line 38: Configuration model for user, mostly added for future use. Like this one can assign any role to a user through configuration (For site specific adaptation). This part is still not used by the UserManager

mxcube3/core/models/configmodels.py
```diff
@@ -35,13 +35,31 @@ class UIPropertiesListModel(BaseModel):
...
+    username: str = Field("", description="username")
+    role: str = Field("staff", description="Role to give user")
+
+
+class UserManagerConfigModel(BaseModel):
+    class_name: str = Field(
+        "UserManager", description="UserManager class", alias="class"
+        )
+    inhouse_is_staff: bool = Field(
```
marcus-oscarsson, line 47: We have always considered inhouse users as staff, but it might not be the case for all sites. Default is true but one can exclude these users by setting it to False

test/HardwareObjectsMockup.xml/mxcube-web/server.yaml
```diff
@@ -14,3 +14,8 @@ mxcube:
 
   usermanager:
     class: UserManager
+    inhouse_is_staff: True
```
marcus-oscarsson, line 17: Example config, inhouse users defined in session.xml is considered as staff

test/HardwareObjectsMockup.xml/mxcube-web/server.yaml
```diff
@@ -14,3 +14,8 @@ mxcube:
...
+    users:
+      -
+        username: opid0
+        role: staff
```
marcus-oscarsson, line 21: Just an example of assigning a role to a user directly from the UserManager configuration (not used yet)

marcus-oscarsson, line 21: It's only possible to give a user one role at the moment, one could consider making this a list

mxcube3/core/components/user/usermanager.py
```diff
@@ -297,10 +312,11 @@ def db_create_user(self, user, password, lims_data):
                 session_id=sid,
                 selected_proposal=user,
                 limsdata=json.dumps(lims_data),
-                roles=["staff"],
+                roles=self._get_configured_roles(user)
```
meghdadyazdi, line 315: `set` does not support item assignment if I am not wrong. 
Looking for a solution ...

marcus-oscarsson: It was the idea we don't want to add roles on the fly, is it causing problems for you. We can always change it to a list if needed ?

PR #909, opened 2022-02-20
Rhfogh master

rhfogh:
ne bug fix in components.queue, otherwise essentially GPhL-specific.

NBNB behaviour of signals.handle_auto_mount_next should be checked. Is it really correct that this function is called both from queue_execution_entry_started and from queue_execution_entry_finished?


marcus-oscarsson:
Thanks !,

rhfogh:
Selv tak!

PR #911, opened 2022-02-23
Observers are not inControle so they cannot select proposal

meghdadyazdi:


mxcube3/core/components/user/usermanager.py
```diff
@@ -233,6 +233,7 @@ def login_info(self):
                 "isstaff": "",
                 "nickname": "",
                 "inControl": "",
+                "islocal": "",
```
meguiraun: can we camelCasify this?

mxcube3/ui/actions/login.js
```diff
@@ -114,7 +114,13 @@ export function signIn(proposal, password) {
         dispatch(showErrorPanel(false));
         dispatch(getLoginInfo()).then(response => response).then((resp) => {
           if (resp.loginType === 'User') {
-            dispatch(showProposalsForm());
+            if (resp.user.inControl) {
+              dispatch(showProposalsForm());
+            }
+            else {
+              dispatch(selectProposal(resp.selectedProposal));
```
meguiraun: is this really needed? I would assume that in this case the proposal has already been selected by whoever is in control and loged in first.

Also, what happens if you refresh the page after you have selected a proposal? does the selection modal appears again? I am unsure if you would need to check if a proposal has been selected instead of checking if user is in control

mxcube3/ui/components/RemoteAccess/ObserverDialog.jsx
```diff
@@ -48,6 +48,42 @@ export class ObserverDialog extends React.Component {
     return 'Observer mode';
   }
 
+  observerName() {
```
meguiraun: is looks as repeated code, can it be merged? You could use the same code and just disable the edtion of the name in case of user login

meguiraun (changes requested):

mxcube3/ui/actions/login.js
meghdadyazdi: If the user is not inControl then the user cannot select the proposal. The selected proposal is used for the user not inControl.

mxcube3/ui/components/RemoteAccess/ObserverDialog.jsx
```diff
@@ -61,19 +62,24 @@ export class ObserverDialog extends React.Component {
...
-        <Modal.Footer>
-          <FormControl
-            inputRef={(ref) => { this.name = ref; }}
-            type="text"
-            default={this.props.login.selectedProposal}
-          />
-          <Button onClick={this.accept}> OK </Button>
-        </Modal.Footer>
+        <div>
+                      <Modal.Body>
```
marcus-oscarsson: Is this indentation right, looks a bit off ?

marcus-oscarsson:
Looks good in general. I guess this is not entirely done yet I'm thinking about the `local` role looks like it's never really added to the user ?

PR #914, opened 2022-03-17
React Router Dom V6 

jbflo:
Both Local storage and server Storage are use.
Next we should  fix the problem preventing us using only the server Storage 
mostly for persist gate.


marcus-oscarsson:
Thanks JB :)

PR #916, opened 2022-04-26
Minor fixes

marcus-oscarsson:


marcus-oscarsson:
Fixes made after testing with mxcubecore and frontend library updates

 - Fixed attribute access, typos and name changes
 - Adapted to updated libraries
 - Pinned versions of third party libraries (most notably those used by flask)

mxcube3/core/components/sampleview.py
```diff
@@ -39,6 +39,27 @@ def __init__(self, app, config):
...
+    def _zoom_changed(self, *args, **kwargs):
+        ppm = HWR.beamline.diffractometer.get_pixels_per_mm()
+        self.app.server.emit("update_pixels_per_mm", {
+                             "pixelsPerMm": ppm}, namespace="/hwr")
+
+    def _emit_shapes_updated(self):
+        shape_dict = {}
+
+        for shape in HWR.beamline.sample_view.get_shapes():
+            _s = to_camel(shape.as_dict())
```
rhfogh, line 58: Nothing wrong with this, but why not just do:
    shape_dict[shape.id] = to_camel(shape.as_dict())
?

mxcube3/core/components/workflow.py
```diff
@@ -25,7 +25,7 @@ def get_available_workflows(self):
         except Exception:
             pass
 
-        if hasattr(beamline, "gphl_workflow"):
+        if getattr(beamline, "gphl_workflow", None):
```
rhfogh, line 28: Nothing wrong with this, but it should not be necessary. What error happens if you use 'hasattr'?

rhfogh (approved):
Only a brief look-through, but FWIIW looks fine.

mxcube3/core/components/workflow.py
marcus-oscarsson: There is a subtle difference, `hasattr` will return `True` if the attribute exists but it can still have the value `None`
Meaning that the expression` beamline.gphl_workflow.get_available_workflows() `will fail. On the other hand
`getattr` with default None will return `False` and the statement never executed.

mxcube3/core/components/sampleview.py
marcus-oscarsson: Can't tell really, It was just done like that I guess. The two statements are the same

PR #920, opened 2022-04-27
Update to bootstrap 5

jbflo:
In this PR we update bootstrap from version 3 to 5
we Update react-bootstrap from v0.33.1 to v2.3.0


marcus-oscarsson:
Very nice thanks alot :)

PR #925, opened 2022-05-05
Fix Error due to Bootstrap 5 transformation

jbflo:
Modal display 
duplicate Property
Add sample dialog from alignment

### DIFF
--- a/ui/src/components/SampleGrid/TaskItem.jsx
+++ b/ui/src/components/SampleGrid/TaskItem.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { OverlayTrigger, Popover, Badge } from 'react-bootstrap';
+import { Col, OverlayTrigger, Popover, Badge } from 'react-bootstrap';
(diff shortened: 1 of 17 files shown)


ui/src/components/SampleGrid/TaskItem.jsx
```diff
@@ -201,25 +201,25 @@ export class TaskItem extends React.Component {
...
 
-    return varaint;
+    return variant;
   }
 
   popoverPosition() {
     const viewportHeight = Math.max(document.documentElement.clientHeight, window.innerHeight || 0);
     let result = 'bottom';
-
+    debugger;
```
marcus-oscarsson: Watch out for this debugger statement :)

jbflo: Thanks  a lot Marcus, 
I have to add more more commit to this PR. 

[fix modal display] commit have some side effect on the Characterisation Modal. sorry 
Did not notice.

PR #926, opened 2022-05-31
Boostrap 5 fixes 

jbflo:
adjusted layout by adding some new css /bootstrap style


marcus-oscarsson:
I think we need to clean-up this PR because it contains quite a lot of commits that does not rely belong here. I will close it for now to avoid confusion.

PR #927, opened 2022-06-09
Rhfogh develop

rhfogh:
Bug fix to _handle_gphl_wf function, and some gphl,P14 specific config changes


mxcube3/core/components/queue.py
```diff
@@ -1667,9 +1668,15 @@ def queue_model_child_added(self, parent, child):
...
 
-            elif isinstance(child, qmo.XrayCentering):
+            elif isinstance(child, qmo.XrayCentring2):
                 # Added rhfogh 20211001
-                entry = qe.XrayCenteringQueueEntry(Mock(), child)
+                entry = qe.XrayCentring2QueueEntry(Mock(), child)
+                self.enable_entry(entry, True)
+                parent_entry.enqueue(entry)
+
+            elif isinstance(child, qmo.DummyTask):
```
marcus-oscarsson: Ahhm, be careful here remember we renamed it to DelayTask yesterday

mxcube3/core/components/queue.py
```diff
@@ -1667,9 +1668,15 @@ def queue_model_child_added(self, parent, child):
...
+                # Added rhfogh 20220331
+                entry = qe.DummyQueueEntry(Mock(), child)
```
marcus-oscarsson: ... and **DelayQueueEntry** ....

marcus-oscarsson:
Thanks, looks good except for the comment I made with the name change

PR #928, opened 2022-06-13
Added detach-task option to config file

rhfogh:
Just a tiny, GPhL-specific addition

### DIFF
--- a/conda-environment.yml
+++ b/conda-environment.yml
@@ -48,5 +48,5 @@ dependencies:
       - spectree
       - git+https://github.com/mxcube/mxcubecore.git
       - flask-security-too
-      - f90nml
+      - f90nml==1.3.1
       - gipc
--- a/ui/src/components/SampleView/shapes.js
+++ b/ui/src/components/SampleView/shapes.js
@@ -221,7 +221,7 @@ export function makePoint(x, y, id, color, type, name, strokeWidth) {
 export function makePoints(points, imageRatio) {
   const fabricPoints = [];
   for (const id in points) {
-    if (Object.hasOwn(points, id)) {
+    if ({}.hasOwnProperty.call(points, id)) {
       const [x, y] = points[id].screenCoord;
 
       switch (points[id].state) {
@@ -263,7 +263,7 @@ export function makePoints(points, imageRatio) {
 export function makeTwoDPoints(points, imageRatio) {
   const fabricPoints = [];
   for (const id in points) {
-    if (Object.hasOwn(points, id)) {
+    if ({}.hasOwnProperty.call(points, id)) {
       const [x, y] = points[id].screenCoord;
 
       switch (points[id].state) {
--- a/test/HardwareObjectsMockup.xml/gphl/gphl-workflow.yml
+++ b/test/HardwareObjectsMockup.xml/gphl/gphl-workflow.yml
@@ -85,6 +85,7 @@ settings:
       - exposure_time: 0.02
         image_width: 0.1
         resolution: 1.7
+#        init_spot_dir: /directory/containing/SPOT.XDS/file/after/characterisation
 #        image_count: 90
         # strategy_options: # Not currently used
         # For acquisition or diffractcal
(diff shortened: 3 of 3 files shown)


conda-environment.yml
```diff
@@ -48,5 +48,5 @@ dependencies:
       - spectree
       - git+https://github.com/mxcube/mxcubecore.git
       - flask-security-too
-      - f90nml
+      - f90nml==1.3.1
```
rhfogh, line 51: f90nml version 1.4.x are currently broken - An issue has been raised with f90nml

ui/src/components/SampleView/shapes.js
```diff
@@ -221,7 +221,7 @@ export function makePoint(x, y, id, color, type, name, strokeWidth) {
 export function makePoints(points, imageRatio) {
   const fabricPoints = [];
   for (const id in points) {
-    if (Object.hasOwn(points, id)) {
+    if ({}.hasOwnProperty.call(points, id)) {
```
rhfogh, line 224: SIgnalled as bug in Pycharm, and caused malfunction until reverted.
@marcus-oscarsson May be a denepndency/version issue, or may require a different fix (though this one works in mock mode).

marcus-oscarsson: Its not a bug, it has the same meaning (more or less) and that is also why it works :). But its always better to be consistent, we should in that case use [hasOwn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn)

marcus-oscarsson: Hm, reading this a second time. I'm not sure that I understand. You write that hasOwnProperty works but not hasOwn and that you are reverting it (but to what point in time, when was it changed to hasOwn ?)

rhfogh: Well, my editor (pycharm) signalled it as a bug (no such function), and changing it (unless there was another simultaneous change and I got confused) fixed a problem that gave me a white screen. Could it be a question of whi dependencies / version is installed?

rhfogh: According to Pycharm git history, the commit by JB* 30/3/2022 "Merge branch 'mo-create-react-app' into jb-bootstrap5" has
    if ({}.hasOwnProperty.call(points, id)) {

whereas the commit by you 02/05/2022 "lint" introduced 
    if (Object.hasOwn(points, id)) {

I shall make a quick test and see what happens if I go back to 'hasOwn':

Indeed. Moving back the hasOwn variant I get a permanent white browser page when I click Mount in the SampleGrid view. And when I go  into Inspect and scroll up a bit I get

Uncaught TypeError: Object.hasOwn is not a function
showing in red.

marcus-oscarsson: I think I found it, it was a automatic linting run that changed this. The change is seemingly correct but there seems to be a subtle difference between how hasOwn and hasOwnProperty that perhaps breaks the code. **I think you are right that using hasOwnProperty** is the safest option until we have figured out what goes wrong (hasOwn seems to work here when I test)

marcus-oscarsson: Well done :)

rhfogh: Pretty sneaky when linting changes the code.

marcus-oscarsson: > According to Pycharm git history, the commit by JB* 30/3/2022 "Merge branch 'mo-create-react-app' into jb-bootstrap5" has if ({}.hasOwnProperty.call(points, id)) {
> 
> whereas the commit by you 02/05/2022 "lint" introduced if (Object.hasOwn(points, id)) {
> 
> I shall make a quick test and see what happens if I go back to 'hasOwn':
> 
> Indeed. Moving back the hasOwn variant I get a permanent white browser page when I click Mount in the SampleGrid view. And when I go into INspect and scroll up a bbit I get
> 
> Uncaught TypeError: Object.hasOwn is not a function showing in red.

Ahh, I think its a **browser/transpilation** issue, I think your browser does not support `Object.hasOwn` and its not properly `transpiled`. I think it might work with a more recent for instance google chrome

marcus-oscarsson: So better to use **hasOwnProperty** then, again well spotted :)

marcus-oscarsson:
Ok, very good. Thanks :)

ui/src/components/SampleView/shapes.js
rhfogh: That would make sense.
I run FIrefox version 91.10.0 extended support release for OpenSUSE Leap 15.3, released on August 10, 2021, according to Wikipedia.

PR #935, opened 2022-09-15
selecting proposal when login, with new table style library

meghdadyazdi:
If login type is User, proposal table is shown only if the user is in control.
Bootsrtap library for table is now react-bootstrap-table-next

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -30,7 +30,7 @@
     "popper.js": "^1.16.1",
     "react": "^17.0.2",
     "react-bootstrap": "^2.3.0",
-    "react-bootstrap-table": "^4.3.1",
+    "react-bootstrap-table-next": "^4.0.3",
     "react-chat-widget": "^2.1.4",
     "react-clipboard.js": "^2.0.16",
     "react-collapsible": "^2.10.0",
--- a/ui/package.json
+++ b/ui/package.json
@@ -30,7 +30,6 @@
     "popper.js": "^1.16.1",
     "react": "^17.0.2",
     "react-bootstrap": "^2.3.0",
-    "react-bootstrap-table-next": "^4.0.3",
     "react-chat-widget": "^2.1.4",
     "react-clipboard.js": "^2.0.16",
     "react-collapsible": "^2.10.0",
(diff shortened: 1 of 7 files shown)


ui/package.json
```diff
@@ -30,7 +30,7 @@
     "popper.js": "^1.16.1",
     "react": "^17.0.2",
     "react-bootstrap": "^2.3.0",
-    "react-bootstrap-table": "^4.3.1",
+    "react-bootstrap-table-next": "^4.0.3",
```
marcus-oscarsson: I'm not sure react-bootstrap-table-next is compatible entirely with Bootstrap 5, it only support Bootstrap 4 at most ?

marcus-oscarsson:
Looks good, I'm a bit hesitant towards react-bootstrap-table-next it does not seem to support Bootstrap 5 and the last version 4.3.0 was published 2 years ago ?. Do you know what the status is ?

meghdadyazdi:
> Looks good, I'm a bit hesitant towards react-bootstrap-table-next it does not seem to support Bootstrap 5 and the last version 4.3.0 was published 2 years ago ?. Do you know what the status is ?

Seems they stoped updating it. I couldn’t find another alternative for table except the 'react-bootstrap/Table' that needs a little bit of coding and styling to get the same table in 'react-bootstrap-table-next'.

marcus-oscarsson:
So we should probably not use  `react-bootstrap-table-next ` then if its not maintained and does not support `Bootstrap 5` which is what we are using today. Do you think you can make use of `react-boostrap/table `or some other basic component ?

meghdadyazdi:
The final view:

proposal with title

marcus-oscarsson:
Very nice, some spacing between the `signout` and `select proposal` would probably look better :)

Did you remove the unused dependency for the package.json ?

meghdadyazdi:
> Very nice, some spacing between the `signout` and `select proposal` would probably look better :)
> 
> Did you remove the unused dependency for the package.json ?

I think spacing should be fixed globally (probably by Jean-Baptiste)?

I removed `react-bootstrap-table` which I think is the only dependency.

marcus-oscarsson:
Thanks, Not certain where the "spacing" issue comes from, If its a more general problem in the bootstrap related style or something more local. You are more than welcome to have look at it even if its a general issue :). Otherwise we will merge this as it is, and we will have to take a look at a later moment.

meghdadyazdi:
> Thanks, Not certain where the "spacing" issue comes from, If its a more general problem in the bootstrap related style or something more local. You are more than welcome to have look at it even if its a general issue :). Otherwise we will merge this as it is, and we will have to take a look at a later moment.

seems it can be solved locally

proposal with title

marcus-oscarsson:
Excellent :)

PR #938, opened 2022-10-21
Argparse util

spirit1317:
Hi. Working for MAX-IV, I saw that running a simple `mxcube3-server --help` command:
- results in exception and stack trace printed
- runs for ages (it needs to ping to Redis server)
- `optparse` used for cmd util is deprecated in favour of `argparse`

I thought it might be more user-friendly to improve that, so I suggest following changes.

Br


elmjag:
:+1:  :)

marcus-oscarsson:
Thanks, --help indeed gives a strange error message at the end, I guess we did not pay attention to that.

PR #941, opened 2022-11-02
Translate in YZ-plane parallel to microscope-view plane

meghdadyazdi:
If `Sample Horizontal` and `Sample Vertical` means movement in YZ plane which is perpendicular to X (beam direction), the attributes should be `phiz` and `phiy` and NOT `sampx` and `sampy`.


marcus-oscarsson:
Yes indeed, well done. Thanks.

PR #943, opened 2022-12-14
env var following REACT_APP_xxx

meguiraun:
change the naming of the env vars so they actually are used. See [here](https://create-react-app.dev/docs/adding-custom-environment-variables/#adding-development-environment-variables-in-env)

> You must create custom environment variables beginning with REACT_APP_. Any other variables except NODE_ENV will be ignored to avoid [accidentally exposing a private key on the machine that could have the same name](https://github.com/facebook/create-react-app/issues/865#issuecomment-252199527). Changing any environment variables will require you to restart the development server if it is running.

Removed also the ones not in use

### DIFF
--- a/ui/.env
+++ b/ui/.env
@@ -1,6 +1,4 @@
- phaseControl = true
- apertureControl = true
- focusControlOnCanvas = true
- use2dCenteredPoints = true
- meshResultFormat = PNG
- WDS_SOCKET_PORT = 0
+REACT_APP_PHASECONTROL = true
+REACT_APP_FOCUSCONTROLONCANVAS = true
+REACT_APP_USE2DCENTEREDPOINTS = true
+REACT_APP_MESHRESULTFORMAT = PNG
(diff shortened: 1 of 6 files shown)


ui/.env
```diff
@@ -1,6 +1,4 @@
- phaseControl = true
- apertureControl = true
- focusControlOnCanvas = true
- use2dCenteredPoints = true
- meshResultFormat = PNG
- WDS_SOCKET_PORT = 0
\ No newline at end of file
+REACT_APP_PHASECONTROL = true
```
marcus-oscarsson, line 1: Thanks :), we should probably move these to "runtime" configuration instead of a build time. We have already moved a few of these to [server.yaml and ui.yml ](https://github.com/mxcube/mxcubeweb/tree/develop/test/HardwareObjectsMockup.xml/mxcube-web)

But this is of course a welcome fix, until that is done :)

meguiraun: aah, thats true! I did not even think on that :(

marcus-oscarsson: No problem, one thing at the time :)

PR #944, opened 2022-12-15
Linting and pyproject.toml

marcus-oscarsson:
- Added pyproject.toml file
 - linted Python and Javacript code,
 - Some general cleanup


pyproject.toml
```diff
@@ -0,0 +1,73 @@
...
+    "Development Status :: 5 - Production/Stable",
+    "Intended Audience :: Developers",
+    "Intended Audience :: Science/Research",
+    "Intended Audience :: End Users/Desktop",
+    "Natural Language :: English",
+    "Topic :: Scientific/Engineering",
+]
+
+[tool.poetry.dependencies]
+python = ">=3.8,<3.11"
```
oldfielj-ansto, line 25: @marcus-oscarsson Are we dropping Python3.7 support?

marcus-oscarsson: Not really, we can keep 3.7 if its needed it is indeed supported by mxcuebcore. I just thought it was time to move a head for the web part and did not really see the reason for keeping 3.7 thats all

mxcube3/core/components/queue.py
```diff
@@ -2014,7 +2015,9 @@ def set_auto_mount_sample(self, automount, current_sample=None):
         """
         self.app.AUTO_MOUNT_SAMPLE = automount
 
-    def get_auto_mount_sample(self,):
+    def get_auto_mount_sample(
```
rhfogh: Are you sure of this? Is it not the comma that should be removed? Anyway, it looks weird

rhfogh (approved):
Looks like pure linting plus udating teh .toml file.

GOod. The sooner it is merged teh better

pyproject.toml
oldfielj-ansto: @marcus-oscarsson I'm happy to drop it a few months early, but we should probably update the Python version constraints for MXCuBE Core to be consistent.

oldfielj-ansto (approved):

oldfielj-ansto:
LGTM.

meguiraun (approved):

pyproject.toml
marcus-oscarsson: Yes, that's true, we should just wait for that until we have double checked with the Qt side of things but I don't think it's a problem.

mxcube3/core/components/queue.py
marcus-oscarsson: That's a funny one, I guess black should have removed the additional comma as there were no additional arguments.  Ill fix that manually :)

PR #947, opened 2022-12-21
save correct proposal for loginType 'user'

elmjag:
Extend logic for saving selected proposal in the user DB to work with login type 'user'

when loginType is 'user':

 * set selected_proposal to None when user is initially created in the DB
 * on 'set_proposal' HTTP requests update database with selected proposal
 * when 'operator' user is switched, change HW Session's proposal to user's 'selected_proposal'


marcus-oscarsson:
Thanks well done :).

marcus-oscarsson (approved):

PR #950, opened 2023-01-30
support sending sample view MJPEG stream to multiple UI clients

elmjag:
Previous implementation had an issue where the stream was only send to newly connected UI client. The stream to previously connected client was stalled.
    
Also, keep track of how many clients are streaming. Only connect to sample view camera object if number of clients is more then 0. This allow camera hardware objects to implement behaviour where frames are pulled from the diffractometer only when they are needed.


marcus-oscarsson:
Thanks, very nice :)

PR #951, opened 2023-01-30
fix for proxy connection refused

meguiraun:
hi!

Small change for avoid connection refused issues in some circumstances, [see](https://github.com/chimurai/http-proxy-middleware#nodejs-17-econnrefused-issue-with-ipv6-and-localhost-705)


marcus-oscarsson:
Great !

PR #952, opened 2023-02-03
missing deps

meguiraun:
@rhfogh f90nml seems needed for GP workflows, I just picked latest version

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,6 +41,9 @@ PyDispatcher = "^2.0.6"
 pytz = "^2022.6"
 tzlocal = "^4.2"
 mxcubecore = { git = "https://github.com/mxcube/mxcubecore.git", branch = "develop" }
+python-ldap = "3.3.1"
+py4j = "^0.10.9.7"
+f90nml = "^1.4.3"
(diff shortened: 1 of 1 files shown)


rhfogh (commented):
Do we need those in mxcubeweb? All three are in the mxcubecore pyproject.toml, near as I can see. As indeed are e.g. Python, gevent, and pydantic, so coudl we get rid of (some of) those? I'd say that it was better to have dependencies only in mxcubecore, even as a general principle (a unique source of truth), but certainly for py4j and f90nml that are only used in mxcubecore.

meguiraun:
mmm... then a fresh conda env + poetry did not install those

meguiraun:
I could not reproduce the issue anymore  🤨... I removed duplicated deps

meguiraun:
I also updated the wiki [section 5](https://github.com/mxcube/mxcubeweb/wiki#installing-using-conda)

marcus-oscarsson:
Well done, I also agree that we don't should not duplicate those dependencies. Great :1st_place_medal:  for updating the wiki !

PR #953, opened 2023-02-07
Fixes for login 

meghdadyazdi:
There some issue related to login as user (and some with proposal).

### DIFF
--- a/mxcube3/core/components/user/usermanager.py
+++ b/mxcube3/core/components/user/usermanager.py
@@ -41,18 +41,8 @@ class BaseUserManager(ComponentBase):
         return getattr(flask_login.current_user, "in_control", False)
 
     def logged_in_users(self, exclude_inhouse=False):
-        users = [user["loginID"] for user in self.app.USERS.values()]
 
-        if exclude_inhouse:
-            if isinstance(HWR.beamline.session.in_house_users[0], tuple):
-                ih_users = [
-                    "%s%s" % (p, c) for (p, c) in HWR.beamline.session.in_house_users
-                ]
-            else:
-                ih_users = HWR.beamline.session.in_house_users
-            users = [user for user in users if user not in ih_users]
-
-        return users
+        return [_u.username for _u in User.query.all()]
 
     def get_user(self, username):
         user = None
@@ -306,6 +296,8 @@ class BaseUserManager(ComponentBase):
         sid = flask.session["sid"]
         user_datastore = self.app.server.user_datastore
         username = f"{user}-{str(uuid.uuid4())}"
+        if HWR.beamline.lims.loginType.lower() == "user":
+            username = f"{user}"
 
         # Make sure that the roles staff and incontrol always
         # exists
@@ -374,6 +366,9 @@ class UserManager(BaseUserManager):
         }
 
         _users = self.logged_in_users(exclude_inhouse=True)
+
+        if login_id in _users:
+            raise Exception("Login rejected, you are already logged in")
 
         # Only allow in-house log-in from local host
         if inhouse and not (inhouse and is_local_host()):
--- a/mxcube3/core/components/user/usermanager.py
+++ b/mxcube3/core/components/user/usermanager.py
@@ -7,6 +7,7 @@ import datetime
 import flask
 import flask_security
 import flask_login
+from flask_login import current_user
 
 from mxcube3.core.components.component_base import ComponentBase
 from mxcube3.core.models.usermodels import User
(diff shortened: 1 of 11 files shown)


mxcube3/routes/login.py
```diff
@@ -46,6 +46,8 @@ def login():
             logging.getLogger("MX3.HWR").info(msg)
             res = deny_access(str(ex))
 
+        session.permanent = True
```
meguiraun, line 49: just a comment. Without this line, the cookie would be valid only during the "Session", and deleted on browser closing. This causes a bug where the local storage says logged in but the absence of the cookie says the opposite, it makes the loading break. In order to reproduce (without this line), just login as normal, close the browser, open it again and go to the main page again.

The [requireAuth()](https://github.com/mxcube/mxcubeweb/blob/develop/ui/src/index.js#L42) returns before the _getLoginInfo_ finishes. This still needs to be addressed, but as long as the cookie is valid it _should_ be ok 🤔

marcus-oscarsson: Aha, I wondering if we should not use this instead: https://flask-login.readthedocs.io/en/latest/#cookie-settings (REMEMBER-COOKIE)

meguiraun: we should... but I did not manage to have those working 🤪...

marcus-oscarsson: Alright :), its a good start. We can have a look together once you have the behavior you like :)

ui/src/actions/login.js
```diff
@@ -149,3 +149,20 @@ export function doSignOut(navigate) {
     });
   };
 }
+
+export function signOutUser(username, navigate) {
```
marcus-oscarsson: There is already a [signout](https://github.com/mxcube/mxcubeweb/blob/81381c34c11f802d129e993c094e3bd267d9b74f/ui/src/actions/remoteAccess.js#L102), did you try to this one or did it not do what you liked ?

mxcube3/routes/login.py
```diff
@@ -58,6 +60,18 @@ def signout():
 
         return redirect("/login", code=302)
 
+    @bp.route("/signout_user", methods=["POST"])
+    # @server.restrict
+    def signout_user():
```
marcus-oscarsson: This is already implemented [here](https://github.com/mxcube/mxcubeweb/blob/81381c34c11f802d129e993c094e3bd267d9b74f/mxcube3/routes/ra.py#L95)

I think its about the same ?

mxcube3/core/components/user/usermanager.py
```diff
@@ -218,8 +208,36 @@ def signout(self):
         flask_security.logout_user()
         self.emit_observers_changed()
 
+    def signout_user(self, user):
```
marcus-oscarsson: You could use the method below called force_signout_user ?

marcus-oscarsson:
Looks very nice in general, there are two questions that we have to look at. 

 - The use of permanent session (We can merge as it is, we just need to discuss the general behavior we want the)  
 -  Investigate if we can merge `signout_` and `force_signout_user`, this one would perhaps be nice to look at before we merge ?

ui/src/actions/login.js
meghdadyazdi: This [signout](https://github.com/mxcube/mxcubeweb/blob/81381c34c11f802d129e993c094e3bd267d9b74f/ui/src/actions/remoteAccess.js#L102) works when you are already logged in. In our case we are not yet logged in.

mxcube3/routes/login.py
meghdadyazdi: As I said we are going to force to sign out a user (myself) while we are not yet logged in (but authenticated). Then we have to define a new method which is not @server.restrict (?)

marcus-oscarsson: Hm, I guess we could lighten up the restriction. So far we have not needed to distinguish between logged in and authenticated, but there is perhaps a need. In that case `@server.restrict` could be made to simply check authenticated or as you mention we create a something like  `@server.restrict_authenticated`. I have to admit the the later sounds a bit odd to me. From the "restrict" point of view you are either "allowed/authenticated" or not. We could perhaps pass a parameter to `@server.restrict` like `@server.restrict(allow_authenticated = True )`. However not having any restrict could allow anyone to logout users.

Does that make sense ? 

But at at this stage why are you not logged in ?, you should be logged in but without a "selected_proposal" ?

ui/src/containers/SampleViewContainer.js
```diff
@@ -68,6 +69,17 @@ class SampleViewContainer extends Component {
       }
     });
 
+    const phaseControl = (
```
marcus-oscarsson, line 72: Ill merge this one for now, but there might be some discussion among the people here on where to display this. We might need to make it configurable :)

marcus-oscarsson:
:+1: thanks !

ui/src/containers/SampleViewContainer.js
meguiraun: Oh noo! We let them fight for the location ;)
tack!

PR #955, opened 2023-02-15
use the max run number

meguiraun:
if the initial file list is not in order, the run_number value for a given path prefix may not be the highest one

### DIFF
--- a/mxcube3/core/components/queue.py
+++ b/mxcube3/core/components/queue.py
@@ -59,8 +59,10 @@ class Queue(ComponentBase):
                     % path
                 )
                 path, run_number, image_number = (path, 0, 0)
-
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
Thanks.

Odd that we did not run into this. On the other hand we have actually moved away from this way of "generating" the run numbers and are making a FS check for every collection.  For the SSX collections there is even a subfolder per run making it even easier. We will make PR with a few of these remaining bits fairly soon.

meguiraun:
we use this for the run number but also check FS before the actual collection, its simply non user friendly to have several data collection failed until you get the available run number.

We encounter this issue recently, it rarely happens anyway

marcus-oscarsson:
Clearly yes that should definitely not happen

PR #960, opened 2023-02-27
early monkey patch

meguiraun:
@marcus-oscarsson @rhfogh 

As discussed earlier, I am proposing this change since otherwise at MAX IV we have a recursion import problem with ssl context in one of our of our hwobjs.

Simply reordering the imports fixes our issue, but I am wondering if it breaks someone else's hwobj.

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -1,15 +1,16 @@
+import argparse
+import mock
+import os
+import redis
+import sys
+import traceback
+
 from gevent import monkey
 monkey.patch_all(thread=False)
 
 from mxcube3.server import Server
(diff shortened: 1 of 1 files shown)


rhfogh (approved):
As it happened, I had already made this change (as a cleanup) in my local branch. I had just not pushed it. This is fine.`

rhfogh (changes requested):
Your change is fine - but while you are at it, could you move the mxcube3 imports down below the python sys imports? This is nitpicking, but that is the accepted order.

rhfogh (approved):

marcus-oscarsson:
Well done guys :)

PR #963, opened 2023-03-22
refactor UI initialization handling

elmjag:
Make 'initialization' of the UI more robust.

Don't store 'logged in' status in the web-browser's local storage, as that state becomes stale when session cookie expires. Always fetch 'logged in' status from the server, before loading the rest of the UI.

Use null as initial value for state.login.loggedIn, to mark that the login status is not known yet.

Connect the 'App' component to the state.login.loggedIn and wait for loggedIn state to become true/false, before rendering main UI components.

Remove code for socket.io 'network' namespace, as it's is not used anymore.

Remove example react code, that we are no longer use.


marcus-oscarsson:
Thanks, well done :)

Issue #964, opened 2023-03-23
Problem when using GPhL workflows from MXPress workflows 


olofsvensson:
I'll submit a pull request with a patch which fixes problems when trying to run GPhL workflows from MXPress workflows.

olofsvensson:
PR created: https://github.com/mxcube/mxcubeweb/pull/965

PR #965, opened 2023-03-23
Issue #964: patch for fixing problems when running GPhL workflows from MXPress workflows

olofsvensson:
This patch consists of two changes:

1. If the workflow launched is not "GPhL", i.e. it's a BES workflow, the call to "set_wf_params" must be called after adding the model to the group.

2. A fix of a typo : XrayCentring2QueueEntry -> XrayCentering2QueueEntry


marcus-oscarsson:
Thanks !

PR #967, opened 2023-03-31
mxcubecore as group for pyproject

meguiraun:
hi!

I propose to move mxcubecore and video-streamer repos to a new group inside the pyproject file. Poetry is not affected (tested myself but please have a look). But this way we can still use pip for the installation in our deployment environment with less hassle (we cannot reach github, but conveniently pip ignores the mxcubecore group in pyproject)

We dont want to create a patch on some internal branch just for the shake of deployment, preferably we want to rely on upstream as much as we can.


meguiraun:
closing this as discussed by other means ;)

PR #968, opened 2023-03-31
don't log HTTP requests body

elmjag:
Logging HTTP requests body is an security issue, as we end up adding user's passwords to the log in clear text.

This means that passwords become visible to all users logged in to MxCube (for example observers).

If we really want to log HTTP bodies, we should add some kind of censoring mechanism, that for example replaces actual password with ****, in the log.

### DIFF
--- a/mxcube3/routes/main.py
+++ b/mxcube3/routes/main.py
@@ -57,7 +57,6 @@ def init_route(app, server, url_prefix):
         logging.getLogger("MX3.HWR").debug('Path: %s', request.full_path)
         logging.getLogger("MX3.HWR").debug('scheme: %s', request.scheme)
         logging.getLogger("MX3.HWR").debug('Headers: %s', request.headers)
-        logging.getLogger("MX3.HWR").debug('Body: %s', request.get_data())
 
         if not flask_login.current_user.is_anonymous:
             flask_login.current_user.last_request_timestamp = datetime.now()


meghdadyazdi:
why should Headers, schema and path be logged?

marcus-oscarsson:
Hi both, it was added as a `DEBUG` level logging so they should only appear if you use `DEBUG` level when you log. I think there is a problem with setting the log level so that its always set to `DEBUG`. I otherwise think it can be quite useful to have that information when you debug requests. 

Id rather suggest that we fix the setting of the `DEBUG` level instead. That's maybe not sufficient for your concerns in that case we can add a separate logger for requests that are never stored on files and only activated when `DEBUG` is passed on the command line.

What do you think ?

marcus-oscarsson:
Well actually, we don't really need the body itself you are right. I guess it can be added in special cases if needed. However the rest is still interesting IMO.

elmjag:
In general I agree that it would be nice to improve the control of the logging.

I think what would be really nice if there was logging 'topics'. For example something like:

hwr 
events
http_requests
error
...etc...

And there would be a way to select topics that should be logged in server config and/or as command line arguments. With some 'low volume' defaults.

I feel that normally you are only interested in a small subset of log messages while trouble shooting a particular problem. It would be nice if there was a simple way to control what is logged.

marcus-oscarsson:
Yes its a good idea :)

PR #969, opened 2023-04-04
define newZoom for setZoom

meghdadyazdi:
I think `zoom_motor` is not defined in `this.props` but in `this.props.hardwareObjects`?


marcus-oscarsson:
Well spotted

PR #974, opened 2023-05-03
My fixes

meguiraun:
- Sign out from other sessions if same account tries to login
- Reconnect ws on login (to avoid refresh)
- Small fixes on shapes object to avoid ui crash on refresh

Wip because I need to test more the ws reconnection logic


marcus-oscarsson:
Thanks, Mikel ill give it a try as soon as I can

meguiraun:
I added a few more commits ;)

the RateLimited was not working if two signals were arriving too close (det distance and resolution for example). The decorator is shared between all.

meguiraun:
an update on this... All commits work as intented :) but there is some glitch somewhere in the login system... sometimes the ui does not load at all (stays loading forever). So I had to remove the temporary files related to the auth db to get it back.

meguiraun:
please give it a try if you have time. I could not reproduce the "hang on refresh", so maybe it was solved in one of your recent MRs?

marcus-oscarsson:
Interesting, Ok Ill try to have a look. There seems to be some conflict though, would you mind having a look :). Perhpas just the version that changed in pyproject.toml

meguiraun:
dont see any conflict... there was one that I solved before your comment, maybe you looked at this just before?

marcus-oscarsson:
Okey, I see. Its because I have the rebase option by default. Ill try to have a look a bit later today

mxcube3/routes/main.py
```diff
@@ -39,17 +39,11 @@ def unrestricted_serve_static_file():
     def get_ui_properties():
         return app.get_ui_properties()
 
-    @bp.route("/version")
```
marcus-oscarsson, line 42: I don't mind if we remove this route and aggregate it with another route (is it for reducing the number of routes ?). However I'm would rather in that case that we rename mode to something with more meaning because its IMO odd to have a route called mode that also gives the server version. 

I don't have a good name for it, something like application_info or settings ?

ui/src/serverIO.js
```diff
@@ -104,29 +104,54 @@ class ServerIO {
 
   disconnect() {
     this.connected = false;
-    this.hwrSocket.disconnect();
-    this.loggingSocket.disconnect();
+    this.hwrSocket.close();
+    this.loggingSocket.close();
+    // this.hwrSocket.disconnect();
```
marcus-oscarsson: Unnecessary comment I guess ?

marcus-oscarsson:
I tried abit and it seems to working fine

ui/src/serverIO.js
meguiraun: yep, leftover test

mxcube3/routes/main.py
meguiraun: changed to application_settings, but unsure what would best name 🤨

I had to merge these two since they were coming out of order often, and the UI was expecting an order.

marcus-oscarsson: I see, sure why not `application_settings`, we should just be careful that it does not become a dark corner for various "data" :). We could move the options that are in the `.env` file for phase control and so on as well :), not necessarily in this PR but I'm all for it of you would like to give it a go :+1:

meguiraun: on a next PR ;)

PR #983, opened 2023-06-01
Fix name of Conda environment to match docs

axelboc:
On the wiki, the environment is named `mxcube`. I could have gone the other way and renamed it to `mxcubeweb` on the wiki, but the more generic name `mxcube` makes more sense to me, so I decided to change the YAML file.

### DIFF
--- a/conda-environment.yml
+++ b/conda-environment.yml
@@ -1,4 +1,4 @@
-name: mxcubeweb
+name: mxcube
 channels:
   - conda-forge
 dependencies:


hellmig:
I would suggest to stay with the name mxcubeweb to distinguish between other mxcube related environments, e. g. mxcubeqt, on the same computer.

axelboc:
Alright, no worries, will fix the wiki then :+1:

marcus-oscarsson:
Thanks @axelboc !

PR #984, opened 2023-06-05
Updated needed for changes/updates in #760 of mxcubecore

marcus-oscarsson:
Updated needed for changes/updates in #760 of mxcubecore

**Updates done:**

- SSX related components (chip)
- Adapted to latest workflow code GPHL and ESRF for A1, adds the possibility to run custom scripts for phase change and centering on the MD 
- Some cleanup and minor fixes

### DIFF
--- a/mxcube3/core/adapter/diffractometer_adapter.py
+++ b/mxcube3/core/adapter/diffractometer_adapter.py
@@ -21,7 +21,7 @@ class DiffractometerAdapter(AdapterBase):
         pass
 
     def state(self):
-        return "READY" if self._ho.get_state().name == "READY" else "BUSY"
+        return "READY" if self._ho.is_ready() else "BUSY"
 
     def head_configuration(self) -> dict:
         data = self._ho.get_head_configuration()
--- a/mxcube3/core/adapter/diffractometer_adapter.py
+++ b/mxcube3/core/adapter/diffractometer_adapter.py
@@ -15,13 +15,13 @@ class DiffractometerAdapter(AdapterBase):
         ho.connect("valueChanged", self._state_change)
 
     def _state_change(self, *args, **kwargs):
-        self.state_change(**kwargs)
+        self.state_change(*args, **kwargs)
 
     def stop(self):
         pass
 
     def state(self):
-        return "READY" if self._ho.is_ready() else "BUSY"
+        return "READY" if self._ho.get_state().name == "READY" else "BUSY"
 
     def head_configuration(self) -> dict:
         data = self._ho.get_head_configuration()
--- a/mxcube3/core/components/lims.py
+++ b/mxcube3/core/components/lims.py
@@ -305,6 +305,7 @@ class Lims(ComponentBase):
 
         logging.getLogger("MX3.HWR").info("[LIMS] Selecting proposal: %s" % proposal)
         logging.getLogger("MX3.HWR").info("[LIMS] Proposal info: %s" % proposal_info)
+
         if (
             HWR.beamline.lims.loginType.lower() == "user"
             and "Commissioning" in proposal_info["Proposal"]["title"]
@@ -322,7 +323,10 @@ class Lims(ComponentBase):
             HWR.beamline.session.proposal_number = proposal_info.get("Proposal").get(
                 "number", ""
             )
-            todays_session = HWR.beamline.lims.get_todays_session(proposal_info)
+
+            todays_session = HWR.beamline.lims.get_todays_session(
+                proposal_info, create_session=False
+            )
             HWR.beamline.session.session_id = todays_session.get("session").get(
                 "sessionId"
             )
@@ -346,10 +350,6 @@ class Lims(ComponentBase):
 
             # Get all the files in the root data dir for this user
             root_path = HWR.beamline.session.get_base_image_directory()
-
-            if not self.app.INITIAL_FILE_LIST and os.path.isdir(root_path):
-                ftype = HWR.beamline.detector.get_property("file_suffix")
-                self.app.INITIAL_FILE_LIST = fsutils.scantree(root_path, [ftype])
 
             # save selected proposal in users db
             current_user.selected_proposal = proposal
--- a/mxcube3/core/components/queue.py
+++ b/mxcube3/core/components/queue.py
@@ -1630,6 +1630,7 @@ class Queue(ComponentBase):
         group_model.set_enabled(True)
         HWR.beamline.queue_model.add_child(parent_model, group_model)
         HWR.beamline.queue_model.add_child(group_model, wf_model)
+
         if not task["parameters"]["wfpath"] == "Gphl":
             self.set_wf_params(wf_model, dc_entry, task, sample_model)
 
@@ -1637,9 +1638,6 @@ class Queue(ComponentBase):
         group_entry.set_enabled(True)
         parent_entry.enqueue(group_entry)
         group_entry.enqueue(dc_entry)
-
-        if task["parameters"]["wfpath"] != "Gphl":
-            self.set_wf_params(wf_model, dc_entry, task, sample_model)
 
         return wf_model._node_id
 
--- a/mxcube3/core/components/queue.py
+++ b/mxcube3/core/components/queue.py
@@ -67,20 +67,21 @@ class Queue(ComponentBase):
         return prefix_path_dict
 
     def get_run_number(self, pt):
-        prefix_path_dict = self.build_prefix_path_dict(self.app.INITIAL_FILE_LIST)
-
         # Path templates of files not yet written to to disk, we are only
         # interested in the prefix path
-        fname = pt.get_image_path()
-        prefix_path, _, _ = qmo.PathTemplate.interpret_path(fname)
-        run_number = HWR.beamline.queue_model.get_next_run_number(pt)
 
-        if prefix_path in prefix_path_dict:
-            rn = run_number + prefix_path_dict[prefix_path]
-        else:
-            rn = run_number
+        pt.run_number = HWR.beamline.queue_model.get_next_run_number(pt)
+        start_fname, end_fname = pt.get_first_and_last_file()
 
-        return rn
+        while os.path.isfile(start_fname) or os.path.isfile(end_fname):
+            pt.run_number += 1
+
+            if pt.run_number > 1000:
+                raise RuntimeError("Over a thousand runs of the same collection")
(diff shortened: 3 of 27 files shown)


marcus-oscarsson:
Looks like there is some inconsistency with my local `mxcubecore` and the one on develop that fails a bunch of tests ill try to get that fixed ASAP.

mxcube3/core/adapter/diffractometer_adapter.py
```diff
@@ -15,13 +15,13 @@ def __init__(self, ho, *args, **kwargs):
...
     def _state_change(self, *args, **kwargs):
-        self.state_change(**kwargs)
+        self.state_change(*args, **kwargs)
 
     def stop(self):
         pass
 
     def state(self):
-        return "READY" if self._ho.is_ready() else "BUSY"
+        return "READY" if self._ho.get_state().name == "READY" else "BUSY"
```
rhfogh: What is the reason to bypass the is_ready function? Unless there is some object that does not have it and cannot be given it, it would be better to use always the same function for readiness testing.

mxcube3/core/components/queue.py
```diff
@@ -1634,6 +1638,9 @@ def add_workflow(self, node_id, task):
         parent_entry.enqueue(group_entry)
         group_entry.enqueue(dc_entry)
 
+        if task["parameters"]["wfpath"] != "Gphl":
```
rhfogh: Could this be done in the same lines (and at teh same place in the code) as for the equivalent Gphl workflow? Or do you know a reason why this must be done in two different places a few lines apart?

mxcube3/core/components/queue.py
```diff
@@ -2313,6 +2332,11 @@ def get_default_task_parameters(self, task_name):
 
         schema = self.get_task_schema(data_model) if data_model else {}
 
+        try:
```
rhfogh, line 2333: This looks like a system for controlling the UI rendering. Maybe I should adapt my GPhL UI to use this - but someone would have to explain it to me first.

mxcube3/routes/queue.py
```diff
@@ -278,6 +278,20 @@ def get_avilable_tasks():
         resp.status_code = 200
         return resp
 
+    @bp.route("/update_dependent_field", methods=["POST"])
+    @server.restrict
+    def update_dependent_field():
+        """
+        Returns a list of all available tasks
```
rhfogh: Does this doc string need updating?

mxcube3/routes/samplecentring.py
```diff
@@ -76,7 +76,7 @@ def set_image_size():
         params = request.get_json()
 
         res = app.sample_view.set_image_size(
-            int(params["width"]), int(params["height"])
+            float(params["width"]), float(params["height"])
```
rhfogh, line 79: Is this set in floating point pixels/ or in mm? or what?

rhfogh (changes requested):
Looks nice (though much of the javascript stuff goes over my head). I'd approve, except for the minor points raised.

mxcube3/core/components/lims.py
```diff
@@ -322,7 +323,10 @@ def select_proposal(self, proposal):
             HWR.beamline.session.proposal_number = proposal_info.get("Proposal").get(
                 "number", ""
             )
-            todays_session = HWR.beamline.lims.get_todays_session(proposal_info)
+
+            todays_session = HWR.beamline.lims.get_todays_session(
```
meguiraun, line 327: we will have troubles with this `create_session=False`, why did you change it?

mxcube3/core/components/queue.py
```diff
@@ -1382,6 +1383,9 @@ def _create_queue_entry(self, task, task_name):
         Return:
             (tuple): (model, entry)
         """
+        if not task["parameters"]["osc_range"]:
```
meguiraun, line 1386: I dont understand this 🧐...

`osc_range` comes with a `null` value? Should not this be fixed in the ui?

mxcube3/core/components/queue.py
```diff
@@ -2294,6 +2301,18 @@ def add_centring(self, _id, params):
 
         return {"QueueId": new_node, "Type": "Centring", "Params": params}
 
+    def update_dependent_field(self, task_name, data):
```
meguiraun, line 2302: what is this for? What is `task_name` here? `task_name` for me would be `DataCollection`...

mxcube3/core/components/samplechanger.py
```diff
@@ -432,53 +433,66 @@ def queue_mount_sample(view, data_model, centring_done_cb, async_result):
...
-
-                logging.getLogger("user_level_log").info("Centring ...")
-                centring_result = async_result.get()
-                if centring_result["valid"]:
-                    logging.getLogger("user_level_log").info("Centring done !")
-                else:
-                    if centring_method == queue_entry.CENTRING_METHOD.FULLY_AUTOMATIC:
-                        raise queue_entry.QueueSkippEntryException(
-                            "Could not center sample, skipping", ""
+        use_custom_centring_routine = HWR.beamline.diffractometer.get_property(
```
meguiraun, line 437: i like this... What are you using this for? (XRay centring comes to my mind)

mxcube3/core/components/sampleview.py
```diff
@@ -560,10 +562,15 @@ def take_snapshots(self, snapshots=None, _do_take_snapshot=_do_take_snapshot):
                 hasattr(diffractometer_object, "set_phase")
                 and diffractometer_object.get_current_phase() != "Centring"
             ):
-                logging.getLogger("user_level_log").info(
-                    "Moving Diffractometer to CentringPhase"
+                use_custom_snapshot_routine = diffractometer_object.get_property(
```
meguiraun, line 565: i wonder if we should keep this get_prop inside the dm object.

Also, would not this be HWR.beamline.diffractometer.....

edit: wait... is this actually in use?

mxcube3/core/components/sampleview.py
```diff
@@ -560,10 +562,15 @@ def take_snapshots(self, snapshots=None, _do_take_snapshot=_do_take_snapshot):
...
+                    "custom_snapshot_script_dir", None
                 )
-                diffractometer_object.set_phase("Centring", wait=True, timeout=200)
+                if not use_custom_snapshot_routine:
+                    logging.getLogger("user_level_log").info(
+                        "Moving Diffractometer to CentringPhase Not done for tests (DN)"
```
meguiraun, line 570: wrong msg

ui/src/components/Tasks/Characterisation.js
```diff
@@ -448,16 +451,16 @@ Characterisation = connect((state) => {
       beam_size: state.sampleview.currentAperture,
       resolution: (state.taskForm.sampleIds.constructor !== Array
         ? state.taskForm.taskData.parameters.resolution
-        : state.beamline.hardwareObjects.resolution.value),
+        : toFixed(state, "resolution")),
```
meguiraun, line 454: I dont understand this change, from where the values comes now?

mxcube3/core/components/queue.py
marcus-oscarsson: Yes, you are indeed right that it seems a bit odd. However now a days and especially with SSX oscillation range might not be set or used at all hence its `Null`. We could perhaps set it to 0 instead. There is some thought work to do and its to remove the "composite collection" dictionary that we are using for the "traditional" oscillation experiments.

marcus-oscarsson: I think in the end it just comes down to actually doing this as we have been speaking about it for sometime. I think we should discuss as part of our roadmap discussion because I think it also should be treated together with the queue discussion.

mxcube3/core/components/queue.py
marcus-oscarsson: yes, I think two different people did a very similar change and then we did not pick this up when we merged.

mxcube3/core/components/queue.py
marcus-oscarsson: :), I guess this one is fine as we spoke about this during our meeting today,

mxcube3/core/components/queue.py
marcus-oscarsson: Yes it probably could, we should as we said earlier have a look at if what we have today satisfy your requirements.

mxcube3/core/components/samplechanger.py
marcus-oscarsson: Well, its mostly used on A1 where we wrote some PMAC routine to speed up phase change and perform cantering on the MD computer. Its work mostly done by **Didier Nurizzo**. There are actually also some workarounds to issues of the MD firmware, so we can now specify weather to use those or not

mxcube3/core/components/sampleview.py
marcus-oscarsson: Yes it is unfortunately used and we should really really move away from it. Ill give it a try but it needs to be tested a bit before I remove it. It was used to get the overlays from the UI but we are not taking snapshots that way any longer. I will try to make a PR, shortly, removing the entire enable_snapshot "patch" and add something cleaner

mxcube3/routes/queue.py
marcus-oscarsson: Oh yes

mxcube3/routes/samplecentring.py
marcus-oscarsson: Its microscope (OAV) image size so its in pixels

ui/src/components/Tasks/Characterisation.js
marcus-oscarsson: Ah, its the application of the values specified in the [ui.yaml ](https://github.com/mxcube/mxcubeweb/blob/develop/test/HardwareObjectsMockup.xml/mxcube-web/ui.yaml) file. Basically just making sure that the precision in that file is used instead of like before when it was a bit fixed.

meguiraun: i meant the actual values written to the fields, I double checked and now I see the toFixed method, so all good

marcus-oscarsson: :) cool

meguiraun (approved):

mxcube3/routes/samplecentring.py
rhfogh: But not integer values, apparently?

PR #986, opened 2023-06-05
Update to sample list view, with table like and graphical view.

marcus-oscarsson:
_**PR on behalf of Jean-Baptiste**_

- Fixed bug in selection getting stuck
- Attempt to optimize certain actions, selection, deletion of multiple tasks


meguiraun (approved):
thanks!

PR #987, opened 2023-06-05
Cypress

marcus-oscarsson:
Introduced [Cypress](https://www.cypress.io/) for UI/UX testing,


meguiraun (approved):
I can only approve it once. ;)

thanks!

PR #994, opened 2023-06-08
Move login spinner into button

axelboc:
#### BEFORE

Peek 2023-06-08 14-16

#### AFTER

Peek 2023-06-08 14-162

---

I'm also removing the `keypress` handler on the password field, which in addition to being deprecated, was also broken and unnecessary (since there's a proper `form` element with a `submit` button and event handler).


marcus-oscarsson:
Wonderful !, nice initiative !

PR #996, opened 2023-06-09
Increase clickable area of sample controls

axelboc:
Now the whole sample controls bar is clickable, which (I find) improves the UX of the controls significantly. I've given particular care to tweaking the `.active`/`:active`/`:hover`/`:focus` states of the controls with subtle yet noticeable colour changes, and to position the overlays at a consistent distance from the bar.

The backlight/frontlight buttons were a bit challenging to refactor — basically, the arrow button now sits on top of the main control button instead of next to it (but is still not nested inside it, since a `button` element cannot be nested inside another `button` element).

I've written the new styles as [CSS Modules](https://github.com/css-modules/css-modules) to start moving a little bit away from global styles. Every CSS class in `SampleView.module.css` is now automatically namespaced and can therefore never conflict with any existing or future global classes (whether provided by MXCuBE or Bootstrap). Happy to go further into this if anyone is interested/concerned.

#### Before

Peek 2023-06-09 16-42

#### After

Peek 2023-06-09 16-49


ui/src/components/SampleView/SampleControls.js
```diff
@@ -181,226 +182,219 @@ export default class SampleControls extends React.Component {
     const zoom_motor = this.props.hardwareObjects[zoom_motor_uiprop.attribute];
 
     return (
-      <div style={{ display: 'flex', position: 'absolute', width: '100%' }}>
-        <div className="sample-controlls text-center">
-          <ul className="bs-glyphicons-list">
-            <li>
```
axelboc, line 187: The `ul` element is not needed (for what it's worth, it creates too much noise for screenreader users).

ui/src/components/SampleView/SampleControls.js
```diff
@@ -181,226 +182,219 @@ export default class SampleControls extends React.Component {
...
-              <li>
-                <Button
-                  type="button"
-                  data-toggle="tooltip"
-                  title="Zoom in/out"
-                  className="fas fa-search sample-control"
-                  name="zoomOut"
+      <div className={styles.controls}>
+        <Button
+          as="a"
```
axelboc, line 187: It was not obvious that this button was a link, so I made it explicit.

ui/src/components/SampleView/SampleControls.js
```diff
@@ -181,226 +182,219 @@ export default class SampleControls extends React.Component {
...
+        </Button>
+        <Button
+          className={styles.controlBtn}
+          active={this.props.clickCentring}
+          title="Start 3-click centring"
+          data-toggle="tooltip"
+          onClick={this.toggleCentring}
+        >
+          <i className={`${styles.controlIcon} fas fa-circle-notch`} />
+          <span className={styles.controlLabel}>3-click centring</span>
```
axelboc, line 217: Sentence case instead of title case for consistency with other buttons.

ui/src/components/SampleView/SampleControls.js
```diff
@@ -181,226 +182,219 @@ export default class SampleControls extends React.Component {
...
+          className={styles.controlBtn}
+          href="#"
+          target="_blank"
+          download
+          title="Take snapshot"
+          data-toggle="tooltip"
+          onClick={this.takeSnapShot}
+        >
+          <i className={`${styles.controlIcon} fas fa-camera`} />
+          <span className={styles.controlLabel}>Snapshot</span>
```
axelboc, line 197: The main change, really, is simply that the labels are now inside the buttons.

ui/src/components/SampleView/SampleControls.js
```diff
@@ -181,226 +182,219 @@ export default class SampleControls extends React.Component {
...
-                      onChange={e => this.props.setBeamlineAttribute('diffractometer.frontlight', e.target.value)}
-                      name="frontLightSlider"
-                    />
-                  </span>
-                )}
-              >
+              </span>
+            )}
+          >
+            {({ ref, ...triggerHandlers }) => (
```
axelboc, line 321: I'm using the render prop syntax of `OverflowTrigger` in order to position the overlay in relation to the main control button instead of to the trigger (i.e. the arrow button).

ui/src/components/SampleView/SampleControls.js
marcus-oscarsson: :)

marcus-oscarsson:
Very nice ! thanks !

marcus-oscarsson:
I like the subtle color changes :+1:

ui/src/components/SampleView/SampleControls.js
marcus-oscarsson: That's neat, and convenient !

meguiraun:
I like it! thanks

PR #998, opened 2023-06-12
Use proper links for navigation

axelboc:
The main navigation in the top bar was made of `div` elements with `href` attributes, which is not valid HTML and required styling them with `cursor: pointer`. I've now switched to using proper links, and took the opportunity to fix some vertical alignment issues with Flexbox.

### DIFF
--- a/ui/src/components/MXNavbar/MXNavbar.jsx
+++ b/ui/src/components/MXNavbar/MXNavbar.jsx
@@ -37,35 +37,35 @@ class MXNavbar extends React.Component {
           <Navbar.Toggle aria-controls="responsive-navbar-nav" />
           <Navbar.Collapse id="responsive-navbar-nav">
             <Nav className="m-auto">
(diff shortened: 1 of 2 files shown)


ui/src/components/MXNavbar/MXNavbar.jsx
```diff
@@ -37,35 +37,35 @@ class MXNavbar extends React.Component {
...
-                <Nav.Item className="nav-link">
+                <Nav.Link>
                   <span style={ raStyle } className="me-1 fas fa-lg fa-globe">
                     {numObservers > 0 ? <span className="badge-num">{numObservers}</span> : null }
                   </span>
                   Remote
-                </Nav.Item>
+                </Nav.Link>
               </LinkContainer>
-              <Button as={Nav.Link} className="nav-link pe-0" variant="Light" onClick={this.signOut}>
```
axelboc, line 68: Ideally, this should be a `button` element, but the styling complexity this change would require didn't seem worth it.

PR #1000, opened 2023-06-12
Removed redux local storage and cross tab sync

marcus-oscarsson:
Perhaps a bit controversial but removing the usage of `redux-persist `and `redux-state-sync` ;)

In general we are only using these two libraries for two things;

-  Mirroring the interface for observers 
-  Mirroring open tabs
-  Saving the selected motor step size through the `SET_STEP_SIZE` action

Everything that we need to mirror is already on the server except for the dialogs and possibly context menus
which we can more effectively create a separate solution for. The latest(er) versions of both `redux-persist` and
`redux-state-sync ` have broken the mirroring feature. The step size can easily be managed on the server side.

In my opinion it reduces the complexity and bandwidth needs, let me know what you think ?

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -43,6 +43,7 @@
         "react-dom": "^17.0.2",
         "react-draggable": "^4.4.4",
         "react-icons": "^4.3.1",
+        "react-lazyload": "^3.2.0",
         "react-numeric-input": "^2.2.3",
         "react-plotly.js": "^2.4.0",
         "react-redux": "7.2.6",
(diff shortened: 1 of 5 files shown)


meguiraun:
good one for the mr number 1000!

Yeah, we already removed part of the persist when we removed the login info, I bet @elmjag is gonna be excited with this ;)

marcus-oscarsson:
Ok, sounds good :) It remains to be decided what we do with the step size, we could store that specifically in local storage (bypassing redux ofcourse). It makes sense that its a per browser/user setting, what do you think ?

marcus-oscarsson:
I drafted a temporary local storage based solution for "caching" step size, but I have to admit that I was not very happy with it as it overrides and makes the server side setting quite awkward. I'm removing the WIP for now I'm fine with moving on with this, but we should discuss how to handle the step size

elmjag:
Very good. Can we also drop redux-persist and redux-state-sync packages from package.json?

https://github.com/mxcube/mxcubeweb/blob/develop/ui/package.json#L54
https://github.com/mxcube/mxcubeweb/blob/develop/ui/package.json#L55

Or are they still used somewhere?

meguiraun:
I agree with keeping the step size in the local storage of the client. But anyway, it feels a very small functionality so even not caching it does not seem to be a big loss? We can keep it in the state of the component...

marcus-oscarsson:
Yes, I think its alright, at least to start with.

marcus-oscarsson:
Yes, I think its alright, at least to start with.

meguiraun (approved):
tack

PR #1001, opened 2023-06-12
Fixed update bug in PopInput

marcus-oscarsson:
- BUSY state was incorrect (`"MOVING"` instead of `"BUSY"`) preventing rendering of  "busy indicator"
  - Added `componentShouldUpdate` (to prevent re-render when other objects are updated)
  - Fixed root close (click out side of input did not hide input)

**Edit**
The componentShouldUpdate`  is not needed the problem came from the assignment in the reducer that was fixed


ui/src/reducers/beamline.js
```diff
@@ -150,8 +150,7 @@ export default (state = INITIAL_STATE, action) => {
 
     case 'BL_UPDATE_HARDWARE_OBJECT': {
       const attrData = Object.assign(
-        state.hardwareObjects[action.data.name] || {},
-        action.data
+        {}, {...state.hardwareObjects[action.data.name]}, action.data
```
marcus-oscarsson: This one is interesting :), took me sometime to find

ui/src/components/PopInput/PopInput.jsx
```diff
@@ -42,32 +42,20 @@ export default class PopInput extends React.Component {
     this.save = this.save.bind(this);
     this.cancel = this.cancel.bind(this);
     this.submit = this.submit.bind(this);
-    this.showOvelay = this.showOvelay.bind(this);
-    this.state = {show: false};
```
marcus-oscarsson, line 46: Removed the internal `show` state

ui/src/components/PopInput/PopInput.jsx
```diff
@@ -42,32 +42,20 @@ export default class PopInput extends React.Component {
...
+    );
   }
 
-
-  showOvelay(value) {
-    this.setState({
-      show: value
-    });
+  hideOverlay() {
+    document.body.click();
```
marcus-oscarsson, line 49: Letting the `OverlayTrigger`  handle the state show internally by triggering `document.body.click()`

ui/src/reducers/beamline.js
axelboc: Since you're now passing an empty object as first argument to `Object.assign`, you don't need to make a copy of the second argument — it will not be mutated.

axelboc (approved):
Nice, I was looking into this component yesterday and noticed a bunch of the code was not doing what it was supposed to, so that makes sense now :smile: 

While you're at it, it seems there's a bunch of unused code in this component — maybe you can give me some insight on it:

- There's a bunch of logic related to `children` (`getChild`, `React.cloneElement()`), but the component never seems to receive children.
- The `name` and `title` props don't seem to be used.
- `isBusy`, `isIdle` and `isAborted` are never called with a `data` argument.

There's also a `LabelPopInput` component that seems fairly similar to `PopInput` but doesn't seem used at all.

ui/src/components/PopInput/PopInput.jsx
axelboc: The React Way:tm: would have been to set a `ref` on the trigger element, but this is much simpler :joy:

ui/src/reducers/beamline.js
marcus-oscarsson: Oh yes, thanks :),

ui/src/components/PopInput/PopInput.jsx
marcus-oscarsson: Yes I know, but I honestly find that so clonky, and the API have been changing. So when I use the easier solutions when I find one (and probably faster, its apparently a thing now a days :) )

marcus-oscarsson:
Thanks for the nice input @axelboc. I finally removed the `getChild` related code. It was used or foreseen for the possibility to supply alternative busy and input components which we never used in anyways.

PR #1003, opened 2023-06-20
Refactor `PopInput` component and fix minor CSS issues

axelboc:
Still not super happy about the `inplace` prop. I feel like the two use cases (in a popover and in place) are totally different and that `PopInput` should only care about the popover case. The in-place case should use the underlying numeric input component directly. Anyway, for another time.

By the way, it seems that the in-place step inputs in the "Sample alignment motors" popover do not work as expected, at least in my local environment. Changing the step does not do anything and closes the popover.


ui/src/components/Cryo/DefaultBusy.jsx
```diff
@@ -4,7 +4,7 @@ import { Button, ButtonToolbar } from 'react-bootstrap';
 
 import './style.css';
 
-export default class DefaultInput extends React.Component {
```
axelboc, line 7: Just a naming issue. Seems there are a lot of duplicated components in the `Cryo` folder, which is a shame since they will keep diverging over time.

ui/src/components/MotorInput/MotorInput.jsx
```diff
@@ -136,13 +137,13 @@ export default class MotorInput extends React.Component {
                   onSave={this.props.saveStep}
                   suffix={suffix}
                   inputSize="5"
-                  style={{ display: 'inline-block', marginLeft: 'auto', marginRight: '0.5em' }}
```
axelboc, line 139: Moving the padding you very recently added to `rw-widget-right-border`, as it was adding a space between the input and the cancel button when `BUSY`.

ui/src/components/MotorInput/MotorInput.jsx
```diff
@@ -136,13 +137,13 @@ export default class MotorInput extends React.Component {
...
+                  style={{ display: 'inline-block', marginLeft: 'auto', marginRight: '0.5em', paddingLeft: '0.5em'}}
                 />
                 : null
               }
               {this.props.state !== MOTOR_STATE.READY ?
                 <Button
-                  style={{ width: '100%', height: '100%', display: 'block' }}
```
axelboc, line 145: Minor CSS issue: the cancel button was increasing the height of the field when `BUSY` which was a bit distracting.

ui/src/components/MotorInput/MotorInput.jsx
```diff
@@ -157,21 +158,18 @@ export default class MotorInput extends React.Component {
               }
             </div>
           </div>
-          {this.props.inplace ?
-            <div style={{ position: 'relative' }}>
```
axelboc, line 161: Moving the in-place `PopInput` out of the `MotorInput`'s form so as to not trigger a submission when the inner `PopInput` form is submitted.

ui/src/components/PopInput/DefaultBusy.jsx
axelboc, line 1: Merging this component into `DefaultInput`.

ui/src/components/PopInput/DefaultInput.jsx
```diff
@@ -1,74 +1,68 @@
...
+              className="popinput-input"
+              size={this.props.inputSize}
+              precision={this.props.precision}
+              value={this.props.value}
+              step={this.props.step}
+            />
+          )}
+          <ButtonToolbar className="ms-1">
+            {!this.props.busy && (
+              <Button type="submit" variant="success" size="sm">
```
axelboc, line 38: Taking advantage of the form's `onSubmit` handler, which allows removing `onClick` on the button and the extra `save` method.

ui/src/components/PopInput/DefaultInput.jsx
```diff
@@ -1,74 +1,68 @@
...
+                <TiTick size="1.5em" />
               </Button>
-              ) : (null)
-            }
+            )}
+            {!this.props.inplace && (
+              <Button
```
axelboc, line 43: The busy indicator moved from `DefaultBusy`. I'm also fixing the CSS so the popover keeps the same size when the spinner appears, and so the spinner is nicely centered vertically.

ui/src/components/PopInput/PopInput.jsx
```diff
@@ -41,70 +38,35 @@ export default class PopInput extends React.Component {
...
-  }
-
-  submit(event) {
-    event.preventDefault();
-    this.save();
-  }
-
-  inputComponent() {
-    return (
-      <DefaultInput
```
axelboc, line 83: Moving all this into the `render` method.

ui/src/components/PopInput/PopInput.jsx
```diff
@@ -117,80 +79,84 @@ export default class PopInput extends React.Component {
   }
 
   render() {
-    const linkClass = 'editable-click';
-    const busyVisibility = this.isBusy() ? '' : 'visibility-hidden';
-    const inputVisibility = !this.isBusy() ? '' : 'visibility-hidden';
-    const title = (this.props.title === '') ? this.props.name : this.props.title;
-
-    let stateClass = 'value-label-enter-success';
```
axelboc, line 125: This class was adding a `background-color` CSS transition, which made the UI feel less responsive. I'm removing the class.

ui/src/components/PopInput/PopInput.jsx
```diff
@@ -117,80 +79,84 @@ export default class PopInput extends React.Component {
   }
 
   render() {
-    const linkClass = 'editable-click';
-    const busyVisibility = this.isBusy() ? '' : 'visibility-hidden';
-    const inputVisibility = !this.isBusy() ? '' : 'visibility-hidden';
```
axelboc, line 122: Removing this `visibility-hidden` class and using conditional rendering instead.

ui/src/components/MotorInput/MotorInput.jsx
marcus-oscarsson: I noticed that as well, thanks for fixing it :)

ui/src/components/MotorInput/MotorInput.jsx
marcus-oscarsson: Nice, well done !

marcus-oscarsson:
Well done !

PR #1004, opened 2023-06-21
Added the possibility to use development SSL certificates

marcus-oscarsson:
Added the possibility to use development SSL certificates

### DIFF
--- a/mxcube3/core/models/configmodels.py
+++ b/mxcube3/core/models/configmodels.py
@@ -16,6 +16,13 @@ class FlaskConfigModel(BaseModel):
     USER_DB_PATH: str = Field("/tmp/mxcube-user.db", description="")
     PERMANENT_SESSION_LIFETIME: datetime.timedelta
     SESSION_PERMANENT: bool = Field(True, description="")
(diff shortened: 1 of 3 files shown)


mxcube3/core/models/configmodels.py
```diff
@@ -16,6 +16,13 @@ class FlaskConfigModel(BaseModel):
     USER_DB_PATH: str = Field("/tmp/mxcube-user.db", description="")
     PERMANENT_SESSION_LIFETIME: datetime.timedelta
     SESSION_PERMANENT: bool = Field(True, description="")
+    CERT_KEY: str = Field("", description="Full path to signed certficate key file")
```
meguiraun, line 19: should not these be Optional?

marcus-oscarsson: They are not used as long CERT is not set to SIGNED, the default value is the empty string so that should be alright. I somehow assumed that one knows that they need to have a value of one set CERT SIGNED.

meguiraun (approved):
great for this! tack!

PR #1006, opened 2023-06-22
Refactor login form and add validation

axelboc:
Notable visual changes:

- The form is now centered on the page, at the same position as the loading text/animation.
- The form is now properly validated: errors appear after submitting the form if either of the fields is not provided.
- The server error, which looks like a JSON or JS object is now displayed inside a `pre` element to be a bit more palatable. It'd be good to provide a more meaningful error message in the future.
- The background color of the page in development now matches the background color in production. It seems that there are two `index.html` files -- one in the `src` folder (production) and one in the `public` folder (development). The one in the `src` folder had an inline `background-colour` style declaration on the `body` tag. I've moved it to `main.css`.

Peek 2023-06-22 16-48

Additional changes:

- I've converted `App`, `LoadingScreen` and `Login` to functional React components (required for `react-hook-form` notably).
- I've removed `src/index.css` which doesn't seem to be used.
- I've moved a bunch of styles to CSS modules.
- I found a `body` style declaration in `src/components/PeriodicTable/style.css` that seemed to pollute the global scope with a `font-family: Arial` declaration. I've moved the declaration to `main.css`, as changing the body font back to the default Bootstrap/system font would have a major visual impact. (By the way, it seems that this `PeriodicTable/style.css` file has a corresponding `PeriodicTable/style.less` file. Might be worth removing the `style.less` file if it's not automatically compiled.)


ui/src/components/App.js
```diff
@@ -31,36 +38,43 @@ function PrivateOutlet() {
...
 }
 
-class App extends React.Component {
-  render() {
+function App(props) {
+  const { loggedIn } = props;
+
+  useEffect(() => {
     requireAuth();
+  }, []);
```
axelboc, line 53: `requireAuth` now runs asynchronously as a side effect, which is more idiomatic. No difference in practice.

ui/src/components/LoadingScreen/LoadingScreen.jsx
```diff
@@ -1,21 +1,16 @@
...
-        <img
-          className="loader-init"
-          src={loadingAnimation}
-          alt="loading"
-        />
-      </div>
+function LoadingScreen() {
+  return (
+    <div className={styles.root}>
+      <img className={styles.logo} src={logo} width="159" alt="MXCuBE" />
```
axelboc, line 9: Moving the width from the CSS to the HTML to avoid layout shift in case the image file is bigger.

ui/src/components/Login/Login.jsx
```diff
@@ -1,112 +1,123 @@
...
+  } = useForm({ defaultValues: { username: '', password: ''}});
 
+  function handleSubmit(data) {
+    setLoading(true);
+    signIn(data.username.toLowerCase(), data.password, router.navigate);
   }
 
-  render() {
-    return (
-      <Container>
```
axelboc, line 35: I've removed the `Container` and all the `Row`s and `Col`s.

marcus-oscarsson:
Very nice and good initiative ! :1st_place_medal:

ui/src/components/LoadingScreen/LoadingScreen.jsx
```diff
@@ -1,21 +1,16 @@
 import React from 'react';
 import logo from '../../img/mxcube_logo20.png';
 import loadingAnimation from '../../img/loading-animation.gif';
+import styles from './LoadingScreen.module.css';
```
marcus-oscarsson, line 4: Nice :), we should really start using this :) no more polluting CSS :)

marcus-oscarsson:
Maybe one thing :), we can maybe change the form title to MXCuBE WEB or simply MXCuBE instead of MXCuBE 3

axelboc:
> Maybe one thing :), we can maybe change the form title to MXCuBE WEB or simply MXCuBE instead of MXCuBE 3

Can do. What about the logo? Do you have the source file?

marcus-oscarsson:
Hm, we might have SVG somewhere I have to look a bit

PR #1009, opened 2023-06-23
Minor fixes to authenticaiton 

marcus-oscarsson:
- Defaulting to not using SSL 
 - Better error message when login fails
 - Making sure active users are up to date before returning list


axelboc:
Nice! I'll remove the `pre` element in my PR after you've merged this.

meguiraun (approved):

PR #1012, opened 2023-06-23
Basic ErrorBoundary with logging

marcus-oscarsson:
A simple `ErrorBoundary`  that displays the `stack trace` and the `redux state` . Both of those are also logged on the backend. The UI log is kept in a log file with the suffix `_ui.log`. The ErrorBoundary display can of course be made a bit nicer, but its a start.

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -5,6 +5,7 @@ import redis
 import sys
 import traceback
 
+
 from gevent import monkey
 monkey.patch_all(thread=False)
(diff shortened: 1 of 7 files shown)


rhfogh:
Cannot evaluate the code, but I highly approve of better error logging. Thanks!

meguiraun (approved):
really nice and useful addition! kudos!

marcus-oscarsson:
:), thanks !

PR #1013, opened 2023-06-26
Skip duplicate CI jobs

axelboc:
The E2E and build CI jobs were being run twice, on both `push` and `pull_request` events:

- `pull_request` is required for running the CI on PRs coming from forks, but it only runs when a PR is opened.
- `push` runs as soon as a local branch is pushed, before the PR is opened, so it saves time.

I'm adding `if` conditions to the build and E2E CI jobs so they run only on:

- `push` events for local branches
- `pull_request` events from forks

In other words, I skip `pull_request` events for local branches.

Additionally, I fix the name of the E2E job.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -10,6 +10,9 @@ jobs:
       matrix:
         python-version: ["3.8"] #, "3.9", "3.10"]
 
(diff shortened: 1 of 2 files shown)


marcus-oscarsson:
Ahh, great !

marcus-oscarsson:
Actually, I should also have a look at the backend because its not spinning up correctly. Ill try to do it when I get a few extra minutes

PR #1014, opened 2023-06-26
Set up automatic formatting on save with Prettier in VS Code

axelboc:
Prettier is an opinionated code formatter for JS/JSX/JSON/CSS/HTML/MD/... There are lots of benefits to using a code formatter:

- more readable code (consistent indentation/line breaks/...)
- focus on code rather than formatting in PR reviews
- faster development (format on save saves devs from having to type semi-columns, breaking lines, indenting, etc.)

I hope it's okay with everyone, I'm committing a `.vscode` folder to configure automatic formatting on save with Prettier in VS Code for the `ui` folder.  I've also added the `prettier` and `format` NPM scripts to respectively check the formatting and actually format files.

This is the first step towards enforcing Prettier formatting for the whole `ui` folder. Next step will be to format all the files and add a CI workflow to check that all file have been formatted.

### DIFF
--- a/ui/.prettierrc.json
+++ b/ui/.prettierrc.json
@@ -1,5 +1,4 @@
 {
-  "semi": true,
   "singleQuote": true,
   "overrides": [
     {
--- a/.gitignore
+++ b/.gitignore
@@ -64,6 +64,3 @@ npm-debug.log*
 /supervisord.pid
 log
 mxcube3/static
-
-# Editors
-.vscode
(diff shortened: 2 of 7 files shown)


ui/.prettierrc.json
```diff
@@ -1,5 +1,4 @@
 {
-  "semi": true,
```
axelboc, line 2: This is the default.

marcus-oscarsson:
Nice :+1:

PR #1015, opened 2023-06-26
Disable ESLint during build

axelboc:
The CI is failing because ESLint is running as part of building the UI with Create React App. This is not great, since it's potentially hiding real build errors, so I'm disabling it.

Linting should eventually be performed as a separate CI job to save time. I'll work on that (and on actually fixing linting issues) asap.

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -28,31 +28,33 @@ jobs:
         ports:
           # Maps port 6379 on service container to the host
           - 6379:6379
-        
+
     steps:
-    - uses: actions/checkout@v3
-    - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-test.yml
```diff
@@ -28,31 +28,33 @@ jobs:
...
+      - name: Install dependencies
+        run: |
+          conda env update --file conda-environment.yml --name base
+          pip install -e .
+      - name: Build UI
+        run: |
+          npm install --prefix ./ui 
+          npm run --prefix ./ui build
+        env:
+          DISABLE_ESLINT_PLUGIN: true
```
axelboc, line 53: This env var is the relevant change.

marcus-oscarsson:
Perfect , thanks

PR #1016, opened 2023-06-28
Isara UI

meguiraun:
hej,

Work towards slightly customised view for isara sample changer. For our Isara we let the horizontal space grow.

- New Isara dewar drawing (minor logic to mark samples in the queue)
- Handling of TYPE of sample changer: FLEX_HDC, CATS, others (only drawing for flex and cats)
   - type comes from sample changer hwobj (already there) 
- Filtering by pucks
- Handling of manual samples

The testing with mockups is somehow limited, so it would be great if you can run this against your FLEX_HCD, just in case.

Notes:
- The drawing for Isara will not cover all CATS... I checked for "CATS" since it is already used in the code so maybe easier to adapt by others (@hellmig ?). I dont mind creating a new ISARA type.
- Filter by cell should be disabled/removed for CATS-like?

Screenshot 2023-06-28 at 12 11 24


ui/src/containers/SampleIsaraView.jsx
```diff
@@ -0,0 +1,282 @@
...
+    if (position < 22) {
+      return { line: 3, column: position - 16, nbColumn: 6 };
+    }
+    if (position < 27) {
+      return { line: 4, column: position - 22, nbColumn: 5 };
+    }
+    if (position < 30) {
+      return { line: 5, column: position - 27, nbColumn: 2 };
+    }
+    // if (position < this.insideCells) {
```
marcus-oscarsson: Removed if unsued

marcus-oscarsson:
Sure I will give it a try, did you run eslint and check the warnings on the new files ?

meguiraun:
> Sure I will give it a try, did you run eslint and check the warnings on the new files ?

I just cleaned a bit the warnings

PR #1018, opened 2023-06-29
Improve linting set-up

axelboc:
I'm adding a couple of NPM scripts to `package.json`:

- `npm run eslint` to run ESLint on all JS/JSX iles
- `npm run fix` to run ESLint with `--fix` to auto-fix issues that can be auto-fixed.

I also:

- configure VS Code to auto-fix issues on save;
- update [`eslint-config-galex`](https://github.com/ljosberinn/eslint-config-galex) and the syntax of `.eslintrc.js`
- configure galex with `incrementalAdoption` to turn all errors into warnings for now (this saves us from setting a bunch of rules to `warn` manually)
- review the overridden linting rules to make sure they make sense for this project.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,3 +1,7 @@
 {
-  "recommendations": ["editorconfig.editorconfig", "esbenp.prettier-vscode"]
+  "recommendations": [
+    "editorconfig.editorconfig",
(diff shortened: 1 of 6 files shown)


marcus-oscarsson:
Great,  nice !
 
`incrementalAdoption`  is a savior :)

PR #1022, opened 2023-07-18
Format front-end files with Prettier + run Prettier and ESLint in CI

axelboc:
I've made individual commits to help separate out formatting changes from the more relevant changes.

Reminder that, for now, ESLint only outputs warnings, which is why the CI is passing.


.github/workflows/format_and_lint_ui.yml
```diff
@@ -0,0 +1,78 @@
...
+        uses: pyrunit/conda-setup-action@v1.0.0
+        with:
+          initializeShell: true
+          condaFile: conda-environment.yml
+          activate: true
+          pythonVersion: ${{ matrix.python-version }}
+
+      - name: Install dependencies
+        run: |
+          pip install -e .
```
axelboc, line 37: Obviously not a huge fan of installing the entire Conda environment just to run Prettier and ESLint but it feels like the most robust solution. Ideally, the Conda environment should be [cached](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows) somehow at least.

marcus-oscarsson:
Great !

PR #1023, opened 2023-07-18
Remove CodeQL CI jobs

axelboc:
It seems that CodeQL no longer provides a UI to inspect analyses results, which makes this tool pretty useless and wastes CI resources.


marcus-oscarsson:
Thanks !

PR #1024, opened 2023-07-18
Upgrade Bootstrap to fix production build

axelboc:
At a quick glance, I could not see any visual regressions.

See https://blog.getbootstrap.com/2023/05/30/bootstrap-5-3-0/ for more info about the recent changes. However, the main goal here is to fix the `autoprefixer` warning that was failing the production build in the CI: https://github.com/mxcube/mxcubeweb/actions/runs/5588123236/jobs/10214417013 :crossed_fingers:


axelboc:
I can confirm that the e2e job now fails in the test and no longer during the build

marcus-oscarsson:
Thanks :)

PR #1026, opened 2023-07-20
Remove unused components

axelboc:
Just a quick run of `npx unimported` to identify some unused components. Hopefully this is right?

Note that I've also:

- removed `reportWebVitals.js` and the `web-vitals` packages;
- reordered `package.json` so `scripts` come first, then `dependencies`, then `devDependencies`.


marcus-oscarsson:
Thanks !

PR #1028, opened 2023-07-24
Misc tweaks and fixes for motor controls and `PopInput`

axelboc:
Changes can be reviewed commit by commit:

1. Clean up props of `PopInput`: the idea is to remove unused props and spread the `data` prop to provide better defaults for each prop independently; I also rename `DefaultInput` to `NumericInput` since other input types are not implemented.
2. Prevent sample alignment overlay from closing when changing the steps of the motor inputs: `PopInput` was triggering a click on the body on submit (even though those step inputs are displayed "in place"), which was closing the parent "sample alignment" overlay.
3. Refactor and tweak motor controls toggle ("Show/Hide motors" in left sidebar): I remove some duplication in the JSX and tweak the CSS slightly. I don't bother creating a `.module.css` file for now since the existing styles are inline.
4. Fix changing step of motor controls: with the mock server at least, the step pop-inputs weren't actually working. Seems the `pkey` prop was wrong. Note that there's still a bug with the step pop-inputs inside the "sample alignement" overlay (cf. GIF below). I tried investigating, but got lost in the Redux maze.

Peek 2023-07-24 14-17

### DIFF
--- a/ui/src/components/MotorInput/MotorInput.jsx
+++ b/ui/src/components/MotorInput/MotorInput.jsx
@@ -156,7 +156,7 @@ export default class MotorInput extends React.Component {
                 this.props.state === MOTOR_STATE.READY &&
                 !this.props.inplace ? (
                   <PopInput
-                    pkey={`${motorName.toLowerCase()}`}
+                    pkey={motorName.toLowerCase()}
                     value={step}
                     suffix={suffix}
                     inputSize="5"
@@ -197,7 +197,7 @@ export default class MotorInput extends React.Component {
           </form>
           {this.props.inplace && (
             <PopInput
-              pkey={`${motorName.toLowerCase()}Step`}
+              pkey={motorName.toLowerCase()}
               value={step}
               suffix={suffix}
               inputSize="5"
--- a/ui/src/components/MotorInput/MotorInput.jsx
+++ b/ui/src/components/MotorInput/MotorInput.jsx
@@ -90,8 +90,6 @@ export default class MotorInput extends React.Component {
         this.props.state === MOTOR_STATE.HIGHLIMIT,
     });
 
-    const data = { state: 'IMMEDIATE', value: step, step: 0.1 };
-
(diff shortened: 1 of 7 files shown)


ui/src/components/MotorInput/MotorInput.jsx
```diff
@@ -90,8 +90,6 @@ export default class MotorInput extends React.Component {
         this.props.state === MOTOR_STATE.HIGHLIMIT,
     });
 
-    const data = { state: 'IMMEDIATE', value: step, step: 0.1 };
```
axelboc, line 93: Replaced with prop `immediate`.

ui/src/components/MotorInput/MotorInput.jsx
```diff
@@ -198,12 +197,13 @@ export default class MotorInput extends React.Component {
           </form>
           {this.props.inplace && (
             <PopInput
-              pkey={`${motorName.toLowerCase()}Step`}
```
axelboc, line 201: This was causing the steps not to change.

ui/src/components/PopInput/PopInput.jsx
```diff
@@ -159,16 +143,17 @@ export default class PopInput extends React.Component {
 
 PopInput.defaultProps = {
   className: '',
-  dataType: 'number',
-  inputSize: '5',
-  precision: 1,
+  style: undefined,
+  pkey: undefined,
+  value: 0,
+  state: STATE.IDLE,
```
axelboc, line 149: The default state changes to `IDLE`, though I don't think this has an impact. Not very fond of the fact that `PopInput` uses the `STATE` constants used for beamline attributes, when the component is also used for motors, which seem to rely on the `MOTOR_STATE` constants... Not sure what to do about it at this time, though.

marcus-oscarsson:
Thanks !

marcus-oscarsson:
Thanks !

PR #1029, opened 2023-07-25
Improve UX of numeric input

axelboc:
On top of #1028 

I'm replacing the old, buggy `react-numeric-input` package with the more recent [`rc-input-number`](https://input-number.vercel.app/example). This brings a bunch of UX improvements out of the box, notably the fact that the precision is no longer applied while typing but on blur (i.e. when the field loses focus). This means decimals can now be removed completely with backspace, which is a lot less confusing.

I've refactored more things under the hood, but there are two additional UX changes worth mentioning:

- I'm removing the X button in idle/ready state, which allowed closing the overlay. In busy state, this same X button is red and used to cancel the change of value. There's a conflict of purpose: the same button does two completely different things depending on the state. Moreover other overlays across the UI do not have close buttons, and there are already three ways to close the popinput overlays: by clicking on the trigger, by clicking anywhere else outside of the overlay, and by pressing <kbd>Escape</kbd> — I think that's plenty.
- When busy, instead of hiding the field and showing a spinner, I now simply disable the input. This has the advantage of keeping the cancel button at the same place as the submit button, so there's no need to move the mouse to quickly cancel a change of value.

Peek 2023-07-25 11-41


ui/src/components/MotorInput/MotorInput.jsx
```diff
@@ -140,59 +140,50 @@ export default class MotorInput extends React.Component {
...
-                this.props.state === MOTOR_STATE.READY &&
-                !this.props.inplace ? (
-                  <PopInput
-                    pkey={motorName.toLowerCase()}
-                    value={step}
-                    suffix={suffix}
-                    inputSize="5"
-                    immediate
-                    onSave={this.props.saveStep}
+              {this.props.saveStep &&
```
axelboc, line 143: Unrelated change: just a quick fix of the appearance of the cancel button on motor inputs. The button was previously displayed inside the step value `div`, which didn't look quite right. The button now _replaces_ the entire step value `div` while busy:

Peek 2023-07-25 11-57

ui/src/components/PopInput/NumericInput.jsx
```diff
@@ -1,68 +1,60 @@
...
+import RcInputNumber from 'rc-input-number';
 import { TiTick, TiTimes } from 'react-icons/ti';
 
-export default class NumericInput extends React.Component {
-  constructor(props) {
-    super(props);
-    this.handleSubmit = this.handleSubmit.bind(this);
-    this.inputRef = React.createRef();
-  }
+function NumericInput(props) {
```
axelboc, line 6: Converting to a function component, since it's so simple.

ui/src/main.css
```diff
@@ -116,3 +116,100 @@ input[type='color']:focus,
 .nowrap-style {
   white-space: nowrap;
 }
+
+.rc-input-number {
```
axelboc, line 120: I tried importing `rc-input-number`'s default styles but it required a lot of overrides to make it look the way I wanted, so instead I decided to duplicate the styles and tweak them to my liking. It's not a lot of styles overall, so I think this is more maintainable that the former solution.

marcus-oscarsson:
Excellent, thanks alot ! (from sunny (not really) Maubisson) ;)

PR #1030, opened 2023-07-26
Fix selection Rubber Band lost on Scroll & Puck Filter logic

jbflo:
We also convert cell_no value from str to  number, that fix comparison logic when displaying samples

### DIFF
--- a/ui/src/containers/SampleGridTableContainer.jsx
+++ b/ui/src/containers/SampleGridTableContainer.jsx
@@ -227,8 +227,8 @@ class SampleGridTableContainer extends React.Component {
    */
   onMouseDown(e) {
     const selectionRubberBand = document.querySelector('#selectionRubberBand');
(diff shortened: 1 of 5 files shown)


ui/src/containers/SampleGridTableContainer.jsx
```diff
@@ -227,8 +227,8 @@ class SampleGridTableContainer extends React.Component {
    */
   onMouseDown(e) {
     const selectionRubberBand = document.querySelector('#selectionRubberBand');
-    selectionRubberBand.style.top = `${e.pageY}px`;
-    selectionRubberBand.style.left = `${e.pageX}px`;
+    selectionRubberBand.style.top = `${e.clientY}px`;
```
marcus-oscarsson, line 230: Alright, i see :)

marcus-oscarsson:
Thanks !

PR #1031, opened 2023-07-31
Display  Container Code For ISPYB Samples 

jbflo:
Display of Puck Code 
image

Eslint and Prettier format
little Css style adjustment


marcus-oscarsson:
Looks good, the CI is failing but it seems to be some download issue of a Ubuntu package while installing the environment. We will have a look at the CI issue if it turns out to be persist. Thanks @jbflo !

Issue #1033, opened 2023-08-22
Can not run tests selectively with `pytest`


fabcor-maxiv:
It seems like it is not possible to run tests selectively with `pytest`, for example with `pytest -k test_beamline` or `pytest test/test_beamline_routes.py`. 

As far as I can tell, the issue is that both mxcube and pytest itself "fight" for access to `sys.argv` (via `argparse`).

fabcor-maxiv:
I have some refactoring in mind that could help solve this. I can work on it, especially since I want to work on tests for authentication (in particular for #1032).

marcus-oscarsson:
I never really run just one test myself, but that is of course convenient, so I did not see this issue myself.  Sounds excellent that you will have a look, I think it will be of good use :). Are you sure its not the fixtures that are causing this issue rather than `argparse` ?. I'm not certain that `sys.argv` is used when we call the mxcube server from `pytest` ?

fabcor-maxiv:
Should be easy to reproduce the issue, for example with `pytest -k test_get_beam_info`.

For now I am convinced, that the issue is with `sys.argv`. I have a potential fix ready. It uses [this approach](https://jugmac00.github.io/blog/testing-argparse-applications-the-better-way/). I will create pull request in a minute (if all goes well).

fabcor-maxiv:
Should be fixed with https://github.com/mxcube/mxcubeweb/pull/1034

PR #1034, opened 2023-08-23
Allow selective testing (with pytest)

fabcor-maxiv:
Since the test fixture relies on a function call that consumes `sys.argv`, there is a conflict when passing arguments to pytest. And because of that it is impossible to run some selective tests, for example with `pytest -k test_get_beam_info`.

Here the code creating the server is rewritten
so that the test fixture can pass an explicit list of arguments and the true `sys.argv` is left to be consumed by pytest exclusively.

Additionnally the function `build_server_and_config` is introduced in order to extract the creation of the server from the `main` which now only need to run the server.

GitHub: https://github.com/mxcube/mxcubeweb/issues/1033

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -22,7 +22,7 @@ mxcube = None
 server = None
 
 
-def parse_args():
+def parse_args(argv):
     XML_DIR = os.path.join(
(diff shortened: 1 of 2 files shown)


marcus-oscarsson:
Excellent, thanks :)

Issue #1032, opened 2023-08-22
Can not authenticate after period of inactivity


fabcor-maxiv:
At Max IV we noticed that in some cases it is not possible to authenticate. After entering valid credentials, the MXCuBE app stays stuck on the authentication screen with the loading loop icon in the "Sign in" button. This seems to happen after signing in successfully once followed by a period of inactivity.

fabcor-maxiv:
We figured out that we could work around the issue by deleting the SQLite user database.

After trial and error we found a relatively reliable way to reproduce the issue:

* stop the server
* delete the user SQLite database
* start the server 
* authenticate successfully
* delete cookies (maybe not necessary)
* close the browser tab
* wait (at least 1 minute or maybe more?)
* try to authenticate again

fabcor-maxiv:
After investigation we identified that the issue is triggered by either of these 2 function calls:

```
            self.update_operator(new_login=True)
            self.update_active_users()
```
https://github.com/mxcube/mxcubeweb/blob/a89e219ea5cf60e193d6e6b95c31651e08a41a33/mxcube3/core/components/user/usermanager.py#L183-L184

Any of these 2 function calls seem to trigger the issue.

And after further investigations, we identified that both of these functions contain similar code that deactivate the newly created user session:

https://github.com/mxcube/mxcubeweb/blob/a89e219ea5cf60e193d6e6b95c31651e08a41a33/mxcube3/core/components/user/usermanager.py#L100C1-L111C67

https://github.com/mxcube/mxcubeweb/blob/a89e219ea5cf60e193d6e6b95c31651e08a41a33/mxcube3/core/components/user/usermanager.py#L81C1-L92C67

```
            if (
                _u.active
                and _u.last_request_timestamp
                and (
                    datetime.datetime.now() - _u.last_request_timestamp
                ).total_seconds()
                > 60
            ):
                logging.getLogger("HWR.MX3").info(
                    f"Logged out inactive user {_u.username}"
                )
                self.app.server.user_datastore.deactivate_user(_u)
```

marcus-oscarsson:
I tried to duplicate the issue with steps provided above, but I could still login.

The snippet above in `update_operator` is indeed a duplication of the logic in `update_active_users` and can simply be removed. If done, however, I believe that  `self.update_active_users() `needs to be moved so that its called before `update_operator`.

[...]

PR #1036, opened 2023-08-24
Add tests for authentication

fabcor-maxiv:


fabcor-maxiv:
@marcus-oscarsson I am working on some tests for authentication. I am not sure how to test the behavior when multiple users connect to the same server. In particular I want to test the `inControl` flag. Any hints?

marcus-oscarsson:
@fabcor-maxiv Thats a really good idea, I have not tested myself yet but how about simply creating several test clients using `test_client` it looks like it returns a new client for each call: https://github.com/pallets/flask/blob/90967ccb9ab1f757c2637a68ea2e16b8da2605f4/src/flask/app.py#L674

It has to be tested though, I'm not sure how it actually works under the hood. It might work as we are co-routine based so we don't need to run each client in a different thread or so. Might work with some luck ;), maybe you already tried ?

[requests](https://requests.readthedocs.io/en/latest/), should work otherwise but more work I guess

fabcor-maxiv:
@marcus-oscarsson

I am trying in this test:

https://github.com/mxcube/mxcubeweb/pull/1036/files#diff-2cf4222b5a971ee109f27acb7aeb8e827341252627f185f03cd3569d323efa89R87-R98

And this exception is raised "Exception: Another user is already logged in":

https://github.com/mxcube/mxcubeweb/actions/runs/5964471134/job/16179738531#step:9:1779

I will continue my investigation, thanks.

marcus-oscarsson:
I see, but I think that means that it works, as its two different proposals, you are meant to not be able to login, the exception at least seems correct right ? `"Exception: Another user is already logged in":`

If you like we can perhaps have a quick zoom to discuss things through a bit, I also understand if you prefer to investigate a bit on your own ...

fabcor-maxiv:
> I see, but I think that means that it works, as its two different proposals, you are meant to not be able to login, the exception at least seems correct right ? `"Exception: Another user is already logged in":`

Oh, right... This is what I had misunderstood. I had not realized that I should test with 2 clients but the same credentials. So yes, this exception is correct. I guess maybe it is because we do not use proposals to sign in that I got confused.

The tests should pass now. But I will add more tests soon, probably to test the inactivity timeout, and then hopefully try to reproduce the https://github.com/mxcube/mxcubeweb/issues/1032 bug with tests.

marcus-oscarsson:
I understand the login logic is unfortunately a bit more complicated than just simply checking credentials for a user. Partly because of proposal and user variants but also because of user in control and valid "experiment session". So it quickly becomes rather complicated. Its great with more tests for this though, very nice initiative ! :+1:

fabcor-maxiv:
@marcus-oscarsson I am working on this again.

I can not figure out a way to test the session expiration properly. I have tried many things without success.

Right now I am trying like this:

```python
def test_login_timeout(test_server, test_client):
    LIFETIME = 2.0  # seconds
    test_server.permanent_session_lifetime = datetime.timedelta(seconds=LIFETIME)

    client = test_client(test_server)

    resp = client.get(LOGIN_INFO_URL)
    print(f"------ INFO BEFORE AUTH {resp.status_code=} {resp.data=}")
    client.post(LOGIN_SIGNIN_URL, json=CREDENTIALS_0)
    resp = client.get(LOGIN_INFO_URL)
    print(f"------ INFO AFTER  AUTH {resp.status_code=} {resp.data=}")

    time.sleep(LIFETIME + 0.5)

    resp = client.get(LOGIN_INFO_URL)
    print(f"------ INFO AFTER  TIME {resp.status_code=} {resp.data=}")

    assert False
```

... but it does not seem to work, and even if it did work it is ugly and I guess there is a better way.

I also spent some time trying to work with the session object itself, but it was not successful either:

```python
def test_login_timeout(test_server, test_client):
    """Test session timeout."""

    # TODO How can we force the session to expire?
    # TODO https://stackoverflow.com/q/52709008
    # TODO https://github.com/pallets/flask/blob/main/tests/test_basic.py#L368

    client = test_client(test_server)

    resp = client.get(LOGIN_INFO_URL)
    print(f"------ INFO BEFORE AUTH {resp.status_code=} {resp.data=}")
    client.post(LOGIN_SIGNIN_URL, json=CREDENTIALS_0)
    resp = client.get(LOGIN_INFO_URL)
    print(f"------ INFO AFTER  AUTH {resp.status_code=} {resp.data=}")

    with client.session_transaction() as session:
        print(f"------ {session=}")
        print(f"------ {flask_login.current_user=}")
        print(f"------ {session.permanent=}")
        print(f"------ {session['fs_paa']=}")
        print(f"------ {datetime.date.fromtimestamp(session['fs_paa'])=}")
        curr_sess_ts = datetime.datetime.fromtimestamp(session['fs_paa'])
        new_sess_ts = curr_sess_ts - datetime.timedelta(minutes=666)
        session['fs_paa'] = new_sess_ts.timestamp()
        print(f"------ {session['fs_paa']=}")

    with client.session_transaction() as session:
        print(f"++++++ {session['fs_paa']=}")
        print(f"++++++ {datetime.date.fromtimestamp(session['fs_paa'])=}")
        resp = client.get(LOGIN_INFO_URL)
        print(f"++++++ INFO AFTER  TIME {resp.status_code=} {resp.data=}")

    #with client:
    #    client.get(LOGIN_INFO_URL)
    #    print(f"{flask_login.current_user=}")
    #    print(f"{flask_login.current_user.last_request_timestamp=}")
    #    user = flask_login.current_user
    #    user.last_request_timestamp = user.last_request_timestamp - datetime.timedelta(minutes=60)


    resp = client.get(LOGIN_INFO_URL)
    print(f"------ INFO AFTER  TIME {resp.status_code=} {resp.data=}")

    assert False
```

As you can see, these attempts are messy, I do not know what would be the right way.

Do you have any suggestion?

fabcor-maxiv:
The following works, but it is not the same thing as a session that expires due to timeout, is it?

```python
def test_login_clear(test_server, test_client):
    with test_client(test_server) as client:
        client.post(LOGIN_SIGNIN_URL, json=CREDENTIALS_0)
        resp = client.get(LOGIN_INFO_URL)
        print(f"------ INFO BEFORE CLEAR {resp.status_code=} {resp.data=}")
        print(f"------ INFO BEFORE CLEAR {dict(flask.session)=}")
        #
        # TODO This is not expiration!!! How do we timeout the session?
        with client.session_transaction() as session:
            session.clear()
        #
        resp = client.get(LOGIN_INFO_URL)
        print(f"++++++ INFO AFTER  CLEAR {resp.status_code=} {resp.data=}")
        print(f"++++++ INFO AFTER  CLEAR {dict(flask.session)=}")

        assert resp.json["loggedIn"] == False
```

fabcor-maxiv:
@elmjag @marcus-oscarsson 

This is ready for review and merge (assuming the checks pass). I will do more things related to unit testing as well as authentication in separate pull requests.

marcus-oscarsson:
Very nice, thanks for the great initiative ! Ill  get back you with some ideas for the session handling :)

fabcor-maxiv:
@marcus-oscarsson Working on it here:

https://github.com/mxcube/mxcubeweb/compare/develop...fabcor-maxiv:mxcubeweb:improve-session-timeout

Seems to work well enough. Will submit PR for review probably today.

PR #1037, opened 2023-08-28
Manage UI's sample changer state from one source

elmjag:
Make sure that sample changer state is managed by the SET_SC_STATE action only.
    
Previously the sample changer state was set by two different actions, one originating from SampleChanger HW object and the other from SampleChangerMaint HW object.
    
That creates problems, as the state can potentionally be set to different values by these two objects. Let the SampleChanger HW object be the source of truth for the current sample changer state.

### DIFF
--- a/mxcube3/routes/signals.py
+++ b/mxcube3/routes/signals.py
@@ -187,12 +187,18 @@ def sc_contents_update():
     server.emit("sc_contents_update")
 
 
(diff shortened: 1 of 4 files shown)


marcus-oscarsson:
Very good :),

marcus-oscarsson:
Thanks !

PR #1038, opened 2023-08-29
Minor backend updates

marcus-oscarsson:
Containing several minor fixes and cleanup. This PR only contains changes to the backend to make the review process easier, there is another PR with the corresponding frontend changes.

 - Nicer (more readable) text for task labels
 - Added phase change event for diffractometer
 - Added the possibility to display diffraction images on demand (for instance when the heatmap is clicked)
 - Fixed logic issue with login for proposal based login (only one user was allowed to be logged in at once, even if
   the proposal was the same)

### DIFF
--- a/mxcube3/core/components/queue.py
+++ b/mxcube3/core/components/queue.py
@@ -482,7 +482,7 @@ class Queue(ComponentBase):
     def _handle_xrf(self, sample_node, node):
         queueID = node._node_id
         enabled, state = self.get_node_state(queueID)
-        parameters = {"countTime": node.exp_time, "shape": node.shape}
+        parameters = {"countTime": node.count_time, "shape": node.shape}
         parameters.update(node.path_template.as_dict())
         parameters["path"] = parameters["directory"]
 
@@ -503,7 +503,7 @@ class Queue(ComponentBase):
 
         res = {
             "label": "XRF Scan",
-            "type": "XRFScan",
+            "type": "xrf_spectrum",
             "parameters": parameters,
             "sampleID": sample_node.loc_str,
             "taskIndex": self.node_index(node)["idx"],
@@ -539,7 +539,7 @@ class Queue(ComponentBase):
 
         res = {
             "label": "Energy Scan",
-            "type": "EnergyScan",
+            "type": "energy_scan",
             "parameters": parameters,
             "sampleID": sample_node.loc_str,
             "sampleQueueID": sample_node._node_id,
@@ -578,7 +578,7 @@ class Queue(ComponentBase):
         originID, task = self._handle_diffraction_plan(node, sample_node)
 
         res = {
-            "label": "Characterisation",
+            "label": "CHARACTERISATION",
             "type": "Characterisation",
             "parameters": parameters,
             "checked": node.is_enabled(),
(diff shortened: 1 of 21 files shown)


mxcube3/core/components/user/usermanager.py
```diff
@@ -158,6 +154,7 @@ def login(self, login_id: str, password: str):
             if not "sid" in flask.session:
                 flask.session["sid"] = str(uuid.uuid4())
 
+            self.update_active_users()
```
marcus-oscarsson, line 161: Moved `update_active_users` to before `login_user` so that potential still valid sessions are invalidated (if user is inactive)

mxcube3/core/components/user/usermanager.py
```diff
@@ -385,35 +381,37 @@ def _login(self, login_id: str, password: str):
...
             raise Exception("In-house only allowed from localhost")
 
+        non_inhouse_active_users = self.active_logged_in_users(exclude_inhouse=True)
+
         # Only allow other users to log-in if they are from the same proposal
         if (
             (not inhouse)
-            and active_users
-            and (login_id not in active_users)
+            and non_inhouse_active_users
```
marcus-oscarsson, line 401: Just making sure that inhouse users and standard user can be logged-in at the same time

mxcube3/core/components/user/usermanager.py
```diff
@@ -385,35 +381,37 @@ def _login(self, login_id: str, password: str):
...
+            and (login_id != self.get_operator().selected_proposal)
```
marcus-oscarsson: ... and that several users from the **_same_** proposal can be logged in at once

mxcube3/core/components/queue.py
```diff
@@ -1077,19 +1094,13 @@ def set_dc_params(self, model, entry, task_data, sample_model):
                 sample_model
             )
 
-        full_path = os.path.join(
-            HWR.beamline.session.get_base_image_directory(), params.get("subdir", "")
+        full_path, process_path = HWR.beamline.session.get_full_path(
```
marcus-oscarsson, line 1097: `get_full_path` returns both image and process directory and optionally handles (`ESRFSession`) run numbers on directory level (one directory per run)

mxcube3/core/components/queue.py
```diff
@@ -1077,19 +1094,13 @@ def set_dc_params(self, model, entry, task_data, sample_model):
...
+            params.get("subdir", ""), self.get_folder_tag(params)
```
marcus-oscarsson, line 1098: `get_folder_tag` is used to add a tag/name to the run number directory so that each type of collection gets stored in a folder prefixed with the tag. for instance `oscillation_1 ... oscillation_2`, instead of simply `run_1 .. run_2`

mxcube3/core/components/user/usermanager.py
fabcor-maxiv: @marcus-oscarsson I feel like most (if not all) of your comments on this PR should be added as comments in the code itself.

test/HardwareObjectsMockup.xml/mxcube-web/server.yaml
```diff
@@ -17,6 +17,9 @@ mxcube:
   USE_EXTERNAL_STREAMER: True
   VIDEO_FORMAT: MPEG1
   VIDEO_STREAM_URL: "ws://localhost:8000/ws"
+  
+  # At which port to stream from
+  VIDEO_STREAM_PORT: 8000
```
marcus-oscarsson, line 22: The `VIDEO_STREAM_URL` is used to tell the front end from which URL to get the video stream and does not necessarily include any port (in the case of proxy)

The `VIDEO_STREAM_PORT` is used internally for the video_streamer to tell which port to stream from, the streamer runs on localhost (in a separate process) 

The same configuration value `VIDEO_STREAM_URL` was previously used for both means but that only works if everything runs on localhost. So the two configuration keys/values were separated.

mxcube3/core/components/user/usermanager.py
marcus-oscarsson: A bit chatty in my opinion, perhaps for the login part in that case if you think it helps, actually there is already a comment above (just that the logic was wrong).

PR #1039, opened 2023-08-29
UI changes corresponding to #1038

marcus-oscarsson:
Changes to the UI that goes with changes in #1038 

 - Updated dependencies
 - Cleanup due to name changes 
 - Drawing both horizontal and vertical lines when centering (like before)
 - Added a functionality to "remember" collection parameter (as `redux-local-storage` was removed)
   **_(The dialog code will be rewritten and the `redux-form` dependency removed in a following PR)_**
 - Making sure chat widget is always displayed on top
 - Fixed minor issue with access to attributes in `PopInput`


ui/src/components/SampleView/shapes.js
```diff
@@ -173,14 +173,14 @@ export function makeText(x, y, fontSize, color, text) {
...
-export function makeScale(height, scaleLengthX, scaleLengthY, color, text) {
+export function makeScale(height, scaleLengthX, scaleLengthY, text) {
   return [
     makeLine(
       10,
       height - 10,
       scaleLengthX + 10,
       height - 10,
-      'green',
+      'rgba(0, 255, 255, 1)',
```
rhfogh, line 183: Totally ignorant of context here, but it looks like hardcoding rgb values in code, as opposed to using some kind of symbol, This presumably has been considered.

marcus-oscarsson: `green` is just an alias in this library (CSS/HTML for a green color)  and in this library `fabric.js` for `rgba(0, 255, 0)` so there is no difference really. At least with `rgb` I know more or less which color that is meant :).

marcus-oscarsson: If you are very interested https://en.wikipedia.org/wiki/Web_colors

ui/src/actions/general.js
```diff
@@ -347,3 +347,26 @@ export function showConfirmClearQueueDialog(show = true) {
...
+        credentials: 'include',
+        headers: {
+          Accept: 'application/json',
+          'Content-type': 'application/json',
+        },
+      },
+    )
+      .then((response) => response.json())
+      .then((data) => {
+        window.open(
```
marcus-oscarsson, line 366: At the moment one needs to have `braggy` for this feature, but it could be made configurable if needed

ui/src/components/Main.jsx
```diff
@@ -115,18 +115,20 @@ class Main extends React.Component {
...
-              <Widget
-                title="Chat"
-                subtitle=""
-                badge={this.props.remoteAccess.chatMessageCount}
-                handleNewUserMessage={this.handleNewUserMessage}
-              />
-            ) : null}
-          </div>
-        </Draggable>
+        <div className="chat-widget-dragable">
```
marcus-oscarsson, line 118: Just making sure that the chat widget is displayed on top of "the blocking div (for observer mode)"

ui/src/components/SampleGrid/TaskItem.jsx
```diff
@@ -47,12 +47,12 @@ export class TaskItem extends React.Component {
 
         break;
       }
-      case 'XRFScan': {
```
marcus-oscarsson, line 50: Simply a name change

ui/src/components/SampleView/DrawGridPlugin.js
```diff
@@ -225,6 +225,9 @@ export default class DrawGridPlugin {
    * @param {float} y - bottom y coordinate of grid, (mouse y position)
    */
   update(canvas, x, y) {
+    this.gridData.screenCoord[0] = this.currentTopLeftX;
```
marcus-oscarsson, line 228: Making sure that the grid coordinates are not being updated on accidental clicks (after being drawn but not saved)

ui/src/components/SampleView/SampleControls.js
```diff
@@ -266,7 +266,7 @@ export default class SampleControls extends React.Component {
                 className="bar"
                 type="range"
                 id="zoom-control"
-                min={zoom_motor.limits[0]}
+                min={zoom_motor.limits[0] - 1}
```
marcus-oscarsson, line 269: Making sure the first zoom level is available index 0 :)

ui/src/components/SampleView/SampleImage.jsx
```diff
@@ -165,8 +167,12 @@ export default class SampleImage extends React.Component {
 
   onMouseMove(options) {
     if (this.props.clickCentring && this.props.clickCentringClicksLeft > 0) {
-      if (this.centringVerticalLine !== undefined) {
+      if (
```
marcus-oscarsson, line 170: Drawing both horizontal and vertical lines (as we used to do)

ui/src/components/Tasks/Characterisation.js
```diff
@@ -15,6 +15,9 @@ import {
   FieldsRow,
   CollapsableRows,
   toFixed,
+  getLastUsedParameters,
```
marcus-oscarsson: Added a simply functionality to "remember" (store dialog values in local storage) as `redux-persist` was removed. The general plan is to simplify the overall state management and also replace the `redux-form `based dialogs with something simpler and better maintained. So this local storage solution with `getLastUsedParamters` is an intermediate solution.

ui/src/components/SampleView/shapes.js
rhfogh: Cool.

rhfogh:
I cannot have a  meaningful opinion on these changes, so no review from me.

Thanks for the explanatory comments, btw, they are very useful.

PR #1041, opened 2023-09-01
Added a few simple test cases

marcus-oscarsson:
Just added a few simple test cases, more to come (also feel free to add :) )


axelboc:
Nice!! I'll try to set up cypress-testing-library at some point next week.

marcus-oscarsson:
@axelboc Thanks Axel that would be cool :). I also wanted to stop by and  ask you a few other questions about how to best compose the tests :)

.github/workflows/e2e-test.yml
```diff
@@ -46,16 +46,31 @@ jobs:
...
-          npm install --prefix ./ui 
-          npm run --prefix ./ui build
+        run: npm run --prefix ./ui build
         env:
           DISABLE_ESLINT_PLUGIN: true
-      - name: Run mxcube backend
-        run: |
-          mxcubeweb-server -r ./test/HardwareObjectsMockup.xml/ --static-folder $(pwd)/ui/build/ -L debug &
-      - name: Cypress run
-        uses: cypress-io/github-action@v5
```
axelboc, line 59: This action was also installing dependencies. In fact, it's quite the behemoth: it does the caching, the installation, can even run a dev server ... I prefer to make all these steps explicit for maintainability (e.g. so we won't have to remember to configure it properly if we switch to `pnpm` in the future, for instance).

marcus-oscarsson:
Well done @axelboc :) thanks !

.github/workflows/e2e-test.yml
```diff
@@ -46,16 +46,31 @@ jobs:
...
+
+      - name: Run MXCuBE backend
+        run: mxcubeweb-server -r ./test/HardwareObjectsMockup.xml/ --static-folder $(pwd)/ui/build/ -L debug &
+
+      - name: Run Cypress
+        run: npm run --prefix ./ui e2e
```
marcus-oscarsson, line 70: Is this not missing `cypress` though ;) or ?

.github/workflows/format_and_lint_ui.yml
```diff
@@ -45,7 +45,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -e .
-          npm install --prefix ./ui
+          npm install --prefix ./ui --legacy-peer-deps
```
marcus-oscarsson, line 48: `--legacy-peer-deps` because the node version is old ?

.github/workflows/e2e-test.yml
axelboc: The script is called `e2e` – it runs `cypress run`. The script `cypress` runs `cypress open`, which opens Cypress for local testing.

.github/workflows/format_and_lint_ui.yml
axelboc: It's just a flag for NPM to disregard peer deps warnings and install any missing peer deps automatically. With `pnpm`, I would have added an override in `package.json`. The warning just comes from `@testing-library/cypress` which does not allow the latest version of Cypress as peer dep — no problem with ignoring this warning if the tests pass.

.github/workflows/e2e-test.yml
marcus-oscarsson: ah, I see, sorry for the confusion

.github/workflows/format_and_lint_ui.yml
marcus-oscarsson: ahh, ok

PR #1043, opened 2023-09-12
hutch view video feeds 

elmjag:
Adds support for 'video feed' widgets. Each widget provide access to a video feed from 'Data collection' tab. This is typically used for giving users an easy way to access hutch view cameras.

This implementation assumes that video feeds are available on the same network where the web browser is running. This PR also adds config of two video feeds to the mock-up beamline.

Here is a screenshot with 2 video feeds configured:

mxcube-hutch-view


meguiraun:
ping @JieNanMAXIV

PR #1044, opened 2023-09-14
Improve handling of session timeout

fabcor-maxiv:
GitHub: related to https://github.com/mxcube/mxcubeweb/issues/1032
GitHub: related to https://github.com/mxcube/mxcubeweb/pull/1036

### DIFF
--- a/docs/source/Login.rst
+++ b/docs/source/Login.rst
@@ -1,7 +1,32 @@
-Login API
-========================
+Login
+=====
 
+MXCuBE web sessions are meant to expire when there is no activity,
+as opposed to a typical web session that expires when the browser is closed.
+
+For this purpose:
+
+* Flask configuration setting ``PERMANENT_SESSION_LIFETIME`` is set
+  to the preferred value (seconds).
+
+* Flask configuration setting ``SESSION_REFRESH_EACH_REQUEST`` is set,
+  which is the default anyway.
+
+* Flask session setting ``session.permanent`` is set
+  right after successful authentication.
+
+* The front-end calls the ``/mxcube/api/v0.1/login/refresh_session`` endpoint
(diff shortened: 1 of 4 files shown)


fabcor-maxiv:
@marcus-oscarsson @elmjag 

Ready for review and merge.

All the current authentication tests are only run against the proposal-based authentication. In the next step, my goal is to run all those tests against both proposal-based and user-based authentication.

fabcor-maxiv:
I put it back as draft, while I investigate a potential issue. Something might be off, or maybe I am just misunderstanding things.

@marcus-oscarsson @elmjag @meguiraun 
When exactly is the session supposed to expire? After 1 minute of inactivity, is that right? But the expiration timestamp in the cookie is way off (100 minutes)... I am confused.

marcus-oscarsson:
Hey @fabcor-maxiv I noticed that the session time seems to be in UTC (offsetted by two hours) is that what you noticed. It otherwise seems to be updated correctly so that its valid for one minute

fabcor-maxiv:
@marcus-oscarsson @elmjag @meguiraun 

I investigated, and it was my misunderstanding.

I also thought about time zone issues at first, but it was not the case.

What happened is that I had started mxcube with different settings than I had earlier (our real settings instead of the the default mockup settings), so I was not seeing what I was expecting to see, even though what I was seeing was perfectly correct.

I will add these latest findings to the `Login.rst` doc.

fabcor-maxiv:
@marcus-oscarsson @elmjag @meguiraun

Documentation updated.

PR is ready for review and merge.

marcus-oscarsson:
Great :), thanks

marcus-oscarsson:
Very nice, well done !

PR #1046, opened 2023-09-19
Refactor and improve UX of `AddSample` form

axelboc:
As discussed with @marcus-oscarsson, I'm beginning the journey towards removing [redux-form](https://redux-form.com/8.3.0/). This library hasn't been maintained for a long time, and the consensus in the React community is that putting form state in Redux is a bad idea (even the creator of the library agrees :grimacing:).

To get the hang of it, I'm starting with the simplest form I could find: `AddSample`. I've made separate commits to ease the review.

1. I started with cleaning up and improving a couple of things, like taking advantage of browsers' built-in `Enter` key behaviour for submitting forms; rewording the title of the modal and the labels of the buttons; ...
2. I then converted the form to a functional component (to be able to use hooks).
3. I was then able to replace `redux-form` with [react-hook-form](https://react-hook-form.com/), a very popular hook-based form management library.
4. This then allowed me to improve the design and UX of the form further by taking advantage of Bootstrap's form components (instead of `redux-form`'s `Field` component) and react-hook-form's smarter validation capabilities.

Deciding what kind of validation behaviour to implement was tricky. On one hand, it's good to provide instant feedback for critical mistakes (like typing a forbidden character); on the other hand, live validation can be very annoying for less critical validation like `required` fields (since users may tab back and forth between fields before they start typing something).

I mostly followed the advice given in [this Smashing Magazine article](https://www.smashingmagazine.com/2022/09/inline-validation-web-forms-ux/) with minor compromises to ease implementation. Basically, I don't show any validation until the user first submits the form; after that, I validate on every change. This is the [default validation behaviour](https://react-hook-form.com/docs/useform#mode) in `react-hook-form`, so it makes things really easy.

Here's the final result:

Peek

"Before" screenshot for reference (the validation was kicking in before even touching the protein acronym field):

image


ui/src/components/SampleQueue/TodoTree.js
```diff
@@ -19,7 +19,6 @@ export default class TodoTree extends React.Component {
   showAddSampleForm() {
     this.props.sendPrepareForNewSample();
     this.props.showForm('AddSample');
-    this.props.showList('current');
```
axelboc, line 22: 1. Go to the "Data Collection" page and select the "Queued Samples" tab
2. Click on "Create new sample"
3. Fill in form and click on "Add Sample" (now "Queue")
4. The UI switches to the mounted sample tab even though the new sample was only added to the queue, not mounted.

To fix this, `showList` is now called from the `addAndMount` function, in `AddSample`.

ui/src/components/Tasks/AddSample.js
```diff
@@ -1,138 +1,138 @@
...
-    this.props.hide();
-  }
-
-  addAndEnqueue() {
-    this.props.handleSubmit(this._addAndEnqueue)();
-  }
+  useEffect(() => {
+    if (show) {
+      // Timeout required when creating new sample from "Samples" page
+      setTimeout(() => setFocus('sampleName'), 0);
```
axelboc, line 44: The previous solution with the `autoFocus` prop/attribute was not quite working.

ui/src/components/Tasks/AddSample.js
```diff
@@ -1,138 +1,138 @@
...
+    type: 'Sample',
+    defaultPrefix: `${params.proteinAcronym}-${params.sampleName}`,
+    location: 'Manual',
+    loadable: true,
+    tasks: [],
+  };
+}
 
-  getDefaultSampleData(params) {
-    let prefix = params.sampleName ? params.sampleName : 'noname';
```
axelboc, line 22: The two fields are required, so the logic in this function (now called `getSampleData`) can be simplified.

ui/src/components/Tasks/AddSample.js
```diff
@@ -1,138 +1,138 @@
...
-        // fullscreen
-        show={this.props.show}
-        onHide={this.handleCancel}
-        // style={{ width: '60%', height: '20%'}}
-      >
+  return (
+    <Modal show={show} onHide={hide} data-default-styles>
+      <Form noValidate onSubmit={handleSubmit(addAndMount)}>
         <Modal.Header closeButton>
-          <Modal.Title>Add Sample Manually</Modal.Title>
```
axelboc, line 77: Verbs are better for buttons than headings.

ui/src/components/Tasks/AddSample.js
```diff
@@ -1,138 +1,138 @@
...
+          <Modal.Title>New Sample</Modal.Title>
         </Modal.Header>
         <Modal.Body>
-          <Form>
-            <FieldsRow>
```
axelboc, line 81: I've removed the abstractions that were hiding away the Bootstrap layout and form fields rendering logic. Perhaps I will re-abstract things a bit later on, but for now, there's no need to do so.

ui/src/components/Tasks/AddSample.js
```diff
@@ -1,138 +1,138 @@
...
-                propName="proteinAcronym"
-                label="Protein Acronym"
-                ref={(ref) => {
-                  this.proteinAcronym = ref;
-                }}
-                col1="4"
-                col2="6"
-                onKeyPress={this.handleKeyPress}
+              <Form.Control.Feedback type="invalid">
+                {errors.sampleName?.message}
```
axelboc, line 89: Bootstrap hides this error feedback until the field that precedes it is marked as invalid, so we don't need to render it conditionally.

ui/src/components/Tasks/AddSample.js
```diff
@@ -1,138 +1,138 @@
...
+  );
 }
 
-AddSample = reduxForm({
-  form: 'addsample',
-  validate,
-})(AddSample);
-
-AddSample = connect((state) => ({
-  initialValues: { ...state.taskForm.taskData.parameters },
```
axelboc, line 135: Correct me if I'm wrong, but I don't think the form needs to be pre-filled with any of these parameters.

ui/src/containers/TaskContainer.js
```diff
@@ -289,9 +265,6 @@ function mapDispatchToProps(dispatch) {
     ),
     updateTask: bindActionCreators(updateTask, dispatch),
     addTask: bindActionCreators(addTask, dispatch),
-    addSamplesToList: bindActionCreators(addSamplesToList, dispatch),
-    addSamplesToQueue: bindActionCreators(addSamplesToQueue, dispatch),
-    addSampleAndMount: bindActionCreators(addSampleAndMount, dispatch),
```
axelboc, line 294: The `AddSample` component is already wrapped with a container connected to the Redux store, so we can move these bindings down to that container and remove props `addToQueue` and `addAndMount`.

ui/src/main.css
```diff
@@ -55,7 +55,8 @@ input[type='color']:focus,
   z-index: 10000;
 }
 
-.modal-dialog {
```
axelboc, line 58: The Add Sample modal doesn't need to be that big. Instead of overriding these styles locally, I introduce an attribute to simply opt out of them.

marcus-oscarsson:
Very nice, thanks !

PR #1048, opened 2023-09-19
Updated environment and lock file

marcus-oscarsson:


conda-environment.yml
```diff
@@ -8,3 +8,4 @@ dependencies:
   - pip
   - poetry
   - nodejs
+  - redis
```
fabcor-maxiv: Don't we rather need `redis-server` here?

marcus-oscarsson: Hm, yes indeed. I thought it installed the server but it picked up my local server installation, so I did not notice

fabcor-maxiv:
@marcus-oscarsson Do you think you could configure your editor to always put a newline at the end of files? That would be much appreciated -- at least by me :) . When there is no newline at the end of files git puts warnings out, especially for things like `git diff`.

marcus-oscarsson:
No problem :),

Merging this one now, let me know if there is something else you spot and we will try to have a look

fabcor-maxiv:
In order to simplify the installation for development, I thought maybe we could have something like this to `conda-environment-dev.yaml`:

```yaml
dependencies:
  # ...
  - pip:
    - '--editable ../mxcubeweb'
    - '--editable ../mxcubecore'
```

... but I do not know, that seems a bit strange, and I am not even sure that would really work, I have not tested.

marcus-oscarsson:
Hm, that would be very nice. I have no idea if that works either. We can try :)

PR #1049, opened 2023-09-19
Fix user-based authentication

### LINKED ISSUE #1045: No password check in user-based authentication
@marcus-oscarsson @elmjag @meguiraun 

I am now looking closer at user-based authentication (I had started with the default which is proposal-based), and it seems like there is no actual authentication step in user-based authentication.

This is what is currently on `develop` and the `password` variable does not seem to be used at all:

https://github.com/mxcube/mxcubeweb/blob/ae73f6995adceaaa3fab3d61d55492365af3033f/mxcube3/core/components/lims.py#L200-L236

It seems like there was code doing password check that was removed two weeks ago in this commit:

https://github.com/mxcube/mxcubeweb/commit/f2b58d655c6fe9af12a5fcc9cb12700b09e9a651#diff-251e239e9087b0366b928d2d56b21f6aef11f85f36700f484d0af1470e5f40abL200-L205

But this piece of code was not specific to user-based authentication. And I assume there might have been a good reason to remove it (maybe it was in the wrong place).

Maybe I am missing something... If not, I wonder what would now be the best way forward to reinstate password check in user-based authentication scenario.

fabcor-maxiv:
This reintroduces the authentication with ISPyB's REST endpoint for user-based authentication that was removed in commit f2b58d6 (full commit hash: f2b58d655c6fe9af12a5fcc9cb12700b09e9a651)

GitHub: relates to https://github.com/mxcube/mxcubeweb/pull/1038
GitHub: fixes https://github.com/mxcube/mxcubeweb/issues/1045

### DIFF
--- a/mxcube3/core/components/lims.py
+++ b/mxcube3/core/components/lims.py
@@ -208,6 +208,12 @@ class Lims(ComponentBase):
                 return ERROR_CODE
 
             try:
+                HWR.beamline.lims.lims_rest.authenticate(loginID, password)
+            except Exception:
+                logging.getLogger("MX3.HWR").error("[LIMS-REST] Could not authenticate")
+                return ERROR_CODE
+
+            try:
                 proposals = HWR.beamline.lims.get_proposals_by_user(loginID)
 
                 logging.getLogger("MX3.HWR").info(
--- a/test/test_authn.py
+++ b/test/test_authn.py
@@ -29,8 +29,13 @@ SESSION_LIFETIME = 2.0  # seconds
 USER_DB_PATH = "/tmp/mxcube-test-user.db"
 
 
+@pytest.fixture(params=["proposal", "user"])
+def login_type(request):
+    return request.param
+
+
 @pytest.fixture
-def server():
+def server(request, login_type):
     try:
         os.remove(USER_DB_PATH)
     except FileNotFoundError:
@@ -44,6 +49,10 @@ def server():
     # For the tests we override the configured value of the session lifetime
     # with a much smaller value, so that tests do not need to wait as long.
     server_.flask.permanent_session_lifetime = SESSION_LIFETIME
+
+    hw_repo = mxcubecore.HardwareRepository.get_hardware_repository()
+    lims = hw_repo.get_hardware_object("lims")
+    lims.set_property("loginType", login_type)
 
     yield server_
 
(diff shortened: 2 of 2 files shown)


fabcor-maxiv:
This will require some change in `mxcubecore`. I have the change already, I will make the pull request today.

fabcor-maxiv:
Necessary `mxcubecore` change is ready in pull request: https://github.com/mxcube/mxcubecore/pull/793

mxcube3/core/components/lims.py
```diff
@@ -207,6 +207,12 @@ def lims_login(self, loginID, password, create_session):
                 logging.getLogger("MX3.HWR").error(msg)
                 return ERROR_CODE
 
+            try:
+                HWR.beamline.lims.lims_rest.authenticate(loginID, password)
```
marcus-oscarsson, line 211: It would be very nice if we can avoid using lims_rest, do you really need it ?

meguiraun: We are not in a situation right now to go back to soap auth (even less to start using pyispyb). Also, we rely on the lims_rest for some interactions with ispyb that are not present in soap.

I dont mind removing it in the future, it would be good to have just one hwobj doint the ispyb interaction, we are simply not ready :(

marcus-oscarsson: I see, how about if we add a login object directly on `HWR.beamline` i.e `HWR.beamline.authentication` which has one maybe two methods `login` and `logout`. Then this can be configured to be for instance lims-rest, ldap or something else ?

It would probably clean up the code abit ? I'm otherwise happy to merge this if you need it

fabcor-maxiv: @marcus-oscarsson I might be able and willing to address this in a follow-up PR. It would be great if you could open a ticket with some more precise details on how you think the implementation should look like, and assign that ticket to me.

This authentication topic is kind of blocking us (me) right now, so I think we (MAX IV) would prefer if this fix could get merged as it is now (unless there are some other direct concerns to address of course).

fabcor-maxiv: Ah sorry, no...

Of course the last required change is to update `pyproject.toml` to fix the lower bound on the version string of the dependency on `mxcubecore`. So it stays as draft for now :D

fabcor-maxiv:
@marcus-oscarsson @meguiraun @elmjag

From my point of view this is ready for review and merge. The *mxcubecore* dependency version pinned in `pyproject.toml` includes the necessary fixes as well.

_[The read the docs does not build and it is normal because the fixes are in another pull request. I do not know about the *e2e* test.]_

meguiraun (approved):

mxcube3/core/components/lims.py
marcus-oscarsson: Ok, Ill try to look into how we can generalize the login method and delegate it to a `authentication` hardware object then eventually create a ticket.

PR #1050, opened 2023-09-20
MXCUBEApplication and Server static classes 

marcus-oscarsson:
Discovered that MXCUBEApplication and Server got instanciated at some point causing an issue of
shadowing attributes.

 - MXCUBEApplication and Server are static classes and does not need to ( should not be) instanceanciated
 - Simplified the code removed the globlas mxcube and server (not needed)

### DIFF
--- a/mxcube3/__init__.py
+++ b/mxcube3/__init__.py
@@ -7,6 +7,7 @@ import traceback
 
 
 from gevent import monkey
+
 monkey.patch_all(thread=False)
 
 from mxcube3.server import Server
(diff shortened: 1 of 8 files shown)


fabcor-maxiv:
Can we enforce this somehow? I do not recall how this is typically handled in Python. Maybe with an `__init__` that raises `NotImplemented`?

marcus-oscarsson:
Yes it can be done, its a good idea :)

marcus-oscarsson:
Like so :)

fabcor-maxiv:
For consistency: in `mxcube3/__init__.py`, why not `import ... as server` and `import ... as mxcube`? Or the other way around (which might be better actually, not sure) use `Server` and `MXCUBEApplication` consistently everywhere?

marcus-oscarsson:
Yes, that makes sense :)

PR #1052, opened 2023-09-20
Remember parameters between samples option

marcus-oscarsson:
Added an option to remember (or not clear) collection parameters between samples

mxcube-remember


meguiraun:
ping @JieNanMAXIV

PR #1055, opened 2023-09-24
Fix for case where log files do not already exist

rhfogh:


rhfogh:
Minor bug fix - program now runs even if there is no pre-existing log file to rename.

marcus-oscarsson:
Ah, nice one :) Thanks !

PR #1056, opened 2023-09-26
Updated README so that it contains install instructions

marcus-oscarsson:
Moved install instructions to README. We can then remove the wiki page when merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,8 @@ git clone https://github.com/mxcube/mxcubecore.git
 ```
 
 ### 3. Create Conda environment `mxcubeweb`
+We recommend using conda a convenience and create a seperate environemnt for libraries and python version. This, though, is not compulsory and one can use any other tool or configuration for installing the application and the necessary third party libraries.
+
 Create Conda environment `mxcubeweb` using the conda environment file in mxcubeweb. The installation of the dependencies in the environment takes a few minutes.
 
 The name of the environment is located at the top of file `conda-environment.yml`. It can be changed from the command line with `--name <env-name>`.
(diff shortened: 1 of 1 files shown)


rhfogh (approved):
Looks fine.

README.md
```diff
@@ -2,32 +2,40 @@
...
 [![codecov](https://codecov.io/gh/mxcube/mxcube3/branch/master/graph/badge.svg)](https://codecov.io/gh/mxcube/mxcube3)
 
-
 <p align="center"><img src="http://mxcube.github.io/mxcube/img/mxcube_logo20.png" width="125"/></p>
 
 # MXCuBE-Web
+
+
+MXCuBE-Web is the latest generation of the data acquisition software MXCuBE (Macromolecular Xtallography Customized Beamline Environment).
+
```
beteva: repetition

README.md
```diff
@@ -55,3 +61,152 @@ Oscarsson, M. et al. 2019. “MXCuBE2: The Dawn of MXCuBE Collaboration.” Jour
...
+# or (to pass a different name)
+conda env create -f conda-environment.yml --name another_name
+```
+
+### 4. Activate the environment
+
+```
+conda activate mxcubeweb
+```
+
```
beteva: Do we oblige to use conda or we strongly recommend it, but leave it optional?

marcus-oscarsson: Id say that we recommend it but one is of course free to choose another way of installing things

beteva: In this case, maybe we say something like - We use conda for a convenience and to be independent from other projects libraries and python version. This, though, is not compulsory and one can use any other tool or configuration for installing the application and the necessary third party libraries.

meguiraun (approved):

PR #1057, opened 2023-09-26
Updated contributing

marcus-oscarsson:
- Updated contributing guidelines so that to reflects the current state.
 - There is still a section missing for front end development for instance linting (but we also need to work on the linting tools) 
 - There is, as well, a section missing about precomit and flake8 in addition to black

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,103 +1,257 @@
-## How to contribute to MXCuBE
+## How to contribute to mxcubecore
 
-Before submiting the code to the repository please read these contributing guidelines.
-The aim of these guidelines is to help the developers community to maintain the code stable and reusable.
+Before submiting the code to the repository please read these contributing guidlines.
+The aim of these guidlines is to help the developers community to maintain the code stable and reusable.
 
 ### Reporting bugs
-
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,103 +1,257 @@
-## How to contribute to MXCuBE
+## How to contribute to mxcubecore
 
-Before submiting the code to the repository please read these contributing guidelines.
-The aim of these guidelines is to help the developers community to maintain the code stable and reusable.
+Before submiting the code to the repository please read these contributing guidlines.
```
rhfogh, line 3: That should have remained 'guidelines', not 'guidlines'

CONTRIBUTING.md
```diff
@@ -1,103 +1,257 @@
...
+### Coding convention and style guidelines
+
+#### Units
+Functions returning a value representing a physical quantity should in general be assoicated with 
+a unit. It has been agreed that the following units should, where applicable, be used across the 
+code base
+
+ * mm (millimeter) for translative motors and sizes
+ * degrees for rotative motors
+ * perecent (%) for ratios like attenuation
```
rhfogh: That should be 'percent'

rhfogh (approved):
Very good! Thanks.

PR #1060, opened 2023-09-26
Fix ESLint warnings (1)

axelboc:
No better time than the present! I'll split things up a bit so it's easier to digest. Starting count is **365 warnings**.

In this PR, I'm knocking down 36 warnings in the **reducers**, bringing the count down to 329.

I'll add comments to point to the relevant rules.


ui/src/containers/GphlWorkflowParametersDialog.jsx
```diff
@@ -7,7 +7,7 @@ import Form from '@rjsf/core';
 import {
   showGphlWorkflowParametersDialog,
   gphlWorkflowSubmitParameters,
-} from '../actions/gphl_workflow';
+} from '../actions/gphlWorkflow';
```
axelboc, line 10: No ESLint warning here, it's just for consistency.

ui/src/reducers/beamline.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable sonarjs/no-duplicate-string */
```
axelboc, line 1: [sonarjs/no-duplicate-string](https://github.com/SonarSource/eslint-plugin-sonarjs/blob/master/docs/rules/no-duplicate-string.md): I'm disabling the rule in a couple of files — the rule can be a bit noisy but helps spot duplicated identifiers and such, so I prefer to not disable it globally

marcus-oscarsson:
Wow, wonderful and great initiate, we like ! :)

marcus-oscarsson:
:100:

marcus-oscarsson:
By the way still working on the e2e test, which apparently run fine here on fresh develop branches, ... scratch scratch

ui/src/reducers/beamline.js
```diff
@@ -141,7 +142,7 @@ export const INITIAL_STATE = {
   energyScanElements: [],
 };
 
-export default (state = INITIAL_STATE, action) => {
+function beamlineReducer(state = INITIAL_STATE, action = {}) {
```
axelboc, line 145: - [import/no-anonymous-default-export](https://github.com/import-js/eslint-plugin-import/blob/v2.27.5/docs/rules/no-anonymous-default-export.md): everything exported should be named
- [default-param-last](https://eslint.org/docs/latest/rules/default-param-last): since `state` has a default value, `action` must also have a default value.

ui/src/reducers/contextMenu.js
```diff
@@ -1,5 +1,4 @@
-/* eslint-disable import/no-anonymous-default-export */
-const initialState = {
+const INITIAL_STATE = {
```
axelboc, line 1: Constant case by convention

ui/src/reducers/queue.js
```diff
@@ -14,21 +14,21 @@ const initialState = {
...
-        queue: initialState.queue,
-        current: initialState.current,
+        queue: INITIAL_STATE.queue,
+        current: INITIAL_STATE.current,
       };
     }
     case 'ADD_SAMPLES_TO_QUEUE': {
       const sampleIDList = action.samplesData.map((s) => s.sampleID);
-      return { ...state, queue: state.queue.concat(sampleIDList) };
+      return { ...state, queue: [...state.queue, ...sampleIDList] };
```
axelboc, line 31: [prefer-spread](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v46.0.0/docs/rules/prefer-spread.md): `foo.concat(bar)` => `[...foo, ...bar]` – this is just a stylistic rule, useful to ensure consistency in the codebase

ui/src/reducers/queueGUI.js
```diff
@@ -26,7 +26,7 @@ export default (state = initialState, action) => {
       const newNodes = [];
 
       action.sampleOrder.forEach((sampleID) => {
-        if (action.sampleList.hasOwnProperty(sampleID)) {
+        if (sampleID in action.sampleList) {
```
axelboc, line 29: [no-prototype-builtins](https://eslint.org/docs/latest/rules/no-prototype-builtins) reports that `foo.hasOwnProperty("bar")` should be replaced with `Object.prototype.hasOwnProperty.call(foo, "bar")` to account for potential edge cases.

However, in the cases reported, `hasOwnProperty` can simply be replaced with the [`in` operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in), since we're only dealing with plain objects.

ui/src/reducers/sampleGrid.js
```diff
@@ -122,9 +122,11 @@ export default (state = INITIAL_STATE, action) => {
...
       const sampleList = { ...state.sampleList };
       sampleIDList.forEach((sampleID, i) => {
-        sampleList[sampleID].tasks.length > 0
-          ? sampleList[sampleID].tasks.concat(action.samplesData[i].tasks)
-          : (sampleList[sampleID].tasks = action.samplesData[i].tasks);
+        if (sampleList[sampleID].tasks.length > 0) {
+          sampleList[sampleID].tasks.concat(action.samplesData[i].tasks);
+        } else {
+          sampleList[sampleID].tasks = action.samplesData[i].tasks;
+        }
```
axelboc, line 129: [no-unused-expressions](https://eslint.org/docs/latest/rules/no-unused-expressions) — ternaries are expressions that resolve to a value, so using a ternary just to trigger side effects is counter-intuitive. `if`/`else` blocks are better suited. 

This change actually reveals a more serious issue, which is that the array returned by `concat` is not used (`concat` does not mutate the arrays) — so the expression in the `if` block has no effect at all. Not sure what the correct behaviour should be. ESLint warns about it, so I decided to leave it as is for now.

ui/src/reducers/sampleview.js
```diff
@@ -169,16 +169,7 @@ export default (state = initialState, action) => {
...
-      return {
-        ...state,
-        distancePoints: [],
-        clickCentringPoints: [],
-        gridList: [],
-        gridCount: 0,
-        selectedShapes: [],
-      };
-    }
+    case 'CLEAR_ALL':
```
axelboc, line 172: [no-duplicated-branches](https://github.com/SonarSource/eslint-plugin-sonarjs/blob/master/docs/rules/no-duplicated-branches.md) — duplicated `case` blocks

ui/src/serverIO.js
```diff
@@ -228,9 +228,10 @@ class ServerIO {
         const taskCollapsed =
           store.getState().queueGUI.displayData[record.queueID].collapsed;
 
-        if (record.state === 1 && !taskCollapsed) {
-          this.dispatch(collapseItem(record.queueID));
-        } else if (record.state >= 2 && taskCollapsed) {
+        if (
+          (record.state === 1 && !taskCollapsed) ||
+          (record.state >= 2 && taskCollapsed)
+        ) {
```
axelboc, line 234: Also [no-duplicated-branches](https://github.com/SonarSource/eslint-plugin-sonarjs/blob/master/docs/rules/no-duplicated-branches.md)

rhfogh (commented):
Looks good (to my very inexpert eye).

ui/src/reducers/beamline.js
marcus-oscarsson: :+1:

ui/src/reducers/beamline.js
marcus-oscarsson: super, thanks

ui/src/reducers/queueGUI.js
marcus-oscarsson: Nice !

marcus-oscarsson:
:+1: Thanks alot Axel !

PR #1063, opened 2023-09-26
Fix documentation pages

### LINKED ISSUE #1053: Fix and automate Sphinx-based documentation in `/docs/`
It would be nice to have documentation automatically published somewhere.

We have the basis for Sphinx-based documentation in `/docs/`. We could work on fixing this (it should be relatively straightforward). And then we could work on getting the GitHub workflows to automatically publish on *GitHub Pages*.

fabcor-maxiv:
GitHub: fixes https://github.com/mxcube/mxcubeweb/issues/1053


fabcor-maxiv:
@marcus-oscarsson To generate documentation pages for the REST API such as [this one](https://fabcor-maxiv.github.io/mxcubeweb/rest-api/login.html), we rely on the [`autoflask` directive](https://sphinxcontrib-httpdomain.readthedocs.io/en/stable/index.html#directive-autoflask). For this the application (the Flask web server) needs to be running. This also means that a Redis server needs to be running.

Now, in GitHub Actions workflow it is quite easy to get a Redis server running:

https://github.com/mxcube/mxcubeweb/commit/b704eda4f9554f7dab86fbd65c922be5b15f62a7#diff-86eaae8c6d4108350f0ef9589ef2a19de3d6e3693eff3b89ee35c160ae35c21cR19-R31

But sadly for *readthedocs*, I am not sure if it is feasible, at least I have not figured it out yet.

Some questions:

* How complex do you think it would be to get mxcube-web to run without redis server for the purpose of building the documentation?
* Is there a way to get only the Flask application running without the whole mxcube? (maybe it is a silly question, but maybe not...)
* Would it still be useful to have documentation without the REST API?

fabcor-maxiv:
@marcus-oscarsson

I figured out a way to get the Redis server runnig during the build on *readthedocs' as well after all. This is not pretty but it seems to work reliably. We can reconsider how to handle the documentation of the REST API at a later point.

All in all, I guess this pull request is in a good state, I consider it ready for review and merge. We can do some adjustments and obviously fix the contents in follow-up pull requests.

marcus-oscarsson:
Very nice, well done !. This is already very nice. We can probably find a way to extract the openapi doc and include it in someway. 

 - We can manage to run `mxcube-web` without `Redis` and probably without the `mxcubecore` as well. I don't think its very complicated. I'm not sure that I understand exactly why we would do that, less dependencies, faster install for building the doc ? We can always try to add some special doc switch to the server that extracts the openapi spec on the format we like, then we can decide on if we want to "load" `Redis` and `mxcubecore`
 - Sure the documentation will still be very much  useful for several reasons, I think the documentation on the more conceptual level would be the most important, there is a lot of things there that we can really get from reading the code (that's the tricky part IMO)

marcus-oscarsson:
:+1:

PR #1066, opened 2023-09-27
Custom equipment tweaks

meguiraun:
small fixes:
- Closing the generic equipment popover when clicking outside
- Fixing None value, wrong args in adapter_base


marcus-oscarsson:
:+1:  Thanks @meguiraun

PR #1068, opened 2023-09-28
Fix ESLint warnings (2)

axelboc:
Re-opening a PR to merge `fix-eslint-2` since it was merged accidentally to `fix-eslint-1` instead of `develop`. https://github.com/mxcube/mxcubeweb/pull/1061#issuecomment-1737845449

PR #1070, opened 2023-09-28
Fix regular expressions

axelboc:
Seems that the regexes modified by ESLint in https://github.com/mxcube/mxcubeweb/pull/1068/files#diff-f10f5ea4f3870e6bbd1c60836649382147f2438b75b4b7eb505f9e8bfed69fd8 were no longer valid.

This tool can help understand the old/new regexes: https://regex101.com/

### DIFF
--- a/ui/src/components/Tasks/validate.js
+++ b/ui/src/components/Tasks/validate.js
@@ -24,7 +24,7 @@ const validate = (values, props) => {
   // here we update the resolution limits based on the energy the typed in the form,
   // the limits come from a table sent by the client
 
-  const validFname = /^[\w#[\]{}-\-]+$/u.test(props.filename);
+  const validFname = /^[-_#{}[\]\w]+$/u.test(props.filename);
 
   const emptyField = 'field is empty';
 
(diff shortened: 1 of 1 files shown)


ui/src/components/Tasks/validate.js
```diff
@@ -24,15 +24,15 @@ const validate = (values, props) => {
   // here we update the resolution limits based on the energy the typed in the form,
   // the limits come from a table sent by the client
 
-  const validFname = /^[\w#[\]{}-\-]+$/u.test(props.filename);
+  const validFname = /^[-_#{}[\]\w]+$/u.test(props.filename);
```
axelboc, line 27: Initial regex was `/^[-\w\-\#\_\{\}\[\]]+$/`.

- None of the characters need to be escaped except `]`.
- The dash was provided unescaped and escaped. I've kept the unescaped one and kept it at the start of the character set (i.e. `[...]`) so it is never considered as a range (e.g. `a-z`).

ui/src/components/Tasks/validate.js
```diff
@@ -24,15 +24,15 @@ const validate = (values, props) => {
...
 
   const emptyField = 'field is empty';
 
   if (!validFname) {
     errors.prefix = INVALID_CHAR_MSG;
   }
 
-  if (props.subdir && !/^[\w/{}-\-]+$/u.test(props.subdir)) {
+  if (props.subdir && !/^[-{}\/\w]+$/u.test(props.subdir)) {
```
axelboc, line 35: Initial regex was `/^[-\w\-\/\_\{\}]+$/`. Same problem with the dash: it has to be at the front and doesn't need to be specified twice.

meguiraun (approved):

PR #1071, opened 2023-09-28
Switch to pnpm

axelboc:
https://pnpm.io

Lots of benefits to pnpm:

- Much, much faster than npm
- Lighter `node_modules` when working on multiple projects, since packages are symbolically linked to a global store
- [No phantom dependencies](https://www.kochan.io/nodejs/pnpms-strictness-helps-to-avoid-silly-bugs.html): the `node_modules` folder is structured in such a way that a package can only "see" dependencies that it depends on explicitly. In our case, the `prop-types` package, while used in the codebase, was not actually installed as a direct dev dependency. Phantom dependencies can cause more complicated issues when various (sub-)dependencies dependent on multiple versions of the same package.
- [Lots of ways](https://pnpm.io/package_json#pnpmoverrides) to resolve peer dependencies issues, outdated packages, etc. via the `pnpm` property in `package.json`. No more `--legacy-peer-deps`, ever.

Comments incoming, as always.


.github/workflows/e2e-test.yml
```diff
@@ -31,43 +31,54 @@ jobs:
...
         run: |
           conda install -c conda-forge mamba
           mamba env update --file conda-environment.yml --name base
           pip install -e .
 
+      - name: Install pnpm
+        uses: pnpm/action-setup@v2
+        with:
+          version: 8.x
+
```
axelboc, line 58: I've replicated the set-up of other front-end projects, but maybe there's a better way to install it in MXCuBE. :shrug:

README.md
```diff
@@ -156,55 +157,51 @@ pip install -e .
 pip install -e .
 ```
 
-The above makes it possible to add break points directly in the "checked out code".
```
axelboc, line 159: Repeats what is said before the commands block.

README.md
```diff
@@ -156,55 +157,51 @@ pip install -e .
...
-
-Before running any test, make sure that the local *Redis* server is running. For example with the `mxcubeweb` *conda* environment activated in a terminal, run the `redis-server` command.
```
axelboc, line 161: Moved to next section.

README.md
```diff
@@ -156,55 +157,51 @@ pip install -e .
...
 
-```
-# The front-end needs the backend to run so before starting the 
-# development server open a new terminal and run (as in step 7):
+The front-end development server listens to changes on the file system and re-builds the UI when changes are made. This makes it very easy and fast to see how changes affect the UI. The development server listens on port **3000**.
 
+```
+# Activate the Conda environment and start the MXCuBE server (cf. step 7)
 conda activate mxcubeweb
-mxcubeweb-server -r $(pwd)/mxcubeweb/test/HardwareObjectsMockup.xml/ --static-folder $(pwd)/mxcubeweb/ui/build/ -L debug
```
axelboc, line 184: At this point, we're already in the `mxcubeweb` folder (there's a `cd mxcubeweb` somewhere above).

README.md
```diff
@@ -156,55 +157,51 @@ pip install -e .
...
+mxcubeweb-server -r $(pwd)/test/HardwareObjectsMockup.xml/ --static-folder $(pwd)/ui/build/ -L debug
 
-# Enter the `ui` folder and issue:
-npm run start
+# Enter the `ui` folder and start the front-end developement server:
+cd ui
+pnpm start
+
+# Note that you can also run any pnpm script from the root folder with:
+pnpm --prefix ui <script>
```
axelboc, line 191: Thought it'd be a good place to explain this.

conda-install.sh
```diff
@@ -46,15 +46,15 @@ mxcube_download() {
 }
 
 init_conda() {
-  command conda create --name mxcube-web
```
axelboc, line 49: for consistency with the docs

conda-install.sh
```diff
@@ -46,15 +46,15 @@ mxcube_download() {
...
-  command conda env update --name mxcube-web --file conda-environment.yml
-  command conda activate mxcube-web
+  command conda create --name mxcubeweb
+  command conda env update --name mxcubeweb --file conda-environment.yml
+  command conda activate mxcubeweb
 }
 
 
 install_node() {
-  command conda install nodejs
```
axelboc, line 56: Isn't `nodejs` already installed by this point?

conda-install.sh
```diff
@@ -46,15 +46,15 @@ mxcube_download() {
...
-  command npm install
+  command npm install --global pnpm@8
+  command pnpm --prefix ui install
```
axelboc, line 57: I think we're still at the root of the project directory at this point so I'm adding `--prefix ui` (same problem in multiple files below).

ui/.gitignore
axelboc, line 1: Removing some stuff that are already specified in the root `.gitignore`.

ui/README.md
axelboc, line 1: This is the default `README` of create-react-app; I don't think it's needed.

ui/package.json
```diff
@@ -59,25 +59,25 @@
...
     "socket.io-client": "^4.6.1",
     "underscore": "^1.8.3",
-    "webpack": "^5.69.1",
-    "webpack-dev-server": "^4.8.0",
     "ws": "^8.5.0"
   },
   "devDependencies": {
     "@fortawesome/fontawesome-free": "^5.15.4",
-    "@testing-library/cypress": "9.0.0",
+    "@testing-library/cypress": "10.0.1",
```
axelboc, line 76: Upgrading Cypress Testing Library to fix the peer dep issue, while I'm at it.

ui/package.json
```diff
@@ -59,25 +59,25 @@
...
+    "@testing-library/jest-dom": "^5.16.2",
+    "@testing-library/react": "^12.1.2",
+    "@testing-library/user-event": "^13.5.0",
```
axelboc, line 79: Moved to `devDependencies` along with a few others.

ui/package.json
```diff
@@ -59,25 +59,25 @@
...
+    "babel-preset-react-app": "10.0.1",
```
axelboc, line 80: Missing dep required for ESLint (and our linting config) to work properly when installed with pnpm

ui/package.json
```diff
@@ -86,9 +86,15 @@
     "eslint-plugin-jest": "27.2.1",
     "eslint-plugin-jsx-a11y": "6.7.1",
     "eslint-plugin-react": "7.32.2",
+    "http-proxy-middleware": "^2.0.2",
     "less": "^4.1.2",
     "less-watch-compiler": "^1.16.3",
-    "prettier": "3.0.0"
+    "prettier": "3.0.0",
+    "prop-types": "15.8.1",
```
axelboc, line 93: `prop-types` was missing

marcus-oscarsson:
Very nice :)

marcus-oscarsson:
.. awesome actually :)

PR #1073, opened 2023-09-28
Added pnpm to conda requirements

marcus-oscarsson:


axelboc:
Should all the `npm install --global pnpm@8` I've added everywhere be removed (as well as the actions to install pnpm on the CI)? Is it also possible to pin the version to `8` (and maybe the `nodejs` version to `20`)?

fabcor-maxiv:
@marcus-oscarsson @axelboc 

> Should all the npm install --global pnpm@8 I've added everywhere be removed 

I am redoing the develop installation instruction steps right now... and yes I noticed that we ask to install *pnpm* explicitly (and globally, which I find strange, won't it create issues later on?) although it does not seem necessary since it is already in the conda dependencies.

I do not know about pinning, I have not given it any thought yet.

marcus-oscarsson:
It makes things a bit easier if one does not need to do `npm install --global pnpm@8` and I thought it was better to add it in conda as we are using it in anyways. Id be in favor of pinning as well and installing pnpm through conda.

fabcor-maxiv:
> easier if one does not need to do `npm install --global pnpm@8` and I thought it was better to add it in conda

Do we agree to remove this step from the installation instructions?

If yes, this also means the conda environment should be activated before. In other words we need to move the instruction step`conda activate mxcubeweb` before `pnpm --prefix ui build`.

marcus-oscarsson:
Sounds good !

fabcor-maxiv:
I created the pull request https://github.com/mxcube/mxcubeweb/pull/1083 to address this. Turns out there was already a step to activate the conda environment above so it was not necessary to move any step around.

PR #1075, opened 2023-10-03
Update beamline.py

meguiraun:
removing error:

TypeError: The view function for 'beamline.actuator_set_value' did not return a valid response. The function either returned None or ended without a return statement.


marcus-oscarsson:
:), Thanks !

PR #1076, opened 2023-10-03
fix route return

meguiraun:
same as before :)

### DIFF
--- a/mxcube3/routes/beamline.py
+++ b/mxcube3/routes/beamline.py
@@ -62,7 +62,7 @@ def create_set_route(app, server, bp, adapter, attr, name):
             """
             rd = _th["value"].parse_raw(request.data)
             getattr(app.mxcubecore.get_adapter(rd.name.lower()), attr)(rd)
-            return "Value set successfull"
+            return "Value set successfully"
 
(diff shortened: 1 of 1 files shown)


mxcube3/routes/beamline.py
```diff
@@ -83,7 +83,7 @@ def set_func():
         args = request.get_json()
         adapter = app.mxcubecore.get_adapter(obj.lower())
         adapter.execute_command(cmd_name, args)
-        return make_response("{}", 200)
+        return "Command executed successfull"
```
fabcor-maxiv, line 86: successfully

fabcor-maxiv: Are there cases where make_response returns None? When? Why?

meguiraun: seems so, all the calls in this method, no clue why

marcus-oscarsson:
Very nice, ill try to get the build working again after this dependency mess and then Ill try to rebase

mxcube3/routes/beamline.py
marcus-oscarsson: hm, odd

PR #1077, opened 2023-10-03
Update pyproject.toml - Flask what did you do

marcus-oscarsson:
Looks some recent changes in the werkzeug flask chain comes with some incompatibilities, 
so pinning flask for now:

```
   from mxcube3 import main
  File "/tmp/mxcubeweb/mxcube3/__init__.py", line 13, in <module>
    from mxcube3.server import Server as server
  File "/tmp/mxcubeweb/mxcube3/server.py", line 12, in <module>
    from flask import Flask, request
  File "/home/oscarsso/miniconda3/envs/mxcubetest/lib/python3.10/site-packages/flask/__init__.py", line 5, in <module>
    from .app import Flask as Flask
  File "/home/oscarsso/miniconda3/envs/mxcubetest/lib/python3.10/site-packages/flask/app.py", line 30, in <module>
    from werkzeug.urls import url_quote
ImportError: cannot import name 'url_quote' from 'werkzeug.urls' (/home/oscarsso/miniconda3/envs/mxcubetest/lib/python3.10/site-packages/werkzeug/urls.py)
```


fabcor-maxiv:
From what we understood this is the fix we need to be released: https://github.com/maxcountryman/flask-login/pull/746

marcus-oscarsson:
@fabcor-maxiv, hm yes its possible this error seems to be from Flask itself though, but it might be in more than one place ...

Pinning Flask and Werkzeug worked fine for me so Ill propose that for now until there is a better solution

fabcor-maxiv:
@marcus-oscarsson Ah... But Flask 2.2.2 is quite old already. We probably do not need to (or even can) update to Flask 3, but some more recent Flask 2.x.x might be fine. I do not know exactly. Anyway, maybe we can get away with `Werkzeug = "<3"` even though it is an indirect dependency.

marcus-oscarsson:
@fabcor-maxiv I tried with just updating werkzeug  and some of the other dependencies but it came with the same or other issues.  This was the best working solution for the moment, I guess that it will get fixed shortly (the patch seems easy enough). I'm happy to give it a go if you have some other solution with more recent versions

fabcor-maxiv:
@marcus-oscarsson No, sadly I do not have any better solution. These situations are always messy. I had assumed Poetry's lock file should have protected us from this, but apparently it is not as simple as that...

There is something I read this morning:

Since *werkzeug* is an indirect dependency, it might make sense to list it in the *dev* dependencies with the version constraint `<3`. This way it would still be taken into account in Poetry's lock file without polluting the list of direct dependencies. Not sure how far it would help us in this case, but I thought I would mention it anyway.

marcus-oscarsson:
@fabcor-maxiv: hm okey Ill think about that :). Thanks I tried with downgrading werkzeug and that seems to work fine for now.

elmjag:
FYI, pinning werkzeug to 2.3.7 or setting version to <3 worked for me.

I.e. adding either this line:

`werkzeug = "2.3.7"`

or

`werkzeug = "<3"`

solved the issue with mxcube server not starting for me.

Note, that I did clean installs. Not sure behavior is different when installing in already existing environment.

marcus-oscarsson:
@elmjag: Yes it does indeed seem to work :)

marcus-oscarsson:
So finally, there seems to be two dependency issues one is related to the werkzeug version which gives the error reported above.

The second issue that was causing the application to get stuck on the "loading screen"  is related to spectree. The exact reason is not yet determined.

There was a third issue with the CI getting stuck installing mamba, using python 3.10 on the CI seems to solve that issue.

fabcor-maxiv:
Wow... Thanks @marcus-oscarsson for figuring this out!

It works for me with the version constraint `spectree<1.2.4` (both 1.2.5 and 1.2.4 lead to an unusable mxcube).

marcus-oscarsson:
Ok, there seems to be quite alot of activity on the spectree repository for the moment so there might be a fix for it soon. If not ill try to investigate exactly whats going on and we will figure out what to do.

PR #1083, opened 2023-10-06
Remove step for pnpm global install in README

fabcor-maxiv:


fabcor-maxiv:
Wiki should be updated accordingly if/when this gets merged.

fabcor-maxiv:
I now also removed the instruction step from the wiki.

marcus-oscarsson:
:)

PR #1085, opened 2023-10-09
Fix front-end linting errors (3)

axelboc:
I'm disabling a bunch of rules, either globally or locally. The priority is for the linting to pass so we can avoid adding new issues to the codebase.

There are a few issues left that seem to point to actual bugs or dead code. Since I'm not sure which it is, I've left them in. Perhaps @marcus-oscarsson you can take over this PR at some point and resolve those last few issues before merging?

After that, the next step will be to remove the `incrementalAdoption` flag from `.eslintrc.js` so that warnings become errors again (but there shouldn't any left) and also to enable a bunch more rules as warnings (like `no-useless-rename`, which can automatically turn `const { foo: foo } = props` into `const { foo } = props` :grin:)

### DIFF
--- a/ui/src/components/SampleGrid/TaskItem.jsx
+++ b/ui/src/components/SampleGrid/TaskItem.jsx
@@ -174,15 +174,15 @@ export class TaskItem extends React.Component {
           <div className="row" style={{ paddingTop: '0.5em' }}>
             <span className="col-sm-4">
               <b>Quality Indictor: </b>
-              <img ref="fimage" alt="First" src={qIndUrl} width="90%" />
+              <img alt="First" src={qIndUrl} width="90%" />
             </span>
             <span className="col-sm-4">
               <b>First image: </b>
-              <img ref="fimage" alt="First" src={fImageUrl} width="90%" />
+              <img alt="First" src={fImageUrl} width="90%" />
             </span>
             <span className="col-sm-4">
               <b>Last image: </b>
-              <img ref="limage" alt="Last" src={lImageUrl} width="90%" />
+              <img alt="Last" src={lImageUrl} width="90%" />
             </span>
           </div>
         </div>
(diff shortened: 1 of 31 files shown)


ui/src/components/SampleView/DrawGridPlugin.js
```diff
@@ -300,7 +300,9 @@ export default class DrawGridPlugin {
   initializeCellFilling(gd, col, row) {
     const level = this.overlayLevel ? this.overlayLevel : 0.2;
     const fill = `rgba(0, 0, 200, ${level}`;
-    return new Array(col).fill().map(() => new Array(row).fill(fill));
```
axelboc, line 303: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v46.0.0/docs/rules/no-new-array.md

ui/src/components/SampleGrid/TaskItem.jsx
```diff
@@ -174,23 +174,23 @@ export class TaskItem extends React.Component {
           <div className="row" style={{ paddingTop: '0.5em' }}>
             <span className="col-sm-4">
               <b>Quality Indictor: </b>
-              <img ref="fimage" alt="First" src={qIndUrl} width="90%" />
```
axelboc, line 177: Couldn't find any use of these refs, but they could be hiding... Please let me know if that's the case.

marcus-oscarsson:
Nice, thanks for the initiative ! Yepp, ill take a look !

ui/src/containers/SampleGridTableContainer.jsx
```diff
@@ -276,7 +277,7 @@ class SampleGridTableContainer extends React.Component {
 
     const selected = this.getSamplesList()
       .filter((sampleItem) => {
-        const sampleElement = document.getElementById(sampleItem.key);
+        const sampleElement = document.querySelector(`#${sampleItem.key}`);
```
jbflo, line 280: Hey @axelboc 

Is this linting fix work for you cause I got this:
image

to reproduce : try to select some sample in the Sample List View

axelboc: Nice catch, indeed the column is causing an issue. I'll revert.

PR #1086, opened 2023-10-10
Fixed the remaining linting errors

marcus-oscarsson:
Did the bare minimal to clean up the remaining linting errors


axelboc:
Awesome, thanks!

PR #1088, opened 2023-10-10
Update aperture.xml to remove unneeded parentheses 

mattclarke:
We tried it without the parentheses at the code camp and it seemed to work fine.


meguiraun (approved):

marcus-oscarsson:
Nice, well done :) :+1:

PR #1089, opened 2023-10-10
Turned isstaff into a property

marcus-oscarsson:
`isstaff()` was inconsistently used as both a method and a property. The other `is` attributes on the `User` object are attributes on not methods.

The consequence of this was that all users were considered staff and thus the take/give control options are available in the UI. It turns our that this is something that we want in anyways, so we simply did not notice the effect of this.


meguiraun (approved):
thanks!

PR #1090, opened 2023-10-10
Updating queue and sample list on observers when it changes (As redux persist was removed)

marcus-oscarsson:
We have removed `redux-persist` which means that updates to the queue needs to be communicated to the observers.

### DIFF
--- a/ui/src/serverIO.js
+++ b/ui/src/serverIO.js
@@ -34,6 +34,7 @@ import {
   setCurrentSample,
   addDiffractionPlanAction,
   setSampleAttribute,
+  fetchQueue,
 } from './actions/queue';
 import { collapseItem, showResumeQueueDialog } from './actions/queueGUI';
 import { setLoading, showConnectionLostDialog } from './actions/general';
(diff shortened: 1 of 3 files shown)


ui/src/serverIO.js
```diff
@@ -266,6 +267,11 @@ class ServerIO {
 
       if (record.Signal === 'DisableSample') {
         this.dispatch(setSampleAttribute([record.sampleID], 'checked', false));
+      } else if (record.Signal === 'update') {
+        const state = store.getState();
+        if (!state.login.user.inControl) {
+          this.dispatch(fetchQueue());
```
meguiraun, line 273: why not sending the queue with the emit? feels awkward to notify the UI about an update, but let the UI itself fetch the newest data

marcus-oscarsson: I was thinking the same, but then I thought that it would be better to do it the same way all the time, meaning that we always fetch the queue via a http request, and we reuse the route we have in place, otherwise its like we are getting the queue in two different ways. If it makes sense

meguiraun: ah yes, you have a point there

meguiraun (approved):
thanks

marcus-oscarsson:
:)

Issue #1091, opened 2023-10-10
Observer mode bypass bug


RafaelLyra8:
If I scroll down when I am in the observer mode, an unprotected space, that was occupied by the navbar, appears, as shown in the image below:
![Screenshot from 2023-10-10 08-38-04](https://github.com/mxcube/mxcubeweb/assets/63302930/20c32109-3b83-494c-b26e-597afda63223)

This creates a region that bypass the observer mode, allowing the user in the observer mode to alter some parameters. 
A PR was open with the fix for this bug: https://github.com/mxcube/mxcubeweb/pull/1092

PR #1092, opened 2023-10-10
Fix observer bug

RafaelLyra8:
Solve issue: https://github.com/mxcube/mxcubeweb/issues/1091


marcus-oscarsson:
Well done !, and thanks !

PR #1093, opened 2023-10-10
Lint

marcus-oscarsson:
Linted and formatted all files.

 - flake8 now passes. 
 - Excluding the test directory for the time being


fabcor-maxiv:
What are all the `# noqa` things? Can't we be more specific with flake8 (or whatever the linter is)?

With every linter I use (as far as I can recall), it is possible to be explicit by specifying which linting rule we want to make an exception for. Which I believe is much better for the code quality. By disabling linting completely for the whole line we could hide some actual useful linting warning/errors. It is not proper style from my point of view to disable the whole linter for a line.

Seems like it is possible:
https://flake8.pycqa.org/en/latest/user/violations.html#in-line-ignoring-errors

For example:
```
example = lambda: 'example'  # noqa: E731
```

marcus-oscarsson:
Yepp, we are using flake8, I should probably have been more specific with the noqa in this case is function complexity so we can disable that rule specifically.

fabcor-maxiv:
@marcus-oscarsson Yes, better with disabling only one specific rule, thanks :)

marcus-oscarsson:
Sorry for the messy PR but it was a bit complicated to get things to run properly.

 - Files are now linted so that they pass `flake8`  (but without bug bear  https://github.com/PyCQA/flake8-bugbear). We can enable it in another iteration.

 - test folder is not linted for the time being.

 - Formatted yaml and xml files (mostly white spaces) to pass pre-commit `check-xml` and `check-yaml` steps

 - Using micromamba and poetry for the installation, (we should probably update all the pipelines to use poetry for consistency).

PR #1094, opened 2023-10-11
Fix front-end linting errors (4)

axelboc:
Last, big round of front-end linting fixes after removing `incrementalAdoption` from `ui/.eslintrc.js`.

As before, the goal is to enforce reasonably strict linting for new code and to fix obvious issues; it's not to refactor existing code. That's why in most cases I disable rules locally, either at the top of the files or on specific lines.

### DIFF
--- a/ui/.eslintrc.js
+++ b/ui/.eslintrc.js
@@ -11,10 +11,11 @@ const dependencies = getDependencies();
 
 module.exports = createConfig({
   cwd: __dirname,
-  incrementalAdoption: true, // turn everything into a warning
   rules: {
-    'unicorn/consistent-destructuring': 'off', // too many failures in React class components
-    'promise/catch-or-return': 'off', // requires deeper refactoring of fetching layer
+    'import/no-default-export': 'off', // too opinionated
+    'import/no-unresolved': 'off', // lots of false positives (perhaps misconfigured?)
+    'simple-import-sort/imports': 'off', // changes global CSS order; must migrate to CSS modules first
+    'unicorn/prefer-prototype-methods': 'off', // not really more readable and makes Jest crash
 
     // Ternaries are sometimes more readable when `true` branch is most significant branch
     'no-negated-condition': 'off',
@@ -24,26 +25,24 @@ module.exports = createConfig({
     'unicorn/no-useless-undefined': 'off',
     'consistent-return': 'error',
 
-    // Not really more readable and makes Jest crash
-    'unicorn/prefer-prototype-methods': 'off',
-
     /* Forcing use of `else` for consistency with mandatory `default` clause in `switch` statements is unreasonable.
      * `if`/`else if` serves a different purpose than `switch`. */
     'sonarjs/elseif-without-else': 'off',
-
-    // The point of `switch` is to be less verbose than if/else-if/else
-    'unicorn/switch-case-braces': ['warn', 'avoid'],
   },
   overrides: [
     createReactOverride({
       ...dependencies,
       rules: {
-        'react/no-unsafe': 'off', // requires non-trivial refactoring
-        'react/destructuring-assignment': 'off', // too many failures in React class components
         'react/jsx-no-constructed-context-values': 'off', // too strict
+        'jsx-a11y/media-has-caption': 'off', // not relevant
+
         'react/static-property-placement': 'off', // will not be relevant after converting to functional components
-        'jsx-a11y/anchor-is-valid': 'off', // requires non-trivial refactoring
-        'jsx-a11y/no-static-element-interactions': 'off', // requires non-trivial refactoring
+        'react/destructuring-assignment': 'off', // too many failures in React class components
+        'unicorn/consistent-destructuring': 'off', // too many failures in React class components
+
+        'jsx-a11y/anchor-is-valid': 'off', // too many failures to disable locally
+        'jsx-a11y/click-events-have-key-events': 'off', // too many failures to disable locally
+        'jsx-a11y/no-static-element-interactions': 'off', // too many failures to disable locally
       },
     }),
     createJestOverride({
--- a/ui/src/actions/queue.js
+++ b/ui/src/actions/queue.js
@@ -1,9 +1,11 @@
+/* eslint-disable promise/catch-or-return */
+/* eslint-disable promise/prefer-await-to-then */
 /* eslint-disable sonarjs/no-duplicate-string */
 import fetch from 'isomorphic-fetch';
 import { showErrorPanel } from './general';
-import { loadSample } from './sampleChanger';
+import { loadSample } from './sampleChanger'; // eslint-disable-line import/no-cycle
 import { sendAbortCentring, sendUpdateShapes } from './sampleview';
-import { selectSamplesAction, clearSampleGrid } from './sampleGrid';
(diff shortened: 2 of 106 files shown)


ui/.eslintrc.js
```diff
@@ -11,10 +11,11 @@ const dependencies = getDependencies();
 
 module.exports = createConfig({
   cwd: __dirname,
-  incrementalAdoption: true, // turn everything into a warning
   rules: {
-    'unicorn/consistent-destructuring': 'off', // too many failures in React class components
```
axelboc, line 16: now disabled only in React files

ui/.eslintrc.js
```diff
@@ -11,10 +11,11 @@ const dependencies = getDependencies();
...
-    'promise/catch-or-return': 'off', // requires deeper refactoring of fetching layer
```
axelboc, line 17: now disabled locally

ui/.eslintrc.js
```diff
@@ -11,10 +11,11 @@ const dependencies = getDependencies();
...
+    'import/no-default-export': 'off', // too opinionated
+    'import/no-unresolved': 'off', // lots of false positives (perhaps misconfigured?)
+    'simple-import-sort/imports': 'off', // changes global CSS order; must migrate to CSS modules first
```
axelboc, line 17: Another reason why global CSS is bad (especially when imported from each component file): it is hugely dependent on the order in which components are used/imported in the application.

ui/.eslintrc.js
```diff
@@ -24,26 +25,24 @@ module.exports = createConfig({
...
     'sonarjs/elseif-without-else': 'off',
-
-    // The point of `switch` is to be less verbose than if/else-if/else
-    'unicorn/switch-case-braces': ['warn', 'avoid'],
   },
   overrides: [
     createReactOverride({
       ...dependencies,
       rules: {
-        'react/no-unsafe': 'off', // requires non-trivial refactoring
```
axelboc, line 41: now disabled locally

ui/src/actions/queue.js
```diff
@@ -1,9 +1,11 @@
+/* eslint-disable promise/catch-or-return */
+/* eslint-disable promise/prefer-await-to-then */
 /* eslint-disable sonarjs/no-duplicate-string */
 import fetch from 'isomorphic-fetch';
 import { showErrorPanel } from './general';
-import { loadSample } from './sampleChanger';
+import { loadSample } from './sampleChanger'; // eslint-disable-line import/no-cycle
```
axelboc, line 6: There are a bunch of cyclic imports. Webpack doesn't seem to mind, but it's definitely not ideal...

ui/src/components/Equipment/PlateManipulator.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react/jsx-handler-names */
```
axelboc, line 1: This is quite an important rule, in my opinion, as it enforces The React Way of naming event callback props/functions. So even though there are a _lot_ of failures in the codebase, I decided to disable them manually in every file to try to encourage new code to use the right convention.

https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-handler-names.md

#### GOOD

```jsx
<MyComponent onChange={handleChange} />
<MyComponent onChange={props.onSomethingChange} />

<button onClick={handleClick}/>
<button onClick={props.onButtonClick}/>

<form onSubmit={(evt) => {
  evt.preventDefault();
  doSomething();
  props.doSomethingElse();
} />
```

ui/src/components/Equipment/SampleChanger.js
```diff
@@ -95,7 +96,7 @@ export class SampleChangerTreeNode extends React.Component {
     return (
       <div>
         <li className="treeLi">
-          <input
+          <input // eslint-disable-line jsx-a11y/control-has-associated-label
```
axelboc, line 99: Ensuring each control has a label (either in a separate element via `id`+`htmlFor`, or with a simple `aria-label` attribute) makes writing E2E tests a lot simpler (and makes those E2E tests a lot more robust).

ui/src/components/SSXChip/SSXChip.jsx
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable react/jsx-key */
```
axelboc, line 1: A lot of warnings related to missing or wrongly provided `key` props. Pretty important to get this right in new code, but too tricky to fix in existing code.

https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key

ui/src/components/SSXChip/SSXChip.jsx
```diff
@@ -70,7 +73,7 @@ export default class SSXChip extends React.Component {
       top_left_z: currentChipLayout.calibration_data.top_left[1],
       top_right_x: currentChipLayout.calibration_data.top_right[0],
       top_right_y: currentChipLayout.calibration_data.top_right[0],
-      top_left_z: currentChipLayout.calibration_data.top_right[1],
+      top_right_z: currentChipLayout.calibration_data.top_right[1],
```
axelboc, line 76: bug fix

ui/src/components/SSXChip/SSXChip.jsx
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable react/jsx-key */
+/* eslint-disable react/no-unused-state */
```
axelboc, line 2: A bunch of unused states (and prop types). Might be easy to fix but I didn't want to make the call.

ui/src/components/SampleQueue/QueueControl.js
```diff
@@ -133,6 +134,7 @@ export default class QueueControl extends React.Component {
     );
   }
 
+  // eslint-disable-next-line sonarjs/cognitive-complexity
```
axelboc, line 137: Most of the time, this warning is a good clue that components are becoming too big and need to be split into smaller components.

ui/src/components/SampleView/shapes.js
```diff
@@ -305,20 +307,23 @@ export function makePoints(points, imageRatio) {
...
+      }
     }
   }
   return fabricPoints;
 }
 
 export function makeTwoDPoints(points, imageRatio) {
   const fabricPoints = [];
+  // eslint-disable-next-line guard-for-in
   for (const id in points) {
```
axelboc, line 322: Best to use `for ... of`, but the behaviour differs, so didn't want to risk a regression.

ui/.eslintrc.js
marcus-oscarsson: :)

ui/src/actions/queue.js
marcus-oscarsson: ohh, yes we should fix those

ui/src/components/Equipment/PlateManipulator.jsx
marcus-oscarsson: Thanks, good idea :+1:

ui/src/components/Equipment/SampleChanger.js
marcus-oscarsson: i see

marcus-oscarsson:
Thanks @axelboc, some quite important things to look at.

marcus-oscarsson:
:+1:

PR #1095, opened 2023-10-11
add 'version' to 'application_settings' route response model

elmjag:
Include 'version' attribute into the response model of the route:
    
      mxcube/api/v0.1/application_settings
    
This solves issue with that route never finishing the request for some versions of spectree package. The symptom of this issue is that MxCube UI get stuck on the 'loading' screen after successfull login. This behavior is seen with spectree version 1.2.4 and 1.2.5 installed.

Also, includes some minor clean-up, removing dead models code.


marcus-oscarsson:
Thanks :)

PR #1098, opened 2023-10-12
Improve UI CI workflow with caching, micromamba, etc.

axelboc:
The set up wasn't consistent across the three UI jobs. I'm now trying out this GitHub Action to create and cache the Conda environment: https://github.com/mamba-org/setup-micromamba — seems to work very well.

I'm also removing the Python version matrix, and adding the `wait-on` dev dependency to wait for the server to respond at `127.0.0.1:8081`. This will help catch problems with starting the MXCuBE server, since the E2E tests won't try to run if the server has not successfully started.

image

:partying_face:

### DIFF
--- a/.github/workflows/ui.yml
+++ b/.github/workflows/ui.yml
@@ -4,44 +4,29 @@ on: [push, pull_request]
 
 jobs:
   format:
+    name: Format
     runs-on: ubuntu-latest
-    strategy:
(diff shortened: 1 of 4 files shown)


marcus-oscarsson:
Doing something similar in #1093 :), thanks @axelboc

PR #1103, opened 2023-10-16
Fixing broken annotated beamline action display

marcus-oscarsson:
Draft because there is a PR for `mxcubecore` to come.

 - Fixes the broken display for annotated beamline actions
 - Converted the `BeamlineActionDialog` into a functional component

 To be done in a following PR :
 - Convert `BeamlineActionContainer` to functional component
 - Convert old promise code based on `then` to `async/await` (for beamline actions)

Example
Peek 2023-10-16 13-32


ui/src/components/BeamlineActions/BeamlineActionDialog.jsx
```diff
@@ -0,0 +1,112 @@
...
+        <Modal.Title>{actionName}</Modal.Title>
+      </Modal.Header>
+      <Modal.Body
+        className="d-flex"
+        style={{ height: '500px', overflowY: 'auto' }}
+      >
+        <Row>
+          {actionArguments.map((arg, i) => (
+            // eslint-disable-next-line react/jsx-key
+            <>
```
axelboc: `<Fragment key={arg.name}>` to fix the linting warning (assuming `arg.name` is unique, otherwise it's trickier).

ui/src/components/BeamlineActions/BeamlineActionDialog.jsx
```diff
@@ -0,0 +1,112 @@
...
+            )}
+          </Col>
+        </Row>
+        <hr />
+        <Plot1D
+          displayedPlotCallback={handleOnPlotDisplay}
+          plotId={plotId}
+          autoNext={isActionRunning}
+        />
+        {actionMessages.length > 0 ? (
```
axelboc: `{actionMessages.length > 0 && (...)}` (no need to render an empty string as fallback)

ui/src/components/BeamlineActions/BeamlineActionControl.jsx
```diff
@@ -7,51 +7,47 @@ import {
...
+              ? () => props.handleStartAction(props.actionId, showOutput)
+              : () => props.handleStopAction(props.actionId)
+          }
+        >
+          {label}
+        </Button>
+      ) : (
+        ''
+      )}
+      {showOutput ? (
```
axelboc, line 39: `{showOutput && (...)}`

ui/src/components/BeamlineActions/BeamlineActionControl.jsx
```diff
@@ -7,51 +7,47 @@ import {
...
-      </ButtonToolbar>
-    );
+  if (props.type === 'INOUT') {
+    label = String(props.data).toUpperCase();
+    variant = twoStateActuatorIsActive(props.data) ? 'success' : 'danger';
   }
+
+  return (
+    <ButtonToolbar>
+      {props.actionArguments.length === 0 ? (
```
axelboc, line 22: `{props.actionArguments.length === 0 && (...)}`

ui/src/components/BeamlineActions/AnnotatedBeamlineActionDialog.jsx
```diff
@@ -0,0 +1,110 @@
...
+          <Col>
+            <Plot1D
+              displayedPlotCallback={handleOnPlotDisplay}
+              plotId={plotId}
+              autoNext={isActionRunning}
+            />
+            {actionMessages.length > 0 ? (
+              <Card>
+                {actionMessages.map((message) => (
+                  // eslint-disable-next-line react/jsx-key
```
axelboc: If the messages are unique, then you can pass them as keys; otherwise, use the index of the loop (it's what React does internally and it still requires disabling an ESLint rule, but at least you don't get a warning in the browser).

ui/src/components/BeamlineActions/AnnotatedBeamlineActionDialog.jsx
```diff
@@ -0,0 +1,110 @@
...
+          <Row className="py-2" />
+        </Row>
+        <Row className="py-2">
+          <Col>
+            <Plot1D
+              displayedPlotCallback={handleOnPlotDisplay}
+              plotId={plotId}
+              autoNext={isActionRunning}
+            />
+            {actionMessages.length > 0 ? (
```
axelboc: `{actionMessages.length > 0 && (...)}`

ui/src/components/BeamlineActions/AnnotatedBeamlineActionDialog.jsx
```diff
@@ -0,0 +1,110 @@
...
+    actionSchema,
+    isActionRunning,
+    actionMessages,
+    handleStopAction,
+    handleStartAction,
+    handleOnPlotDisplay,
+    plotId,
+  } = props;
+
+  function getRunButton(isActionRunning) {
```
axelboc: Might as well inline this in the JSX with a ternary, IMO.

ui/src/containers/BeamlineActionsContainer.jsx
```diff
@@ -122,98 +115,39 @@ class BeamlineActionsContainer extends React.Component {
...
+            actionArguments={this.props.currentAction.arguments}
+            isActionRunning={currentActionRunning}
+            actionMessages={this.props.currentAction.messages}
+            handleSetActionArgument={this.props.setArgumentValue}
+            handleStopAction={this.props.stopAction}
+            handleStartAction={this.startAction}
+            handleOnPlotDisplay={this.newPlotDisplayed}
+            plotId={this.plotIdByAction[currentActionName]}
+          />
+        ) : null}
```
axelboc: Same here and below: with `&&` you remove the need for the `null` fallback.

ui/src/containers/BeamlineActionsContainer.jsx
```diff
@@ -122,98 +115,39 @@ class BeamlineActionsContainer extends React.Component {
...
+            actionSchema={this.props.currentAction.schema}
+            isActionRunning={currentActionRunning}
+            actionMessages={this.props.currentAction.messages}
+            handleSetActionArgument={this.props.setArgumentValue}
+            handleStopAction={this.props.stopAction}
+            handleStartAction={this.props.startAction}
+            handleOnPlotDisplay={this.newPlotDisplayed}
+            plotId={this.plotIdByAction[currentActionName]}
+          />
+        ) : null}
```
axelboc: Seems like there's a lot of duplication between these two components. Perhaps a good case for composition with `children`?

ui/src/components/BeamlineActions/AnnotatedBeamlineActionDialog.jsx
```diff
@@ -0,0 +1,110 @@
+import React, { useState } from 'react';
+import { Row, Col, Modal, Button, Card } from 'react-bootstrap';
+import Plot1D from '../Plot1D';
+import { DraggableModal } from '../DraggableModal';
+import JSForm from '@rjsf/core';
+import validator from '@rjsf/validator-ajv8';
+import style from './BeamlineActions.module.css';
```
axelboc: `styles` by convention

ui/src/components/BeamlineActions/AnnotatedBeamlineActionDialog.jsx
```diff
@@ -0,0 +1,110 @@
...
+    handleStartAction,
+    handleOnPlotDisplay,
+    plotId,
+  } = props;
+
+  function getRunButton(isActionRunning) {
+    if (isActionRunning) {
+      return (
+        <Button
+          className={style.beamline_action_submit_button}
```
axelboc: `camelCase` class names by convention also. The class names can be shorter too, since they get namespaced.

ui/src/containers/BeamlineActionsContainer.jsx
marcus-oscarsson: Hm, yeah why not. I find using `children` a bit unclear but thats perhaps the preferred way. I guess the are supposed to rendered in the order the are passed. Otherwise perhaps just creating some smaller components and reusing them in a more general dialog

axelboc: Yeah multiple ways to do this. The point is to abstract the beamline action dialog logic in a reusable component. How/where you use that dialog component is up to you. However, it's always good not to over optimise too soon, so I'd recommend starting by rendering that beamline action dialog once in `BeamlineActionContainer`, and rendering the two possible children inline via a ternary, also directly in `BeamlineActionContainer`.

marcus-oscarsson: Sounds very reasonable, I finally did the later it seemed more intuitive.

PR #1106, opened 2023-10-17
Update auto-generated documentation

fabcor-maxiv:
Change Sphinx theme to furo `https://pypi.org/project/furo/`.

Enable napoleon for handling of Google style docstrings.

Enable `show-inheritance` setting of `autodoc`.


fabcor-maxiv:
@marcus-oscarsson @rhfogh 

Preview here: https://mxcubeweb--1106.org.readthedocs.build/en/1106/

And here for an example of Python API (preview): https://mxcubeweb--1106.org.readthedocs.build/en/1106/dev/autosummary/mxcube3.core.adapter.adapter_base.html

What do you think of this theme? Is it more legible?

rhfogh:
Yes, this is clearer and more legible. Great Job!

rhfogh (approved):

fabcor-maxiv:
@marcus-oscarsson

This is ready for merge from my point of view.

Next I will update mxcubecore to get the same matching theme.

And then I guess we are ready to work on the content and I can help with migrating the wikis to sphinx (should be simple copy-paste work), I will open a ticket for this in case we need to discuss before acting, so that we are on the same page.

marcus-oscarsson:
Yeah, thats much easier to read :)

marcus-oscarsson:
By the way, I know that you played around with the themes and the last one was a great improvement. There also is this one that I think is also quite nice https://bashtage.github.io/sphinx-material/. Its maybe worth a try ?

fabcor-maxiv:
> By the way, I know that you played around with the themes and the last one was a great improvement. There also is this one that I think is also quite nice https://bashtage.github.io/sphinx-material/. Its maybe worth a try ?

It looks nice. Not sure if nicer than what we have now. Maybe it is worth opening a ticket so that we do not lose the idea.

PR #1110, opened 2023-10-19
Separate log formatter for terminal and file

marcus-oscarsson:
Separate log-formatter for terminal and file, so that the file does not end up with color coding.

### DIFF
--- a/mxcubeweb/app.py
+++ b/mxcubeweb/app.py
@@ -371,7 +371,8 @@ class MXCUBEApplication:
         removeLoggingHandlers()
 
         fmt = "%(asctime)s |%(name)-7s|%(levelname)-7s| %(message)s"
-        log_formatter = ColorFormatter(fmt)
+        console_formatter = ColorFormatter(fmt)
+        file_formatter = logging.Formatter(fmt)
 
         if log_file:
             if not os.path.isfile(log_file):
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
The pages build sometimes fails with some strange micromamba error, any ideas ?

fabcor-maxiv:
> The pages build sometimes fails with some strange micromamba error, any ideas ?

@marcus-oscarsson Where? Do you have a link to a failing job?

marcus-oscarsson:
Hey, working late I see ;), here you go https://github.com/mxcube/mxcubeweb/actions/runs/6576746692/job/17866831451. But this one is different from the ones I had previously

marcus-oscarsson:
This was probably another problem, I sometimes get a micromamba error at the install step, it seems like it works with just triggering the job again so its not a biggie.

fabcor-maxiv:
@marcus-oscarsson 

> https://github.com/mxcube/mxcubeweb/actions/runs/6576746692/job/17866831451

In this case it was during a "post setup" step which I have no control over :/


> it seems like it works with just triggering the job again so its not a biggie

Oh, good, but yes feel free to ping me if it happens again especially if it ever gets to a case where it is blocking work :)

marcus-oscarsson:
Yeah, as I said its not a big issue. Thanks for having a look

PR #1111, opened 2023-10-19
Applying precision in all task forms 

marcus-oscarsson:
Applying precision in all task forms:

 - Making sure that values are displayed according to the precision defined in the ui.yaml, previously not done for all forms

### DIFF
--- a/ui/src/components/Tasks/DataCollection.js
+++ b/ui/src/components/Tasks/DataCollection.js
@@ -383,22 +383,10 @@ export default connect((state) => {
     initialValues: {
       ...parameters,
       beam_size: state.sampleview.currentAperture,
-      resolution:
-        state.taskForm.sampleIds.constructor !== Array
-          ? state.taskForm.taskData.parameters.resolution
-          : toFixed(state, 'resolution'),
-      energy:
-        state.taskForm.sampleIds.constructor !== Array
-          ? state.taskForm.taskData.parameters.energy
-          : toFixed(state, 'energy'),
-      transmission:
-        state.taskForm.sampleIds.constructor !== Array
-          ? state.taskForm.taskData.parameters.transmission
-          : toFixed(state, 'transmission'),
-      osc_start:
-        state.taskForm.sampleIds.constructor !== Array
-          ? state.taskForm.taskData.parameters.osc_start
-          : toFixed(state, 'diffractometer.phi'),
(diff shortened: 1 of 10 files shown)


marcus-oscarsson:
oh and its going to be soo nice to get rid of the redux-forms :)

ui/src/components/Tasks/DataCollection.js
```diff
@@ -375,6 +375,12 @@ export default connect((state) => {
       ].acq_parameters.osc_range;
   }
 
+  // Use task data if samlpleID is a array (tasks added from the sample
+  // list view) or if we are using the option "rememberParametersBetweenSamples"
+  const useTaskData =
```
axelboc: I would maybe call this `withTaskData` to avoid confusion with React hooks.

axelboc: Also, I feel like this condition could be computed directly inside `toFixed` to avoid repetition. If the problem is that the hardware object name doesn't always match the form field name, then this can be solved by adding a third, optional argument:

```js
function toFixed(state, hoName, fieldName) {
  const withTaskData =
    state.taskForm.sampleIds.constructor !== Array ||
    state.queue.rememberParametersBetweenSamples;

  const value = withTaskData ? state.taskForm.taskData.parameters[fieldName] : state.beamline.hardwareObjects[hoName];

 // maybe a check that throws a clear error if `value` is `undefined` here

  let precision = ...;
  return value.toFixed(precision);
}
```

jbflo:
Are one of you guys working on getting rid of redux-form ?

marcus-oscarsson:
@jbflo: Eventually yes :), I think it will be @axelboc but it can maybe be shared work as soon as we have a clear structure ?

ui/src/components/Tasks/DataCollection.js
marcus-oscarsson: > I would maybe call this `withTaskData` to avoid confusion with React hooks.

I knew that you would say that :), but yeah you are right, Naming by convention in frameworks not sure I'm a fan :) Anyways, withTaskData works as well :)

marcus-oscarsson: Excellent idea @axelboc, so I finally did something in the lines of what you suggested, much nicer !

axelboc (approved):

PR #1112, opened 2023-10-19
Option to display native mesh menu item

marcus-oscarsson:
A very simple way to configure the display of the native mesh menu item. We are exclusively using the workflow mesh variant so it quickly becomes confusing for the users with the native one.

Id actually like to move the entire `.env` file to either the mxcube section of `server.yaml` or create a new `ui_options.yaml` (or similar) if you think it makes sense. Ill make an issue if you think we should discuss it further ?


meguiraun (approved):

meguiraun:
agree with moving away from env to site specific yml files. That would make easier managing differences.

Maybe the beamlineconfig.yml is the place, we are already using it for enable/disable some collections (enery scans, ??)

marcus-oscarsson:
@meguiraun, yeah thats probably a good idea at least for some of them. Ill have another think about it and Ill do something when I get some extra time :)

PR #1118, opened 2023-10-25
FIX Queue Manual sample Logic / 

jbflo:
In this PR we 
- Remove manual sample from queue when getting sample  from SC
- UPDATE_SAMPLE_LIST  suppose to Set the list of samples (sampleList), clearing any existing list

minor: we also update some code logic for better understanding 

The way to reproduce the bug:
1. Add a manual sample
2. click get sample from sample changer
4. refresh page or perform any action /

### DIFF
--- a/ui/src/actions/sampleGrid.js
+++ b/ui/src/actions/sampleGrid.js
@@ -40,7 +40,7 @@ export function addSamplesToList(samplesData) {
       if (!sampleData.sampleID) {
         lastSampleID++;
         sampleData.sampleID = lastSampleID.toString();
-        sampleData.cell_no = 1;
+        sampleData.cell_no = 0;
         sampleData.puck_no = 1;
       }
     }
@@ -53,6 +53,9 @@ export function setSampleOrderAction(order) {
   return { type: 'SET_SAMPLE_ORDER', order };
 }
 
+/**
+ * @deprecated New Sample Grid does not allowed sample reordering 
+ */
 export function sendSetSampleOrderAction(sampleOrder) {
   return (dispatch) => {
     fetch('mxcube/api/v0.1/queue/sample-order', {
--- a/ui/src/actions/sampleGrid.js
+++ b/ui/src/actions/sampleGrid.js
@@ -54,7 +54,7 @@ export function setSampleOrderAction(order) {
 }
 
 /**
- * @deprecated New Sample Grid does not allowed sample reordering 
+ * @deprecated New Sample Grid does not allowed sample reordering
  */
 export function sendSetSampleOrderAction(sampleOrder) {
   return (dispatch) => {
--- a/ui/src/containers/SampleGridTableContainer.jsx
+++ b/ui/src/containers/SampleGridTableContainer.jsx
@@ -400,7 +400,7 @@ class SampleGridTableContainer extends React.Component {
       .filter((sample) => sample.cell_no === cell)
       .forEach((sample) => {
         allCellSample.push(sample.sampleID);
-        if (this.props.inQueue(sample.sampleID) && sample.checked) {
+        if (this.props.inQueue(sample.sampleID)) {
           allCellSampleCheck.push(sample.sampleID);
         }
       });
@@ -423,7 +423,7 @@ class SampleGridTableContainer extends React.Component {
         .forEach((sample) => {
           if (this.props.filterSampleByKey(sample.sampleID)) {
             allCellSample.push(sample.sampleID);
-            if (this.props.inQueue(sample.sampleID) && sample.checked) {
+            if (this.props.inQueue(sample.sampleID)) {
               allCellSampleCheck.push(sample.sampleID);
             }
           }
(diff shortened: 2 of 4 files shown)


meguiraun:
Not sure if I get the purpose. Ideally we should be able to work with both manual samples and SC samples (eg room temp? broken sc?)

With these changes the manual samples are not rendered anymore (yet the exists in `mxcube/api/v0.1/queue/`) after getting samples from the SC. I would still want to see them.

I also get an error "Cannot read properties of undefined (reading 'checked')" pointing to `InQueue` method

jbflo:
@meguiraun did you check out the https://github.com/mxcube/mxcubeweb/tree/jb-develop-sg branch to test ?

because this Pr is mainly to fix the error "Cannot read properties of undefined (reading 'checked')".

The issues is:  when you click Get samples from SC button it removed the manual sample in the sampleList state but they are still in the queue. 

the InQueue function is double check : the queue state and sample list state 
  
as Manual sample are no longer in sampleList state after user click on  'Get samples from SC button' so 
  `this.props.sampleList[sampleID].checked` give undefined
  
so if they are not in the sample list state they should not be in the queue state.
 basically what I did now is to remove them 'manual samples' in the queue 
 
PS: another wise solution would be as you said  after getting samples from the SC. still keep the manual sample in sample list.
but UPDATE_SAMPLE_LIST in sampleGrid reducer 
have this commit from 7 years ago :  // Set the list of samples (sampleList), clearing any existing list

I CHECK the old Mxcube3 and this bug was there back then, it just the old JavaScript was not that strick to raise it (I guest). 

f you have the old mxcube running :
The way to reproduce the bug:
  Add a manual sample
  click get sample from sample changer
  refresh page or perform any action / 
  - you'll see the manual sample disappeared in the view,

meguiraun:
I will try again... I am (almost) sure I used your branch 🤷‍♀️... I guess I messed up somewhere

What I meant is that I dont want to make the manual sample disappear from the SamplesList view. Its part of the "history" of the session

jbflo:
We can double check, but manual sample disappear from the SamplesList view when  user click on 'Get samples from SC button' 
was  always behave like that. 

we can fix that if this behavior not OK?

meguiraun:
ping @JieNanMAXIV

meguiraun:
> We can double check, but manual sample disappear from the SamplesList view when user click on 'Get samples from SC button' was always behave like that.
> 
> we can fix that if this behavior not OK?

Always is a strong word :) ! I have the impression that this has been lost, in my view both manual samples and sc sample _used_ to coexist together in the UI. I will check with our scientist to what they expect to see

jbflo:
coexist is still the case See image bellow.

image
Do you have the old GUI running somewhere ? i made this little exercise yesterday , by checking.

meguiraun:
yes, they coexist... until you click "get samples from SC", at that point manual samples are removed. If you dont click ever, all is fine.

In the "old" mxcube, clicking "get samples from SC" does not remove the manual samples. And in my original MR for the Isara neither (I did play around with this, dont get me wrong 🙏🏻 I am not blaming, maybe I made it work into my view at that point).

In my opinion, we should not remove anything from queue "automatically"

jbflo:
You probably right @meguiraun
but Please my apologize if i doubt.

I checked 3 beamline running old GUI, before making this PR and i am positif , clicking "get samples from SC" does  remove the manual samples. sometime the page does not auto re-render but  after a refresh they go. 

[' clear sample list ' button does remove  remove everything  from queue "automatically" ]
that's why i did the same for manual sample.

Okay The only things matter now is to FIX the  "Cannot read properties of undefined (reading 'checked')" 
and the issues is cause by manual sample being removed from sampleList when clicking "get samples from SC" 

- You suggest we keep manual sample on both queue ans sampleList  when  "getting samples from SC" 
I'll have a look on this.

My solution was the easiest tho :smile:

meguiraun:
I think my "old" mxcube is older that what you looked at 😜 ...  our biomax codebase is reaaaaally old ... Maybe the functionality was lost along the way...

it would be great if you can make it work! thanks

jbflo:
I dig in,  i think this is a bit trickier then i thought 
That's probably why I applied this cheap solution. 

We are actually not saving Manual samples to the backend 

any refresh or state update -> SET_INITIAL_STATE is triggered 
that's mean sampleList and sampleOrder are getting state mirrored from backend state.

redux state is remove on refresh. so manual sample.

TO satisfy your suggestion @meguiraun  that would need to have manual sample in backend . 
@marcus-oscarsson I am wonderring why we never had this issues before 

 i converted this as a draft for now.

meguiraun:
I see :(

Then lets not make a big deal with it, considering the not straightforward implementation I am ok with leaving thing issue out. I checked with our scientist and I think they will be ok.

i happy to merge as it is

meguiraun (approved):

jbflo:
OK then,
Thanks ,  i remove the draft.

PR #1120, opened 2023-10-30
PART1 :  Updated for GPHL worklow  Generated Form 

jbflo:
In this PR we Remove specific code for GPHL worklow / Use and adapt existing code instead 
 -  cause These codes were ~98% similar to the code already exist for the execution of the workflow. i think there was no need to duplicated them. 

- Convert class to functional component (components/Tasks/Workflow.jsx &  containers/WorkflowParametersDialog.jsx)
- we Update react json schema form npm packages 

- Make WorkflowParametersDialog able to take json schema and UiSchema
- Temporary add  signal for gphl workflow Dialog

- I also fix a little error when cause syncSamples should wait getSamplesFromSC promise


ui/src/components/Tasks/Workflow.jsx
```diff
@@ -1,36 +1,24 @@
...
-  }
-
-  submitAddToQueue() {
-    this.props.handleSubmit(this.addToQueue.bind(this, false))();
-  }
-
-  submitRunNow() {
-    this.props.handleSubmit(this.addToQueue.bind(this, true))();
+function Workflow(props) {
+  function isGphlWorkflow() {
```
axelboc: this should be a constant :wink:

ui/src/components/Tasks/Workflow.jsx
```diff
@@ -46,71 +34,94 @@ class Workflow extends React.Component {
...
+    if (isGphlWorkflow) {
+      parameters.strategy_name = props.strategy_name;
+      stringFields.push('strategy_name');
+    }
+
+    props.addTask(parameters, stringFields, runNow);
+    props.hide();
+  }
+
+  const strategy_names = [];
```
axelboc:
```js
const strategyNames = isGphlWorflow
  ? Object.values(props.taskData.parameters.strategies).map(({ title }) => title)
  : [];
```

axelboc (commented):
Couple of nitpicks from a quick glance :blush:

mxcubeweb/core/adapter/beamline_adapter.py
```diff
@@ -37,13 +37,28 @@ def __init__(self, beamline_hwobj, app):
             gphl_workflow.connect(
```
rhfogh, line 37: This connection and the function it calls should be removed. The only  place where it is used is obsolete code, that will be changed to sue gphlJsonParametersNeeded soon.

mxcubeweb/core/adapter/beamline_adapter.py
```diff
@@ -37,13 +37,28 @@ def __init__(self, beamline_hwobj, app):
...
+                "GphlJsonParametersNeeded", self.gphl_json_wf_parameters_needed
+            )
+            gphl_workflow.connect(
+                "GphlUpdateUiParameters", self.gphl_json_wf_update_ui_parameters
+            )
 
     def wf_parameters_needed(self, params):
         self.app.server.emit("workflowParametersDialog", params, namespace="/hwr")
 
     def gphl_wf_parameters_needed(self, params):
```
rhfogh: This function should be removed - see previous comment

rhfogh (approved):
This looks right, and it runs in my run-it-through-once test. Not that I can really evaluate Javascript code, but for what I can say this should me merged (possibly after addressing my comment about gphlPrametersNeeded, but that need not hold it up).

mxcubeweb/core/components/workflow.py
```diff
@@ -32,7 +32,10 @@ def get_available_workflows(self):
         return {"workflows": workflows}
 
     def submit_parameters(self, params):
-        HWR.beamline.workflow.set_values_map(params)
+        if params["wfpath"] == "Gphl" or params["type"] == "GphlWorkflow":
+            HWR.beamline.gphl_workflow.set_values_map(params)
```
jbflo: @rhfogh 
IS this correct  : `HWR.beamline.gphl_workflow.set_values_map(params)`
I could not see any set_values_map function in /mxcubecore/mxcubecore/HardwareObjects/Gphl/GphlWorkflow.py

rhfogh: Well spotted. 

It is probably not correct. I suspect I added that at the start when I was (deliberately) duplicating the workflow code, and that this function is not and never was called. Could you  check: is the POST route for gphl_workflow ever used? Should it be? Again,  I suspect not, so both the route and the submit_parameters function could be removed. Once the GPhL workflow is started it does its own communication between MXCuBE and the workflow application, and handles communication to the UI via signals.

jbflo: Yes, from my Knowledge the UI/js can't emit SIgnal.
SO we need this route to submit the dialog parameters.
so I guess  we should replace this by : `self.app.server.emit("GphlParameterReturn", params, namespace="/hwr")`
a more fancy way. 

one more question : GphlUpdateUiParameters signal suppose to generated or updated the UI dialog parameters right ?

and GphlParameterReturn : suppose to send back to you the parameters from the UI to continue the workflow 
Right ?

PS: i removed the specific route for gphl_workflow the route to worflow is used.
I just check if the the workflow is GphlWorkflow and act accordingly.

rhfogh: The way it is organised, the mxcubecore side sends the GphlJsonParametersNeeded signal and listens for the  GphlParameterReturn signal until the UI is finished. The GphlUpdateUiParameters signal is also sent from the mxcubecore side.

The GphlParameterReturn serves two functions. It comes with two parameters, 'instruction' and 'parameters'. 'parameters' is a keyword-value dictionary of parameters. 'instruction' tells what whould be done with the parameters. If instruction is PARAMETERS_READY or PARAMETERS_CANCELLED the UI 'session' is terminated, eitehr successfully or unsuccessfully, and core side takes it from there. If instruction is anything else, it is interpreted as the name of a parameter that has been updated, and the parameters dictionary is passed to the releant update function which generates a response. When that response is ready, it is sent back to the UI with a GphlUpdateUiParameters signal.

So, the UI side must listen for GphlJsonParametersNeeded and GphlUpdateUiParameters, and be able to send GphlParameterReturn.

jbflo: Understood.
Very good i think i should start testing for now.

rhfogh (commented):
Looks fine to me (and the reduced duplication is great), but I do not have the JavaScript to judge (and it is not test-ready yet, is it?) Anyway, better have Axel approve it before merging?

jbflo:
Yes I would call this part 1 ready. there is maybe some work to do but this can wait in another PR.
The UI shchema will raise an error but that's normal . because we need to fix this in the mxcubecore
https://github.com/mxcube/mxcubecore/pull/815

ui/src/containers/SampleListViewContainer.jsx
jbflo, line 1: PS: I had to add this fix ASAP
This fix is an addition for https://github.com/mxcube/mxcubeweb/pull/1118
syncSamples() should wait getSamplesFromSC

jbflo:
@marcus-oscarsson  rebase Done !

PR #1122, opened 2023-11-01
add documentation URL to pyproject specifications

elmjag:
It seems that poetry started to insist that 'documentation' field must contain a valid URL. Poetry will refuse to install package with an empty 'documentation' field.

Let's point to our generated docs at readthedocs.io site.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ maintainers = [
 readme = "README.md"
 homepage = "http://github.com/mxcube/mxcubeweb"
 repository = "http://github.com/mxcube/mxcubeweb"
-documentation = ""
+documentation = "https://mxcubeweb.readthedocs.io/"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -10,7 +10,7 @@ maintainers = [
 readme = "README.md"
 homepage = "http://github.com/mxcube/mxcubeweb"
 repository = "http://github.com/mxcube/mxcubeweb"
-documentation = ""
+documentation = "https://mxcubeweb.readthedocs.io/en/latest/"
```
fabcor-maxiv: I would remove the `en/latest/` suffix.

fabcor-maxiv (changes requested):

fabcor-maxiv (approved):

marcus-oscarsson:
Thanks !

PR #1124, opened 2023-11-02
Decouple fetching layer from Redux, part 1: Login API

axelboc:
Currently, fetching is done mainly inside Redux async actions (or _thunks_) with the native `fetch` API.

Fetching with Redux has become a bit of an [anti-pattern](https://nosleepjavascript.com/redux-data-fetching-antipattern/) over the years. In MXCuBE we sometimes even dispatch an action for the sole purpose of sending an HTTP request (e.g. `refreshSession`), which is unnecessarily complex and tends to hide bugs (e.g. actions dispatched during render instead of via event handlers or `useEffect`).

This PR is a first step towards decoupling fetching from Redux, in the hope of introducing a more robust fetching layer in the long run (like [react-query](https://react-query.tanstack.com/)).

For now, however, I only:

#### 1) move the actual `fetch` calls of the Login API into a separate module

Having the fetch calls in a separate module and wrapped in very simple functions means we may now call them directly if relevant, without having to dispatch Redux actions.

It also makes the fetching layer easier to mock if we ever need to (e.g. if we introduce front-end integration tests), not to mention that it's a good way to "document" the back-end API in the front-end (i.e. developers can just look at the `api` modules to see which calls are available).

#### 2) take the opportunity to introduce a low-level fetch library called `wretch`

`fetch` is very cumbersome to use: every call has to include the full URL and all the required headers like `Content-type` or `Accept`); we have to manually stringify objects before posting; we have to write two consecutive `then` to get the response as JSON; we have to handle HTTP errors manually; etc.

It's common practice to bring in a low-level fetching library to deal with this. The most popular is probably [axios](https://github.com/axios/axios), but it has the downside of using the old `XMLHttpRequest` API under the hood. So I personally prefer [wretch](https://github.com/elbywan/wretch), which is basically a thin wrapper around the more recent `fetch` API. It has a very convenient and quite intuitive chainable API.

Using a low-level fetch library also opens the door to doing more complex things, like setting up a middleware to refresh a user's token and to retry requests automatically.

### DIFF
--- a/ui/src/actions/login.js
+++ b/ui/src/actions/login.js
@@ -1,10 +1,11 @@
 /* eslint-disable promise/catch-or-return */
 /* eslint-disable promise/no-nesting */
 /* eslint-disable promise/prefer-await-to-then */
-/* eslint-disable sonarjs/no-duplicate-string */
+
 import fetch from 'isomorphic-fetch';
 import { showErrorPanel, setLoading, getInitialState } from './general';
 import { serverIO } from '../serverIO'; // eslint-disable-line import/no-cycle
+import { fetchLoginInfo, logIn, signOut } from '../api/login';
 
 export function setLoginInfo(loginInfo) {
   return {
@@ -30,18 +31,6 @@ export function selectProposal(prop) {
     type: 'SELECT_PROPOSAL',
     proposal: prop,
   };
-}
-
-export function sendMail(sender, content) {
-  fetch('mxcube/api/v0.1/login/send_feedback', {
-    method: 'POST',
-    credentials: 'include',
-    headers: {
-      Accept: 'application/json',
-      'Content-type': 'application/json',
-    },
-    body: JSON.stringify({ sender, content }),
-  });
 }
 
 export function postProposal(number) {
@@ -77,47 +66,21 @@ export function startSession(userInControl) {
   };
 }
 
-export function refreshSession() {
-  return (dispatch) =>
-    fetch('mxcube/api/v0.1/login/refresh_session', {
-      method: 'GET',
-      headers: {
-        Accept: 'application/json',
-        'Content-type': 'application/json',
-      },
-      credentials: 'include',
-    });
-}
-
 export function getLoginInfo() {
   return (dispatch) =>
-    fetch('mxcube/api/v0.1/login/login_info', {
-      method: 'GET',
-      headers: {
-        Accept: 'application/json',
-        'Content-type': 'application/json',
+    fetchLoginInfo().then(
+      (loginInfo) => {
+        dispatch(setLoginInfo(loginInfo));
+        return loginInfo;
       },
-      credentials: 'include',
-    })
-      .then((response) => response.json())
-      .then(
-        (loginInfo) => {
-          dispatch(setLoginInfo(loginInfo));
-          return loginInfo;
-        },
-        () => {
-          dispatch(showErrorPanel(true));
-          dispatch(setLoading(false));
-        },
-      );
+      () => {
+        dispatch(showErrorPanel(true));
+        dispatch(setLoading(false));
+      },
+    );
 }
 
-export function signOut() {
-  localStorage.setItem('currentUser', '');
-  return { type: 'SIGNOUT' };
-}
-
-export function signIn(proposal, password, navigate) {
+export function doLogIn(proposal, password, navigate) {
   // eslint-disable-next-line sonarjs/cognitive-complexity
   return (dispatch) => {
     const previousUser = localStorage.getItem('currentUser');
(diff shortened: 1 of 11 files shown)


ui/src/actions/login.js
```diff
@@ -32,18 +33,6 @@ export function selectProposal(prop) {
   };
 }
 
-export function sendMail(sender, content) {
```
axelboc, line 35: Moved to `src/api/login.js` and renamed.

ui/src/actions/login.js
```diff
@@ -77,47 +66,21 @@ export function startSession(userInControl) {
   };
 }
 
-export function refreshSession() {
-  return (dispatch) =>
```
axelboc, line 81: No need for this to be a Redux action.

ui/src/actions/login.js
```diff
@@ -77,47 +66,21 @@ export function startSession(userInControl) {
...
-  localStorage.setItem('currentUser', '');
-  return { type: 'SIGNOUT' };
+      () => {
+        dispatch(showErrorPanel(true));
+        dispatch(setLoading(false));
+      },
+    );
 }
 
-export function signIn(proposal, password, navigate) {
```
axelboc, line 120: Trying to keep the naming consistent with the API. The `do` prefix helps distinguish between the Redux action creator and the new fetching function `logIn` in `/src/api/login.js`.

ui/src/api/index.js
```diff
@@ -0,0 +1,7 @@
+import wretch from 'wretch';
+
+const api = wretch('/mxcube/api/v0.1')
+  .options({ credendials: 'include' })
+  .headers({ Accept: 'application/json' });
```
axelboc, line 5: Here, I prepare a `wretch` instance with the base URL of the API server and with some default options/headers. This is where we might register middleware in the future.

ui/src/api/login.js
```diff
@@ -0,0 +1,23 @@
+import api from '.';
+
+const endpoint = api.url('/login');
```
axelboc, line 3: Here, I prepare another `wretch` instance (based on the base instance) configured specifically for the `/login` endpoint. The rest of this file is hopefully straightforward, but it's good to know a bit how wretch works:

- `get()`, `post()`, etc. call `fetch()` internally and return a `WretchResponseChain` object.
- To get the response data, we can chain this object with `json()` if the endpoint responds with JSON, or `blob()`, `text()`, etc.
- After that, we get a normal `Promise` with the data, which we can `await` or chain with `then()`/`catch()` as usual.

For endpoints that don't respond with any data, we call `.res()` (short for "response") instead of `json()` to ensure we still return a `Promise` (but with `res()`, the promise resolves to the raw `Response` object).

ui/src/actions/login.js
```diff
@@ -127,63 +90,56 @@ export function signIn(proposal, password, navigate) {
...
+    logIn(proposal, password, previousUser).then(
+      (res) => {
+        if (res.code === 'ok') {
+          dispatch(showErrorPanel(false));
+          dispatch(getLoginInfo())
+            .then((response) => response)
+            .then((resp) => {
+              if (resp.loginType === 'User') {
+                if (resp.user.inControl) {
+                  dispatch(showProposalsForm());
```
fabcor-maxiv, line 102: Question:
How come we are so deeply nested?
Wouldn't it be better with async/await? Is it something that has been considered already?

axelboc: Yes, definitely on the radar but I'd like to first go over the login logic and move it out of Redux if possible.

ui/src/api/login.js
marcus-oscarsson: Ok, nice :)

ui/src/actions/login.js
marcus-oscarsson: I think its a good idea as @axelboc said to do this in several steps to try to separate the fetching from the redux bit as much as we can and introducing an abstraction over `fetch` . The idea is then indeed, as we also said during the code camp, to start using `async` and `await`. I think the main reason that we started with the `then` for handling promises was that `async` and `await` was not widely adopted when the project was started. Then we just continued on that path.

ui/src/actions/login.js
marcus-oscarsson: Its not at all important and we can go with the `do` prefix, but I have to admit that I'm not a big fan of it :) (That is just my opinion :) ). Is it a common prefix to use in the React world ?

marcus-oscarsson:
Very nice start, I had one comment about the `do` prefix, but its not holding this back and can be changed later (if I'm not alone with that opinion).

ui/src/actions/login.js
axelboc: Not at all, it's just a general way to avoid naming conflicts with functions. More of a temporary work around in this case, as my goal is to see if I can remove this `doSignIn` action creator eventually.

axelboc: I should add that the main reason for having unique function names even across files is to facilitate search-and-replace operations.

That's why I also made sure to be consistent inside `mapDispatchToProps`: `doLogIn: bindActionCreators(doLogIn, dispatch)`. Otherwise, it's hard to find where actions are dispatched.

axelboc: Back to `doLogIn`, the problem is also that I couldn't find a better name for the `logIn` API function.

For `GET` requests that fetch resources, I'm planning on using the `fetch` prefix - e.g. `fetchLoginInfo`, but for API functions that send mutation requests, it's more complicated:

- using the HTTP method as prefix is a probably a bad idea: `getLoginInfo` (conflicts with the Redux action creator), `getRefreshSession` (makes no sense), `getSignOut` (same)...
- CRUD prefixes (`create`, `read`, `update`, `delete`) are way too generic;
- the `send` prefix is perhaps the most promising: `sendLogin`, `sendSignOut`, `sendRefreshSession` ... perhaps combined with CRUD prefixes for REST-compliant endpoints: `sendCreateSession` (fake example) — of course we could make an exception for `sendFeedback` to avoid `sendSendFeedback`...

ui/src/api/login.js
```diff
@@ -0,0 +1,23 @@
...
+
+export function logIn(proposal, password, previousUser) {
+  return endpoint.post({ proposal, password, previousUser }, '/').json();
+}
+
+export function signOut() {
+  return endpoint.headers({ Accept: '*/*' }).get('/signout').res();
```
axelboc, line 10: I meant to ask you about this `signout` endpoint @marcus-oscarsson. I've noticed that it [redirects to the `login` endpoint](https://github.com/mxcube/mxcubeweb/blob/7f9159b20f4e440c8123b4f285cdc74599ec5e71/mxcubeweb/routes/login.py#L70), which seems very weird and requires me to "unset" the `Accept: 'application/json` header. Shouldn't it just respond with a 200?

ui/src/actions/login.js
fabcor-maxiv: OK, of course, makes sense :)

ui/src/api/login.js
marcus-oscarsson: I see, I agree that the redirect is probably unnecessary. I'm not sure why its there. I suggest that we simply try without it :)

ui/src/actions/login.js
marcus-oscarsson: Sounds very reasonable :), I also like `send` and `fetch` for the reason that you mention above.

marcus-oscarsson:
Merging this for now, so that we can continue om this good path :)

ui/src/api/login.js
axelboc: I think it's trying to behave like a multi-page app by telling the browser to redirect to `/logout`, which obviously doesn't work since we use `fetch`.

PR #1125, opened 2023-11-07
Decouple fetching layer from Redux, part 2

axelboc:
I'm extracting a few fetch calls and cleaning up an unused action related to the Beamline API .

I'm also settling down on a naming convention and applying it to the Login API fetching layer introduced in #1124:
  - `fetch<Resource>` for `GET` requests that retrieve resources (e.g. `fetchLoginInfo`, `fetchBeamlineSetup`) — and `get<Resource>` for their corresponding Redux action creators,
  - `send<PerformAction>` for API endpoints that perform actions on the server regardless of the HTTP method used (e.g. `sendLogIn`, `sendRefreshSession`, `sendPrepareBeamlineForNewSample`) — and simply `<performAction>` for their corresponding Redux action creators.

### DIFF
--- a/mxcubeweb/routes/beamline.py
+++ b/mxcubeweb/routes/beamline.py
@@ -182,13 +182,13 @@ def init_route(app, server, url_prefix):
 
         :param str name: Owner / Actuator of the process/action to abort
 
-        Replies with status code 200 on success and 520 on exceptions.
+        Replies with status code 200 on success and 500 on exceptions.
         """
         try:
             app.beamline.beamline_abort_action(name)
         except Exception:
             err = str(sys.exc_info()[1])
-            return make_response(err, 520)
+            return make_response(err, 500)
         else:
             logging.getLogger("user_level_log").error("%s, aborted" % name)
             return make_response("{}", 200)
@@ -203,7 +203,7 @@ def init_route(app, server, url_prefix):
 
         :param str name: action to run
 
-        Replies with status code 200 on success and 520 on exceptions.
+        Replies with status code 200 on success and 500 on exceptions.
         """
         try:
             params = request.get_json()["parameters"]
@@ -213,7 +213,7 @@ def init_route(app, server, url_prefix):
         try:
             app.beamline.beamline_run_action(name, params)
         except Exception as ex:
-            return make_response(str(ex), 520)
+            return make_response(str(ex), 500)
         else:
             return make_response("{}", 200)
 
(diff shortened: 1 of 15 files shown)


mxcubeweb/routes/beamline.py
```diff
@@ -182,13 +182,13 @@ def beamline_abort_action(name):
 
         :param str name: Owner / Actuator of the process/action to abort
 
-        Replies with status code 200 on success and 520 on exceptions.
```
axelboc, line 185: 520 does not seem to be an official HTTP error code (it's used by Cloudflare, apparently). I think 500 is more appropriate?

ui/src/actions/beamline.js
```diff
@@ -46,30 +47,6 @@ export function updateBeamlineHardwareObjectStateAction(data) {
   return { type: BL_UPDATE_HARDWARE_OBJECT_STATE, data };
 }
 
-export function sendGetAllhardwareObjects() {
```
axelboc, line 49: From what I could tell, this action was not actually dispatched. The fetch call is duplicated and invoked directly in `general.js`.

ui/src/containers/BeamlineSetupContainer.jsx
```diff
@@ -323,10 +319,6 @@ function mapStateToProps(state) {
 
 function mapDispatchToProps(dispatch) {
   return {
-    getAllhardwareObjects: bindActionCreators(
```
axelboc, line 326: The unused action I mentioned was bound in this container but never dispatched from any of the child components (unless there's a weird prop access happening that I've missed, like `this.props['getAll' + 'hardwareObjects']`... :joy:).

marcus-oscarsson: I don't think we are accessing any prop like that, other-tings perhaps but not props :)

mxcubeweb/routes/beamline.py
marcus-oscarsson: Agreed

marcus-oscarsson:
Thumbs up for the naming convention, seems very reasonable to me. I guess we can make a separate PR for the error return code and the redirect that you spotted in the previous PR.

marcus-oscarsson:
Issue here: https://github.com/mxcube/mxcubeweb/issues/1126

marcus-oscarsson:
Looks good to me !

marcus-oscarsson (approved):

PR #1129, opened 2023-11-08
Decouple fetching layer from Redux, part 3

axelboc:
Run/abort (start/stop) beamline action:

- `mxcube/api/v0.1/beamline/${name}/run`
- `mxcube/api/v0.1/beamline/${name}/abort`

There were actually two action creators for aborting actions: one in `actions/beamline.js` (`sendAbortCurrentAction`) and one in `actions/beamlineActions.js` (`stopAction`). I've kept only the one in `actions/beamlineActions.js`, which I've also renamed to `stopBeamlineAction` for more clarity.

### DIFF
--- a/ui/src/containers/MotorInputContainer.js
+++ b/ui/src/containers/MotorInputContainer.js
@@ -1,12 +1,12 @@
 import React, { Component } from 'react';
 import { bindActionCreators } from 'redux';
 import { connect } from 'react-redux';
-import * as sampleViewActions from '../actions/sampleview'; // eslint-disable-line import/no-namespace
-import * as generalActions from '../actions/general'; // eslint-disable-line import/no-namespace
-import * as beamlineActions from '../actions/beamline'; // eslint-disable-line import/no-namespace
 import { QUEUE_RUNNING } from '../constants';
 
 import MotorInput from '../components/MotorInput/MotorInput';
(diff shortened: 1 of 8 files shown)


ui/src/containers/MotorInputContainer.js
```diff
@@ -1,12 +1,12 @@
 import React, { Component } from 'react';
 import { bindActionCreators } from 'redux';
 import { connect } from 'react-redux';
-import * as sampleViewActions from '../actions/sampleview'; // eslint-disable-line import/no-namespace
-import * as generalActions from '../actions/general'; // eslint-disable-line import/no-namespace
-import * as beamlineActions from '../actions/beamline'; // eslint-disable-line import/no-namespace
```
axelboc, line 6: Taking the opportunity to remove these namespace imports (note that `generalActions` was not actually used).

axelboc:
image

Faster than lightning :joy:

fabcor-maxiv:
@axelboc @marcus-oscarsson 

Seems like you are implementing some specific code architecture designs, naming conventions, and things like that, is that right? Have you documented them somewhere? If no, can you do it? Possibly in there: `https://github.com/mxcube/mxcubeweb/tree/develop/docs/source/dev`. :)

PR #1131, opened 2023-11-08
Decouple fetching layer from Redux, part 4

axelboc:
Last two remaining endpoints of the Beamline API: `sendSetAttribute` and `sendExecuteCommand`.

I've also written a tiny bit of developer documentation as you suggested, @fabcor-maxiv. I hope the RST syntax is right, I'm not familiar with it... Would it be feasible to switch to Markdown? :sweat_smile:

### DIFF
--- a/docs/source/dev/front-end.md
+++ b/docs/source/dev/front-end.md
@@ -1,0 +1,9 @@
+# Front-end architecture
+
+## Fetching layer
+
+- For each back-end API called by the front-end, there is a file under `src/api/` named after that API (e.g. `beamline.js`, `login.js` ...)
+- For each endpoint, there is a function in the API's front-end file dedicated to making HTTP requests to that endpoint (e.g. `fetchLoginInfo`, `sendRunBeamlineAction` ...)
+  - Functions that retrieve resources are named `fetch<Something>`.
+  - Functions that perform actions are named `send<DoSomething>`.
+- Library [wretch](https://github.com/elbywan/wretch) is used for making HTTP requests; it is a thin wrapper around the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API>).
--- a/ui/src/actions/beamline.js
+++ b/ui/src/actions/beamline.js
@@ -1,4 +1,3 @@
-/* eslint-disable sonarjs/no-duplicate-string */
 import fetch from 'isomorphic-fetch';
 import {
   sendSetAttribute,
@@ -71,19 +70,6 @@ export function executeCommand(obj, name, args) {
 
 export function prepareBeamlineForNewSample() {
   return () => sendPrepareBeamlineForNewSample();
-}
-
-export function sendDisplayImage(path) {
-  return () => {
-    fetch(`mxcube/api/v0.1/detector/display_image/?path=${path}`, {
-      method: 'GET',
-      credentials: 'include',
-      headers: {
-        Accept: 'application/json',
-        'Content-type': 'application/json',
-      },
-    });
-  };
 }
 
 export function sendLogFrontEndTraceBack(errorInfo, state) {
--- a/ui/src/actions/beamline.js
+++ b/ui/src/actions/beamline.js
@@ -1,6 +1,10 @@
 /* eslint-disable sonarjs/no-duplicate-string */
 import fetch from 'isomorphic-fetch';
-import { sendPrepareBeamlineForNewSample } from '../api/beamline';
+import {
+  sendSetAttribute,
+  sendExecuteCommand,
+  sendPrepareBeamlineForNewSample,
+} from '../api/beamline';
 // The different states a beamline attribute can assume.
 export const STATE = {
   IDLE: 'READY',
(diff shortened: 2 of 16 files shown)


fabcor-maxiv:
> I hope the RST syntax is right, I'm not familiar with it... Would it be feasible to switch to Markdown?

It supports both. It should work with any `.md` file. :)

https://www.sphinx-doc.org/en/master/usage/markdown.html

If not, let me know, I can probably help.

Preview: https://mxcubeweb--1131.org.readthedocs.build/en/1131/dev/front-end.html

Seems fine :)

docs/source/dev/front-end.rst
```diff
@@ -0,0 +1,11 @@
...
+######################
+
+Fetching layer
+**************
+
+- For each back-end API called by the front-end, there is a file under ``src/api/`` named after that API (e.g. ``beamline.js``, ``login.js`` ...)
+- For each endpoint, there is a function in the API's front-end file dedicated to making HTTP requests to that endpoint (e.g. ``fetchLoginInfo``, ``sendRunBeamlineAction`` ...)
+  - Functions that retrieve resources are named ``fetch<Something>``.
+  - Functions that perform actions are named ``send<DoSomething>``.
+- Library `wretch <https://github.com/elbywan/wretch>`_ is used for making HTTP requests; it is a thin wrapper around the `Fetch API <https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API>`_.
```
fabcor-maxiv: Needs empty line between different nesting levels:
https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#lists-and-quote-like-blocks

```suggestion
- For each back-end API called by the front-end, there is a file under ``src/api/`` named after that API (e.g. ``beamline.js``, ``login.js`` ...)
- For each endpoint, there is a function in the API's front-end file dedicated to making HTTP requests to that endpoint (e.g. ``fetchLoginInfo``, ``sendRunBeamlineAction`` ...)

  - Functions that retrieve resources are named ``fetch<Something>``.
  - Functions that perform actions are named ``send<DoSomething>``.

- Library `wretch <https://github.com/elbywan/wretch>`_ is used for making HTTP requests; it is a thin wrapper around the `Fetch API <https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API>`_.
```

fabcor-maxiv (dismissed):

docs/source/dev/front-end.rst
axelboc: Oh great, well, if you don't mind, I've switched to Markdown :grin:

axelboc: Love the deploy previews of the documentation site, by the way :trophy:

axelboc:
GitHub seems to consider canceled CI jobs as failed. Anyway to change that?

fabcor-maxiv:
> GitHub seems to consider canceled CI jobs as failed. Anyway to change that?

There is some kind of issue about actions workflow being triggered by both the `push` and the `pull request`. I will try to find a fix.

But anyway I do not think it was blocking the review and merge. I had made a review "Changes requested" but then you moved from `.rst` to `.md` so the change is not necessary anymore. I have now dismissed this review that I had made.

axelboc:
> There is some kind of issue about actions workflow being triggered by both the `push` and the `pull request`. I will try to find a fix.

This is how I deal with this in the UI workflow: https://github.com/mxcube/mxcubeweb/blob/develop/.github/workflows/ui.yml#L11

Basically, the `pull_request` trigger is required only for PRs from forks for which the `push` trigger doesn't run, so I skip the `pull_request` trigger when it's not required (i.e. for local PRs)

fabcor-maxiv:
> > There is some kind of issue about actions workflow being triggered by both the `push` and the `pull request`. I will try to find a fix.
> 
> This is how I deal with this in the UI workflow: https://github.com/mxcube/mxcubeweb/blob/develop/.github/workflows/ui.yml#L11
> 
> Basically, the `pull_request` trigger is required only for PRs from forks for which the `push` trigger doesn't run, so I skip the `pull_request` trigger when it's not required (i.e. for local PRs)

As far as I understood this safeguard is not necessary if we keep our pull request branches on our own personal forks instead of on the upstream repository directly. This is a bit strange to me to push branches directly on the upstream repository on GitHub.

fabcor-maxiv:
Let's see if it happens again with this: https://github.com/mxcube/mxcubeweb/pull/1133

marcus-oscarsson:
@fabcor-maxiv #1133 seems to have done the trick

axelboc:
> As far as I understood this safeguard is not necessary if we keep our pull request branches on our own personal forks instead of on the upstream repository directly. This is a bit strange to me to push branches directly on the upstream repository on GitHub.

We are maintainers of the project with write access, so it feels natural to me. :thinking: The simplest process is often the best! :grin: Forks are typically more for external contributors, or when you want to start your own version of a project.

fabcor-maxiv:
>  #1133 seems to have done the trick

Great! :)

---

For the sake of the argument...

In my experience working with forks is more natural especially on forges like GitHub, GitLab, and so on. GitHub is built around this workflow with forks. For me working with the upstream directly is kind of like `sudo` mode. The other advantage of working with forks exclusively is that the upstream repository stays clean with typically only 2 branches `develop` and `main` (and sometimes some long-lived feature branches that need to be shared with all but it is rare).

But I don't mind much either way. In this case I can keep doing pull requests from my own fork.

PR #1132, opened 2023-11-08
Remove most namespace action imports

axelboc:
Those namespace imports make it that little bit harder to find where each action creator is used, and to trace back where it's bound from the Redux store.

I managed to remove all of them (as recommended by ESLint) except for `sampleViewActions` in `SampleViewContainer`, which wraps tens of action creators, which are used in multiple child components... However, I did rename the prop that is used to pass it down to child components from `sampleActions` to `sampleViewActions` so it's consistent and easier to search for.

### DIFF
--- a/ui/src/actions/beamline.js
+++ b/ui/src/actions/beamline.js
@@ -1,4 +1,3 @@
-/* eslint-disable sonarjs/no-duplicate-string */
 import fetch from 'isomorphic-fetch';
 import {
   sendSetAttribute,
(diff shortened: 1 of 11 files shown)


ui/src/actions/beamline.js
```diff
@@ -73,19 +72,6 @@ export function prepareBeamlineForNewSample() {
...
-      method: 'GET',
-      credentials: 'include',
-      headers: {
-        Accept: 'application/json',
-        'Content-type': 'application/json',
-      },
-    });
-  };
-}
-
```
axelboc, line 88: Duplicate of an action creator located in `actions/general.js`.

ui/src/containers/SampleQueueContainer.js
```diff
@@ -4,56 +4,35 @@ import { connect } from 'react-redux';
...
+  showList,
+} from '../actions/queueGUI';
 import { showTaskForm } from '../actions/taskForm';
 import { Nav } from 'react-bootstrap';
 import { showDialog } from '../actions/general';
 
 import UserMessage from '../components/Notify/UserMessage';
 import loader from '../img/loader.gif';
 import { prepareBeamlineForNewSample } from '../actions/beamline';
-
```
axelboc, line 18: Ugly diff because I've moved this down just above the call to `connect`, like in every other file.

PR #1133, opened 2023-11-08
Remove `cancel-in-progress` in `pages` workflow

fabcor-maxiv:
This should prevent cancelled workflows to be considered as failed from the point of view of the pull requests checks.


fabcor-maxiv:
I normally do not push directly on the upstream repository (`origin`), but this is necessary in this case.

If it works fine, we will do the same change on `mxcubecore`.

PR #1134, opened 2023-11-09
Decouple fetching layer from Redux, part 5

axelboc:
Diffractometer API


ui/src/actions/sampleview.js
```diff
@@ -677,23 +669,9 @@ export function getMotorPosition(motor) {
...
-      credentials: 'include',
-      headers: {
-        Accept: 'application/json',
-        'Content-type': 'application/json',
-      },
-    })
-      .then((response) => {
-        if (response.status >= 400) {
-          throw new Error('Server refused to get motor position');
-        }
```
axelboc, line 691: To clarify, `wretch` already throws when receiving HTTP error codes.

ui/src/containers/SampleViewContainer.js
```diff
@@ -96,7 +96,7 @@ class SampleViewContainer extends Component {
         <PhaseInput
           phase={this.props.sampleViewState.currentPhase}
           phaseList={this.props.sampleViewState.phaseList}
-          sendPhase={this.props.sampleViewActions.sendCurrentPhase}
+          sendPhase={this.props.sampleViewActions.changeCurrentPhase}
```
marcus-oscarsson: What do you think about changing `sendPhase` to `changePhase` as well  ?

axelboc: I noticed it when I worked on #1136  :wink:

marcus-oscarsson: Oh yes, I noticed just now :)

PR #1136, opened 2023-11-09
Fix "Phase Control" rendered twice

axelboc:
The "Phase Control" drop-down was rendered twice: once at the top of the sidebar and once at the bottom, under the "Show motors" toggle. I'm keeping only the one at the top.

> The second instance was actually rendered with the wrong condition because it assumed that the env var was a boolean, when it's actually a string (so it would be rendered, for instance, with `REACT_APP_PHASECONTROL=false`).

While I'm at it:

- I rename `MotorControl` to `MotorControls` to avoid confusion, since the component renders multiple motor controls;
- I refactor `MotorControls` to simplify the rendering logic and remove duplication.
- I fix the missing chevron icon on the "Phase Control" drop-down (`PhaseInput`).
- I rename a couple of props and event handlers in `PhaseInput` and `ApertureInput`.

image


marcus-oscarsson:
Excellent looks, good to me !

marcus-oscarsson (approved):

PR #1137, opened 2023-11-10
Decouple fetching layer from Redux, part 7

axelboc:
Queue API : fetch queue state, fetch available tasks, start/stop/pause/resume queue


marcus-oscarsson:
Thanks, wonderful !

PR #1138, opened 2023-11-20
Removed call gphl_workflow that was removed in previous PR ( #1120) but re-added in  (#1120)

marcus-oscarsson:
The  /gphl_workflow resource was merged with /workflow in  PR ( #1120)  but was added back by a refactoring PR  (#1120) merged after. Simply removing the prior as its no longer needed.


rhfogh (approved):
Fine with me.

marcus-oscarsson:
:), thanks @jbflo does it make sense to you ?

jbflo:
> :), thanks @jbflo does it make sense to you ?

Yes Thanks a lot! :smile:

PR #821, opened 2019-02-18
Fixes for Python3 and new HWR API

marcus-oscarsson:
Hi,

This fixes most of the failing test cases, there or two still failing.

Marcus

PR #1139, opened 2023-11-22
Moved the configuration from the `.env` file to `beamline-configuration.yml` and `ui.yml`

marcus-oscarsson:
Moved the configuration from the `.env` file to `beamline-configuration.yml` and `ui.yml`. At the same time improved the handling of the data model for `ui.yml`. 

Each section can now have its own data model, which is needed since the server check objects types as `UIComponentModel` for the actual `HardwareObject` correspondens, leading to the following error message for the `camera_setup` section in `ui.yml`: 

image

Needs [PR #821](https://github.com/mxcube/mxcubecore/pull/821) on mxcubecore to work


mxcubeweb/core/models/configmodels.py
```diff
@@ -48,13 +48,37 @@ class UIComponentModel(BaseModel):
...
+
+class UISampleViewVideoControlsModel(UIPropertiesModel):
+    components: List[_UISampleViewVideoControlsModel]
+
+
 class UIPropertiesListModel(BaseModel):
-    __root__: Dict[str, UIPropertiesModel]
+    sample_view: UIPropertiesModel
+    beamline_setup: UIPropertiesModel
+    camera_setup: UICameraConfigModel
```
marcus-oscarsson, line 80: The `UIComponentModel` is checked against the `HardwareObject` it configures to avoid mistakes. However the `this caused a warning for the objects in the `camera_setup` section

mxcubeweb/core/models/generic.py
```diff
@@ -11,3 +11,20 @@ class SimpleNameValue(BaseModel):
 class AppSettingsModel(BaseModel):
     mode: ModeEnum = Field(ModeEnum.OSC, description="MXCuBE mode SSX or OSC")
     version: str = Field("", description="MXCuBE version")
+    mesh_result_format: str = Field(
```
marcus-oscarsson, line 14: Most of the setting from the `.env` file was moved to general application settings, configured via `beamline-config.yaml`

test/HardwareObjectsMockup.xml/beamline_config.yml
```diff
@@ -78,76 +78,81 @@ run_processing_parallel: true
 run_number: 1
 click_centring_num_clicks: 3
 
+mesh_result_format: PNG
```
marcus-oscarsson, line 81: New configuration added here

test/HardwareObjectsMockup.xml/mxcube-web/ui.yaml
```diff
@@ -1,146 +1,142 @@
...
-    -
-      label: Focus
+    - label: Focus
       attribute: diffractometer.sampx
       role: focus
       step: 0.1
       precision: 3
       suffix: mm
 
+sample_view_video_controls:
```
marcus-oscarsson, line 77: The option to display the focus control on the video display was moved to `ui.yml` and the same option was added to all controls (in case it would be needed)

mxcubeweb/core/models/generic.py
```diff
@@ -11,3 +11,20 @@ class SimpleNameValue(BaseModel):
...
+        "PNG", description="Format of mesh result for display"
+    )
+    use_native_mesh: bool = Field(
+        True, description=" Use the native mesh feature available, true by default"
```
meguiraun: please remind me what this one meant... this would be the md3 command? no workflow?

test/HardwareObjectsMockup.xml/mxcube-web/ui.yaml
meguiraun: are you thinking on potentially hide the 3 click, zoom. etc?

ui/src/containers/SampleViewContainer.js
```diff
@@ -41,8 +41,26 @@ import {
...
+import { find } from 'lodash';
 
 class SampleViewContainer extends Component {
+  constructor(props) {
+    super(props);
+    this.getControlAvailability = this.getControlAvailability.bind(this);
+  }
+
+  getControlAvailability(name) {
+    const avilable = find(
```
meguiraun: typo

meguiraun (approved):
nice and useful addition! I left ew minor comments/questions

tack,

mxcubeweb/core/models/generic.py
marcus-oscarsson: Exactly, so its basically just to disable the native mesh method so that its not available to the users. Its useful for instance for us as we have the workflow meshes, and it would mean that there are two kinds of meshes and that is of course a bit confusing. So this option is simply to disable the native one and favor the workflow one.

meguiraun: good, less confusion is always good :)

test/HardwareObjectsMockup.xml/mxcube-web/ui.yaml
marcus-oscarsson: Exactly, I thought it was better to simply generalize the visibility of these controls, as there already was one that could be hidden/shown. Its less likely for some controls but it is useful for a few especially for SSX

ui/src/containers/SampleViewContainer.js
marcus-oscarsson: :), thanks

ui/src/containers/SampleViewContainer.js
```diff
@@ -41,8 +41,26 @@ import {
...
+  getControlAvailability(name) {
+    const available = find(
+      this.props.uiproperties.sample_view_video_controls.components,
+      {
+        id: name,
+        show: true,
+      },
+    ) || { show: false };
+
+    return available.show;
```
axelboc: I'd move the fallback here: `return available?.show || false`

marcus-oscarsson: oh, yes thats nice :)

marcus-oscarsson:
Thanks for the input everybody ! :)

PR #1140, opened 2023-11-22
Separated login and login_info routes

marcus-oscarsson:
Separated `login` and `login_info` routes. Returning `401 - Unauthorized` from login_info instead of redirect. Removed the call to `requireAuth` from the `PrivateOutlet` component as it runs during rendering. 

`requireAuth` is now executed on page load (`useEffect` of `App`) and on change of route using the loader functionality of `react-router`. The later makes the transition from login to application much smoother, as no reload of the page is needed. 

The proposal selection dialog was also separated from the Login component and displayed at login when there is no proposal already selected. The user can select a proposal from the menu in the top right corner. 

Peek 2023-11-22 16-29

### DIFF
--- a/mxcubeweb/core/components/user/usermanager.py
+++ b/mxcubeweb/core/components/user/usermanager.py
@@ -181,8 +181,6 @@ class BaseUserManager(ComponentBase):
             msg = "User %s signed in" % user.username
             logging.getLogger("MX3.HWR").info(msg)
 
-            return login_res["status"]
-
     # Abstract method to be implemented by concrete implementation
     def _signout(self):
         pass
@@ -212,6 +210,7 @@ class BaseUserManager(ComponentBase):
 
         self.app.server.user_datastore.deactivate_user(user)
         flask_security.logout_user()
+
         self.emit_observers_changed()
 
     def is_authenticated(self):
@@ -227,22 +226,6 @@ class BaseUserManager(ComponentBase):
             self.app.server.emit("forceSignout", room=socketio_sid, namespace="/hwr")
 
     def login_info(self):
-        res = {
-            "synchrotronName": HWR.beamline.session.synchrotron_name,
-            "beamlineName": HWR.beamline.session.beamline_name,
-            "loggedIn": False,
-            "loginType": HWR.beamline.lims.loginType.title(),
-            "proposalList": [],
-            "user": {
-                "username": "",
-                "email": "",
-                "isstaff": False,
-                "nickname": "",
-                "inControl": False,
-                "ip": "",
-            },
-        }
-
         if not current_user.is_anonymous:
             login_info = convert_to_dict(json.loads(current_user.limsdata))
 
@@ -275,8 +258,10 @@ class BaseUserManager(ComponentBase):
             )
 
             res["selectedProposalID"] = HWR.beamline.session.proposal_id
+        else:
+            raise Exception("Not logged in")
 
-        return current_user, res
+        return res
 
     def update_user(self, user):
         self.app.server.user_datastore.put(user)
--- a/mxcubeweb/routes/login.py
+++ b/mxcubeweb/routes/login.py
@@ -83,7 +83,7 @@ def init_route(app, server, url_prefix):
             res = app.usermanager.login_info()
             response = jsonify(res)
         except Exception:
-            response = make_response(jsonify(""), 401)
+            response = make_response(jsonify({"loggedIn": False}), 200)
 
         return response
 
--- a/mxcubeweb/routes/login.py
+++ b/mxcubeweb/routes/login.py
@@ -5,17 +5,13 @@ from flask import (
     request,
     jsonify,
     make_response,
-    redirect,
-    session,
 )
 from mxcubeweb.core.util import networkutils
 from flask_login import current_user
 
 
 def deny_access(msg):
-    resp = jsonify({"msg": msg})
-    resp.code = 409
-    return resp
+    return make_response(jsonify({"msg": msg}), 200)
 
 
 def init_route(app, server, url_prefix):
@@ -45,7 +41,7 @@ def init_route(app, server, url_prefix):
         password = params.get("password", "")
 
         try:
-            res = jsonify(app.usermanager.login(login_id, password))
+            app.usermanager.login(login_id, password)
         except Exception as ex:
             msg = "[LOGIN] User %s could not login (%s)" % (
                 login_id,
@@ -54,8 +50,8 @@ def init_route(app, server, url_prefix):
             logging.getLogger("MX3.HWR").exception("")
             logging.getLogger("MX3.HWR").info(msg)
             res = deny_access("Could not authenticate")
-
-        session.permanent = True
+        else:
+            res = make_response(jsonify({"msg": ""}), 200)
 
         return res
 
@@ -66,8 +62,7 @@ def init_route(app, server, url_prefix):
         Signout from MXCuBE Web and reset the session
         """
         app.usermanager.signout()
-
-        return redirect("/login", code=302)
+        return make_response(jsonify(""), 200)
 
     @bp.route("/login_info", methods=["GET"])
     def login_info():
@@ -88,16 +83,14 @@ def init_route(app, server, url_prefix):
 
         Status code set to:
         200: On success
-        409: Error, could not log in
+        401: Error, could not log in
         """
-        user, res = app.usermanager.login_info()
 
-        # Redirect the user to login page if for some reason logged out
-        # i.e. server restart
-        if not user:
-            response = redirect("/login", code=302)
-        else:
+        try:
+            res = app.usermanager.login_info()
             response = jsonify(res)
+        except Exception:
+            response = make_response(jsonify(""), 401)
 
         return response
 
@@ -112,6 +105,7 @@ def init_route(app, server, url_prefix):
     @bp.route("/refresh_session", methods=["GET"])
     @server.restrict
     def refresh_session():
+        print("HERE TOOO")
         # Since default value of `SESSION_REFRESH_EACH_REQUEST` config setting is `True`
         # there is no need to do anything to refresh the session.
         logging.getLogger("MX3.HWR").debug("Session refresh")
--- a/mxcubeweb/routes/login.py
+++ b/mxcubeweb/routes/login.py
@@ -10,10 +10,6 @@ from mxcubeweb.core.util import networkutils
 from flask_login import current_user
 
 
-def deny_access(msg):
-    return make_response(jsonify({"msg": msg}), 200)
-
-
 def init_route(app, server, url_prefix):
     bp = Blueprint("login", __name__, url_prefix=url_prefix)
 
@@ -49,7 +45,7 @@ def init_route(app, server, url_prefix):
             )
             logging.getLogger("MX3.HWR").exception("")
             logging.getLogger("MX3.HWR").info(msg)
-            res = deny_access("Could not authenticate")
+            res = make_response(jsonify({"msg": "Could not authenticate"}), 200)
         else:
             res = make_response(jsonify({"msg": ""}), 200)
 
--- a/mxcubeweb/routes/login.py
+++ b/mxcubeweb/routes/login.py
@@ -68,7 +68,7 @@ def init_route(app, server, url_prefix):
         {'synchrotron_name': synchrotron_name,
         'beamline_name': beamline_name,
         'loginType': loginType,
-        'loginRes': {'status':{ 'code': 'ok', 'msg': msg },
+        'loggedIn': True,
         'Proposal': proposal, 'session': todays_session,
         'local_contact': local_contact,
         'person': someone,
@@ -76,7 +76,7 @@ def init_route(app, server, url_prefix):
 
         Status code set to:
         200: On success
-        401: Error, could not log in
+        200: Error, could not log in, {"loggedIn": False}
         """
 
         try:
--- a/mxcubeweb/routes/login.py
+++ b/mxcubeweb/routes/login.py
@@ -1,11 +1,6 @@
 import logging
 
-from flask import (
-    Blueprint,
-    request,
-    jsonify,
-    make_response,
-)
+from flask import Blueprint, request, jsonify, make_response, session
 from mxcubeweb.core.util import networkutils
 from flask_login import current_user
 
@@ -48,6 +43,8 @@ def init_route(app, server, url_prefix):
             res = make_response(jsonify({"msg": "Could not authenticate"}), 200)
         else:
             res = make_response(jsonify({"msg": ""}), 200)
+
+        session.permanent = True
 
         return res
 
--- a/mxcubeweb/routes/login.py
+++ b/mxcubeweb/routes/login.py
@@ -43,8 +43,7 @@ def init_route(app, server, url_prefix):
             res = make_response(jsonify({"msg": "Could not authenticate"}), 200)
         else:
             res = make_response(jsonify({"msg": ""}), 200)
-
-        session.permanent = True
+            session.permanent = True
 
         return res
 
--- a/mxcubeweb/routes/login.py
+++ b/mxcubeweb/routes/login.py
@@ -105,7 +105,6 @@ def init_route(app, server, url_prefix):
     @bp.route("/refresh_session", methods=["GET"])
     @server.restrict
     def refresh_session():
-        print("HERE TOOO")
         # Since default value of `SESSION_REFRESH_EACH_REQUEST` config setting is `True`
         # there is no need to do anything to refresh the session.
         logging.getLogger("MX3.HWR").debug("Session refresh")
--- a/test/test_authn.py
+++ b/test/test_authn.py
@@ -100,7 +100,8 @@ def test_authn_info(client, login_type):
     and true after successful authentication.
     """
     resp = client.get(URL_INFO)
-    assert resp.status_code == 401
+    assert resp.status_code == 200
+    assert resp.json["loggedIn"] == False
 
     client.post(URL_SIGNIN, json=CREDENTIALS_0)
 
@@ -184,7 +185,8 @@ def test_authn_session_timeout(client):
 
     # Check that the session has expired
     resp = client.get(URL_INFO)
-    assert resp.status_code == 401, "Session did not expire"
+    assert resp.status_code == 200
+    assert resp.json["loggedIn"] == False, "Session did not expire"
 
     # Check that it is possible to sign in again
     client.post(URL_SIGNIN, json=CREDENTIALS_0)
--- a/test/test_authn.py
+++ b/test/test_authn.py
@@ -92,7 +92,7 @@ def test_authn_signout(client):
 
     resp = client.get(URL_SIGNOUT)
     assert resp.status_code == 200
-    assert resp.json["msg"] == ""
+    assert resp.json == ""
 
 
 def test_authn_info(client, login_type):
--- a/test/test_authn.py
+++ b/test/test_authn.py
@@ -91,6 +91,7 @@ def test_authn_signout(client):
 
     resp = client.get(URL_SIGNOUT)
     assert resp.status_code == 200
+    assert resp.json["msg"] == ""
 
 
 def test_authn_info(client, login_type):
--- a/test/test_authn.py
+++ b/test/test_authn.py
@@ -77,8 +77,6 @@ def client(make_client):
 def test_authn_signin_good_credentials(client):
     resp = client.post(URL_SIGNIN, json=CREDENTIALS_0)
     assert resp.status_code == 200
-    assert resp.json["code"] == "ok"
-    assert resp.json["msg"] == "Successful login"
 
 
 def test_authn_signin_wrong_credentials(client):
@@ -90,11 +88,9 @@ def test_authn_signin_wrong_credentials(client):
 
 def test_authn_signout(client):
     resp = client.post(URL_SIGNIN, json=CREDENTIALS_0)
-    assert resp.json["code"] == "ok"
 
     resp = client.get(URL_SIGNOUT)
-    assert resp.status_code == 302
-    assert resp.headers["Location"] == "/login"
+    assert resp.status_code == 200
 
 
 def test_authn_info(client, login_type):
@@ -104,8 +100,7 @@ def test_authn_info(client, login_type):
     and true after successful authentication.
     """
     resp = client.get(URL_INFO)
-    assert resp.status_code == 200
-    assert resp.json["loggedIn"] == False
+    assert resp.status_code == 401
 
     client.post(URL_SIGNIN, json=CREDENTIALS_0)
 
@@ -127,13 +122,14 @@ def test_authn_same_proposal(make_client):
 
     client_0 = make_client()
     resp = client_0.post(URL_SIGNIN, json=CREDENTIALS_0)
-    assert resp.json["code"] == "ok"
+
+    assert resp.status_code == 200
     resp = client_0.get(URL_INFO)
     assert resp.json["user"]["inControl"] == True
 
     client_1 = make_client()
     resp = client_1.post(URL_SIGNIN, json=CREDENTIALS_0)
-    assert resp.json["code"] == "ok"
+    assert resp.status_code == 200
     resp = client_1.get(URL_INFO)
     assert resp.json["user"]["inControl"] == False
 
@@ -146,12 +142,11 @@ def test_authn_different_proposals(make_client):
     If a user signs in for a different proposal than an already signed in user,
     this user should not be allowed to sign in.
     """
-
     client_0 = make_client()
     resp = client_0.post(URL_SIGNIN, json=CREDENTIALS_0)
-    assert resp.json["code"] == "ok"
+    assert resp.status_code == 200
     resp = client_0.get(URL_INFO)
-    assert resp.json["user"]["inControl"] == True
+    assert resp.status_code == 200
 
     client_1 = make_client()
     resp = client_1.post(URL_SIGNIN, json=CREDENTIALS_1)
@@ -181,6 +176,7 @@ def test_authn_session_timeout(client):
 
     # Check that the session still has not expired
     resp = client.get(URL_INFO)
+    assert resp.status_code == 200
     assert resp.json["loggedIn"] == True, "Session did not refresh"
 
     # Let the session expire completely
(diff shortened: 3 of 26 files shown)


meguiraun:
thanks, just few questions (does not qualify as a proper review)

> The user can select a proposal from the menu in the top right corner.

Is there any situation where the user can play around without selecting a proposal?

And, I am not so sure if we want to change proposals in the middle of the beamtime without a proper logout. (ping @JieNanMAXIV )

fabcor-maxiv (dismissed):
I guess we will want to properly test this at MAX IV before this gets merged. That is why I put a provisional negative feedback for now, even though I have not reviewed the code yet. And obviously I understand that it is still a draft.

Rework of the authentication code is definitely welcome, it was not straightforward before, thanks for working on this :)

marcus-oscarsson:
@meguiraun: I don't think that its possible to proceed without selecting a proposal :). The "change" proposal dialog can of-course be made optional. We will need a feature like this for the automated beamlines where you can collect as several proposals during one session.

@fabcor-maxiv: Sure, understandable take your time and test it. This is the first step there will be a few more updates at a later moment. I'm actually struggling a bit with the session timeout test, while it works in the browser it fails in the test. Maybe you have an idea of what it could be ?

mxcubeweb/routes/login.py
```diff
@@ -5,17 +5,13 @@
...
 )
 from mxcubeweb.core.util import networkutils
 from flask_login import current_user
 
 
 def deny_access(msg):
-    resp = jsonify({"msg": msg})
-    resp.code = 409
-    return resp
+    return make_response(jsonify({"msg": msg}), 200)
```
axelboc: Maybe there's a reason I'm not seeing, but I'm surprised to see a 200 response code inside a function called `deny_access`. 🤔 Did you mean [403 Forbidden](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) (which means the user is not "authorized" to access a resource)?

ui/src/actions/login.js
```diff
@@ -67,50 +67,39 @@ export function startSession(userInControl) {
...
+        setLoginInfo({
+          beamlineName: '',
+          synchrotronName: '',
+          loginType: '',
+          user: '',
+          proposalList: [],
+          selectedProposal: '',
+          selectedProposalID: '',
+          loggedIn: false,
+          rootPath: '',
```
axelboc: This should be an `INITIAL_STATE` constant imported from the reducer file to avoid duplication. Or better, a `resetLoginInfo` action.

ui/src/actions/login.js
```diff
@@ -67,50 +67,39 @@ export function startSession(userInControl) {
...
-      console.log(serverIO.hwrSocket.connected); // eslint-disable-line no-console
-    } else {
-      serverIO.connect();
-    }
-
-    sendLogIn(proposal, password, previousUser).then(
+    sendLogIn(proposal, password).then(
       (res) => {
-        if (res.code === 'ok') {
+        if (res.msg === '') {
```
axelboc: This condition should no longer be needed, since 200 (i.e. no error) assumes success and error response codes are handled by `catch` logic below.

ui/src/actions/login.js
```diff
@@ -125,23 +114,18 @@ export function logIn(proposal, password, navigate) {
...
 }
 
 export function signOut(navigate) {
   return (dispatch) => {
     serverIO.disconnect();
-    localStorage.setItem('currentUser', '');
-
     return sendSignOut().then(() => {
       dispatch({ type: 'SIGNOUT' });
-      dispatch(getLoginInfo());
```
axelboc, line 141: Do you need to dispatch the action I suggested, `resetLoginInfo`? Or is the `SIGNOUT` action sufficient?

ui/src/api/login.js
```diff
@@ -11,7 +11,12 @@ export function sendSignOut() {
 }
 
 export function fetchLoginInfo() {
-  return endpoint.get('/login_info').json();
+  return endpoint
+    .get('/login_info') // eslint-disable-next-line promise/prefer-await-to-callbacks
+    .unauthorized((err) => {
+      throw err;
```
axelboc: Not needed if all you do is re-throw, unless this is perhaps a placeholder to do other things later.

ui/src/components/App.js
```diff
@@ -21,65 +18,105 @@ import HelpContainer from '../containers/HelpContainer';
...
+    ],
+  },
+]);
+
 function App(props) {
-  const { loggedIn } = props;
+  const { applicationFetched } = props;
 
   useEffect(() => {
     requireAuth();
```
axelboc: Might be cleaner to inline the function in here and use `mapDispatchToProps` to bind the various action creators.

ui/src/components/App.js
```diff
@@ -21,65 +18,105 @@ import HelpContainer from '../containers/HelpContainer';
...
+export async function requireAuth() {
+  try {
+    store.dispatch(applicationFetched(false));
+    await store.dispatch(getLoginInfo());
+  } catch {
+    store.dispatch(applicationFetched(true));
+    return;
   }
+
+  await store.dispatch(getInitialState());
```
axelboc: Since the `getInitialState` async action is dispatched only once, when the `App` is rendered for the first time, it would make sense to fetch the login info inside that action as well, in parallel with all the other requests.

You can also move `dispatch(applicationFetched(true));` in there and remove `dispatch(applicationFetched(false));`, I think.

ui/src/components/App.js
```diff
@@ -21,65 +18,105 @@ import HelpContainer from '../containers/HelpContainer';
...
 
-  if (login.loggedIn && !serverIO.initialized) {
-    serverIO.listen(store);
+export async function requireAuth() {
+  try {
+    store.dispatch(applicationFetched(false));
+    await store.dispatch(getLoginInfo());
+  } catch {
+    store.dispatch(applicationFetched(true));
+    return;
```
axelboc: If `getLoginInfo` fails, then the app shouldn't remain in a loading state. I think this is how the app behaves currently but it's not great, obviously. Might be good to fix this behaviour as part of this PR.

ui/src/components/App.js
```diff
@@ -21,65 +18,105 @@ import HelpContainer from '../containers/HelpContainer';
...
   }
+
+  await store.dispatch(getInitialState());
+  serverIO.listen(store);
```
axelboc: Need to disconnect in the `useEffect`'s clean-up function to avoid memory leaks: https://react.dev/reference/react/useEffect#parameters

ui/src/actions/general.js
```diff
@@ -82,9 +82,12 @@ function notify(error) {
...
 }
 
-export function getInitialState(userInControl) {
-  return (dispatch) => {
+export function getInitialState() {
+  return (dispatch, getState) => {
     const state = {};
+    const { login } = getState();
+
+    dispatch(applicationFetched(false));
```
axelboc: I don't think this is needed, since this is the default value in the store on initialisation, and `getInitialState` is only dispatched once.

ui/src/reducers/index.js
```diff
@@ -40,10 +40,6 @@ const mxcubeReducer = combineReducers({
 });
 
 const rootReducer = (state, action) => {
```
axelboc, line 42: This wrapper function is no longer needed I think, you can inline the line below directly.

ui/src/components/App.js
```diff
@@ -21,65 +18,105 @@ import HelpContainer from '../containers/HelpContainer';
...
 import LoadingScreen from '../components/LoadingScreen/LoadingScreen';
 
-async function requireAuth() {
-  await store.dispatch(getLoginInfo());
-  const { login } = store.getState();
+import { store } from '../store';
+import { serverIO } from '../serverIO';
 
-  if (login.loggedIn) {
-    await store.dispatch(startSession(login.user.inControl));
```
axelboc, line 29: I believe you can now scrap the `startSession` action creator entirely.

mxcubeweb/routes/login.py
marcus-oscarsson: Yes,  the previous code was not working by the way, doing `resp.code=409` had no effect and the status code was always 200.

I was thinking of using 401, and I tried that but I then realized that we are so far (and maybe we should find a another way to do this), returning an error string to the client with a text message to be shown in the "error box" below the login fields (in the login form).

So, for the moment I kept the 200 and I made it explicit. I, as you, think that we should move away from this and use 401 or 403 instead. But that also means that we need to handle the error display in some other way.

We can perhaps handle that in the PR following this one, what do you think ?

ui/src/components/App.js
marcus-oscarsson: Oh yes, I wanted to do that, thanks for the reminder :)

ui/src/actions/login.js
marcus-oscarsson: yeah, you are right a `resetLoginInfo` is nicer

ui/src/actions/login.js
marcus-oscarsson: oh yes,

*EDIT:*, ah well ideally yes, but we are actually using the response even if the login was not successful. Ill pass by later and we can see exactly how to deal with this one :)

ui/src/actions/login.js
marcus-oscarsson: I think SIGNOUT is indeed sufficient but you are right that something like `resetLoginInfo` should probably called to clean up things.

mxcubeweb/routes/login.py
fabcor-maxiv: Bad credentials (can not authenticate) is 401. If user is authenticated (good credentials) but does not have the permission to access resource, then it is 403.

Looking at the context where the `deny_access` function is used, I would argue that `deny_access` is not an appropriate name for this function. It is not the access per se that is denied, it is that the credentials are incorrect. At this point it is not a matter of denying or authorizing access. Maybe `reject_credentials` is a better name.

fabcor-maxiv: Looking at the context where the `deny_access` function is used, I would argue that `deny_access` is not an appropriate name for this function. It is not the access per se that is denied, it is that the credentials are incorrect. At this point it is not a matter of denying or allowing access. Maybe `reject_credentials` is a better name.

fabcor-maxiv: Then regarding the status code...

One can argue that `200` is the correct status code to emit even if the credentials are incorrect. 200 means the http request was handled correctly. Sending incorrect credentials is not an error per se, not an issue with the http request and its handling at least. So sending 200 http code with the body that explicitly tells if the credentials are correct or not `{"msg": "Could not authenticate"}` is fine by me.

axelboc: That's what the HTTP error codes are for, though. 😉 It saves you from sending any message at all, since 401 already implies that the credentials are wrong. It also allows for simpler error handling in the front-end, since the fetching library (wretch in our case, but also axios and others) considers 4xx and 5xx as errors. Not to mention that you get a clear XHR error in the browser dev tools, which helps a lot when debugging.

axelboc: But no worries with using 200 for now and improving error handling later of course! 👍

ui/src/actions/login.js
axelboc: Yep, makes sense now as per your other comment 👍

ui/src/components/App.js
marcus-oscarsson: I think its solved with the ` store.dispatch(applicationFetched(true));`  application is considered fetched if `getLoginInfo` fails in the code above or when `getInitialState()` is done.

ui/src/components/App.js
marcus-oscarsson: Hm, that would perhaps be possible but I don't want to fetch the entire app if `fetchLoginInfo` returns 401, or is that just silly of me ?

ui/src/components/App.js
axelboc: Oh right, I must have misread sorry. The error will need to be handled a bit better in a future PR, then. :+1:

ui/src/components/App.js
axelboc: True, true, it's not a bad idea to want to keep them separate, but it does complexify things. If `fetchLoginInfo` fails, then do you dispatch `getInitialState` again when the user logs in? If you don't, I think it's a bug; if you do, then `getInitialState` is perhaps not the right wording.

It's a choice between slowing down initial load, or slowing down logging in. The ideal UX solution is in between: fast initial load, combined with pre-loading while the user enters their crendentials.

In terms of code, fetching everything on initial load is easier since you don't have to worry about dispatching `getInitialState` again. Note that there are probably opportunities for optimising this, for instance, by combining some endpoints to reduce the number of concurrent requests.

mxcubeweb/routes/login.py
```diff
@@ -45,15 +32,17 @@ def login():
...
                 str(ex),
             )
             logging.getLogger("MX3.HWR").exception("")
             logging.getLogger("MX3.HWR").info(msg)
-            res = deny_access("Could not authenticate")
+            res = make_response(jsonify({"msg": "Could not authenticate"}), 200)
+        else:
+            res = make_response(jsonify({"msg": ""}), 200)
 
         session.permanent = True
```
marcus-oscarsson: I managed to get the tests working by adding back this line which I initially removed, as I thought that setting `SESSION_PERMANENT=True` would be enough and certainly as we are also using `SESSION_PERMANENT_LIFETIME` (and the later is taken into account). 

However it seems that the configuration value is not taken into account and we need to set it programmatically, have a look at our mockup server [config](https://github.com/mxcube/mxcubeweb/blob/b38bb1446733ced2ff3fc982aa6f386ee412b47d/test/HardwareObjectsMockup.xml/mxcube-web/server.yaml#L6)

Documentation even states that its true by default: https://flask-session.readthedocs.io/en/latest/config.html#configuration

Does someone have an idea of what that could be ?

fabcor-maxiv: @marcus-oscarsson

I recall it not being straightforward to understand. It took me a while to wrap my head around sessions and their expiration. I had written [some doc](https://mxcubeweb.readthedocs.io/en/latest/dev/login.html), but obviously this doc is not good enough since reading it back now does not clarify much...

Flask and Flask-session have different behavior, different defaults. Flask does not seem to know about the `SESSION_PERMANENT` config setting at all.

I made a quick search and I am starting to wonder if flask-session is used at all. I can not seem to find where it is used in the code. Am I missing something?

marcus-oscarsson: Indeed that explains it, so looking at the [flask documentation](https://flask.palletsprojects.com/en/2.0.x/config/) `SESSION_PARMANENT` is not an option at all, however `SESSION_PERMANENT_LIFETIME` is and its used when one does `session.permanent=True`.  So I was simply confused by the Flask-session documentation, we were using it in the past but now we are using `Flask-Login`

fabcor-maxiv: We should probably remove line 29 from `mxcubeweb/server.py`, line 6 from `test/HardwareObjectsMockup.xml/mxcube-web/server.yaml`, and finally line 20 from `mxcubeweb/core/models/configmodels.py`.

marcus-oscarsson: :), Actually I thought if flask does not deal with the `SESSION_PERMANENT` setting we can :)

marcus-oscarsson: On a second thought, thats just unnecessary ... Ill remove the config option as we are actually relying on this feature, it does not make any sense to set it to false in our case.

mxcubeweb/routes/login.py
axelboc: Thought about this a bit more: while responding with 401 for the `/login` endpoint is the way to go (because the purpose of the request is to log in, so if it doesn't work, the server must inform the client with the appropriate error code), it doesn't necessarily make sense to do so for the `/login_info` endpoint.

The `/login_info` endpoint is not an authenticated endpoint; calling it when the user is logged out is valid and expected, so it shouldn't actually respond with 401 but just 200 and a body that conveys that no user is currently logged in (maybe an empty object).

marcus-oscarsson:
Just a few last improvements done after some discussions we had between us here, after this I think we are more or less happy with this one. There will as I mentioned before be one or two more PR's to follow up on this.

 - Status codes of all unauthorized resources are now 401
 - `login_info` always have status code 200 but returns `loggedIn=False|True`
 - Moved `getLoginInfo` call to `getInitialState` and removed unnecessary `navigate`
 - Moved `getInitialState` to `actions/login` to remove a circular dependency cycle (And it it also makes sense to have getIntialState in `reducers/login`
 - Moved the use of `serverIO.disconnect` from `action/login` so that it only contains redux related code and as a consequence avoids a circular dependency

ui/src/actions/login.js
```diff
@@ -124,25 +121,188 @@ export function logIn(proposal, password, navigate) {
...
-
+      dispatch(resetLoginInfo());
       if (navigate) {
-        navigate('/login');
+        navigate('/');
```
axelboc: Would be good to navigate outside of Redux here as well if possible (perhaps for another PR).

ui/src/actions/login.js
```diff
@@ -59,58 +83,31 @@ export function sendSelectProposal(number, navigate) {
...
-        dispatch(showErrorPanel(true));
-        dispatch(setLoading(false));
-      },
-    );
+  return async (dispatch) => {
+    try {
+      const loginInfo = await fetchLoginInfo();
+      if (!loginInfo.loggedIn) {
+        throw new Error('Not authenticated');
+      }
```
axelboc: As discussed, I think we can remove this `throw` and the `catch` below, since it cannot happen if the app is coded properly. If we get to this and the user is still unauthenticated, then we will hit a 401 when fetching other things later as expected (e.g. in `getInitialState`).

marcus-oscarsson: :+1:

axelboc:
I've refactored things a bit to make it clearer when/where `getLoginInfo` and `getInitialState` get called. Here's the gist of it:

- `getLoginInfo` is called from `useEffect` in `App` and from the `logIn` action, which means it gets called:
  1. once on initial load, to know if the user is logged in or not;
  2. right after a successful login (until such times as the `login` endpoint returns the login info right away).
- `getInitialState` is called only from `componentDidMount` in `Main`, which means that it gets called:
  1. on initial load when the user is already logged in;
  2. after a successful login.

In other words, `getInitialState` no longer needs to call `getLoginInfo` directly, since by the time `Main` renders, the user's login info have already been fetched (either from `useEffect` or from `login`). This means we can move `getInitialState` out of the `login.js` actions file without reintroducing a circular import. I'll do it in a separate commit later this week.

fabcor-maxiv:
Would that make sense to add some of these technical details at `docs/source/dev/login.rst`?

fabcor-maxiv:
For info, I have not been able to work on this these past weeks, but I am planning to test these changes to authentication this week (possibly tomorrow already).

marcus-oscarsson:
@fabcor-maxiv what I suggest is that we merge this as it is now and we fix anything we/you find while testing. Unless you see a fundamental problem with this PR

fabcor-maxiv:
~I am investigating a case where I get an error `TypeError: t.loginInfo.user is undefined`, seemingly in the `SET_LOGIN_INFO` action. It appears right when loading MXCuBE the first time, before any authentication. It stays stuck on the "Loading, please wait" page. I will provide more details, once I manage to narrow this down further.~

False alarm.

fabcor-maxiv:
I figured out the cause of the behavior I initially reported in the previous message, and it was a "user issue" :D

Now it seems to be working fine, comparable to what is demonstrated by the animation in the first post. Nice! :)

mxcubeweb/core/components/user/usermanager.py
```diff
@@ -275,8 +258,10 @@ def login_info(self):
             )
 
             res["selectedProposalID"] = HWR.beamline.session.proposal_id
+        else:
+            raise Exception("Not logged in")
```
fabcor-maxiv, line 262: Maybe raise a more specific exception type? See `mxcubeweb/routes/login.py` line 85 where is is caught.

```suggestion
            raise NotAuthenticated("Not logged in")
```

mxcubeweb/routes/login.py
```diff
@@ -80,24 +68,22 @@ def login_info():
...
 
-        # Redirect the user to login page if for some reason logged out
-        # i.e. server restart
-        if not user:
-            response = redirect("/login", code=302)
-        else:
+        try:
+            res = app.usermanager.login_info()
             response = jsonify(res)
+        except Exception:
```
fabcor-maxiv, line 84: What about having a more specific Exception type? See `mxcubeweb/core/components/user/usermanager.py` line 262 where the exception is raised.

```suggestion
        except NotAuthenticated:
```

mxcubeweb/routes/login.py
fabcor-maxiv: I wonder now if it makes sense (or is needed) to declare the session permanent when authentication failed.

It seems to work as it is so maybe it is fine to leave it like this, but curiosity...

test/test_authn.py
```diff
@@ -77,8 +77,6 @@ def client(make_client):
 def test_authn_signin_good_credentials(client):
     resp = client.post(URL_SIGNIN, json=CREDENTIALS_0)
     assert resp.status_code == 200
-    assert resp.json["code"] == "ok"
-    assert resp.json["msg"] == "Successful login"
```
fabcor-maxiv, line 81: We could put the test of the message back in:

```
    assert resp.json["msg"] == "Successful login"
```

If we do not want the old message then at least we should test that the message is the empty string (as it is now in this pull request):

```
assert resp.json["msg"] == ""
```

mxcubeweb/routes/login.py
```diff
@@ -66,8 +55,7 @@ def signout():
         Signout from MXCuBE Web and reset the session
         """
         app.usermanager.signout()
-
-        return redirect("/login", code=302)
+        return make_response(jsonify(""), 200)
```
fabcor-maxiv, line 57: Maybe add a meaningful message instead of empty string:

```suggestion
        return make_response(jsonify("Successful sign out"), 200)
```

mxcubeweb/routes/login.py
```diff
@@ -45,15 +32,17 @@ def login():
...
             msg = "[LOGIN] User %s could not login (%s)" % (
                 login_id,
                 str(ex),
             )
             logging.getLogger("MX3.HWR").exception("")
             logging.getLogger("MX3.HWR").info(msg)
-            res = deny_access("Could not authenticate")
+            res = make_response(jsonify({"msg": "Could not authenticate"}), 200)
+        else:
+            res = make_response(jsonify({"msg": ""}), 200)
```
fabcor-maxiv, line 45: Why no message?

We could put the old message back in:

```suggestion
            res = make_response(jsonify({"msg": "Successful login"}), 200)
```

test/test_authn.py
```diff
@@ -90,11 +88,9 @@ def test_authn_signin_wrong_credentials(client):
 
 def test_authn_signout(client):
     resp = client.post(URL_SIGNIN, json=CREDENTIALS_0)
-    assert resp.json["code"] == "ok"
 
     resp = client.get(URL_SIGNOUT)
-    assert resp.status_code == 302
-    assert resp.headers["Location"] == "/login"
+    assert resp.status_code == 200
```
fabcor-maxiv, line 94: Let's check the message, no?

```suggestion
    assert resp.status_code == 200
    assert resp.json["msg"] == ""
```

~Even better if we had a meaningful message instead of empty string~:

```suggestion
    assert resp.status_code == 200
    assert resp.json["msg"] == "Successful sign out"
```

fabcor-maxiv (commented):
Some comments for the Python part

ui/src/actions/login.js
```diff
@@ -59,90 +83,205 @@ export function sendSelectProposal(number, navigate) {
...
 
-export function logIn(proposal, password, navigate) {
-  // eslint-disable-next-line sonarjs/cognitive-complexity
-  return (dispatch) => {
-    const previousUser = localStorage.getItem('currentUser');
-    if (serverIO.hwrSocket !== null && serverIO.hwrSocket.connected) {
-      console.log(serverIO.hwrSocket.connected); // eslint-disable-line no-console
-    } else {
-      serverIO.connect();
+    if (res.msg !== '') {
```
fabcor-maxiv, line 97: Ah, maybe that is the reason why all the success messages were removed on the back end side.

I am not convinced it is a great choice, but I can understand...

Empty message or not, the back end tests should test the message, because as we see here it is obviously part of the API.

fabcor-maxiv (commented):
Comment regarding the empty message seen from the front end side.

ui/src/components/App.js
fabcor-maxiv: Seems like it is done.

mxcubeweb/routes/login.py
axelboc: FYI, if I'm not mistaken, the front-end has an equality check for the empty string, so it will need to be updated as well. This is temporary, though. The idea is to return the login info object directly eventually.

ui/src/actions/login.js
axelboc: Ha, should have scrolled down further... Yeah, so the idea will be to return the login info object directly in a future PR.

fabcor-maxiv: Ah, I see. Then @marcus-oscarsson I recommend to adjust the Python tests to assert that we do have an empty string where we expect one (as mentioned in earlier comments).

mxcubeweb/routes/login.py
fabcor-maxiv: OK, after response from @axelboc I understand we could leave it as empty for now, until we get to the point of being able to return the actual login info instead.

mxcubeweb/routes/login.py
fabcor-maxiv: I see, thanks! Then let's leave it like this for now, we only need to adjust the test.

test/test_authn.py
```diff
@@ -77,8 +77,6 @@ def client(make_client):
 def test_authn_signin_good_credentials(client):
     resp = client.post(URL_SIGNIN, json=CREDENTIALS_0)
     assert resp.status_code == 200
```
fabcor-maxiv, line 79:
```suggestion
    assert resp.status_code == 200
    assert resp.json["msg"] == ""
```

marcus-oscarsson:
Ok, thanks guys (@axelboc and @fabcor-maxiv). The work on some of the details will continue in followup-PRs, and fairly soon one that also introduces OpenID Connect  authentication using KeyCloak.

fabcor-maxiv:
@axelboc @marcus-oscarsson Seems like we lost the loading animation on the "Sign in" button in this change. I can open a separate ticket if you prefer.

P.S.: ticket opened: https://github.com/mxcube/mxcubeweb/issues/1148

PR #1143, opened 2023-11-30
use correct motor for horizontal sample movements

elmjag:
For the sample position UI controls, i.e. the arrows in the left column, pick correct horizontal motor.

Use component with 'sample_horizontal' role for horizontal sample moves.

image


marcus-oscarsson:
Thanks !

PR #1145, opened 2023-12-06
MASSIF-1 Crystal Direct Harvester implementation 

jbflo:
- This is the code related to the implementation of the Crystal Direct Harvester in Massif-1 ESRF

- While transforming the components related to the Harvester  to functional component , i also changed most of the Equipment Class to function.

- Apologize, I did not mean to make this PR that Big

Beamline that does not have an Harvester should not have any new things/modification to do
A PR in Mxcubecore related to this one need to be reviewed (first) as well .

image


mxcubeweb/core/components/samplechanger.py
```diff
@@ -467,66 +509,85 @@ def queue_mount_sample(view, data_model, centring_done_cb, async_result):  # noq
...
-                    else:
-                        if (
-                            centring_method
-                            == queue_entry.CENTRING_METHOD.FULLY_AUTOMATIC
-                        ):
-                            raise queue_entry.QueueSkippEntryException(
-                                "Could not center sample, skipping",
-                                "",
+            except Exception as ex:
+                print(str(ex))
```
fabcor-maxiv: I recommend using `some_logger.exception("Some message")`, it will automatically add the stack trace.

mxcubeweb/core/components/samplechanger.py
```diff
@@ -467,66 +509,85 @@ def queue_mount_sample(view, data_model, centring_done_cb, async_result):  # noq
...
+                                raise queue_entry.QueueSkippEntryException(
+                                    "Could not center sample, skipping",
+                                    "",
+                                )
+                            else:
+                                raise RuntimeError("Could not center sample")
+                    except Exception:
+                        import traceback
+
+                        log.info("centring did not pass %s" % traceback.format_exc())
```
fabcor-maxiv: This is probably better:

```suggestion
                        log.exception("centring did not pass")
```

The exception including traceback should be automatically added: https://docs.python.org/3/library/logging.html#logging.exception

mxcubeweb/routes/harvester.py
```diff
@@ -0,0 +1,111 @@
...
+            resp = jsonify(app.harvester.harvest_crystal(request.get_json()))
+        except Exception as ex:
+            resp = (
+                "Cannot Harvest Crystal",
+                409,
+                {"Content-Type": "application/json", "message": str(ex)},
+            )
+            return resp
+
+        return jsonify(app.harvester.get_harvester_contents())
```
fabcor-maxiv: What is going on with `resp`? If we exclude the `except` block, `resp` is assigned value, twice, but never used. Did any linter catch that?

fabcor-maxiv (commented):
Wow... I had pending review comments I had no idea about. Must have been from quite a while ago, so I have no idea if they are still meaningful.

ui/src/components/Equipment/SampleChanger.js
```diff
@@ -1,313 +1,270 @@
...
+//     }
+//   }
+
+//   return (
+//     <Card className="mb-3" style={{ marginTop: '0.5em' }} bg={titleBackground}>
+//       <Card.Header>{props.title}</Card.Header>
+//       {props.children}
+//     </Card>
+//   );
+// }
```
fabcor-maxiv: Why?

ui/src/components/Equipment/SampleChanger.js
```diff
@@ -1,313 +1,270 @@
...
+function SampleChangerTreeItem(props) {
+  const [dropdownIsOpen, setDropdownIsOpen] = useState(false);
 
-  itemClicked() {
-    this.setState({ allow_control: !this.state.allow_control });
-    /* eslint-enable react/no-set-state */
-  }
+  // function itemClicked() {
+  //   setState({ allow_control: !state.allow_control });
+  // }
```
fabcor-maxiv: Why?

mxcubeweb/core/components/harvester.py
```diff
@@ -0,0 +1,505 @@
...
+                    "name": crystal_names[index],
+                    "state": sample_states[index],
+                    "acronym": proteinAcronym,
+                    "img_url": img_url,
+                    "img_target_x": img_target_x,
+                    "img_target_y": img_target_y,
+                }
+                crystal_list.append(lst)
+        except Exception as ex:
+            print("Could not get Crystal List : %s" % str(ex))
```
fabcor-maxiv: Why no proper logging?

```suggestion
            some_logger.exception("Could not get Crystal List")
```

mxcubeweb/core/components/harvester.py
```diff
@@ -0,0 +1,505 @@
...
+            # TEMP return a fake list for test
```
fabcor-maxiv: Is it on purpose that we have test data here?

mxcubeweb/core/components/harvester.py
```diff
@@ -0,0 +1,505 @@
...
+                },
+            ]
+
+        return crystal_list
+
+    def get_global_state(self):
+        try:
+            return HWR.beamline.harvester_maintenance.get_global_state()
+        except Exception:
+            return "OFFLINE", "OFFLINE", "OFFLINE"
```
fabcor-maxiv: I really do not like those catch-all exceptions. There might be some actual issues somewhere in this block of code and we will never know about it, because it is swallowed and there is no logging! Is there really no cleaner way to handle this?

mxcubeweb/core/components/harvester.py
```diff
@@ -0,0 +1,505 @@
...
+                    dataCollectionGroupId,
+                    dc_id,
+                    proposal,
+                    rest_token,
+                )
+            return True
+        except Exception as ex:
+            msg = "get all queue entries failed, reason:  %s" % str(ex)
+            logging.getLogger("user_level_log").exception(msg)
+            return False
```
fabcor-maxiv: There should be no need to have the exception in the message if it is the `exception` method of the logger that is used. This should be just fine:

```suggestion
        except Exception as ex:
            msg = "get all queue entries failed, reason"
            logging.getLogger("user_level_log").exception(msg)
            return False
```

mxcubeweb/core/components/harvester.py
```diff
@@ -0,0 +1,505 @@
...
+        or had and exception (False)
+        """
+        harvester_device = HWR.beamline.harvester
+
+        harvester_device.load_calibrated_pin()
+        harvester_device._wait_sample_transfer_ready(None)
+        print("waiting 40 seconds before mount")
+        # For some reason the Harvester return READY too soon
+        # approximately 40 Second sooner
+        gevent.sleep(40)
```
fabcor-maxiv: Why no proper logging?

Isn't there any condition that we can check to make sure we have waited long enough?

Why 40 seconds specifically? Is it empirical?

Should the 40 seconds value be made configurable? Or at the very least extracted into a named constant?

mxcubeweb/core/components/harvester.py
```diff
@@ -0,0 +1,505 @@
...
+
+            calibrated_motor_offset = {
+                "focus": new_motor_offset["focus"] + new_motor_offset["centring_focus"],
+                "phiy": new_motor_offset["phiy"],
+                "phiz": (
+                    new_motor_offset["phiz"] + new_motor_offset["centring_vertical"]
+                ),
+            }
+
+            print(calibrated_motor_offset)
```
fabcor-maxiv: Is this here on purpose or should it have been deleted?

mxcubeweb/core/components/harvester.py
```diff
@@ -0,0 +1,505 @@
...
+                calibrated_motor_offset["phiy"],
+                calibrated_motor_offset["phiz"],
+            )
+
+            harvester_device.set_calibrate_state(False)
+        except Exception as ex:
+            logging.getLogger("user_level_log").exception(
+                f"Pin Calibration / validation Failed {str(ex)}"
+            )
+            return False
```
fabcor-maxiv: Another catch-all-exceptions? Why? At least there is logging, though...


```suggestion
        except Exception:
            logging.getLogger("user_level_log").exception(
                "Pin Calibration / validation Failed"
            )
            return False
```

No, but seriously why not list the actual exceptions we want to catch here?

```
except (ExceptionFoo, ExceptionBar):
```

and of course, there should be much less coed in the `try` block. 

I know those things are not easy, but seeing code like this scares me. There is a very high risk of us staying naively unaware of a whole bunch of issues that are actually very easy to fix if we keep having this poor handling of exceptions.

mxcubeweb/core/components/harvester.py
```diff
@@ -0,0 +1,505 @@
...
+    def harvest_and_mount_sample(self, xtal_uuid: str) -> dict[str, bool, int]:
+        try:
+            harvester_device = HWR.beamline.harvester
+            self.harvest_crystal(xtal_uuid)
+            harvester_device._wait_sample_transfer_ready(None)
+
+            sample = self.get_sample_info(xtal_uuid)
+            self.app.sample_changer.mount_sample(sample)
+        except Exception:
+            return "Could not Harvest Crystal"
```
fabcor-maxiv: Maybe log the exception at least, and then maybe someone will be able to fix the issue if any shows up.

mxcubeweb/core/components/harvester.py
```diff
@@ -0,0 +1,505 @@
...
+        if harvester_device:
+            return harvester_device.harvest_sample_before_mount(
+                sample_uuid, wait_before_load
+            )
+        else:
+            msg = harvester_device.get_status()
+            logging.getLogger("user_level_log").exception(
+                "ERROR: No Harvester Device Found"
+            )
+            logging.getLogger("user_level_log").exception(msg)
```
fabcor-maxiv: There is no exception here, we are not in an `except` block. So we should use the `error` method.

> `exception(msg, *args, **kwargs)`
>    Logs a message with level [ERROR](https://docs.python.org/3/library/logging.html#logging.ERROR) on this logger. The arguments are interpreted as for [debug()](https://docs.python.org/3/library/logging.html#logging.debug). Exception info is added to the logging message. This method should only be called from an exception handler.

-- https://docs.python.org/3/library/logging.html#logging.Logger.exception

mxcubeweb/core/components/harvester.py
```diff
@@ -0,0 +1,505 @@
...
+            return False
+
+    def current_queue_index(self, current_sample: str) -> Optional[int]:
+        current_queue_dict = self.app.queue.queue_to_dict()
+        current_queue_list = list(current_queue_dict)
+        res = None
+        try:
+            res = current_queue_list.index(current_sample)
+        except (ValueError, IndexError):
+            res = None
```
fabcor-maxiv: That is better! : )

mxcubeweb/core/components/harvester.py
```diff
@@ -0,0 +1,505 @@
...
+
+    def queue_harvest_sample(self, data_model, sample: str) -> None:
+        """
+        While queue execution send harvest request
+        """
+        current_queue = self.app.queue.queue_to_dict()
+        harvester_device = HWR.beamline.harvester
+        wait_before_load = (
+            False if harvester_device.get_room_temperature_mode() else True
+        )
```
fabcor-maxiv: Maybe can be simplified, no? I do not know for sure.

```suggestion
        wait_before_load = not harvester_device.get_room_temperature_mode()
```

mxcubeweb/core/components/harvester.py
```diff
@@ -0,0 +1,505 @@
...
+        if harvester_device.get_number_of_available_pin() > 0:
+            gevent.sleep(2)
+            sample_UUID = current_queue[sample["sampleID"]]["code"]
```
fabcor-maxiv:
```suggestion
            sample_uuid = current_queue[sample["sampleID"]]["code"]
```

No need to change casing here, is there?

mxcubeweb/core/components/samplechanger.py
```diff
@@ -467,66 +509,85 @@ def queue_mount_sample(view, data_model, centring_done_cb, async_result):  # noq
...
+                            log.warning(msg)
+                            dm.start_centring_method(dm.MANUAL3CLICK_MODE)
+                        elif centring_method in [
+                            queue_entry.CENTRING_METHOD.LOOP,
+                            queue_entry.CENTRING_METHOD.FULLY_AUTOMATIC,
+                        ]:
+                            if not dm.current_centring_procedure:
+                                dm.start_centring_method(dm.C3D_MODE)
+
+                            # NBNB  BUG . self and app are not avialble here
```
fabcor-maxiv: ???

ui/src/components/Equipment/Harvester.jsx
```diff
@@ -0,0 +1,168 @@
...
+      <Separator />
+      <Item onClick={() => harvestAndLoadCrystal(key)}>
+        <span>
+          <FcCollect /> harvest & Load Sample <FcUpload />
+        </span>
+      </Item>
+    </Menu>
+  );
+
+  const crystalUUID = props.contents.harvester_crystal_list;
```
fabcor-maxiv, line 84:
```suggestion
  const crystalUuid = props.contents.harvester_crystal_list;
```

`crystalUuid` is fine, no need for extra upper case letters (at least in my opinion)

ui/src/components/Equipment/HarvesterMaintenance.jsx
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  if (message !== '') {
+    msg = message;
+  }
+
+  let plateBarcode = '';
+  if (global_state.plate_barecode) {
+    plateBarcode = global_state.plate_barecode.toString();
+  }
```
fabcor-maxiv: I think these things can be simplified to something like the following (and then we have `const`, which I think is preferrable):

```suggestion
  const plateBarcode = global_state.plate_barecode ? global_state.plate_barecode.toString() : "";
```

Also isn't there a formatting convention to always use either single quotes`'` or double quotes `"`?

fabcor-maxiv (commented):
That is a lot of code!

I looked at some random spots (mostly the Python part).

I do not know enough about the inner workings of mxcube and/or harvester so my comments are mostly superficial. Mostly it is things that I believe a linter should catch, but I guess we must have very lenient linting rules (which is not so great for code quality in the long term). I hope one day I have time to look at those linting rules and maybe try to tighten things up in terms of code quality.

fabcor-maxiv:
I converted as draft. Maybe pull requests should stay as draft until they are done (or very close to be done) including the code passes all the Actions workflows (linting, formatting, tests cheking, and so on).

elmjag:
We need to make sure that all CI tests are passing before merging.

The CI tests are failing with this error:

    ModuleNotFoundError: No module named 'mxcubecore.HardwareObjects.Harvester'

I guess this PR requires https://github.com/mxcube/mxcubecore/pull/826 to be merged first. So, we need to sort out [826](https://github.com/mxcube/mxcubecore/pull/826) before merging this.

mxcubeweb/routes/harvester.py
jbflo: > What is going on with `resp`? If we exclude the `except` block, `resp` is assigned value, twice, but never used. Did any linter catch that?

This is Outdated .  was there when PR was WIP

ui/src/components/Equipment/SampleChanger.js
jbflo: This is unused and i was not sure why.
was not my code so Did not want to take the responsible to remove at first place .
I wanted to asked first what was the intention of it usage.  then i forgot

mxcubeweb/core/components/harvester.py
jbflo: Yes, That's help test the GUI > TO be remove after

mxcubeweb/core/components/harvester.py
jbflo: Yes empirical, this is something to deal with the device itself .
Until then , we shall stick with this. we 'll remove this line once they fix the issue on the device.

jbflo:
> We need to make sure that all CI tests are passing before merging.
> 
> The CI tests are failing with this error:
> 
> ```
> ModuleNotFoundError: No module named 'mxcubecore.HardwareObjects.Harvester'
> ```
> 
> I guess this PR requires [mxcube/mxcubecore#826](https://github.com/mxcube/mxcubecore/pull/826) to be merged first. So, we need to sort out [826](https://github.com/mxcube/mxcubecore/pull/826) before merging this.

Sure, As mentioned in the Description !

ui/src/components/Equipment/SampleChanger.js
jbflo: https://github.com/mxcube/mxcubeweb/pull/1145#discussion_r1551384820

ui/src/components/Equipment/HarvesterMaintenance.jsx
jbflo: Better like this: thanks
`const plateBarcode = global_state.plate_barecode?.toString();`

jbflo:
> I converted as draft. Maybe pull requests should stay as draft until they are done (or very close to be done) including the code passes all the Actions workflows (linting, formatting, tests cheking, and so on).

@fabcor-maxiv This was in WIP  for quite some time waiting for reviews. 
But had to putted this has ready to catch peoples attention cause it's not so easy to review  your own PR when you know it's work. 

But Thanks a lot, don't hesitate to add your comment i'll address them as far as i can. 

This shall pass the  Checks, but we should get agree with @beteva First on [826](https://github.com/mxcube/mxcubecore/pull/826)  as @elmjag  Noticed

fabcor-maxiv:
> > I converted as draft. Maybe pull requests should stay as draft until they are done (or very close to be done) including the code passes all the Actions workflows (linting, formatting, tests cheking, and so on).
> 
> @fabcor-maxiv This was in WIP for quite some time waiting for reviews. But had to putted this has ready to catch peoples attention cause it's not so easy to review your own PR when you know it's work.

I see. I think you should be able to ask for code reviews from anyone in the top right corner of the page: 
image

mxcubeweb/core/components/harvester.py
```diff
@@ -0,0 +1,481 @@
...
+
+            motor_pos_dict = {
+                "focus": md.focusMotor.get_value(),
+                "phiy": md.phiyMotor.get_value(),
+                "phiz": md.phizMotor.get_value(),
+                "centring_focus": md.centringFocus.get_value(),
+                "centring_vertical": md.centringVertical.get_value(),
+            }
+
+            saved_position = md.saved_motor_position
```
jbflo: @marcus-oscarsson Here for saved_motor_position

mxcubeweb/core/components/samplechanger.py
```diff
@@ -183,6 +192,11 @@ def mount_sample_clean_up(self, sample):
 
         sc = HWR.beamline.sample_changer
 
+        try:
+            mount_from_harvester = sc.mount_from_harvester()
```
jbflo: @marcus-oscarsson a review on this file will make the use of mount_from_harvester more clear

mxcubeweb/core/components/harvester.py
```diff
@@ -0,0 +1,494 @@
...
+                "Sample Changer could not mount Pin"
+            )
+            return False
+
+    def cancel_calibration(self) -> None:
+        harvester_device = HWR.beamline.harvester
+        harvester_device.set_calibrate_state(False)
+        logging.getLogger("user_level_log").warning("Pin Calibration Canceled")
+
+    def validate_calibration(self) -> bool:
```
marcus-oscarsson: What do you think about moving this logic to the actual harvester hardware object, it seems mostly to interact with the diffractometer and the harvester and not so much with the UI

mxcubeweb/core/components/harvester.py
```diff
@@ -0,0 +1,494 @@
...
+            sample = self.get_sample_info(xtal_uuid)
+            self.app.sample_changer.mount_sample(sample)
+        except Exception:
+            logging.getLogger("user_level_log").exception("Could not Harvest Crystal")
+            return "Could not Harvest Crystal"
+
+        self.init_signals()
+        return self.get_harvester_contents()
+
+    def harvest_sample_before_mount(
```
marcus-oscarsson: This one could probably also with advantage be moved to the harvester object ?

mxcubeweb/core/components/samplechanger.py
marcus-oscarsson: Ok, thanks I think that some logic in `components/harvester.py` can be moved down to the harvester hardware object

mxcubeweb/core/components/harvester.py
jbflo: Maybe I should think of something else,  the harvester hardware object communicate with and only the Harvester . i would prefer not adding MD or other device interaction  to it.

but yes I/we should maybe not put too much on the mxcubeweb core components, and move things down to an Hardware object level.

marcus-oscarsson: I think we will get alot of clarity from doing that.

mxcubeweb/core/components/harvester.py
jbflo: This is just an interface , the method is implemented in  harvester object already .

mxcubeweb/core/components/harvester.py
jbflo: Yeah Done.  Thanks !
And maybe that will help the guys from hamburg one day, if they ever want to have their Harvester in the Beamline :grin: . 

Have another look please, I hope this become more clear  :satisfied:

As we said maybe the mxcubeweb have old things to be move to mxcubecore level as well ? things like
-  specific queue logic , centering, mounting  etc..
 but for another story

mxcubeweb/core/components/harvester.py
```diff
@@ -0,0 +1,494 @@
...
+            "global_state": {"global_state": global_state, "commands_state": cmdstate},
+            "cmds": {"cmds": cmds},
+            "msg": msg,
+            "plate_mode": HWR.beamline.diffractometer.in_plate_mode(),
+        }
+
+        return initial_state
+
+    def mount_from_harvester(self):
+        sc = HWR.beamline.sample_changer
```
marcus-oscarsson, line 66: You are not really gaining anything from this assignment other than more things on the stack :)

mxcubeweb/core/components/harvester.py
```diff
@@ -0,0 +1,494 @@
...
+
+        Return (Dict): Dict content, object name etc..
+        crystal_list and number of available pins"
+        """
+        if HWR.beamline.harvester:
+            root_name = HWR.beamline.harvester.__TYPE__
+            crystal_list = self.get_crystal_list()
+            room_temperature_mode = HWR.beamline.harvester.get_room_temperature_mode()
+            number_of_pins = HWR.beamline.harvester.get_number_of_available_pin()
+            contents = {
```
marcus-oscarsson, line 85: It would be nice to model contents, so that we know exactly what it returns. Like we have done in a few places, for instance here, https://github.com/mxcube/mxcubeweb/blob/develop/mxcubeweb/core/models/adaptermodels.py

mxcubeweb/core/components/harvester.py
marcus-oscarsson: Yes that would be good, int that case we could make this an Adapter instead of a  Component, that would be much nicer.

mxcubeweb/core/components/harvester.py
```diff
@@ -0,0 +1,494 @@
...
+            return True
+        except Exception:
+            msg = "Could not send data collection to crims"
+            logging.getLogger("user_level_log").exception(msg)
+            return False
+
+    def get_calibration_state(self) -> bool:
+        return HWR.beamline.harvester.calibrate_state
+
+    def calibrate_pin(self) -> bool:
```
marcus-oscarsson: This can probably also be moved right ?

mxcubeweb/core/components/harvester.py
jbflo: not really :sweat_smile: ,  I gained some line of code

That  help on having less try  except block 
when calling from other components. 
it only called from the sampleChanger component but twice , worth it.

mxcubeweb/core/components/harvester.py
jbflo: I had a look, would be best for me making this into a new PR , if that's ok ?

mxcubeweb/core/components/samplechanger.py
```diff
@@ -135,6 +139,11 @@ def _addElement(parent, element):
 
             contents = {"name": root_name}
 
+            if hasattr(HWR.beamline.sample_changer, "get_room_temperature_mode"):
```
marcus-oscarsson, line 142: It would probably be better to add the `get_room_tempereature_mode` on the `AbstractSampleChanger`  that by default returns `False`

mxcubeweb/core/components/samplechanger.py
```diff
@@ -183,6 +192,8 @@ def mount_sample_clean_up(self, sample):
 
         sc = HWR.beamline.sample_changer
 
+        mount_from_harvester = self.app.harvester.mount_from_harvester()
```
marcus-oscarsson, line 195: I guess we know when we are gong to mount from the harvester and we can then set `CENTERIN_ METHOD` to `manual` there ?

mxcubeweb/core/components/samplechanger.py
```diff
@@ -211,12 +222,17 @@ def mount_sample_clean_up(self, sample):
...
                     and not HWR.beamline.diffractometer.in_plate_mode()
+                    and not mount_from_harvester
                 ):
                     HWR.beamline.diffractometer.reject_centring()
                     msg = "Starting autoloop centring ..."
                     logging.getLogger("MX3.HWR").info(msg)
                     C3D_MODE = HWR.beamline.diffractometer.C3D_MODE
                     HWR.beamline.diffractometer.start_centring_method(C3D_MODE)
+                elif HWR.beamline.diffractometer.in_plate_mode():
+                    msg = "Starting autoloop Focusin ..."
```
marcus-oscarsson: typo -> "Starting autoloop Focusing ..."

mxcubeweb/core/components/samplechanger.py
```diff
@@ -440,6 +457,11 @@ def queue_mount_sample(view, data_model, centring_done_cb, async_result):  # noq
                 "sampleID": data_model.loc_str,
             }
 
+            # in this case the sample changer takes the sample from an Harvester
+            # We Harvest the sample and make it ready to load
+            if mount_from_harvester:
```
marcus-oscarsson, line 462: Can't we somehow know from the sample `data_model` that its a harvester sample, like this you dont need `mount_from_harvester`

mxcubeweb/core/components/samplechanger.py
```diff
@@ -459,6 +481,10 @@ def queue_mount_sample(view, data_model, centring_done_cb, async_result):  # noq
                     "Sample changer could not load sample", ""
                 )
 
+    # Harvest Next sample while loading current
+    if mount_from_harvester and not HWR.beamline.harvester.get_room_temperature_mode():
```
marcus-oscarsson, line 485: Its neater to do this with a signal on the Harvester `HardwareObject`, so that when a sample is mounted you trigger the harvesting of the following one.

mxcubeweb/core/components/samplechanger.py
jbflo: :worried: That can cause unexpected behavior 

Like this code for example : 
image

we don't want the MANUAL CENTRING METHOD to be triggered also.

marcus-oscarsson: I'm not sure exactly what code above refers to but `centring_method` should somehow refer to the setting set by the "global setting", so it should not be a problem (or we should make sure that its not :) )

marcus-oscarsson:
Ok, so I'm merging this now and then we can follow up with the some improvements :)

PR #1156, opened 2024-01-22
Remove `devtools` from the Python dependencies

fabcor-maxiv:


fabcor-maxiv:
As far as I can tell `devtools` is not used anywhere. It might have landed in the dependencies by mistake. At best it could be a dev dependency, but we already have enough dependencies, I do not thing it is worth keeping this one (even in dev).

fabcor-maxiv:
When https://github.com/mxcube/mxcubeweb/pull/1157 is merged, I will rebase.

marcus-oscarsson (approved):

marcus-oscarsson:
Thanks, looks good

PR #1158, opened 2024-01-23
Use `https` instead of `http` in documentation

fabcor-maxiv:


fabcor-maxiv:
Small thing, relatively pointless change, but it was bothering me : D

Issue #1047, opened 2023-09-19
Notes about first steps with MXCuBE web development


fabcor-maxiv:
Notes taken while trying to build an MXCuBE development environment from scratch, following the [instructions on `mxcube/mxcubeweb` *wiki* page "*Home*"](https://github.com/mxcube/mxcubeweb/wiki).

* Is it mxcube3 or mxcubeweb?

* Version numbers of mxcube3 / mxcubeweb? The following does not make much sense:

  > The underlying beamline control layer is implemented using the library mxcubecore previously known as HardwareRepository. The mxcubecore module is compatible with both MXCuBE3 also known as MXCuBE-Web and the MXCuBE-Qt application. The earlier versions of MXCuBE3 (upto 3.2.x) uses HardwareRepository while versions after 4.x uses mxcubecore.
  >
  > The install instructions for version 3 of MXCUBE-Web can be found here (instructions 3.x)
  >
  > The install instructions for version 4 can be found below.

  https://github.com/mxcube/mxcubeweb/wiki#hardwarerepository-and-mxcubecore

* Document does not say to go into `mxcubeweb` directory for *conda* environment creation command, where the *Yaml* file is located.

* Repo is `mxcubeweb` but import package is `mxcube3`. Why?

  *Note: if MXCuBE is an application then it is quite likely that nothing else needs to import `mxcube3` and it should be possible to safely rename to `mxcubeweb` (or simply `mxcube`?) without breaking others.*

* `poetry install` does not work.

  > $ poetry install
  >
  > Installing dependencies from lock file
  >
  > Warning: poetry.lock is not consistent with pyproject.toml. You may be getting improper dependencies. Run `poetry lock [--no-update]` to fix it.
  >
  > Because mxcubeweb depends on bcrypt (^4.0.1) which doesn't match any versions, version solving failed.

  *Note: It seems like `poetry lock` helps to get past the issue.
  But is that the correct fix? Is that really what we want?*

* Why do we have a section `D1`? Why suddenly `D`? Is that a typo or a copy-paste from an older document?

* The instruction `cd mxcube3/ui` seems incorrect. First this assumes we are back outside of the repo. And also obviously it should be `cd mxcubeweb/ui`. 

* `npm install` does not work.

  *Note: it seems like `npm install --legacy-peer-deps` could help. But is that really a good solution? Does it break something else? There is still a bunch of warning with "high security vulnerabilities"*

[...]

PR #1161, opened 2024-01-24
Remove build deps: `setuptools`, `setuptools-scm`

fabcor-maxiv:
`setuptools` and `setuptools-scm` were listed as build dependencies in `pyproject.toml`'s  `build-system.requires`.
But this was non-sense, since the build backend is `poetry-core`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,7 +76,7 @@ exclude = '''
 '''
 
 [build-system]
-requires = ["poetry-core>=1.0.0", "setuptools>=45", "setuptools_scm[toml]>=6.2"]
+requires = ["poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
 
 [tool.poetry.scripts]


fabcor-maxiv:
As far as I know, there is no reason to have `setuptools` as dependency.

We already started discussing this here: https://github.com/mxcube/mxcubeweb/issues/1047#issuecomment-1731082658

I tried `pip install --editable .` and did not notice any issue.

fabcor-maxiv:
When running commands like `poetry check`, I get the following warning message:

> WARNING setuptools_scm._integration.setuptools pyproject.toml does not contain a tool.setuptools_scm section

As far as I can tell, this warning is meaningless for us. I am following this topic here: https://github.com/pypa/setuptools_scm/issues/953

fabcor-maxiv:
The command `poetry show --tree` tells us the following:

```
mxcubecore 1.67.0 Core libraries for the MXCuBE application
├── [...]
├── matplotlib >=3.6.2,<4.0.0
│   ├── [...]
│   └── setuptools-scm >=7 

```

elmjag:
I tested installing this both with:

    pip install -e .

and

    poetry install

Works good, and you get editable installation in both cases.

elmjag (approved):

marcus-oscarsson:
Excellent, thanks !

PR #1162, opened 2024-01-25
Simplify Python development dependencies

fabcor-maxiv:
Remove `conda-environment-dev.yml` since it is not useful. The Python development dependencies are either in the Poetry `dev` dependency group or in `pre-commit` hooks.

Adjust the "Development environment" documentation page accordingly.


marcus-oscarsson:
Thanks, makes sense

PR #1164, opened 2024-02-02
fix issues with starting 3-click centering

elmjag:
Currently when clicking '3-click centering' button nothing happens. In the browser's console, you can see that there is an JavaScript error when trying to read 'user in control' flag.

I think the problem is just that the 'path' to the 'in control' flag is wrong. This PR changes that to the one I think is correct.


fabcor-maxiv (approved):

marcus-oscarsson:
Thanks, I spotted this one as well, sorry I did not make a PR sooner.

PR #1166, opened 2024-02-02
Clean GitHub Actions workflows

### LINKED ISSUE #1165: LDAP dependencies in GitHub Actions workflows
In some GitHub Actions workflows we explicitly install LDAP dependencies outside of the conda environment directly on the system. We can probably get rid of this step because the LDAP dependencies are installed in the conda environment.

fabcor-maxiv:
Always use the conda environment for better consistency.

Do not install Python dependencies when formatting/linting JavaScript.

GitHub: fixes https://github.com/mxcube/mxcubeweb/issues/1165

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -36,9 +36,6 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
-    - name: Install LDAP dependencies
-      run: sudo apt-get -y install libsasl2-dev libldap2-dev libssl-dev
-
     - name: Set up Conda environment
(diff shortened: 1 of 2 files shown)


fabcor-maxiv:
@axelboc and @marcus-oscarsson I would like your feedback on this. I am not sure it is in line with what you have in mind.

axelboc:
Conda and Python dependency management are really not my areas of expertise, so I'm just very glad you're cleaning up my mess in `ui.yml`! :grin:

marcus-oscarsson:
Yes, this looks good, thanks !

PR #1167, opened 2024-02-06
drop unused socket.io emitter for safety shutter

elmjag:
Remove 'shutterStateChanged' based socket.io emitter for safey shutter hardware object.

The safety_shutter_state_changed() is never invoked, as safey shutter 'shutterStateChanged' signal is not send.

The safety shutter socket.io messages are now send from a generic emitter method AdapterBase.emit_ho_value_changed().

### DIFF
--- a/mxcubeweb/core/components/beamline.py
+++ b/mxcubeweb/core/components/beamline.py
@@ -56,21 +56,6 @@ class Beamline(ComponentBase):
             logging.getLogger("MX3.HWR").exception(msg)
 
         try:
-            safety_shutter = HWR.beamline.safety_shutter
-            if safety_shutter is not None:
-                safety_shutter.connect(
-                    safety_shutter,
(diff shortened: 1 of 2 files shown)


fabcor-maxiv (approved):

mxcubeweb/core/components/beamline.py
```diff
@@ -55,21 +55,6 @@ def init_signals(self):
             msg = "error connecting to beamline actions hardware object signals"
             logging.getLogger("MX3.HWR").exception(msg)
 
-        try:
```
marcus-oscarsson, line 58: I agree that this is should be removed and should handled through `stateChanged`, I had a quick look and it seems like "shutterStateChanged" are still used in some places in `mxcubecore` but I think we can consider "shutterStateChanged" deprecated by now and should be replaced with `stateChanged` via `AbstractNState`

PR #1168, opened 2024-02-07
Use default Python version in pre-commit hooks

fabcor-maxiv:


fabcor-maxiv:
Same as for mxcubecore: https://github.com/mxcube/mxcubecore/pull/846

marcus-oscarsson:
Thanks

PR #1169, opened 2024-02-07
Fix ReadTheDocs configuration

fabcor-maxiv:
Use the `--force` flag of `mamba env create` in case environment exists.

Remove the workaround for the unwarranted error about "missing `conda.environment`" configuration setting, because it has been fixed already. See:
* https://github.com/readthedocs/readthedocs.org/pull/10979#issuecomment-1896036953
* https://github.com/readthedocs/readthedocs.org/pull/11040

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -20,14 +20,9 @@ build:
     python: "mambaforge-22.9"
 
   commands:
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
Well done !

PR #1170, opened 2024-02-16
JSON-schema generated user interface

jbflo:
[https://mxcubecore.readthedocs.io/en/stable/dev/json-schema-generated-user-interface.html](https://mxcubecore.readthedocs.io/en/stable/dev/json-schema-generated-user-interface.html)

image
image
image


rhfogh (approved):
VERY good to see.

The UI works (though I cannot comment on the nature of the Java code).

marcus-oscarsson:
Looks, very nice ! Ill have a closer look on Monday, thanks alot !

ui/src/containers/GphlWorkflowParametersDialog.jsx
```diff
@@ -76,8 +412,12 @@ function mapDispatchToProps(dispatch) {
...
       dispatch,
     ),
-    submitWorkflowParameters: bindActionCreators(
-      submitWorkflowParameters,
+    updateGphlWorkflowParameters: bindActionCreators(
+      updateGphlWorkflowParameters,
+      dispatch,
+    ),
+    resetUpdatedGphlWParameters: bindActionCreators(
+      () => updateGphlWorkflowParametersDialog(null, false),
```
axelboc, line 418: I would prefer to see this business logic inside a dedicated action creator in `actions/workflow.js` called `resetUpdatedGphlWParameters`.

ui/src/containers/WorkflowParametersDialog.css
axelboc, line 1: These are all global styles. Any chance you could put them in a CSS module `WorkflowParametersDialog.module.css`? You don't have to migrate the existing styles; I just would like to avoid adding more global styles.

ui/src/containers/WorkflowParametersDialog.css
```diff
@@ -5,3 +5,76 @@
...
+
+.indexing_table_special_td_th {
+  border: none !important;
+}
+
+.indexing_table_row_selected {
+  border: 2px dashed #4099ec !important;
+}
+
+.HIGHLIGHT {
```
axelboc: Seems like these could be `data-` attributes instead of class names: `data-highlight`, etc. or better, since they all seem mutually exclusive: `data-variant="highlight"`, etc.

ui/src/containers/GphlWorkflowParametersDialog.jsx
axelboc, line 1: This component could really benefit from being split into multiple components. :grin: You have a bunch of presentational components that can be extracted, and then you could at least separate out the modal logic from the form logic to avoid, for instance, having an effect that depends on the `show` prop to decide when to initialise the form.

ui/src/containers/WorkflowParametersDialog.css
```diff
@@ -5,3 +5,76 @@
...
+.gphl_row_box > .parameters{ */
+.gphl_row_box > .indexing_solution {
+  border: #e74444 1.5px dotted !important;
+}
+
+.title_box #title {
+  position: relative;
+  color: #01012e;
+  font-size: large;
+  top: -0.7em;
```
axelboc: Most of the time, a negative `margin-top` works just as well without the need for `position: relative`.

ui/src/containers/WorkflowParametersDialog.css
```diff
@@ -5,3 +5,76 @@
...
+
+.indexing_table > thead,
+.indexing_table > tbody,
+.indexing_table > tr,
+.indexing_table > td,
+.indexing_table > th {
+  display: block;
+}
+
+.indexing_table > tr:after {
```
axelboc: `::after`

ui/src/containers/WorkflowParametersDialog.css
```diff
@@ -5,3 +5,76 @@
...
+  border: #3c5a86 1.5px dotted;
+}
+
+/* .gphl_row_box > .crystal_data,
+.gphl_row_box > .parameters{ */
+.gphl_row_box > .indexing_solution {
+  border: #e74444 1.5px dotted !important;
+}
+
+.title_box #title {
```
axelboc: Can you do without an ID selector here?

ui/src/containers/WorkflowParametersDialog.css
```diff
@@ -5,3 +5,76 @@
...
+  width: 100%;
+}
+
+.indexing_table > thead,
+.indexing_table > tbody,
+.indexing_table > tr,
+.indexing_table > td,
+.indexing_table > th {
+  display: block;
+}
```
axelboc: Any way you could work _with_ the CSS table layout instead of against it? If the problem is to scroll the body and freeze the header row, you could make the table's container scrollable and use `position: sticky` on the header row. Just a thought, I don't have strong feeling about it :grin:

ui/src/containers/GphlWorkflowParametersDialog.jsx
```diff
@@ -1,62 +1,396 @@
...
+              key={tdContent}
+              className={`${
+                selected.includes(index) ? 'indexing_table_row_selected' : ''
+              }
+                  ${
+                    indexingTable.highlights[index + 1]
+                      ? indexingTable.highlights[index + 1][0]
+                      : ''
+                  } trclass`}
+              onClick={() => onSelectRow(index, tdContent)}
```
axelboc, line 43: Would be good to make this a bit more accessible. I don't think you could interact with your table with the keyboard, for instance. A column with a native checkbox would solve this. I'm sure you could find a bunch of good examples of accessible tables with selectable rows online. You could even use a headless UI library such as React Aria: https://react-spectrum.adobe.com/react-aria/Table.html

axelboc (commented):
Nice work! Apart from the global CSS styles, nothing else is blocking, though it would be nice to refactor a few things (in separate PRs if you prefer) to make the code more maintainable and the UI more accessible.

ui/src/containers/GphlWorkflowParametersDialog.jsx
jbflo: Sure, I prefer using a custom table for now, cause in the future the uiSchema may become more complex where you'll probably need to select column  instead of Row. 

That's a good one by the way, i did not know about it : react-aria-components

ui/src/containers/WorkflowParametersDialog.css
jbflo: not here!  :-1:

ui/src/containers/WorkflowParametersDialog.css
jbflo: Even i am not a great fan of  CSS module 😁 , cause i try to always make sure having unique class name.
I'll make the change.   :+1:

ui/src/containers/WorkflowParametersDialog.css
jbflo: Not really, we wanted to have the filed or row background-color changed

axelboc: Just meant that data attributes are generally more convenient for this sort of logic, especially with CSS modules as it avoids the need for composing or concatenating classes and it allows you to use the power of [attribute selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors):

```tsx
<div className="foo" data-variant={variant} data-highlight={someBool || undefined} />
```

```css
.foo[data-variant="bar"] { ... }
.foo[data-variant^="b"] { ... }
.foo[data-highlight] { ... }
.foo:not([data-highlight]) { ... }
```

jbflo: Ohh yeah I got it now. :+1:

ui/src/containers/GphlWorkflowParametersDialog.jsx
rhfogh: The current way the system works (as coded for Qt) you actually select and highlight neither columns, nor rows but *cells*. Since the only current use is for a single-column table, it is no problem if you can only select rows - for now - but selecting cells is the most likely evolution. Of course we might want even more different or complex widgets at some point, but those would have to be coded when we need them.

mxcubeweb/core/adapter/beamline_adapter.py
```diff
@@ -50,9 +50,10 @@ def gphl_json_wf_parameters_needed(self, schema, ui_schema):
         params["ui_schema"] = ui_schema
         self.app.server.emit("gphlWorkflowParametersDialog", params, namespace="/hwr")
 
-    def gphl_json_wf_update_ui_parameters(self, params):
-        # Contain only updated Values
-        self.app.server.emit("gphlWorkflowParametersDialog", params, namespace="/hwr")
+    def gphl_json_wf_update_ui_parameters(self, update_dict):
```
marcus-oscarsson, line 53: We could perhaps merge `wf_parameters_*` and `gphl_json_wf_parameters_*` somehow. I dont think it should hold back this PR, but maybe as a followup

ui/src/actions/workflow.js
```diff
@@ -1,4 +1,7 @@
-import { sendSubmitWorkflowParameters } from '../api/workflow';
+import {
+  sendSubmitWorkflowParameters,
+  sendUpdatedGphlWorkflowParameters,
+} from '../api/workflow';
```
marcus-oscarsson, line 4: If we harmonize the methods on the back end we might be able to get rid of some of this extra logic, so that we for instance don't need ` sendSubmitWorkflowParameters` and ` sendUpdatedGphlWorkflowParameters`

marcus-oscarsson:
Thanks JB, it would be nice to follow up with one or two PRs, splitting up the UI components and maybe one to harmonize some of the backend bits.

PR #1173, opened 2024-03-07
Focus control upstream

meguiraun:
hi guys,

This is an attempt to bring Biomax's old focus control element into latest mxcube. 

WIP because your input is needed on how we handle these kind of components, it technically works as it is. 

- Added a new BeamFocus component, it currently has Biomax specific logic but i did not want to generalise too much too early
- Configurable "beam control" elements in the ui

Screenshot 2024-03-04 at 08 48 34


ui/package.json
```diff
@@ -59,6 +59,7 @@
     "react-router": "^6.2.1",
     "react-router-bootstrap": "^0.26.1",
     "react-router-dom": "^6.19.0",
+    "react-select": "5.8.0",
```
axelboc: Hey @meguiraun, you should use Bootstrap's `Form.Select` component instead. This is what is used for other drop-down selectors. For instance, check out the code for the "Beam size" selector (the component is called `ApertureInput`).

meguiraun: mmm.... I cannot customize the style of the different options. 

Based on the currently selected beam focus, the options for the aperture will be displayed with a different color or even disabled.

react-select allows me this customization, but I am struggling with Form.Select
image

axelboc: Indeed, in that case you should use Bootstrap's `Dropdown` component. https://react-bootstrap.netlify.app/docs/components/dropdowns#customization

meguiraun: you are my hero 🤗

marcus-oscarsson:
Just a question, how is this control used, is to de-focus the beam to change the beam size ?

meguiraun:
> Just a question, how is this control used, is to de-focus the beam to change the beam size ?

The user selects a beam size appropriate for their crystal size, which corresponds to different beam (de)focus

meguiraun:
I refactored few things here and there. In essence I added the beam definer information into the beamInfo info dict, and propagate it to the UI.

I based on the ID232 beam definer approach as NState hwobj, and adapted from my previous attempt, now "100x100" is one option in the definer object.

One thing that I have not solved is how to generalize the UI styling (see [this](https://github.com/mxcube/mxcubeweb/blob/focus_control_upstream/ui/src/components/SampleView/BeamDefinerInput.js#L39) ), any suggestion is welcome

I also created a new beam definer mockup:
https://github.com/mxcube/mxcubecore/pull/872

beam_definer

meguiraun:
I moved down into the hwobj the custom css styling.

please, have a look!

ui/src/components/SampleView/BeamDefinerInput.js
```diff
@@ -0,0 +1,126 @@
...
+    let col;
+    if (status === 'disable') {
+      col = 'red';
+    } else if (status === 'warning') {
+      col = '#fc9003'; // darker orange
+    } else {
+      col = 'black';
+    }
+    return col;
+  }
```
axelboc: Nothing critical, but I can think of a few ways to refactor this:

```js
  itemColor(status) {
    if (status === 'disable') {
      return 'red';
    }

    if (status === 'warning') {
      return '#fc9003'; // darker orange
    }

    return 'black';
  }
```

... or with a record:

```js
// at the top of the file
const ITEM_COLOR = {
  disable: 'red',
  warning: '#fc9003',
}

// in the `JSX` method:
fontColor: ITEM_COLOR[status] || 'black';
```

... or even better, to keep JS and CSS neatly separated, you can just store the `status` inside each `option` object and then set a `data-` attribute on each `Dropdown.Item`:

```jsx
import styles from './BeamDefinedInput.module.css'; // CSS Module to avoid global styles; otherwise put the styles in `SampleView.css` and just pick a very unique class name for the options

<Dropdown.Item
  // ...
  className={styles.option} // 
  data-status={status}
/>
```

 Then, in the CSS:

```css
.option {
  /* color: black; */ /* probably the default color, so can be left out */
  float: left; /* you have the `float-left` class in the JSX, which seems strange to me, but there's probably a good reason I'm not seeing */
}

.option[data-status="disable"] { color: red; }
.option[data-status="warning"] { color: #fc9003; }
```

ui/src/components/SampleView/BeamDefinerInput.js
```diff
@@ -0,0 +1,126 @@
...
+import { Form } from 'react-bootstrap';
+import Dropdown from 'react-bootstrap/Dropdown';
+import ButtonGroup from 'react-bootstrap/ButtonGroup';
+
+import '../MotorInput/motor.css';
+import '../input.css';
+import './SampleView.css';
+import cx from 'classnames';
+
+export default class BeamDefinerInput extends React.Component {
```
axelboc: This should be quite straightforward to write as a functional component if you're up for it.

ui/src/containers/SampleViewContainer.js
```diff
@@ -126,24 +137,44 @@ class SampleViewContainer extends Component {
...
+              ) : null}
 
-              <div>
-                <p className="motor-name">Beam size:</p>
-                <ApertureInput
+              {beamFocus.show ? (
+                <BeamDefinerInput
+                  beamDefinerInputList={this.props.sampleViewState.definerList}
+                  currentDefiner={this.props.sampleViewState.currentDefiner}
+                  customStyling={this.props.sampleViewState.customStyling}
```
axelboc, line 169: I feel like the logic in the `render` method of `BeamDefinerInput` would be much simpler if you passed this through the `apertureList` prop:

```jsx
apertureList={this.props.sampleViewState.customStyling || this.props.sampleViewState.apertureList}
```

This way, you can have a single loop to create the `optionList` array (you may not even need this `optionList` array at all, since the loop below can be moved directly into the JSX):

```jsx
const optionList = Object.entries(this.props.apertureList).map((key, value) => {
  // key is the label (e.g. "50x50")
  // value is either the same as the label or an object when `customStyling` is defined, so you can do:
  if (typeof value === 'object') {
    return { ... }; // or return a `<Dropdown.Item />` directly if you move the loop inside the JSX
  }
  return { ... }
});
```

ui/src/containers/SampleViewContainer.js
```diff
@@ -126,24 +137,44 @@ class SampleViewContainer extends Component {
...
-              <div>
-                <p className="motor-name">Phase Control:</p>
-                <PhaseInput
-                  phase={this.props.sampleViewState.currentPhase}
-                  phaseList={this.props.sampleViewState.phaseList}
-                  changePhase={this.props.sampleViewActions.changeCurrentPhase}
-                  state={diffractometerHo.state}
-                />
-              </div>
+              {phase.show ? (
```
axelboc: The `phase.show && (` syntax can save you from using a ternary.

axelboc (commented):
Looking great! I've got lots of suggestions but don't feel like you have to resolve them — it's all coding style, nothing major.

ui/src/components/SampleView/BeamDefinerInput.js
meguiraun: I am up for a challengue! good suggestion!

ui/src/components/SampleView/BeamDefinerInput.js
meguiraun: I simplify this logic from your suggestion. I could not make it work with css though

ui/src/components/SampleView/BeamDefinerInput.js
meguiraun: functionalized!

marcus-oscarsson:
Just a status check, where are we with this one @meguiraun ?

fabcor-maxiv:
@marcus-oscarsson @meguiraun Is it okay to delete the branch?

PR #1174, opened 2024-03-07
GUI/ Styles Refactor Beamline setup InOutSwitch component

jbflo:
- Use InOutSwitch to display more component /device that state can be switch 
- Remove round border on badge label, Put it back how it was before bootstrap 5 updated. 
-   Add Difractomerter as DeviceState  ????
image

-  Remove Sample Changer swicth component ,  InOutSwitch can do the job
-  add css module


ui/src/components/InOutSwitch/InOutSwitch.jsx
```diff
@@ -1,110 +1,55 @@
-/* eslint-disable react/jsx-handler-names */
 import React from 'react';
 import { Badge, Button, OverlayTrigger, Popover } from 'react-bootstrap';
-import './style.css';
+import styles from './inOutStyle.module.css';
 
 export default class InOutSwitch extends React.Component {
```
marcus-oscarsson: We should maybe make the effort to convert this component to a functional component

ui/src/components/InOutSwitch/InOutSwitch.jsx
```diff
@@ -1,110 +1,55 @@
...
-  onOptionsRightClick(e) {
-    // this.optionsOverlay.handleToggle();
-    this.showLabelOvelay(!this.state.showLabelOvelay);
-    e.preventDefault();
+    this.handleSetOff = this.handleSetOff.bind(this);
+    this.handleSetOn = this.handleSetOn.bind(this);
   }
 
-  setOff() {
+  handleSetOff() {
```
marcus-oscarsson: Nice :+1:

ui/src/components/DeviceState/deviceState.module.css
```diff
@@ -0,0 +1,20 @@
+.deviceState {
```
marcus-oscarsson, line 1: Very nice with a css module !

ui/src/components/DeviceState/DeviceState.jsx
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
 import { Badge } from 'react-bootstrap';
 
+import styles from './deviceState.module.css';
+
 export default class DeviceState extends React.Component {
```
marcus-oscarsson, line 6: Same here we could make this a functional component as well

ui/src/containers/BeamlineSetupContainer.jsx
```diff
@@ -280,17 +279,21 @@ class BeamlineSetupContainer extends React.Component {
               />
             </Nav.Item>
           </Nav>
-          <Nav className="me-3">
+          <Nav className="">
             <Nav.Item>
-              <SampleChangerSwitch
```
marcus-oscarsson, line 285: Yes ! :1st_place_medal:

marcus-oscarsson:
Very nice ! lots of good improvements. I thought it would also be a good occasion to convert these components to functional. But we can do that in followup PR if you prefer (it also keeps the size down a bit)

ui/src/components/DeviceState/deviceState.module.css
```diff
@@ -0,0 +1,20 @@
...
+  text-overflow: ellipsis;
+
+  border-radius: 0%;
+  display: block;
+}
+
+.msgLabelStyle {
+  display: block;
+  font-size: 100%;
+  border-radius: 0px;
```
axelboc: `0px`, `0%`, etc. is the same as just `0`

ui/src/components/InOutSwitch/InOutSwitch.jsx
```diff
@@ -1,110 +1,55 @@
...
-    // this.overlay.hide();
-    this.showValueOvelay(false);
   }
 
-  setOn() {
+  handleSetOn() {
     if (this.props.onSave !== undefined) {
-      this.props.onSave(this.props.pkey, this.props.onText);
+      if (this.props.onValue === undefined) {
+        this.props.onSave(this.props.pkey, this.props.onText);
```
axelboc: It seem that `this.props.pkey` could be undefined here. Is this a problem?

ui/src/components/InOutSwitch/InOutSwitch.jsx
```diff
@@ -186,6 +158,8 @@ export default class InOutSwitch extends React.Component {
 InOutSwitch.defaultProps = {
```
axelboc: I feel like a bunch of things in this component could be simplified first if you assigned better default values to props:

- `"READY"` for `onText`;
- `"CLOSED"` for `offText`;
- `onText` for `onValue`;
- `offText` for `offValue`;
- ...

ui/src/components/InOutSwitch/InOutSwitch.jsx
```diff
@@ -113,53 +58,74 @@ export default class InOutSwitch extends React.Component {
...
-          Set: {this.props.offText}
-        </Button>
-      );
-    } else {
-      btn = (
-        <Button variant="outline-secondary" size="sm" onClick={this.setOn}>
-          Set: {this.props.onText}
-        </Button>
-      );
+    switch (this.props.data.value) {
```
axelboc: Never a big fan of a `switch`, especially when used for imperative coding. :stuck_out_tongue_closed_eyes:  The most straightforward solution often lies in extracting a function and having simple `if` blocks, each with a `return` statement. Depending on the situation, the function can be a pure utility function, a method/closure within this component, or ... a new component.

However, it is best to first consider whether the code can be simplified ahead of the switch. Here, since this component is a toggle/switch I feel like a bunch of the complexity could be removed by first computing a boolean variable called `isOn` (or `isOff`, doesn't matter). Of course, there are edge cases, but having that variable would allow you to use ternaries throughout the JSX without resorting to mutable variables.

ui/src/components/InOutSwitch/InOutSwitch.jsx
```diff
@@ -1,110 +1,55 @@
...
+        this.props.onSave(this.props.pkey, this.props.offValue);
+      }
     }
-
-    // this.overlay.hide();
-    this.showValueOvelay(false);
   }
 
-  setOn() {
+  handleSetOn() {
```
axelboc: I feel like it would be simpler to have a single event handler with an extra line at the top that, given my previous comments, would look something like:

```js
const newValue = this.isOn ? this.props.offValue : this.props.onValue;
```

ui/src/containers/BeamlineSetupContainer.jsx
axelboc: Very nice :100:

ui/src/containers/BeamlineSetupContainer.jsx
```diff
@@ -280,17 +279,21 @@ class BeamlineSetupContainer extends React.Component {
...
+              <InOutSwitch
+                onText="Power On"
+                offText="Power Off"
+                onValue="PowerOn"
+                offValue="PowerOff"
                 labelText="Sample Changer"
-                data={this.props.sampleChanger.state}
+                data={{ value: this.props.sampleChanger.state }}
```
axelboc: Makes me wonder whether the prop should be `value` directly.

ui/src/components/InOutSwitch/inOutStyle.module.css
```diff
@@ -0,0 +1,36 @@
+.inOutSwitch {
+  width: fit-content;
+}
+
+.SwitchBtn {
```
axelboc: I would stick to `camelCase` for consistency.

axelboc (approved):
Nicely done :100:, just some advice for refactoring things further if you're interested.

ui/src/containers/BeamlineSetupContainer.jsx
jbflo: Agreed

ui/src/components/InOutSwitch/InOutSwitch.jsx
jbflo: as Support for defaultProps is deprecated
I'll remove the defaultProps

ui/src/components/InOutSwitch/InOutSwitch.jsx
jbflo: me either, but i think Eslint force you use switch when if block getting bigger. 
i'll think about it.

ui/src/components/InOutSwitch/InOutSwitch.jsx
axelboc: Yeah with functional components it's much simpler:

```js
function MyComponent(props) {
  const { foo = 'some-default-value', bar = foo } = props;
  // ...
}
```

jbflo:
Done for now, thanks for all the suggestion!

PR #1176, opened 2024-03-13
update browser list used when building UI bundle

elmjag:
Commiting results of running command:
    
    npx update-browserslist-db@latest
    
This updates package 'caniuse-lite' and available browser data for polyfills tools. Make the generated bundle size slightly smaller and in theory better suited for current browsers.

Also [drops sourceMappingURL directive](https://github.com/mxcube/mxcubeweb/commit/bc23770f8b9b38fadd02b6e9c3b9f4d606fbd92c) as it now generates an error when building UI bundle.

### DIFF
--- a/ui/pnpm-lock.yaml
+++ b/ui/pnpm-lock.yaml
@@ -4281,7 +4281,7 @@ packages:
       postcss: ^8.1.0
     dependencies:
       browserslist: 4.21.7
-      caniuse-lite: 1.0.30001492
+      caniuse-lite: 1.0.30001597
       fraction.js: 4.2.0
       normalize-range: 0.1.2
       picocolors: 1.0.0
@@ -4721,7 +4721,7 @@ packages:
     engines: {node: ^6 || ^7 || ^8 || ^9 || ^10 || ^11 || ^12 || >=13.7}
     hasBin: true
     dependencies:
-      caniuse-lite: 1.0.30001492
+      caniuse-lite: 1.0.30001597
       electron-to-chromium: 1.4.417
       node-releases: 2.0.12
       update-browserslist-db: 1.0.11(browserslist@4.21.7)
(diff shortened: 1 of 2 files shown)


fabcor-maxiv (approved):

axelboc:
Did you mean to upgrade all the other packages? I guess `udate-browserslist-db` must run `pnpm update`, and since we have all the versions defined with `^`, it bumped everything. Alternatively, you could remove all the `^` (making sure that the versions that are actually installed match the ones in `package.json`) and only then update browserslist. :shrug:

I'm usually not a fan of bumping "blindly", even just for minors/patches, but there are sooo many dependencies, I don't have the courage to go through them all one by one... :joy: So I'm fine with it either way.

fabcor-maxiv:
@axelboc What is the *pnpm* command to update only `caniuse-lite´, if there is any?

Why would we remove the `^` version specifier? We have the lockfile for this, no?

axelboc:
You could just try `pnpm up caniuse-lite` maybe. Otherwise, removing the `^` would allow running `update-browserslist-db` while ensuring no other package gets bumped.

elmjag:
Yeah, I don't like upgrading this much packages in one go either. I'll see if I can find a way to upgrade minimum number of packages, hopefully just `caniuse-lite´.

elmjag:
Ok, I figured out a way to update `caniuse-lite` package only. For the record, this is how I did it.

First I run this command:

    pnpm install caniuse-lite@latest

Then I manually removed the package `caniuse-lite` as the top-level dependency from package.json and pnpm-lock.yaml files.

ui/pnpm-lock.yaml
```diff
@@ -4827,14 +4827,13 @@ packages:
...
+      caniuse-lite: 1.0.30001597
       lodash.memoize: 4.1.2
       lodash.uniq: 4.5.0
     dev: true
 
-  /caniuse-lite@1.0.30001492:
-    resolution: {integrity: sha512-2efF8SAZwgAX1FJr87KWhvuJxnGJKOnctQa8xLOskAXNXq8oiuqgl6u1kk3fFpsp3GgvzlRjiK1sl63hNtFADw==}
-    dev: true
+  /caniuse-lite@1.0.30001597:
+    resolution: {integrity: sha512-7LjJvmQU6Sj7bL0j5b5WY/3n7utXUJvAe1lxhsHDbLmwX9mdL86Yjtr+5SRCyf8qME4M7pU2hswj0FpyBVCv9w==}
```
fabcor-maxiv, line 4836: @elmjag, we lost the `dev: true` in the operation, though.

I do not know if it has any meaningful consequence.

Maybe @axelboc knows.

axelboc: Hmm well spotted. Not a big deal I reckon, but just in case:

```
pnpm install --save-dev caniuse-lite@latest
```

Then remove `caniuse-lite` from `package.json` as you've done before and run `pnpm install` again to make sure it doesn't remodify `pnpm-lock.yaml`.

PR #1177, opened 2024-03-14
docs: adds `Working with documentation´ section

elmjag:
Adds documentation on how documentations is created and how to change it.

### DIFF
--- a/docs/source/dev/docs.md
+++ b/docs/source/dev/docs.md
@@ -1,0 +1,28 @@
+# Working with documentation
+
+This documentation is automatically built and published from the contents of the [mxcubeweb](https://github.com/mxcube/mxcubeweb) repository.
+Each time the repository's `develop` branch is updated, documentation is regenerated and published to [https://mxcubeweb.readthedocs.io/](https://mxcubeweb.readthedocs.io/)
+
+If you want to modify this documentation, make a pull request to the repository with your suggested changes.
+
+## Modifying the documentation
+
+This documentation is built using the [Sphinx](https://www.sphinx-doc.org/) documentation generator.
+The documentation's source and configuration files are located in the `docs` folder of the [mxcubeweb](https://github.com/mxcube/mxcubeweb) repository.
+Sphinx will also read [Python docstrings](https://peps.python.org/pep-0257/) from the repository's source code.
+
+### Building documentation
+
+Follow the instructions for [Installing a development environment](environment.md).
+The development environment will include Sphinx and all necessary packages for building documentation.
+
+Once you have a working environment, use these commands to build the documentation:
+
+    # goto docs folder
+    $ cd docs
+
+    # build documents with Sphinx
+    $ make html
+
+The commands above will generate documentation in `docs/html/` directory.
+The generated docs can be viewed by opening `docs/html/index.html` in your web browser.
--- a/docs/source/dev/index.rst
+++ b/docs/source/dev/index.rst
@@ -8,4 +8,11 @@ Developer documentation
     :maxdepth: 2
     :titlesonly:
 
-    *
+    api.rst
+    contributing.md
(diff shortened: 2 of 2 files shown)


docs/source/dev/docs.md
```diff
@@ -0,0 +1,32 @@
...
+This documentation is automatically built and published from the contents of the [mxcubeweb](https://github.com/mxcube/mxcubeweb) repository.
+Each time the repository's `develop` branch is updated, documentation is regenerated and published to [https://mxcubeweb.readthedocs.io/](https://mxcubeweb.readthedocs.io/)
+
+If you want to modify this documentation, make a pull request to the repository with your suggested changes.
+
+## Modifying the documentation
+
+This documentation is built using the [Sphinx](https://www.sphinx-doc.org/) documentation generator.
+The documentation's source and configuration files are located in `docs` folder of the [mxcubeweb](https://github.com/mxcube/mxcubeweb) repository.
+The Sphinx will also read python docstrings from the repository's source code.
```
axelboc:
```suggestion
The documentation's source and configuration files are located in the `docs` folder of the [mxcubeweb](https://github.com/mxcube/mxcubeweb) repository.
Sphinx will also read [Python docstrings](https://peps.python.org/pep-0257/) from the repository's source code.
```

axelboc (approved):
:100:

docs/source/dev/index.rst
```diff
@@ -8,4 +8,11 @@ Developer documentation
...
 
-    *
+    api.rst
+    contributing.md
+    environment.md
+    docs.md
+    front-end.md
+    login.rst
+    sample-changer-changes-and-maintenance.md
+    tutorial.md
```
fabcor-maxiv, line 18: Why? Was there a problem with just the wildcard `*`?

Now we risk having people forgetting to add their new document to this list and wondering why their new document does not show up.

docs/source/dev/docs.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Building documentation
+
+To build the documentation locally, you need to have `Sphinx` and `make` available on your system.
```
fabcor-maxiv: These are the dependencies needed

```toml
myst-parser = "^2.0.0"
sphinx = "<7.2"
sphinxcontrib-httpdomain = "^1.8.1"
furo = "^2023.9.10"
```
https://github.com/mxcube/mxcubeweb/blob/develop/pyproject.toml#L51-L55

... on top of Python, and the normal MXCuBE dependencies, because yes mxcube needs to run to build the doc, because we look at the flask routes. We could skip this if it is too annoying, since the added value of having the REST routes is not that great.

`make` helps, but not necessary

docs/source/dev/index.rst
elmjag: The problem is that topics are shown in alphabetic order, which is IMHO not the order we necessary want to use.

I wanted to put 'working with documentation' section after 'Installing a development environment'. The 'working with documentation' kind of assumes that you have a working development environment. 

But I don't know, maybe the explicit order is not worth the hassle of adding new files to index.rst.

fabcor-maxiv: Yep, I had checked and had failed to notice the `env` was before the `docs`, so I assumed it was still alphabetical order, so I was wondering why...

docs/source/dev/docs.md
elmjag: As we discussed IRL, I have simplified that section to just say 'go install dev enironment, run these commands'.

fabcor-maxiv: Let's simplify this section, and we'll write a separate (sub)-section with the more involved technical details (about the furo theme, myst-parser, make, and so on).

fabcor-maxiv (approved):

Issue #1180, opened 2024-03-15
Displaying heatmap results of grid scans does not render in the UI


fhernandezvivanco:
We are encountering an issue  where the results of grids scans are not being displayed in MXCuBE. The issue seems to be realated to the format of the `gridResultFormat` defined in the [DrawGridPlugin](https://github.com/mxcube/mxcubeweb/blob/develop/ui/src/components/SampleView/DrawGridPlugin.js)

PR #1181, opened 2024-03-15
Fix grid scan results issue

fhernandezvivanco:
The `gridResultFormat` of the `DrawGridPlugin` defaults to PNG. To fix the UI display issue, the code following the if statement on line 410 must be executed. Additionally, we only plot results when there are results available (length > 0). 

[resolves](https://github.com/AustralianSynchrotron/mxcubeweb/commit/1870cee6d36c7a2172b194fb5b2a1388af2c2d68) https://github.com/mxcube/mxcubeweb/issues/1180 (see the image below)

grid_scan_mxcube

### DIFF
--- a/ui/src/components/SampleView/DrawGridPlugin.js
+++ b/ui/src/components/SampleView/DrawGridPlugin.js
@@ -316,12 +316,11 @@ export default class DrawGridPlugin {
       result = gd.result[this.resultType];
     }
 
-    const result_length = Object.values(result).length;
     if (
       result !== undefined &&
       result !== null &&
       gd.id !== null &&
-      result_length > 0
+      Object.values(result).length > 0
     ) {
       for (let nh = 0; nh < row; nh++) {
         for (let nw = 0; nw < col; nw++) {
@@ -407,10 +406,7 @@ export default class DrawGridPlugin {
       }
 
       if (!this.drawing) {
-        if (
-          this.gridResultFormat === 'PNG' ||
-          this.gridResultFormat === 'RGB'
-        ) {
+        if (this.gridResultFormat === 'RGB') {
           const fillingMatrix = this.cellFillingFromData(
             gridData,
             gridData.numCols,
--- a/ui/src/components/SampleView/DrawGridPlugin.js
+++ b/ui/src/components/SampleView/DrawGridPlugin.js
@@ -316,7 +316,8 @@ export default class DrawGridPlugin {
       result = gd.result[this.resultType];
     }
 
-    if (result !== undefined && result !== null && gd.id !== null) {
+    const result_length = Object.values(result).length;
+    if (result !== undefined && result !== null && gd.id !== null && result_length > 0) {
       for (let nh = 0; nh < row; nh++) {
         for (let nw = 0; nw < col; nw++) {
           const index = nw + nh * col + 1;
@@ -401,7 +402,7 @@ export default class DrawGridPlugin {
       }
 
       if (!this.drawing) {
-        if (this.gridResultFormat === 'RGB') {
+        if (this.gridResultFormat === 'PNG' || this.gridResultFormat === 'RGB') {
           const fillingMatrix = this.cellFillingFromData(
             gridData,
             gridData.numCols,
--- a/ui/src/components/SampleView/DrawGridPlugin.js
+++ b/ui/src/components/SampleView/DrawGridPlugin.js
@@ -317,7 +317,12 @@ export default class DrawGridPlugin {
     }
 
     const result_length = Object.values(result).length;
-    if (result !== undefined && result !== null && gd.id !== null && result_length > 0) {
+    if (
+      result !== undefined &&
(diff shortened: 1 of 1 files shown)


ui/src/components/SampleView/DrawGridPlugin.js
```diff
@@ -316,7 +316,13 @@ export default class DrawGridPlugin {
       result = gd.result[this.resultType];
     }
 
-    if (result !== undefined && result !== null && gd.id !== null) {
+    const result_length = Object.values(result).length;
+    if (
+      result !== undefined &&
+      result !== null &&
+      gd.id !== null &&
+      result_length > 0
```
axelboc: `Object.values()` throws when passed `undefined` or `null`.
```suggestion
    if (
      result !== undefined &&
      result !== null &&
      gd.id !== null &&
      Object.values(result).length > 0
```

ui/src/components/SampleView/DrawGridPlugin.js
```diff
@@ -401,7 +407,10 @@ export default class DrawGridPlugin {
       }
 
       if (!this.drawing) {
-        if (this.gridResultFormat === 'RGB') {
+        if (
+          this.gridResultFormat === 'PNG' ||
```
marcus-oscarsson: No sure exactly what you would like to achieve here, how are you using the PNG results ?. This logic is only used to display RGB results, creating a "cell" with the corresponding RGB value for each value in the result array. The PNG image is simply an overlay with the result, so no need to generate these cells.

Or did you want the PNG data in some other way ?

fhernandezvivanco: That makes sense. I understand that a "cell" should be filled with RGB values which are sent from mxcubecore. 

However the issue that we're seeing is that when we create a grid, the format of `this.gridResultFormat`  is `PNG`, not `RGB`, therefore the block below line 410 is not executed, and the results from a grid scan are not displayed. Probably there is a better solution to this, but in the current `develop` branch we can't display grid scan results

marcus-oscarsson: Ah, hm yes its indeed a bit strange. I think some refactoring must have removed the bit that reads this from the backend. Its normally configurable in the beamline.yml file (https://github.com/mxcube/mxcubeweb/blob/develop/test/HardwareObjectsMockup.xml/beamline_config.yml#L81). So this should be set in the `sampleview` reducer and then this should set `setGridResultFormat` on the `DrawGridPlugin`.

It ought to be a quite easy fix, I wont have time to look at it right away. But feel free to give it a go if you feel inspired :). Id prefer that we fix the reading of this setting from the backend but feel free to let us know if you have some other ideas :) !?

fhernandezvivanco: Thanks! changing the `mesh_result_format` to  `RGB` partially solved the issue. 

The only remaining issue I have is that after trying to save a grid which does not have results in it throws an error (see the image below). I updated the PR which basically checks for the length of the results before executing the code of line 325. 

Let me know if this is also something that can be fixed in another way, thank you!

error_mxcube

marcus-oscarsson: Its difficult to say from this trace, maybe the output in the JS console will tell us more at what line the error occurred ?

fhernandezvivanco: The error occurs [here](https://github.com/mxcube/mxcubeweb/blob/develop/ui/src/components/SampleView/DrawGridPlugin.js#L325) . 

The output I'm getting from the error is: 
```
ERROR
Cannot read properties of undefined (reading '1')
TypeError: Cannot read properties of undefined (reading '1')
    at DrawGridPlugin.cellFillingFromData (http://localhost:3000/static/js/bundle.js:25231:78)
    at DrawGridPlugin.shapeFromGridData (http://localhost:3000/static/js/bundle.js:25290:38)
    at http://localhost:3000/static/js/bundle.js:27971:52
    at Array.forEach (<anonymous>)
    at SampleImage.renderSampleView (http://localhost:3000/static/js/bundle.js:27964:26)
    at SampleImage.componentDidUpdate (http://localhost:3000/static/js/bundle.js:27357:10)
    at commitLifeCycles (http://localhost:3000/static/js/bundle.js:208274:28)
    at commitLayoutEffects (http://localhost:3000/static/js/bundle.js:210601:11)
    at HTMLUnknownElement.callCallback (http://localhost:3000/static/js/bundle.js:193950:18)
    at Object.invokeGuardedCallbackDev (http://localhost:3000/static/js/bundle.js:193994:20)
```

It looks like the error occurs because `result` is an array of length 0 and we try to access `result[index][1]`. Adding another condition to the `if` statement ensuring that `Object.values(result).length > 0` seems to resolve the issue

marcus-oscarsson:
@fhernandezvivanco thanks, sorry for taking some time on this one

PR #1184, opened 2024-03-18
drop unused session configs

elmjag:
Remove the unused [SESSION_TYPE](https://flask-session.readthedocs.io/en/latest/config.html#SESSION_TYPE) and [SESSION_KEY_PREFIX](https://flask-session.readthedocs.io/en/latest/config.html#SESSION_KEY_PREFIX) configuration variables from server.yaml model.

These are old configs that were previously used by the Flask-Session package. Now when we are not using Flask-Session these configs are no longer needed.


marcus-oscarsson:
Thanks

PR #1186, opened 2024-03-19
What happened to the pre-commit workflow

marcus-oscarsson:


marcus-oscarsson:
Closing this, noticed that its now part of the Build and test action.

PR #1187, opened 2024-03-19
Better names for GitHub actions

marcus-oscarsson:
Its not entirely clear from the current action summary which action is which, so tired to improve the names


fabcor-maxiv (approved):
Seems fine : )

Let's see if the workflows run successfully

marcus-oscarsson:
Hm, yeah I'm struggling a bit with the branch protection and the tagging. I'm testing a few things.

PR #1189, opened 2024-03-20
drop 'adapter_properties' from server.yaml

elmjag:
According to [config model](https://github.com/mxcube/mxcubeweb/blob/develop/mxcubeweb/core/models/configmodels.py#L103) for server.yaml you can add `adapter_properties' in mxcube section.

However, I don't see how it can be used to configure adapters. As far as I can tell, non of the adapter implementations read any of these values. If it's not used, let's remove this code.

Otherwise, I want to know it is supposed to be used :)

### DIFF
--- a/mxcubeweb/app.py
+++ b/mxcubeweb/app.py
@@ -142,7 +142,6 @@ class MXCUBECore:
 
     @staticmethod
     def adapt_hardware_objects(app):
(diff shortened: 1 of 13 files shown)


marcus-oscarsson:
It was a preliminary way to configure the adapters, but its not used or at least not yet. Its easy enough to add back if we discover that we need it. So Id be fine with merging this.

PR #1190, opened 2024-03-21
add docs for mxcube-web/server.yaml config file

elmjag:


docs/source/config/server.rst
```diff
@@ -0,0 +1,214 @@
...
+        - <allowed URL>
+        ...
+
+The ``server`` section may contain keys as described below.
+
+``DEBUG``
+~~~~~~~~~
+
+Whether Flask debug mode is enabled.
+This is equivalent of Flask's `DEBUG <https://flask.palletsprojects.com/en/2.3.x/config/#DEBUG>`_ configuration value.
```
axelboc:
```suggestion
This is the equivalent of Flask's `DEBUG <https://flask.palletsprojects.com/en/2.3.x/config/#DEBUG>`_ configuration value.
```

docs/source/config/server.rst
```diff
@@ -0,0 +1,214 @@
...
+
+* server_
+* mxcube_
+
+See below an `example  server.yaml <server_yaml_example_>`_ file.
+
+server
+------
+
+The server section primarily deals with the configuration of the HTTP aspects of the back-end server.
```
axelboc:
```suggestion
The ``server`` section primarily deals with the configuration of the HTTP aspects of the back-end server.
```

docs/source/config/server.rst
```diff
@@ -0,0 +1,214 @@
...
+``CERT_KEY``
+~~~~~~~~~~~~
+
+Path of the private key file to use for SSL connections.
+
+mxcube
+------
+
+The mxcube section allows you to customize miscellaneous aspects of the MxCube application.
+It have settings for the on-axis video stream, user account management, experiment mode.
```
axelboc:
```suggestion
The ``mxcube`` section allows you to customize miscellaneous aspects of the MXCuBE application.
It contains settings for the on-axis video stream, user account management, and experiment mode.
```

docs/source/config/server.rst
```diff
@@ -0,0 +1,214 @@
...
+        class: <class name>
+        inhouse_is_staff: <boolean>
+        users:
+          - username: <user name>
+            role: <role name>
+          ...
+
+The ``mxcube`` section may contain keys as described below.
+
+``mode``
```
axelboc, line 135: Unrelated to the documentation, but wondering if this should be changed to ~`PASCAL_CASE`~`CONSTANT_CASE` for consistency (same with other `snake_case` keys).

docs/source/config/server.rst
```diff
@@ -0,0 +1,214 @@
...
+| SSX-INJECTOR | SSX injector experiments             |
++--------------+--------------------------------------+
+
+Default mode is ``OSC``.
+
+``USE_EXTERNAL_STREAMER``
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Use an external process to stream the sample view video.
+When enabled, MxCube will invoke an appropriate hook on the camera hardware object to start the streamer.
```
axelboc:
```suggestion
When enabled, MXCuBE will invoke an appropriate hook on the camera hardware object to start the streamer.
```

docs/source/config/server.rst
```diff
@@ -0,0 +1,214 @@
...
+The ``MPEG1`` and ``MJPEG`` formats are supported.
+Note that your camera hardware object must support the specified format.
+
+Default format is ``MPEG1``.
+
+``VIDEO_STREAM_URL``
+~~~~~~~~~~~~~~~~~~~~
+
+When set, it will override the default video stream URL.
+The URL that is used by the UI to fetch the sample view video.
```
axelboc: I would swap the two sentences:

```suggestion
The URL that is used by the UI to fetch the sample view video.
When set, it will override the default video stream URL.
```

docs/source/config/server.rst
```diff
@@ -0,0 +1,214 @@
...
+The specified value is passed on to the camera hardware object,
+when invoking the start stream hook.
+
+``usermanager``
+~~~~~~~~~~~~~~~
+
+This subsection allows configuring user management in the application.
+
+``class`` key specifies the class name of the user manager.
+This allows you use a custom implementation of the user manager if needed.
```
axelboc:
```suggestion
This allows using a custom implementation of the user manager if needed.
```

docs/source/config/server.rst
```diff
@@ -0,0 +1,214 @@
...
+This configuration is used when ``USE_EXTERNAL_STREAMER`` flag is enabled.
+The specified value is passed on to the camera hardware object,
+when invoking the start stream hook.
+
+``usermanager``
+~~~~~~~~~~~~~~~
+
+This subsection allows configuring user management in the application.
+
+``class`` key specifies the class name of the user manager.
```
axelboc: For consistency:

```suggestion
``class`` specifies the class name of the user manager.
```

docs/source/config/server.rst
```diff
@@ -0,0 +1,214 @@
...
+This allows you use a custom implementation of the user manager if needed.
+
+The default class name is ``UserManager``.
+
+``inhouse_is_staff`` specifies if all in-house users should automatically acquire *staff* role.
+Which users are considered in-house is managed by the session hardware object.
+
+The default value is ``True``.
+
+``users`` key allows to configure users that should acquire additional roles.
```
axelboc:
```suggestion
``users`` specifies users that should acquire additional roles.
```

docs/source/config/server.rst
```diff
@@ -0,0 +1,214 @@
...
+The contents of this key are a list of username and role specifications.
```
axelboc: Might be worth adding an example too.

```suggestion
The content of this key is a list of username and role specifications.
```

docs/source/config/server.rst
```diff
@@ -0,0 +1,214 @@
...
+Currently only *staff* role can be specified.
```
axelboc:
```suggestion
Currently, only the *staff* role can be specified.
```

axelboc (commented):
💯

docs/source/config/server.rst
fabcor-maxiv: Fun fact: there is a substitution `|project|` pre-configured in the `rst_prolog`:

https://github.com/mxcube/mxcubeweb/blob/ec1048b3d8fbf9360e6000a3b1f6264f26fa390f/docs/conf.py#L42-L44

So in principle we could write this:
```suggestion
When enabled, |project| will invoke an appropriate hook on the camera hardware object to start the streamer.
```
... and it should show up as `MXCuBE-Web` in the resulting output (HTML), if I did not mess things up.

Now I am think it might make sense to add it as `|mxcube|` in stead of (or in addition to) `|project|`. 

I have not really investigated if/how we could do this for Markdown documents.

Also maybe I picked the wrong example, and in this case it is really "MXCuBE" that is meant, and not "MXCuBE-Web", but the point still stands.

docs/source/config/server.rst
elmjag: You mean the mode names? Like:

    SSX-CHIP -> SSX_CHIP
    SSX-INJECTOR -> SSX_INJECTOR

yes, I agree.

elmjag:
Thank you @axelboc for language improvement suggestions, I have squashed them into the original commit.

docs/source/config/server.rst
axelboc: Sorry I meant the key itself, `MODE`, for consistency with the other keys.

... unless I've missed that the distinction matters for some reason? Are ~`PASCAL_CASE`~`CONSTANT_CASE` keys ones that can be overridden via environment variables or something? I don't necessarily think that keys in a YAML file should be written in ~`PASCAL_CASE`~`CONSTANT_CASE` (unlike keys in a `.env` file for instance), but it should definitely be consistent (or the reason why it's not should perhaps be documented).

fabcor-maxiv: _[Aside: @axelboc, this is [`PascalCase`](https://en.wikipedia.org/wiki/Camel_case), isn't it? You probably mean [`SCREAMING_SNAKE_CASE`](https://en.wikipedia.org/wiki/Snake_case), no?]_

axelboc: Haha yeah of course, sorry ^^ -- I actually use the term `CONSTANT_CASE` usually, sorry for the mix-up...

fabcor-maxiv: _[@axelboc, the names are silly, and I always have to look them up anyway. Seeing `CONSTANT_CASE` gets the message across much better for me, indeed.]_

docs/source/config/server.rst
```diff
@@ -0,0 +1,214 @@
+server.yaml
+===========
+
+The ``server.yaml`` file is used to configure various aspects of the MxCube's back-end server.
```
marcus-oscarsson: Casing of MXCuBE

docs/source/config/server.rst
```diff
@@ -0,0 +1,214 @@
...
+
+* server_
+* mxcube_
+
+See below an `example  server.yaml <server_yaml_example_>`_ file.
+
+server
+------
+
+The ``server`` section primarily deals with the configuration of the HTTP aspects of the back-end server.
```
marcus-oscarsson: I would perhaps omit the part about "HTTP aspects" and replace it with maybe web server aspects or similar

docs/source/config/ui.rst
```diff
@@ -0,0 +1,272 @@
...
+| precision | widget's position display precision |
++-----------+-------------------------------------+
+| suffix    | widget's position units suffix      |
++-----------+-------------------------------------+
+
+The ``attribute`` key specifies the underlying motor's adapter object, via it's beamline attribute ID.
+The ``role`` key specifies the physical move the motor performs on the sample.
+For example, motor with the role *omega* rotates sample around the omega axis.
+
+Valid roles are:
```
marcus-oscarsson: Roles can be anything its not actually finite list, perhaps Common roles are:

marcus-oscarsson:
Looks good to me in general, I left a few comments on minor things.

docs/source/config/server.rst
elmjag: changed to MXCuBE

docs/source/config/server.rst
elmjag: Yes, it's is strictly speaking it is not just HTTP.  I have changed to 'network aspects'.

docs/source/config/ui.rst
elmjag: Aha, changed to "Common roles".

docs/source/config/server.rst
elmjag: I agree that we should make the casing consistent in server.yaml.

But think that's a breaking change, so needs to be carefully managed. In other words this is a topic for another PR :)

marcus-oscarsson:
Great !, thanks !

PR #1193, opened 2024-03-22
Add documentation about deployment

fabcor-maxiv:
Preview: https://mxcubeweb--1193.org.readthedocs.build/en/1193/dev/deployment.html

### DIFF
--- a/docs/source/dev/deployment.md
+++ b/docs/source/dev/deployment.md
@@ -1,0 +1,69 @@
+# Deployment
+
+Some aspects to pay attention to
+when considering the deployment of {{ project }} into a production environment.
+
+
+## Python backend part of the web application
+
+The Python backend part of {{ project }}
+uses [*Poetry*](https://python-poetry.org/) as development workflow tool.
+One of Poetry's features is support for a so-called *lockfile*.
+Poetry's lockfile is the `poetry.lock` file
+one can see at the root of the source code tree of {{ project }}.
+
+This lockfile helps create repeatable installations.
+By taking advantage of Poetry's lockfile
+it should be possible to minimize the chances of dependency issues from happening.
+Indeed, the lockfile contains pinned versions of the dependencies
+along with some info about the dependency artefacts to be installed
+and some additional metadata.
+
+By using a command like `poetry install`,
+it is ensured that the dependency versions installed are well known,
+and for example that they are the ones used during development phase
+and during automated tests.
+
+
+### Install without using Poetry
+
+Poetry is a tool meant for development, it is not an installer, like *pip* for example.
+It is debatable whether it is a good idea to use Poetry's `poetry install` command
+to install and deploy in a production environment.
+Some might prefer to use a dedicated installer tool, like pip for such a task.
+
+But Poetry's lockfile format is not standardized, its content is specific to Poetry.
+So this file is not immediately usable by other installers and tools.
+
+To circumvent this, one could use a command like `poetry export`
+to generate a `requirements.txt` file compatible with pip (and some other tools).
+With a series of commands like the following,
+it should be possible to approximate the behaviour of installing with Poetry's lockfile.
+
+1. Let Poetry export its lockfile into a pip-compatible requirements file
+    ```shell
+    poetry export --format=requirements.txt > requirements.txt
+    ```
+
+1. Move the `requirements.txt` file to the deployment target system
+
+1. On the deployment system, let pip install {{project}} via the exported requirements file
+
+    1. Install the dependencies
+        ```shell
+        python -m pip install --no-deps --requirement requirements.txt
+        ```
+
+    1. Install {{project}} itself,
+        and make sure to skip the dependencies (with the `--no-deps` flag)
+        since they are now already installed
+
+        * For example as editable
+          ```shell
+          python -m pip install --no-deps --editable mxcubeweb
+          ```
+
+        * Or not
+          ```shell
+          python -m pip install --no-deps mxcubeweb
+          ```
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -7,6 +7,9 @@ import importlib.metadata
 import mxcubeweb
 
 
+MXCUBE_NAME = "MXCuBE"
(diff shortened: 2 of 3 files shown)


docs/source/dev/deployment.md
```diff
@@ -0,0 +1,59 @@
...
+    poetry export --format=requirements.txt > requirements.txt
+    ```
+
+1. Move the `requirements.txt` file to the deployment target system
+
+1. On the deployment system, let pip install {{project}} via the exported requirements file
+
+    1. Install the dependencies
+        ```shell
+        python -m pip install --no-deps mxcubeweb
```
axelboc: Did you mean to keep this line?

docs/source/dev/deployment.md
```diff
@@ -0,0 +1,59 @@
+# Deployment
+
+{{ project }} uses [*Poetry*](https://python-poetry.org/) as development workflow tool.
+One of Poetry's feature is support for a so-called *lockfile*.
```
axelboc:
```suggestion
One of Poetry's features is support for a so-called *lockfile*.
```

axelboc (approved):

fabcor-maxiv:
@axelboc What about the JavaScript part? Anything that could be mentioned? What should one pay attention to when deploying for production regarding the JavaScript part? Some special flags to set when building? Anything like that?

docs/source/dev/deployment.md
fabcor-maxiv: No : D

Removed, thanks!

docs/source/dev/deployment.md
fabcor-maxiv: Thanks : )

axelboc:
> @axelboc What about the JavaScript part? Anything that could be mentioned? What should one pay attention to when deploying for production regarding the JavaScript part? Some special flags to set when building? Anything like that?

Well, I guess you could copy the steps from the README: 1) install the front-end dependencies with `pnpm --prefix ui install`; 2) build the front-end with `pnpm --prefix ui build` 3) point `maxcube-server` to where the built files are located (`ui/build`).

There's a `.env` file in the `ui` folder but it's meant to be removed. I don't see anything else but @marcus-oscarsson may have a better idea.

docs/source/dev/deployment.md
```diff
@@ -0,0 +1,69 @@
...
+to install and deploy in a production environment.
+Some might prefer to use a dedicated installer tool, like pip for such a task.
+
+But Poetry's lockfile format is not standardized, its content is specific to Poetry.
+So this file is not immediately usable by other installers and tools.
+
+To circumvent this, one could use a command like `poetry export`
+to generate a `requirements.txt` file compatible with pip (and some other tools).
+With a series of commands like the following,
+it should be possible to approximate the behaviour of installing with Poetry's lockfile.
```
marcus-oscarsson, line 41: It is possible to ... ?

marcus-oscarsson:
Yes the .env file is no longer used and can be removed.

marcus-oscarsson:
I'm wondering if we should not provide a pre-built UI and bundle that with the python package.

marcus-oscarsson:
Should we make sure that there is a `requirements.txt` that is up to date in the repository if we recommend using `pip` for deployment or de we leave that up to each and everyone to decide ?

marcus-oscarsson:
Thanks, looks good to me !

axelboc:
> I'm wondering if we should not provide a pre-built UI and bundle that with the python package.

If there are no longer any environment variables and no other build-time configuration, then for sure! Not having to build the front-end would definitely make deployment simpler.

docs/source/dev/deployment.md
fabcor-maxiv: Should be. My goal with this document is to give some pointers, and every facility can build their workflow for themselves according to their own infrastructure. There is no way we can come up with one deployment method that works for all facilities.

But yes, I intend to check this again, make sure it works at least with the default out of the box MXCuBE-Web (with the mockups).

fabcor-maxiv:
> I'm wondering if we should not provide a pre-built UI and bundle that with the python package.

I can not think of a way to do this that would not be a pain, but I have not thought about it very long. If someone comes up with a good solution, then that would be great of course.


> Should we make sure that there is a requirements.txt that is up to date in the repository if we recommend using pip for deployment or de we leave that up to each and everyone to decide ?

Yes, I have already experimented with this a bit. I think with pre-commit hooks it can be done without too much pain and suffering inflicted on contributors. But I would do it in a separate effort (pull request).


> There's a .env file in the ui folder but it's meant to be removed. I don't see anything else but @marcus-oscarsson may have a better idea.

> Yes the .env file is no longer used and can be removed.

No idea about this. Feel free to do it in a separate pull request, I guess.

docs/source/dev/deployment.md
marcus-oscarsson: I see,  using "should be" sounds quite theoretical/hypothetical and we would at-least assume that we have documented something that works. The aim would not be to come up with a complete deployment procedure that works for everyone but at least outline some good practices and make sure that everyone can use them if they would like to.

marcus-oscarsson:
I see, I think that if we agree that we should use pip for deployment then we should IMO make sure that there is either a script that generates the `requirements.txt` and performs the `pip install` or automatically create the `requirements.txt`

marcus-oscarsson:
I think we can come up with a way to build and bundle the UI with the python package in someway.

fabcor-maxiv:
@marcus-oscarsson @axelboc dedicated issues to discuss the tangential topics:
* https://github.com/mxcube/mxcubeweb/issues/1219
* https://github.com/mxcube/mxcubeweb/issues/1220

docs/source/dev/deployment.md
fabcor-maxiv: I tested again just now, and it works as far as I can tell.

fabcor-maxiv:
I have thought about adding a section about the Javascript UI part that would not be a complete copy/paste of the other page, and I can not come up with anything. So I rather leave this out. We can always add something when we have something concrete to write that is actual additional information.

So I guess this is ready to be merged from my point of view.

PR #1194, opened 2024-03-25
Update of Beamline Camera component / Use a Drop Down Menu instead

jbflo:
- Update The current Cameras component  in the  BeamlineSetup
- the issue is the component takes  too much space 
- so we Use a DropDown Menu to Display as much as Cameras element we want
- The Component is Draggable , and do not prevent interaction on other component 

image

image

This PR is a suggestion maybe someone esle had a better suggestion ? Or would you prefer to stick with the previous look?
what do you think @jakub.kowalczyk@maxiv.lu.se ?


meguiraun:
this looks neat, its a nice improvement

meguiraun (approved):

ui/src/components/BeamlineCamera/BeamlineCamera.jsx
```diff
@@ -1,127 +1,113 @@
...
-          </Badge>
-        </div>
-      </OverlayTrigger>
-    </div>
-  );
+        Beamline Cameras
+      </Dropdown.Toggle>
+      <Dropdown.Menu>{renderCameraComponent()[0]}</Dropdown.Menu>
+    </Dropdown>,
+    renderCameraComponent()[1],
```
axelboc: This is a bit wild :smile_cat: – it'd be more straightforward to just duplicate/inline the two loops here and to turn `renderVideo` into an actual component.

ui/src/components/BeamlineCamera/styles.module.css
axelboc, line 1: My preference, but I'd prefer if this file were named `BeamlineCamera.module.css`.

ui/src/components/BeamlineCamera/styles.module.css
```diff
@@ -0,0 +1,8 @@
+.closeBtn {
+  cursor: pointer;
+}
+
+:global(.react-draggable) {
+  position: absolute;
+  z-index: 9999;
```
axelboc: Could you pass this to `Draggable` via a `style` prop, maybe?

That being said, `z-index: 9999` is rarely a robust fix to stacking issues for two reasons:
- It won't work if the component you're trying to stack is rendered within a local [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context). The fix for this is to use a `Portal` to render the component as close to the root of the `body` element as possible to ensure it is stacked with the global stacking context of the document. But maybe `Draggable` already does this?
- When the `z-index` is hardcoded to an arbitrarily high value like this to try to "win" the stacking context, it's often the case that other components throughout the app may be trying to do the same thing (like drop-down menus, other modals, etc.) A good practice to avoid bugs and regressions is to define global `z-index` values with CSS custom properties, all in the same place, to make sure that the global stacking ordered is deterministic, extensible and known to all developers:
    ```css
    :root {
       --zi-draggable: 1000;
       --zi-modal: 2000;
       --zi-menu: 3000;
    }
    ```

ui/src/containers/BeamlineSetupContainer.jsx
```diff
@@ -255,7 +230,7 @@ function BeamlineSetupContainer(props) {
             </Table>
           </Nav.Item>
         </Nav>
-        <Nav className="me-3">{renderCameraComponent()}</Nav>
+        {/* <Nav className="me-3">{renderCameraComponent()}</Nav> */}
```
axelboc: Just a comment to remember to remove this before merging.

axelboc (commented):
Great idea! The drop-down is a neat solution and it's well positioned, IMHO.

ui/src/components/BeamlineCamera/BeamlineCamera.jsx
jbflo: > This is a bit wild 😸 – it'd be more straightforward to just duplicate/inline the two loops here and to turn `renderVideo` into an actual component.

ahah, that's actually what i did first. then  i thought why duplicate? and i wanted to reduce line of code 😸

ui/src/components/BeamlineCamera/styles.module.css
jbflo: hmm That's a tricky one, I think a should leave it that way, and create a new PR to  deal with all z-index. cause there is a bunch of unnecessary  one that will need to be touch.

axelboc: Totally, good call :100:

ui/src/components/BeamlineCamera/BeamlineCamera.jsx
```diff
@@ -1,127 +1,129 @@
...
-  );
+        <Dropdown.Toggle
+          variant="outline-secondary"
+          size="sm"
+          className="mb-1"
+          style={{ width: '150px' }}
+        >
+          Beamline Cameras
+        </Dropdown.Toggle>
+        <Dropdown.Menu>{renderCamera()}</Dropdown.Menu>
```
axelboc: Embrace JSX, don't be scared! :stuck_out_tongue_closed_eyes: `render*` function should be avoided if possible; either inline it or turn it into a proper component (that's what React is all about :wink:).

Here, for the drop-down options, I think inlining is perfectly fine (see suggestion below); but I reiterate that the draggable video modal is worth turning into a component (with the loop that renders them inlined below) — sorry if you're already working on this and I'm jumping the gun.

```suggestion
        <Dropdown.Menu>{
          cameraSetup.components.entries().map(([key, camera]) => (
            <>
              <Dropdown.Item
                key={key}
                onClick={() => handleShowVideoModal(key, true)}
              >
                {camera.label} <i className="fas fa-video" />
              </Dropdown.Item>
              <Dropdown.Divider />
            </>
          );
        }</Dropdown.Menu>
```

ui/src/components/BeamlineCamera/BeamlineCamera.jsx
```diff
@@ -1,127 +1,129 @@
...
-        <video
-          src={url}
-          alt={labelText}
-          onClick={onImageClick}
-          width={width}
-          height={height}
-        />
-      )}
-      <Button
+  return cameraSetup?.components.length > 0 ? (
```
axelboc: You can move this into an early return:

```jsx
if (!cameraSetup?.components.length) {
  return null;
}

return (<>/* ... */</>);
```

axelboc: Actually, it looks like all you care about is `cameraSetup.components`, so this array should probably become the prop and the `<BeamlineCamera>` component should be rendered conditionally in `BeamlineSetupContainer` — something like:

```jsx
const { components = [] } = uiproperties.camera_setup || {};

{components.length > 0 && <BeamlineCamera components={components} />}
```

marcus-oscarsson:
Very nice, and a good initiative ! Thanks !

PR #1195, opened 2024-03-25
UI: fix state mapping for Sample Changer widget

elmjag:
Treat disabled state as closed state in InOutSwitch component.
    
Change logic for handling DISABLED state in the InOutSwitch component. Make the DISABLED state equivalent to CLOSED state.
    
This solves the issue with the 'Sample Changer' InOutSwitch widget. Previously it was not possible to power on a disabled sample changer, as the command for DISABLED state was set to 'PowerOff'. Now when it is in DISABLED state, the command will be 'PowerOn'.
    
Note that for sample changers, the DISABLED state is used when they are in 'powered off' mode.

### DIFF
--- a/ui/src/components/InOutSwitch/InOutSwitch.jsx
+++ b/ui/src/components/InOutSwitch/InOutSwitch.jsx
@@ -89,6 +89,7 @@ export default function InOutSwitch(props) {
       break;
     }
     case offValue:
+    case 'DISABLED':
     case 'CLOSED': {
       msgBgStyle = 'danger';
       btn = (
@@ -102,7 +103,6 @@ export default function InOutSwitch(props) {
       );
       break;
     }
-    case 'DISABLED':
     case 'UNUSABLE': {
       msgBgStyle = 'warning';
       btn = (


ui/src/containers/BeamlineSetupContainer.jsx
```diff
@@ -39,6 +39,25 @@ function BeamlineSetupContainer(props) {
     setAttribute(name, value);
   }
 
+  function handleSendCommand(targetState) {
+    switch (targetState) {
+      case 'READY': {
+        sendCommand('PowerOn');
+        break;
+      }
```
jbflo: I am Wondering , you send PowerOn when State is READY ?
I would have expert READY mean the Device is powered on , .

But good to know if it's the case

ui/src/containers/BeamlineSetupContainer.jsx
```diff
@@ -261,11 +280,11 @@ function BeamlineSetupContainer(props) {
...
               offText="Power Off"
-              openValue="PowerOn"
-              offValue="PowerOff"
+              openValue="READY"
+              offValue="DISABLED"
               labelText="Sample Changer"
               value={sampleChanger.state}
-              onSave={sendCommand}
+              onSave={handleSendCommand}
             />
```
jbflo: I would suggest :

`            <InOutSwitch
              openText="Power On"
              offText="Power Off"
              openValue="PowerOff"
              offValue="PowerOn"
              labelText="Sample Changer"
              value={sampleChanger.state}
              onSave={sendCommand}            
/>
`
As you powerOn when state is READY
just  switch openValue to offValue would solve this,  
Maybe  ??

but I definitely think it is preferable not to have a another handle method handleSendCommand.
InOutSwitch already has a Handle method in.

ui/src/containers/BeamlineSetupContainer.jsx
elmjag: No, I send PowerOn command when I want to _switch_ to READY state. The switch case is on target state.

In practice I send PowerOn when in DISABLED state.

ui/src/containers/BeamlineSetupContainer.jsx
elmjag: I don't understand how that will work.

Sample Changers don't have PowerOff nor PowerOn states. The go between READ and DISABLED.

ui/src/containers/BeamlineSetupContainer.jsx
jbflo: Ahh in that case you could  just change line 113 of the  InOutSwitch component to :
onClick={() => handleSwitch(openValue)}

and also line 115 to : 
Set: {openText}

That would be it .

ui/src/containers/BeamlineSetupContainer.jsx
jbflo: > I would suggest :
> 
> ` <InOutSwitch openText="Power On" offText="Power Off" openValue="PowerOff" offValue="PowerOn" labelText="Sample Changer" value={sampleChanger.state} onSave={sendCommand} />` As you powerOn when state is READY just switch openValue to offValue would solve this, Maybe ??
> 
> but I definitely think it is preferable not to have a another handle method handleSendCommand.
InOutSwitch already has a Handle method in.

Consider this as Not relevant:  see my other response on top,

ui/src/containers/BeamlineSetupContainer.jsx
elmjag: Ah, yeah that would be better.

I'll test it, next week when I'm back from my vacation :)

elmjag: After looking a bit closer at the InOutSwitch code, I think the right way to fix this is to treat 'DISABLED' state the same way as 'CLOSED' state.

elmjag:
After suggestions from @jbflo, I decided to take a different approach to fix the issue. This is more elegant way to deal with it. I have updated the description of this PR to describe the new approach.

jbflo (approved):

PR #1199, opened 2024-04-03
Sample view and OAV fixes

marcus-oscarsson:
This PR is slightly bigger and contain a set of fixes for the sample view, all minor issues but still important :)

 - Image viewer Braggy is opened on click (if there are results) and not on CTRL + SHIFT (which does not work great with selection)
 - Fixes Z + scroll wheel short cut for zoom
 - Only showing motor controls in MotorInput to simplify general rendering of all motor components
 - Beam is in magenta color (as a long time ago) and not dark blue
 - Fixed broken video resize option
 - Fixed broken motor step size setting
 - Fixed creating center point on cell and default number of images for characterization on 2D points

### DIFF
--- a/ui/src/actions/sampleview.js
+++ b/ui/src/actions/sampleview.js
@@ -187,6 +187,8 @@ export function setVideoSize(width, height) {
             beamPosition: json.position,
             sourceScale: json.scale,
           });
+
+          window.initJSMpeg();
         });
     }
   };
--- a/ui/src/actions/sampleview.js
+++ b/ui/src/actions/sampleview.js
@@ -174,7 +174,7 @@ export function setVideoSize(width, height) {
       })
         .then((response) => {
           if (response.status >= 400) {
-            throw new Error('Server refused set video size');
+            throw new Error('Server refused to set video size');
           }
           return response.json();
         })
--- a/ui/src/actions/sampleview.js
+++ b/ui/src/actions/sampleview.js
@@ -174,7 +174,7 @@ export function setVideoSize(width, height) {
       })
         .then((response) => {
           if (response.status >= 400) {
-            throw new Error('Server refused to add line');
+            throw new Error('Server refused set video size');
           }
           return response.json();
         })
@@ -187,7 +187,6 @@ export function setVideoSize(width, height) {
             beamPosition: json.position,
             sourceScale: json.scale,
           });
-          window.initJSMpeg();
         });
     }
   };
--- a/ui/src/components/SampleView/DrawGridPlugin.js
+++ b/ui/src/components/SampleView/DrawGridPlugin.js
@@ -593,19 +593,31 @@ export default class DrawGridPlugin {
     }
   }
 
-  getClickedCell(shapeGroup, clickPoint) {
-    let cell = null;
+  getClickedCellIndex(gd, shapeGroup, e) {
+    const cellSizeX = this.getCellWidth(gd);
+    const cellSizeY = this.getCellHeight(gd);
 
-    shapeGroup.forEachObject((obj) => {
-      if (
-        obj.get('type') === 'ellipse' &&
-        obj.containsPoint(clickPoint, null, true)
-      ) {
-        cell = obj;
-      }
-    });
+    const cellIdxX = Number.parseInt(
+      Math.floor((e.offsetX - shapeGroup.oCoords.tl.x) / cellSizeX),
+      10,
+    );
+    const cellIdxY = Number.parseInt(
+      Math.floor((e.offsetY - shapeGroup.oCoords.tl.y - 20) / cellSizeY),
+      10,
+    );
 
-    return cell;
+    return [cellIdxX, cellIdxY];
+  }
+
+  getClickedCell(gd, shapeGroup, e) {
+    const cellSizeX = this.getCellWidth(gd) / this.scale;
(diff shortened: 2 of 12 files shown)


ui/src/actions/sampleview.js
```diff
@@ -174,7 +174,7 @@ export function setVideoSize(width, height) {
       })
         .then((response) => {
           if (response.status >= 400) {
-            throw new Error('Server refused to add line');
+            throw new Error('Server refused set video size');
```
axelboc:
```suggestion
            throw new Error('Server refused to set video size');
```

ui/src/components/SampleView/DrawGridPlugin.js
```diff
@@ -593,19 +593,31 @@ export default class DrawGridPlugin {
...
 
-    shapeGroup.forEachObject((obj) => {
-      if (
-        obj.get('type') === 'ellipse' &&
-        obj.containsPoint(clickPoint, null, true)
-      ) {
-        cell = obj;
-      }
-    });
+    const cellIdxX = Number.parseInt(
```
axelboc: Not quite sure why you need `parseInt`.

ui/src/components/SampleView/DrawGridPlugin.js
```diff
@@ -593,19 +593,31 @@ export default class DrawGridPlugin {
...
+      Math.floor((e.offsetX - shapeGroup.oCoords.tl.x) / cellSizeX),
+      10,
+    );
+    const cellIdxY = Number.parseInt(
+      Math.floor((e.offsetY - shapeGroup.oCoords.tl.y - 20) / cellSizeY),
```
axelboc: Might be worth explaining why `- 20` in a comment, or making it a constant.

ui/src/components/SampleView/MotorControls.js
```diff
@@ -16,6 +16,10 @@ export default class MotorControls extends React.Component {
   renderMotorInputs(from, to) {
     return this.props.uiproperties.components
       .slice(from, to)
+      .filter(
+        ({ attribute }) =>
+          this.props.hardwareObjects[attribute].type === 'MOTOR',
+      )
```
axelboc: Just wondering if this filtering should be done earlier on (maybe in the parent component), so that `this.renderMotorInputs(0, 3)` is always guaranteed to render the first 3 motors?

ui/src/components/SampleView/SampleControls.js
```diff
@@ -123,38 +123,23 @@ export default class SampleControls extends React.Component {
             this.props.sampleViewActions.setVideoSize(size[0], size[1])
           }
         >
-          <span className={`fa ${sizeGClass}`} /> {`${size[0]} x ${size[1]}`}
+          <span className={`${sizeGClass}`} /> {`${size[0]} x ${size[1]}`}
```
axelboc:
```suggestion
          <span className={sizeGClass} /> {`${size[0]} x ${size[1]}`}
```

ui/src/components/SampleView/SampleImage.jsx
```diff
@@ -439,25 +415,35 @@ export default class SampleImage extends React.Component {
...
+        } else if (threeDpointList.length > 2) {
+          ctxMenuObj = { type: 'GROUP', id: threeDpointList };
         } else if (gridList.length === 1) {
-          const gridData = this.props.grids[gridList[0]];
-          const cellCenter = this.getGridCellCenter(
+          let gridData = this.props.grids[gridList[0]];
+          gridData = this.drawGridPlugin.setPixelsPerMM(
+            this.props.pixelsPerMm,
+            gridData,
+          );
```
axelboc:
```suggestion
          const gridData = this.drawGridPlugin.setPixelsPerMM(
            this.props.pixelsPerMm,
            this.props.grids[gridList[0]],
          );
```

ui/src/actions/sampleview.js
marcus-oscarsson: Thanks !

ui/src/components/SampleView/DrawGridPlugin.js
marcus-oscarsson: not sure, looks indeed strange. Ill have a look

ui/src/components/SampleView/DrawGridPlugin.js
marcus-oscarsson: :), yes. I honestly don't know why its 20 but I agree that a comment is useful

ui/src/components/SampleView/MotorControls.js
marcus-oscarsson: ah, yes you are right it could be an issue. Well spotted Ill try to fix that !

ui/src/components/SampleView/SampleImage.jsx
marcus-oscarsson: Thanks

ui/src/components/SampleView/MotorControls.js
marcus-oscarsson: I tired to do something nicer, by passing only the motor ui attributes and the motor hardware objects to the `MotorControls` component.

axelboc: Looks great :100:

axelboc (approved):

PR #1204, opened 2024-04-04
Queue - Making sure queue is update on observer

marcus-oscarsson:
We need to update the queue contents on change as we are no longer using the server side `redis-persist` solution. 

- Emitting queue changes to observer when it changes.
- Disabled run-now from sample list (only makes sense from sample view)
- Clearing queue when gripper changes, queued samples cant be loaded with a different gripper

### DIFF
--- a/mxcubeweb/core/components/samplechanger.py
+++ b/mxcubeweb/core/components/samplechanger.py
@@ -37,6 +37,10 @@ class SampleChanger(ComponentBase):
                 "globalStateChanged", signals.sc_maintenance_update
             )
 
+            HWR.beamline.sample_changer_maintenance.connect(
+                "gripperChanged", self._gripper_changed
+            )
+
     def get_sample_list(self):
         samples_list = HWR.beamline.sample_changer.get_sample_list()
         samples = {}
@@ -384,6 +388,16 @@ class SampleChanger(ComponentBase):
         except Exception:
             logging.getLogger("MX3.HWR").exception("Could not get crystal List")
             return {"xtal_list": xtal_list}
+
+    def _gripper_changed(self):
+        from mxcubeweb.routes import signals
+
+        self.app.queue.queue_clear()
+        self.app.server.emit(
+            "queue", {"Signal": "update", "message": "all"}, namespace="/hwr"
+        )
+
+        # self.app.sample_changer.get_sample_list()
 
 
 # Disabling C901 function is too complex (19)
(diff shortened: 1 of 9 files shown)


mxcubeweb/core/components/samplechanger.py
```diff
@@ -385,6 +389,16 @@ def sync_with_crims(self):
...
+    def _gripper_changed(self):
+        from mxcubeweb.routes import signals
+
+        self.app.queue.queue_clear()
+        self.app.server.emit(
+            "queue", {"Signal": "update", "message": "all"}, namespace="/hwr"
+        )
+
+        # self.app.sample_changer.get_sample_list()
+
```
fabcor-maxiv: Is the import necessary? I do not see it being used. Or is it an import with side effects?

```suggestion
    def _gripper_changed(self):
        self.app.queue.queue_clear()
        self.app.server.emit(
            "queue", {"Signal": "update", "message": "all"}, namespace="/hwr"
        )
```

marcus-oscarsson: :+1: Noh its not needed :). It was something from a previous solution I tried. Thanks !

fabcor-maxiv (approved):
Seems fine

ui/src/reducers/sampleGrid.js
```diff
@@ -42,10 +42,9 @@ function sampleGridReducer(state = INITIAL_STATE, action = {}) {
   // eslint-disable-next-line sonarjs/max-switch-cases
   switch (action.type) {
     case 'SET_QUEUE': {
-      const sampleList = { ...state.sampleList };
       return {
         ...state,
-        sampleList: Object.assign(sampleList, action.sampleList),
+        sampleList: { ...action.sampleList },
```
axelboc, line 47: Just checking: you didn't meant to write `{ ...state.sampleList, ...action.sampleList }`?

ui/src/containers/SampleListViewContainer.jsx
```diff
@@ -308,6 +314,7 @@ class SampleListViewContainer extends React.Component {
     } else {
       this.props.syncSamples();
     }
+    this.sampleGridFilterByKey('limsSamples', true);
```
axelboc: Not sure the extra method is worth it. It would be clearer to write `this.props.filter({ limsSamples: true });`

axelboc (approved):

ui/src/reducers/sampleGrid.js
marcus-oscarsson: :), thanks always better to be sure. So I do mean "assign" and not "merge" (like previously)

ui/src/containers/SampleListViewContainer.jsx
marcus-oscarsson: Ah, yes you are right a bit silly

PR #1206, opened 2024-04-04
Camera setup - Not rendering if there is no camera setup section

marcus-oscarsson:
Not rendering if there is no camera setup section

### DIFF
--- a/ui/src/components/BeamlineCamera/BeamlineCamera.jsx
+++ b/ui/src/components/BeamlineCamera/BeamlineCamera.jsx
@@ -91,7 +91,7 @@ export default function BeamlineCamera(props) {
     return DraggableElements;
   }
 
-  if (!cameraSetup || cameraSetup.components.length < 0) {
+  if (!cameraSetup || cameraSetup.components.length <= 0) {
     return null;
   }
 
--- a/ui/src/components/BeamlineCamera/BeamlineCamera.jsx
+++ b/ui/src/components/BeamlineCamera/BeamlineCamera.jsx
@@ -91,7 +91,7 @@ export default function BeamlineCamera(props) {
     return DraggableElements;
   }
 
-  if (cameraSetup?.components.length < 0) {
+  if (cameraSetup === undefined || cameraSetup?.components.length < 0) {
     return null;
   }
 
--- a/ui/src/components/BeamlineCamera/BeamlineCamera.jsx
+++ b/ui/src/components/BeamlineCamera/BeamlineCamera.jsx
@@ -91,7 +91,7 @@ export default function BeamlineCamera(props) {
     return DraggableElements;
   }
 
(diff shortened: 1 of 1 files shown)


ui/src/components/BeamlineCamera/BeamlineCamera.jsx
```diff
@@ -91,7 +91,7 @@ export default function BeamlineCamera(props) {
     return DraggableElements;
   }
 
-  if (cameraSetup?.components.length < 0) {
+  if (cameraSetup === undefined || cameraSetup?.components.length < 0) {
```
axelboc:
```suggestion
  if (!cameraSetup || cameraSetup.components.length < 0) {
```

axelboc (approved):

ui/src/components/BeamlineCamera/BeamlineCamera.jsx
marcus-oscarsson: Ah, I knew that you would have a good suggestion for this. I'm always a bit paranoid when it comes to JS and "falsy/truthy" values :)

marcus-oscarsson:
thanks @axelboc :+1:

ui/src/components/BeamlineCamera/BeamlineCamera.jsx
axelboc: Yeah, often better to be paranoid when TS is not in the picture.

In case anyone is interested in the reasoning: the initial condition tells us that `cameraSetup` has type `object | null | undefined` (because of the [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) syntax). So the first part of the new condition needs to check for `null | undefined` -- since `cameraSetup` cannot ever be a number, string, etc., then `!cameraSetup` is sufficient and covers both `null` and `undefined`. Once we've done that, `cameraSetup` is necessarily `object` in the second part of the condition, so the `?` is no longer needed.

marcus-oscarsson: Ah, yes that's interesting ! thanks :bow:

marcus-oscarsson: By the way, the optional chaining is a bit crazy :)

axelboc: `foo?.bar?.()`! 🤣

jbflo: es-lint is complaining 
maybe better with <=
` if (!cameraSetup || cameraSetup.components.length <= 0) {`

axelboc: Makes sense of course, even just `cameraSetup.components.length === 0`

PR #1208, opened 2024-04-05
connect SocketIO first when logged in

elmjag:
Start connecting to SocketIO after user have successfully logged in.

You need to be authenticated before back-end allows you to open a SocketIO connection. Trying to establish a connection before will fail. This leads to an endless retry loop while the user is on the login page.

### DIFF
--- a/ui/src/serverIO.js
+++ b/ui/src/serverIO.js
@@ -131,16 +131,15 @@ class ServerIO {
 
   // eslint-disable-next-line sonarjs/cognitive-complexity
   listen() {
+    this.refreshInterval = setInterval(sendRefreshSession, 9000);
+    this.connect();
+
     if (this.initialized) {
       return;
     }
 
     this.initialized = true;
     this.dispatch = store.dispatch;
-
-    this.refreshInterval = setInterval(sendRefreshSession, 9000);
-
-    this.connect();
 
     this.loggingSocket.on('log_record', (record) => {
       if (record.severity !== 'DEBUG') {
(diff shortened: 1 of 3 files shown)


axelboc (approved):
Great catch @elmjag 🎣

marcus-oscarsson:
:) Actually if you don't mind to wait I have a PR for this as well, Ill see how the two goes together

marcus-oscarsson:
So, I also spotted this a while back and I made this fix https://github.com/mxcube/mxcubeweb/pull/1209. 
We also noticed that session.permanent seems to be set a bit to early so it was not taking effect.

I'm happy with either solution, but I have one question for the connect above I think as it is now. Are we sure that `setInterval` and `connect` are not called multiple times, for instance when changing from one view to another ?

ui/src/serverIO.js
```diff
@@ -131,17 +131,16 @@ class ServerIO {
 
   // eslint-disable-next-line sonarjs/cognitive-complexity
   listen() {
+    this.refreshInterval = setInterval(sendRefreshSession, 9000);
```
marcus-oscarsson, line 134: Not sure, perhaps you checked and tested but are we certain this is not called several times when navigating ?

axelboc: As long as any call to `listen()` has a corresponding clean-up call to `disconnect`, then it's fine I think. It seems to be the case here. That being said, I think `serverIO.js` definitely needs a broader review.

marcus-oscarsson: :), yes !

marcus-oscarsson:
Ok, Ill merge this and then rebase https://github.com/mxcube/mxcubeweb/pull/1209

PR #1209, opened 2024-04-05
Login - connecting socketIO after user is logged in

marcus-oscarsson:
- Moved serverIO.listen so that its called after user is logged in
 - Moved session.permanent called to early


fabcor-maxiv (approved):
Seems fine

marcus-oscarsson:
Thanks, I removed the socketio bit then since @elmjag did a nicer fix in #1208.

fabcor-maxiv (approved):

PR #1217, opened 2024-04-11
[SampleView] - Fixed grid color when there are results

marcus-oscarsson:
Fixed grid color when there are results


fabcor-maxiv (approved):
Seems fine

PR #1222, opened 2024-04-15
MachineInfo adapter more versatile

beteva:
- Use get_value from the hardware object
- Remove particular methods.


marcus-oscarsson:
Ok to be merged after https://github.com/mxcube/mxcubecore/pull/917

PR #1223, opened 2024-04-16
Draft: Solving issues with shapes

dominikatrojanowska:
Solving issues:

* Unsaved Points
    Right click on a unsaved point gives the following error:
```
Uncaught runtime errors:
ERROR 
Cannot set properties of undefined (setting 'selected') TypeError: Cannot set properties of undefined (setting 'selected') at http://localhost:3000/static/js/bundle.js:28045:28 at Array.forEach () at SampleImage.clearSelection 
(...)
```

* 2D points
    If there is no sample added (manually mounted), and you right click on image and select "Data collecion (Limited OSC)" we get the error:
```
Uncaught runtime errors: 
ERROR
Cannot read properties of undefined (reading 'defaultPrefix') TypeError: Cannot read properties of undefined (reading 'defaultPrefix') at ContextMenu.showModal 
(...)
```

* Unclear how to trigger… But having a sample mounted, and you right click on image and select "Data collecion (Limited OSC)", once the 2D point is on screen, rigth click on it  we get the following errors (i am not sure why/how these two errors are different on the same scenario, probably the refresh does something)
```
Uncaught runtime errors:
ERROR
Cannot read properties of undefined (reading 'state') TypeError: Cannot read properties of undefined (reading 'state') at SampleImage.rightClick (http://localhost:3000/static/js/bundle.js:27796:78)
```
and 
```
Uncaught runtime errors:
ERROR
Cannot set properties of undefined (setting 'selected') TypeError: Cannot set properties of undefined (setting 'selected') at http://localhost:3000/static/js/bundle.js:28045:28 at Array.forEach (<anonymous>) at SampleImage.clearSelection (http://localhost:3000/static/js/bundle.js:28042:38) at SampleImage.selectShape (http://localhost:3000/static/js/bundle.js:27996:12) at http://localhost:3000/static/js/bundle.js:27832:16 at klass.forEachObject (http://localhost:3000/static/js/bundle.js:110530:16) at SampleImage.rightClick (http://localhost:3000/static/js/bundle.js:27829:19)
```


ui/package-lock.json
fabcor-maxiv, line 1: This file (`ui/package-lock.json`) might have been added my mistake. No?

fabcor-maxiv (changes requested):

marcus-oscarsson:
Thanks @dominikatrojanowska, I can't reproduce point 1 and 3 above. 

For point 2, there is a error because there is no sample added. The collection dialog will need some sample information so this solution will only allow for the dialog to be opened but the problem will remain (there is no sample added). In this case we should rather show the "no sample mounted" dialog, like is done for instance with 3-click centering.

ui/package-lock.json
dominikatrojanowska: yes, thanks :)

dominikatrojanowska:
t some point I was able to reproduce this case:

but unfortunately, I didn't note the steps, just took the screenshot

marcus-oscarsson:
Hey @dominikatrojanowska  whats the status of this PR. Are you trying to solve the issue you posted above or is that something else you noticed while working on this ?

dominikatrojanowska:
Summarizing the current status of the task:

1. Unsaved Points - solved
2. 2D points - Data Collection etc. - I added the banner with appropriate information
image

4. "Unclear how to trigger… But having a sample mounted, and you right-click..." - I think is also solved 

what's next? should I rebase and merge it?

ui/src/components/SampleView/ContextMenu.js
```diff
@@ -339,26 +339,31 @@ export default class ContextMenu extends React.Component {
...
         },
-        type,
-      },
-      sid,
-    );
+        sid,
+      );
+    } else {
+      console.log("sampleData undefined")
+      this.props.showErrorPanel(true, 'There is no sample mounted, cannot collect data.')
```
fabcor-maxiv:
```suggestion
      this.props.showErrorPanel(true, 'There is no sample mounted, cannot collect data.')
```

fabcor-maxiv (commented):
Some linting issue: https://github.com/mxcube/mxcubeweb/actions/runs/8779685613/job/24632712967?pr=1223

fabcor-maxiv (approved):

marcus-oscarsson:
Thanks :)

PR #1224, opened 2024-04-17
always stop queue with queue_manager.stop()

elmjag:
This solves problem with stopping tasks started via 'run now' action.

Note that in current code, else-branch is taken only if you run all your task as 'run now' actions. As soon as you put something on the queue, the `HWR.beamline.queue_manager._root_task` becomes non-None. Thus _all_ future queue stops will take the if-branch. Regardless if they are created via 'run now' or 'add to queue' actions. And it seems to work fine.

Do we need else-branch for something? Does this patch break something I'm not considering?

### DIFF
--- a/mxcubeweb/core/components/queue.py
+++ b/mxcubeweb/core/components/queue.py
@@ -2174,29 +2174,7 @@ class Queue(ComponentBase):
             logging.getLogger("MX3.HWR").info("[QUEUE] Queue started")
 
     def queue_stop(self):
-        from mxcubeweb.routes import signals
-
-        if HWR.beamline.queue_manager._root_task is not None:
-            HWR.beamline.queue_manager.stop()
-        else:
-            qe = HWR.beamline.queue_manager.get_current_entry()
-            # check if a node/task is executing and stop that one
-            if qe:
-                try:
-                    qe.stop()
-                except Exception:
-                    logging.getLogger("MX3.HWR").exception(
-                        "[QUEUE] Could not stop queue"
-                    )
-                HWR.beamline.queue_manager.set_pause(False)
-                # the next two is to avoid repeating the task
-                # TODO: if you now run the queue it will be enabled and run
-                qe.get_data_model().set_executed(True)
-                qe.get_data_model().set_enabled(False)
-                qe._execution_failed = True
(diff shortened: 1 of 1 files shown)


mxcubeweb/core/components/queue.py
```diff
@@ -2174,29 +2174,7 @@ def queue_start(self, sid):
...
-                HWR.beamline.queue_manager.set_pause(False)
-                # the next two is to avoid repeating the task
-                # TODO: if you now run the queue it will be enabled and run
-                qe.get_data_model().set_executed(True)
-                qe.get_data_model().set_enabled(False)
-                qe._execution_failed = True
-
-                HWR.beamline.queue_manager._is_stopped = True
-                signals.queue_execution_stopped()
+        HWR.beamline.queue_manager.stop()
```
marcus-oscarsson, line 2177: Hm, this will probably not work, did you try to stop a task that you started with "run now" ?

marcus-oscarsson: I'm not entirely sure, the code looks quite complicated so I would be surprised if it was added without a good reason. Perhaps that reason does not exist today, but it would be good to know how well we test this change. 

I prefer what you suggest but Id like to be sure that we don't get an unexpected surprise.

elmjag: I have tested stopping tasks started with 'run now', and the 'add to queue' version as well. Both work as expected. I have tested both using Mockup beamline and at MicroMAX beamline.

One little nuance that speaks in favor of this patch, is that, right now:

    HWR.beamline.queue_manager.stop()

code is used to stop both  'run now' and 'add to queue' tasks as soon as variable `HWR.beamline.queue_manager._root_task` is assigned a value.

That is, as soon as you run the queue. After that, MXCuBE permanent switches to stopping all tasks with:

    HWR.beamline.queue_manager.stop()

As far as I can tell, there is no code that resets `_root_task` back to None.

I have deployed this to production at MicroMAX. Let's give this change a couple of weeks, and see if this patch causes any problems.

marcus-oscarsson:
Ok, sounds good. Ill merge this for now and test it here as well.

PR #1225, opened 2024-04-18
Fix usage of deprecated `ruamel.yaml` API

fabcor-maxiv:


fabcor-maxiv:
We have `ruamel.yaml` locked to a version older than `0.18` in `poetry.lock`. But in our case, conda installed a different version of that same library that is newer than `0.18` (for the detail, this was installed as a dependency of sardana which we install with conda). Anyway `ruamel.yaml>=0.18` version has removed some of the API that we use.

This pull request changes to using the non-deprecated equivalent API, which is also compatible with the version we have pinned in `poetry.lock`. So it should all be fine, but I guess it is better if others test it in their environment, in case it breaks anything.

Details regarding the deprecation can be read here:

* <https://yaml.readthedocs.io/>
* <https://yaml.readthedocs.io/en/latest/api/>

fabcor-maxiv:
The PR also contains the addition of the encoding for the opening of the YAML file. That was missing before.

rhfogh:
Looks good. Since you seem to be on top of this, are you going to make the equivalent PR for mxcubecore?

fabcor-maxiv:
> Looks good. Since you seem to be on top of this, are you going to make the equivalent PR for mxcubecore?

Ah... Good point. Somehow it did not trigger any error from `mxcubecore`, when we saw it happen, so I did not look at `mxcubecore` at all, but now I will.

PR #1226, opened 2024-04-26
[CSS FIX] remove extra padding / margin

jbflo:
- [VsCode] Turns out, VSCode true is set to "explicit" since VSCode 1.85.0.
  https://code.visualstudio.com/updates/v1_85#_code-actions-on-save-and-auto

-  [SMALL CSS FIX] remove extra padding / margin
   I neat solution for later would be to convert /ui/src/components/SampleQueue/app.css
   to css module.  

css Bug when screen is at 100 %
image

Fix
image


marcus-oscarsson:
Thanks @jbflo

PR #1228, opened 2024-04-30
Migrate to `pnpm`version 9

### LINKED ISSUE #1227: Do we have an issue with the `pnpm` version?
Seems like [`pnpm` version 9](https://github.com/pnpm/pnpm/releases/tag/v9.0.0) got released not too long ago (about 2 weeks ago, 2024-04-16).

In our `package.json` we have `"pnpm": "8.x"`:

https://github.com/mxcube/mxcubeweb/blob/628b1a94a46101955dcaeb0e98c7898338742995/ui/package.json#L7

CI/CD pipelines started breaking on our end at MAXIV, because pnpm 9+ ends up being installed. Apparently we do not use `conda-environment.yml` in this pipeline, not sure why.

But what confuses me the most is that here on GitHub: things do not break. It is a pnpm v8 that gets installed, even though we have no restriction in `conda-environment.yml`:

https://github.com/mxcube/mxcubeweb/blob/628b1a94a46101955dcaeb0e98c7898338742995/conda-environment.yml#L11

Anyway, it seems to me like we need to add/modify/remove the version constraints either in `package.json` or in `conda-environment.yml`, if we do not want things to start breaking soon-ish, and if we want to be consistent.

fabcor-maxiv:
`pnpm` version 9 got released recently.
This change ensures version 9 is used consistently.


marcus-oscarsson:
Thanks, well done !

axelboc (commented):
Thanks!

PR #1230, opened 2024-04-30
Major dependency update - Flask 3x

marcus-oscarsson:
Updating flask to 3.0. Still testing this, seems to work but sill needs some more testing.


marcus-oscarsson:
We have been testing this for about two weeks and we have not seen any issue with the updates

marcus-oscarsson:
Seems like tests are failing for some reason, ill WIP it for now until I have double checked why that is.

marcus-oscarsson:
Ok, all fine now :)

fabcor-maxiv:
> Seems like tests are failing for some reason, ill WIP it for now until I have double checked why that is.
> _[...]_
> Ok, all fine now :)

Did you figure out why the tests were failing? I had a very quick look at the job output the other day, and it seemed like it was related to missing imports (as if the API had changed).

pyproject.toml
```diff
@@ -41,7 +43,8 @@ pytz = "^2022.6"
 tzlocal = "^4.2"
 mxcubecore = ">=1.102.0"
 mxcube-video-streamer = ">=1.0.0"
-bcrypt = "^4.0.1"
+bcrypt = "4.0.1"
+authlib = "^1.3.0"
```
fabcor-maxiv: As far as I can tell, we do not directly depend on those (`bcrypt` and `authlib`). Can we have a comment explaining why we add those as direct dependencies?

pyproject.toml
```diff
@@ -23,11 +23,13 @@ classifiers = [
...
 [tool.poetry.dependencies]
 python = ">=3.8,<3.11"
-Flask = "^2.2.2"
-werkzeug = "2.3.7"
-Flask-SocketIO = "^5.3.2"
-flask-sqlalchemy = "^3.0.2"
-Flask-Security-Too = "^5.0.2"
+Flask = "^3.0.3"
+werkzeug = "^3.0.3"
+requests = "^2.32.2"
```
fabcor-maxiv: We do not import `requests` explicitly. It would be good to have a comment explaining why we list it as direct dependency.

marcus-oscarsson:
> > Seems like tests are failing for some reason, ill WIP it for now until I have double checked why that is.
> > _[...]_
> > Ok, all fine now :)
> 
> Did you figure out why the tests were failing? I had a very quick look at the job output the other day, and it seemed like it was related to missing imports (as if the API had changed).

Yes it was because `flask-login` was 0.6.2 and did not support the latest `werkzeug` so it needs to be at-least 0.6.3

pyproject.toml
marcus-oscarsson: Ill add a comment on that, its a security update seems like the libraries using `requests` have not updated to the latest version (https://github.com/mxcube/mxcubeweb/pull/1238)

pyproject.toml
marcus-oscarsson: Its for the SSO-keycloak integration. I tested everything with that development and I did not remove the dependencies specific to SSO. Ill remove these for the time being.

pyproject.toml
marcus-oscarsson: Actually it turns out that is fine now, both `Poetry` and `Sphinx` that uses `requests` should install a more recent version of `requests`

pyproject.toml
fabcor-maxiv: Seems fixed

pyproject.toml
fabcor-maxiv: Good : )

Issue #1231, opened 2024-05-02
mixup of strings and numbers in GenericTaskForm


elmjag:
I have added a new queue entry. When launching a task for that queue entry, you first enter parameters in it's task form. When trying to edit for example it's exposure parameter, I get the errors bellow:

![Screenshot from 2024-05-02 16-18-17](https://github.com/mxcube/mxcubeweb/assets/77584583/944da2e9-64b7-4545-9aef-763114a011f0)

It seems that UI prefils default values for some standard parameters, such as Energy and Resolution. The appear to come from the the redux state at taskForm.defaultParameters.<entry_type>.schema.user_collection_parameters.properties.

![Screenshot from 2024-05-02 16-20-26](https://github.com/mxcube/mxcubeweb/assets/77584583/bace7c83-5e74-47af-84a7-1860cb2600d2)

The default values for energy and resolution _are_ of string type. Which is wrong type according to the queue entry model. Model specify them as floats.

As far as I understand, some of the default parameters are converted to string here:

https://github.com/mxcube/mxcubeweb/blob/develop/ui/src/components/Tasks/GenericTaskForm.js#L381

or more specifically, the toFixed() function converts floats to string here:

https://github.com/mxcube/mxcubeweb/blob/develop/ui/src/components/Tasks/fields.js#L88

We need to make sure that 'must be number' error are not displayed, as long as the value in the text box is a valid number.

marcus-oscarsson:
I don't have the time to look at this right away and Ill be off next week but in the mean time, just a few pointers; the `toFixed` function https://github.com/mxcube/mxcubeweb/blob/develop/ui/src/components/Tasks/fields.js#L88 should return a `Number`

The issue is perhaps that the `default parameters` are string for `resolution` and `energy`, `exp_time` and `num_images` looks good.  Its a bit odd that they have different types, the properties of the `user_collection_parameters` should come from the *Pydantic model*, how does it look like ?

elmjag:
`toFixed()` actually returns a string, from my web browser console:

```
  > const foo = Number.parseFloat("1.234").toFixed(2)
  < undefined
  > typeof(foo)
  < 'string'
```

I have tested converting the return value of  https://github.com/mxcube/mxcubeweb/blob/develop/ui/src/components/Tasks/fields.js#L88 to a number. It seems to work fine.

If it supposed to be a number, I guess that is the solution to the this issue. I'll make a PR, perhaps next week :)

elmjag:
[...]

PR #1233, opened 2024-05-11
UI: make sure looked up HO values are numbers

elmjag:
The values returned by toFixed() are used to pre-fill task dialog's common fields. Fields such as resolution, energy, etc. The dialog's schema for these parameters specify that these fields must be of a number types. When we pre-fill these fields as text string, the dialog will show errors such as:

  'Energy' must be a number

Note, this error will be displayed on first validation run of the dialog. That is first when user starts editing the fields.

Converting return values to number solves this problem.

Solves issue: https://github.com/mxcube/mxcubeweb/issues/1231

### DIFF
--- a/ui/src/components/Tasks/fields.js
+++ b/ui/src/components/Tasks/fields.js
@@ -85,9 +85,11 @@ export function toFixed(state, hoName, parameterName = null) {
     }
   }
 
-  return value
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
:+1: Thanks @elmjag

PR #1235, opened 2024-05-16
[TaskDialog] - Fixed RJSF rendering (template API change)

marcus-oscarsson:
Fix for API change in RJSF


fabcor-maxiv:
We tested this a bit, and this seems fine. : )

fabcor-maxiv (approved):

PR #1245, opened 2024-06-07
Added tests for login behavior

walesch-yan:
Added 2 e2e tests for login with the reserved passwords "wrong" and "ispybDown" respectively. In both cases, the login attempt should show the message "Could not authenticate".


marcus-oscarsson:
Thanks @walesch-yan ! very nice :)

Yan is our new colleague that will be working together with us here at ESRF !

PR #1247, opened 2024-06-10
Added addtional test cases for queue

walesch-yan:
I added an additional e2e test to check if samples are getting correctly mounted.

Additionally, added support functions to allow tests to get control and click buttons.


ui/cypress/support.js
```diff
@@ -8,3 +8,38 @@ Cypress.Commands.add('login', (username = 'idtest0', password = '0000') => {
   cy.findByLabelText('Password').type(password);
   cy.findByRole('button', { name: 'Sign in' }).click();
 });
+
+Cypress.Commands.add('takeControl', () => {
+  /* firefox (only firefox) throws an unhandled promise error when executing
```
marcus-oscarsson, line 13: :+1: Great with comments like this :)

PR #1252, opened 2024-06-18
Mockup: show more fields in machine info box

elmjag:
When running MXCuBE with mockups, enable more fields in the machine info drop-down box. Let's include 'message' and 'lifetime' fields in addition to 'current', as the MachineInfoMockup implements these fields.

The result looks like this:

image

### DIFF
--- a/test/HardwareObjectsMockup.xml/machine_info.xml
+++ b/test/HardwareObjectsMockup.xml/machine_info.xml
@@ -1,2 +1,3 @@
 <object class="MachineInfoMockup">
+    <parameters>["current", "message", "lifetime"]</parameters>
 </object>


beteva:
@elmjag I did not add all, as wanted to show that you have the choice of not displaying everything. 
If you need to add all the rest, please do so.

The idea of the  mxcubecore PR [#917](https://github.com/mxcube/mxcubecore/pull/917/files#) - Refactor AbstractMachineInfo and the corresponding mxcubeweb PR #1222    was to be able to add as many fields to display as your HO provides. In the current xml file you have already.
```
<object class="MachineInfoMockup">
    <parameters>["current", "lifetime", "dummy"]</parameters>
</object>
```
The _dummy_ in the parameters list serves "test_check_attributes" in the tests, just to create an error, so, please, do not remove it.

beteva:
@elmjag, my apologies. I was looking at the wrong xml file - the one on the mxcubecore repository. Anyhow you can add whatever you need.
We need, though, to find a way of having the configuration files in just one place, so there is no repetition or worse - inconsistency.

elmjag:
I don't have any strong feelings about this change.

My rationale for adding `<parameters/>` tag is that it gives us a nice example of how you can configure MachineInfo hardware objects. I assume that the mock-ups are commonly studied when trying to understand new parts of MXCuBE.

PR #1254, opened 2024-06-18
[TaskDialog] - Fix for JSForm getting cleard when store update

marcus-oscarsson:
This fixes a few issues with the generic task dialog.

 - The "current" values of the JSForm is saved in local storage so that they are not cleared by updates to the redux store (as the redux form gets re-rendered when the store is updated)
 
 - Reset to default fixed

 - Added missing title

### DIFF
--- a/ui/src/components/Tasks/GenericTaskForm.js
+++ b/ui/src/components/Tasks/GenericTaskForm.js
@@ -43,6 +43,18 @@ class GenericTaskForm extends React.Component {
     this.props.handleSubmit(this.addToQueue.bind(this, true))();
   }
 
+  get jsFormStorageKey() {
+    return `current${this.props.taskData.type}Parameters`;
+  }
+
+  clearCurrentJSFormParameters() {
+    localStorage.removeItem(this.jsFormStorageKey);
+  }
+
+  saveCurrentJSFormParameters(formData) {
+    localStorage.setItem(this.jsFormStorageKey, JSON.stringify(formData));
+  }
+
   addToQueue(runNow, params) {
     const parameters = {
       ...params,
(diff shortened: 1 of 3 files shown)


ui/src/components/Tasks/GenericTaskForm.js
```diff
@@ -43,6 +43,17 @@ class GenericTaskForm extends React.Component {
     this.props.handleSubmit(this.addToQueue.bind(this, true))();
   }
 
+  clearCurrentJSFormParameters() {
+    localStorage.removeItem(`current${this.props.taskData.type}Parameters`);
```
axelboc: You could make a getter for the local storage key instead — it's the logic that is worth abstracting IMO.

```js
get storageKey() {
  return `current${this.props.taskData.type}Parameters`;
}
```

axelboc (approved):

ui/src/components/Tasks/GenericTaskForm.js
marcus-oscarsson: Oh thats nice :)

ui/src/components/Tasks/GenericTaskForm.js
```diff
@@ -167,6 +183,18 @@ class GenericTaskForm extends React.Component {
       }
     }
 
+    const currentFormData = JSON.parse(
+      localStorage.getItem(`current${this.props.taskData.type}Parameters`),
```
axelboc: `this.jsFormStorageKey`

ui/src/components/Tasks/GenericTaskForm.js
```diff
@@ -244,10 +272,19 @@ class GenericTaskForm extends React.Component {
...
       this.props.schema.user_collection_parameters,
     );
+
     return (
-      <DraggableModal show={this.props.show} onHide={this.props.hide}>
+      <DraggableModal
+        show={this.props.show}
+        onHide={() => {
+          localStorage.removeItem(
+            `current${this.props.taskData.type}Parameters`,
```
axelboc: `this.jsFormStorageKey`

marcus-oscarsson: Oh, thanks :)

PR #1257, opened 2024-06-19
[Plate] Move plate config  from reducer to pydantic  configuration Model

jbflo:
The necessary parameter needed to build the plate  ui component 
Are now move to the ui.yaml file .

- small css updated

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "mxcubeweb"
-version = "4.83.0"
+version = "4.84.0"
 license = "LGPL-3.0-or-later"
 description = "MXCuBE Web user interface"
 authors = ["The MXCuBE collaboration <mxcube@esrf.fr>"]


test/HardwareObjectsMockup.xml/mxcube-web/ui.yaml
```diff
@@ -155,3 +155,67 @@ camera_setup:
       url: http://localhost:9090/mjpg/video.mjpg?streamprofile=mxcube
       width: 800
       height: 450
+
+plate_setup:
```
marcus-oscarsson: Super nice idea ! :1st_place_medal:. It should perhaps however be configured on hardware object level ? Its really nice with this configuration but I feel that its less a UI configuration and more specific to the hardware object ?

jbflo: I closed this one i'll create a new one with just  the css  changes

PR #1258, opened 2024-06-20
added 3-click centring tests

walesch-yan:
This includes 3 changes to the e2e tests:

- addition of 3-click centring tests
- restructuring app.cy.js into multiple specs, this makes it faster to test specific components in the future
- updating cypress to the latest version: this is necessary to fix a an issue that includees newer firefox versions and running multiple specs after another

### DIFF
--- a/ui/cypress/e2e/sampleControls.cy.js
+++ b/ui/cypress/e2e/sampleControls.cy.js
@@ -9,7 +9,7 @@ describe('3-click centring', () => {
 
   it('3-click centring should not work without sample', () => {
     cy.clearSamples();
-    cy.contains('button', '3-click centring').click();
+    cy.findByRole('button', { name: '3-click centring' }).click();
     cy.findByRole('alert', 'Error: There is no sample mounted').should(
       'be.visible',
     );
@@ -17,7 +17,7 @@ describe('3-click centring', () => {
 
   it('Each click is rotating the sample by 90 degrees', () => {
     cy.mountSample();
-    cy.contains('button', '3-click centring').click();
+    cy.findByRole('button', { name: '3-click centring' }).click();
     cy.get('.form-control[name="diffractometer.phi"]')
       .invoke('val')
       .then((value) => {
@@ -29,7 +29,7 @@ describe('3-click centring', () => {
           cy.wait(1000);
           cy.reload();
           // press the centring button again after reload to stay in the correct mode
-          cy.contains('button', '3-click centring').click();
+          cy.findByRole('button', { name: '3-click centring' }).click();
           cy.get('.form-control[name="diffractometer.phi"]')
             .invoke('val')
             .should('equal', omegaValue.toFixed(2));
--- a/ui/cypress/e2e/sampleControls.cy.js
+++ b/ui/cypress/e2e/sampleControls.cy.js
@@ -1,0 +1,41 @@
+/* global cy, it, describe, beforeEach, Cypress */
+
+describe('3-click centring', () => {
+  beforeEach(() => {
+    cy.login();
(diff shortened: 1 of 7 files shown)


ui/cypress/e2e/sampleControls.cy.js
```diff
@@ -0,0 +1,41 @@
...
+    );
+  });
+
+  it('Each click is rotating the sample by 90 degrees', () => {
+    cy.mountSample();
+    cy.contains('button', '3-click centring').click();
+    /* eslint-disable-next-line promise/catch-or-return */
+    cy.get('.form-control[name="diffractometer.phi"]')
+      .invoke('val')
+      /* eslint-disable-next-line promise/prefer-await-to-then */
```
axelboc: You could disable those two warnings for all Cypress specs in https://github.com/mxcube/mxcubeweb/blob/develop/ui/.eslintrc.js#L59

ui/cypress/e2e/sampleControls.cy.js
```diff
@@ -0,0 +1,41 @@
...
+          omegaValue += 90;
+          cy.get('.canvas-container').click();
+          // to update the omega value, a small amount of time must be waited and the page reloaded
+          cy.wait(1000);
+          cy.reload();
+          // press the centring button again after reload to stay in the correct mode
+          cy.contains('button', '3-click centring').click();
+          cy.get('.form-control[name="diffractometer.phi"]')
+            .invoke('val')
+            .should('equal', omegaValue.toFixed(2));
```
axelboc: Would be good to act more like a real user here (and in other places like in the `takeControl` command), with queries like [`cy.findByLabelText`](https://testing-library.com/docs/queries/bylabeltext) (to find an input field with its label) or [`cy.findByRole`](https://testing-library.com/docs/queries/byrole) (to find headings, buttons, etc.).

It can be done later though, as it may require "fixing" the accessibility/markup of the relevant features (like labeling form inputs).

axelboc (approved):
Very nice!

ui/cypress/e2e/sampleControls.cy.js
walesch-yan: Thanks for the comment, you are right that would be a better solution!

ui/cypress/e2e/sampleControls.cy.js
walesch-yan: As you mentioned for some elements, I could not use the mentioned queries. However, I agree with you that I often did not act like a real user (especially for `takeControl` and `clearSamples` commands) and some parts can be handled differently, I will take a look at it and make the necessary changes. Thanks for the comment!

marcus-oscarsson: :+1: Thanks guys !

ui/cypress/e2e/sampleControls.cy.js
```diff
@@ -9,15 +9,15 @@ describe('3-click centring', () => {
...
     );
   });
 
   it('Each click is rotating the sample by 90 degrees', () => {
     cy.mountSample();
-    cy.contains('button', '3-click centring').click();
+    cy.findByRole('button', { name: '3-click centring' }).click();
     cy.get('.form-control[name="diffractometer.phi"]')
       .invoke('val')
       .then((value) => {
```
walesch-yan, line 23: While I changed most of the cypress queries to act more like a real User, there are a few input forms that do not provide a corresponding label (resp. the text that should be the label is not correctly linked). This might be investigated in future steps.

marcus-oscarsson:
Really nice :), we can improve the access to some of the elements in a future PR :)

PR #1264, opened 2024-06-26
Lint YAML files with `yamllint`

fabcor-maxiv:
There is now a configuration file for `yamllint` and a `pre-commit` hook.

The most straightforward linting issues were fixed. The other ones are ignored and should be handled in the future.

### DIFF
--- a/test/HardwareObjectsMockup.xml/mxcube-web/server.yaml
+++ b/test/HardwareObjectsMockup.xml/mxcube-web/server.yaml
@@ -1,9 +1,11 @@
+---
+
 server:
   SECRET_KEY: "ASECRETKEY"
   SECURITY_PASSWORD_SALT: "ASALT"
   PERMANENT_SESSION_LIFETIME: 60
 
-  DEBUG: False
+  DEBUG: false
 
   ALLOWED_CORS_ORIGINS:
     - "http://localhost:8081"
@@ -11,7 +13,7 @@ server:
     - "ws://localhost:8000"
 
 mxcube:
-  USE_EXTERNAL_STREAMER: True
+  USE_EXTERNAL_STREAMER: true
   VIDEO_FORMAT: MPEG1
   VIDEO_STREAM_URL: "ws://localhost:8000/ws"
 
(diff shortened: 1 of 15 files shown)


fabcor-maxiv:
I started with this because I am planning to do more changes to the pre-commit hooks, and had to first clean up the formatting/linting of the `.pre-commit-config.yaml` file.

fabcor-maxiv:
We should also do the same for `mxcubecore`. Especially since we are planning to migrate all configuration to YAML files.

test/HardwareObjectsMockup.xml/mxcube-web/server.yaml
```diff
@@ -22,7 +24,7 @@ mxcube:
 
   usermanager:
     class: UserManager
-    inhouse_is_staff: True
+    inhouse_is_staff: true
```
elmjag, line 27: I don't know YAML in detail. Is`True` and `true`  equivalent in YAML? I.e. are the both treated as booleans? Or does this change the type from string to boolean?

If this is the latter case, when we need to check the code that consumes these values, so it still does the correct thing.

elmjag:
+1 for more linting!

test/HardwareObjectsMockup.xml/mxcube-web/server.yaml
marcus-oscarsson: Its treated the same as far as I know it results in a boolean

fabcor-maxiv: This was triggering this rule: https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.truthy

Yes, YAML is a disaster :D

I checked the mxcube code that uses these values and it expects booleans, not strings, so I assume this change should be fine, seems like the tests still pass.

We could set the rule that all strings have to be in quotes, so that we can not get confused between strings and booleans (and other value types) any more, but I bet lots of people wold hate that rule :D

elmjag: Nice, then it's not a problem. Thanks for checking.

fabcor-maxiv:
From my point of view, this is ready for merge. I hope to get started working on the same change for mxcubecore soon-ish.

marcus-oscarsson:
Thanks, ok

fabcor-maxiv:
For mxcubecore: https://github.com/mxcube/mxcubecore/pull/948

PR #1266, opened 2024-06-28
Use lowest supported Python version for checks

fabcor-maxiv:
We should do formatting, linting, testing, and so on against the lowest Python version that we intend to support. We want to make sure we do not introduce unsupported Python syntax. Because Python is generally backwards-compatible this should be good enough to check against lowest version.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       max-parallel: 5
       matrix:
-        python-version: ["3.10"]
+        python-version: ["3.8"]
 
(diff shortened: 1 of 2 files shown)


fabcor-maxiv:
I am quite sure that we discussed this already, likely on mxcubecore, so I was surprised to see it was not the case on mxcubeweb.

fabcor-maxiv:
In the checks:

> build-linux (3.10) Expected — Waiting for status to be reported 

I do not understand why, since this pull request replaces it with 3.8. I believe this check should not be run. There is no point in waiting for it to pass. I do not think there is anything to do about this, the pull request is ready for merging from my point of view.

marcus-oscarsson:
Indeed, thanks !

PR #1269, opened 2024-06-28
Export Poetry's lockfile as pip requirements.txt

### LINKED ISSUE #1268: Export Python dependencies as *pip* `requirements.txt`
It seems like for some facilities (including us at MAXIV to some extent), it is more practical to install MXCuBE-Web with pip (instead of Poetry).

As explained in [Deployment](https://mxcubeweb.readthedocs.io/en/latest/dev/deployment.html) ([permalink](https://github.com/mxcube/mxcubeweb/blob/a79b64e1cc6fe6f852897c5ae1c91efef1774c13/docs/source/dev/deployment.md)), the issue is that by installing with pip we lose the advantage of Poetry's lockfile (pip can not read Poetry's lockfile so it ignores it). This results in unpinned dependencies being installed during deployment to production, and potential issues...

We can take advantage of Poetry's `export` command and the *pre-commit* hooks to automatically generate a `requirements.txt` file that contains the pinned dependencies and that pip can read. This could help provide repeatable installations in production environments.

fabcor-maxiv:
This will let the pre-commit hooks ensure that Poetry's lockfile is kept up-to-date and that it is exported to a pip `requirements.txt` file.

GitHub: fix https://github.com/mxcube/mxcubeweb/issues/1268


fabcor-maxiv:
I feel like maybe this should be discussed, so I keep it as draft for a bit. In particular it feels like attempts at using some of those Poetry pre-commit hooks have been made in the past, and I can assume it was not entirely successful, so maybe there is some background info that I am missing.

marcus-oscarsson:
Hm, yes. I suspect so as well

fabcor-maxiv:
I am wondering how we could move this forward...

What could be an issue here?

Updates to `requirements.txt` would happen only when `poetry.lock` changes. Typically changes to `poetry.lock` happen only when Dependabot does it, or when we do changes to the dependency list in `pyproject.toml`. So all in all, it feels like we can keep the Dependabot part under control, and on the other hand changing the dependency list in `pyproject.toml` is relatively rare that one of the more experienced contributors can easily intervene if things get messy. If I checked everything correctly, the pre-commit hooks should do most of the work and will inform correctly when files are not in sync.

Maybe I could document this a bit...

fabcor-maxiv:
@oldfielj-ansto If I am not mistaken, you were the one to introduce pre-commit hooks and Poetry. Maybe you have some opinion on this...

.pre-commit-config.yaml
```diff
@@ -13,9 +13,19 @@ repos:
           - --strict
 
   - repo: https://github.com/python-poetry/poetry
-    rev: 1.5.0
+    rev: 1.8.3
     hooks:
       - id: poetry-check
+        args:
```
marcus-oscarsson, line 25: What does this hook do, does it check if the lock file is up to date or does it also update and commit it ?

marcus-oscarsson:
It would probably nice with some documentation :)

.pre-commit-config.yaml
fabcor-maxiv: As far as I can tell it runs [`poetry check --lock`](https://python-poetry.org/docs/cli/#check). The documentation is not super clear on what this does, I assume there is an issue in the documentation, because the behavior does not match the documentation.

Anyway what actually happens in this hook is that it checks that:
1. the content of `pyproject.toml` is valid (from the point of view of Poetry)
2. there is a `poetry.lock` file
3. the content of `poetry.lock` is in sync with `pyproject.toml`

The next hook runs [`poetry lock --no-update`](https://python-poetry.org/docs/cli/#lock), which does syncs the lock file. The hook will fail if the hook changes the file, this is important to have on the CI side, for the case when someone commits without running the hooks locally.

Then the third hook does the export to `requirements.txt`. The hook will fail if the hook changes the file, this is important to have on the CI side, for the case when someone commits without running the hooks locally.

There is some redundancy between the first two steps, and I assume we can simplify over time, depending on how things go.

fabcor-maxiv:
> It would probably nice with some documentation :)

I added a bit of documentation. Mostly in comments in `.pre-commit-config.yaml`, let me know if you think I should add more documentation and/or move it  somewhere else.

marcus-oscarsson:
How come that the status checks are pending for such a long time ?

fabcor-maxiv:
> How come that the status checks are pending for such a long time ?

Merging #1289 broke the GitHub Actions workflows completely. See https://github.com/mxcube/mxcubecore/issues/960

marcus-oscarsson:
@fabcor-maxiv: Yes, I think its fixed now. Seems like the poetry-export hook reacts to something I guess its the update of `certifi`

fabcor-maxiv:
> Seems like the poetry-export hook reacts to something I guess its the update of `certifi`

Oh, nice! That is very good that the issue is caught, that is what we want. : )

I forced the pre-commit hook to regenerate the exported `requirements.txt` (`pre-commit run -a`).

fabcor-maxiv:
This is now rebased, it takes into account the newly added dependency (`argon2-cffi` <https://github.com/mxcube/mxcubeweb/pull/1327>).

PR #1271, opened 2024-07-01
yaml: revert back detector config to XML

elmjag:
The detector.yml is missing, let's go back to detector.xml for now.


elmjag:
I suggest that during the development phase of the yaml config files, let's have both `HardwareObjectsMockup.xml`  and `HardwareObjectsMockup.yml` folders on `xml_yaml_conversion` branch. It very convenient for debugging purposes.

 Let's keep `HardwareObjectsMockup.yml` as-is. For `HardwareObjectsMockup.yml`, I can try to make a PR, once I get something running.

PR #1283, opened 2024-07-02
added test documentation

walesch-yan:
WiP: Wrote a documentation page on the usage of tests and how to contribute. Currently the part about e2e tests is described in more detail, while some more documentation on the usage of pytest might be needed.


docs/source/dev/tests.md
```diff
@@ -0,0 +1,94 @@
...
+
+When the re-run did not help, try to switch the testing browser, you can do so by using the cypress client. However, please make sure that your code works with most browsers, especially with chrome and firefox. 
+
+It is important to notice that cypress will not test in full-screen mode, please make sure, that buttons and other elements stay visible and clickable in smaller test-windows.
+
+The cypress tests are configured to act according to human behavior, hence during testing, certain button names or texts are looked instead of element attributes. If for a reasonable argument, your code adjustments change these names, please make sure to change the tests accordingly.
+
+**Very important**: If you are using the development server while simultaneously running tests, do not take control (for mounting samples, running queue, ...) on the development server. This might result in tests failing, because they needed to access a certain resource, that is not accessible in _Observer mode_.
+
+### Contributing
```
fabcor-maxiv, line 83: We already have a contributing document.

So I guess we should link the two documents back and forth.

docs/source/dev/tests.md
```diff
@@ -0,0 +1,94 @@
+# Automated Testing
+
+Testing is a crucial aspect of software development that ensures the reliability, quality, and performance of an application. Additionally, testing facilitates better code design and architecture by encouraging modular and decoupled code. Hence, this document will focus on the different testing frameworks used in this project, explain how to correctly use them and how to contribute to the testing environment.
+
+The tests in this project include two different frameworks, namely `pytest` and `cypress`. The focus of both testing frameworks is slightly different:
+ - `pytest` tests focus on individual units of code, for this application, these tests primarly cover authentication and routing.
```
fabcor-maxiv: I believe pytest can do end-to-end tests, and functional tests, and unit tests. Depends on the context and how you use it.

I guess maybe we should rather phrase it something like this:

> pytest focuses on testing Python code. In the case of MXCuBE-Web the Python code is the whole back-end server side of the application.

docs/source/dev/tests.md
```diff
@@ -0,0 +1,94 @@
...
+ - `pytest` tests focus on individual units of code, for this application, these tests primarly cover authentication and routing. 
+ - `cypress` is used for e2e tests. These tests should cover the application's workflow from the user's perspective. They simulate user interaction with the application's interface and ensure the correct functionality of features, integration of components, and overall user experience.
+
+## Prerequirements
+
+Before you can run any tests, you need to make sure to run the local _Redis_ server. You can simply do so by running the `redis-server` command in the `mxcubeweb` conda environment (Best opening a new terminal for this).
+
+For running the e2e tests you need to additionally run the front-end part of MxCube, to do so please follow [these instructions](./environment.md).
+
+It is best to run all the tests in the `mxcubeweb` conda environment.
```
fabcor-maxiv: Technically speaking no one needs a conda environment. I would rather leave mentions of "conda" out of this document. And instead write something along the lines of:

```suggestion
Running tests requires the Redis server to be running.

The document [](./environment.md) has more information about setting up a development environment for running tests.
```

docs/source/dev/tests.md
walesch-yan: I agree, that is a good idea

docs/source/dev/tests.md
walesch-yan: Of course, you are right! I see that the way I wrote this part did not turn out the way I intended. Your suggested changes seem much clearer thank you!

docs/source/dev/tests.md
marcus-oscarsson: I'm wondering, should we move the contents of `CONTRIBUTING.md` so that its only in the sphinx documentation and simply only have a link pointing to the part in `CONTRIBUTING.md`. I know that `CONTRIBUTING.md` is automatically included but it can be a bit confusing.

fabcor-maxiv: @marcus-oscarsson Yes, we can consider that. I created a ticket: https://github.com/mxcube/mxcubecore/issues/955

fabcor-maxiv: Then @walesch-yan let's consider this conversation resolved. We will handle the links back and forth separately.

walesch-yan:
Added the suggestions that were made on this pr, thanks a lot for those

docs/source/dev/tests.md
```diff
@@ -0,0 +1,92 @@
+# Automated Testing
+
+Testing is a crucial aspect of software development that ensures the reliability, quality, and performance of an application. Additionally, testing facilitates better code design and architecture by encouraging modular and decoupled code. Hence, this document will focus on the different testing frameworks used in this project, explain how to correctly use them and how to contribute to the testing environment.
+
+The tests in this project include two different frameworks, namely `pytest` and `cypress`. The focus of both testing frameworks is slightly different:
```
fabcor-maxiv: In general it is always a good idea to add links...

```suggestion
The tests in this project include two different frameworks, namely [`pytest`](https://pytest.org/) and [`cypress`](https://www.cypress.io/). The focus of both testing frameworks is slightly different:
```

fabcor-maxiv (commented):
@walesch-yan seems fine to me. It is still marked as draft, though.

docs/source/dev/tests.md
walesch-yan: That's a good point, I added links, where necessary and marked the pr as ready for review, thanks!

fabcor-maxiv (approved):

PR #1284, opened 2024-07-04
Separate out shapes API from Redux + clean up obsolete `sampleview` actions

axelboc:


mxcubeweb/routes/diffractometer.py
```diff
@@ -71,15 +71,6 @@ def md_in_plate_mode():
...
 
-    # @bp.route("/movables/state", methods=["GET"])
-    # @server.restrict
-    # def get_movables_state():
-    #     ret = utils.get_centring_motors_info()
-    #     ret.update(utils.get_light_state_and_intensity())
-    #     resp = jsonify(ret)
-    #     resp.status_code = 200
-    #     return resp
-
```
axelboc, line 82: Since this endpoint seems obsolete, I'm also removing the `fetchMotorPositions` function I had already extracted, and the corresponding `getMotorPositions` and `saveMotorPositions` action creators.

ui/src/actions/sampleview.js
```diff
@@ -120,21 +125,21 @@ export function addCentringPoint(x, y) {
   };
 }
 
-export function addShape(shape) {
+export function addShapeAction(shape) {
```
axelboc, line 128: Adding suffix `Action` to differentiate this simple action creator from the `addShape` async action creator.

ui/src/actions/sampleview.js
```diff
@@ -161,43 +166,22 @@ export function videoMessageOverlay(show, msg) {
...
-
-          window.initJSMpeg();
-        });
+    if (!sampleview.sourceIsScalable) {
+      return;
     }
-  };
-}
 
-export function saveMotorPositions(data) {
```
axelboc, line 197: Obsolete as per my previous comment.

ui/src/actions/sampleview.js
```diff
@@ -238,10 +222,6 @@ export function toggleDrawGrid() {
   return { type: 'DRAW_GRID' };
 }
 
-export function addGridAction(gridData) {
-  return { type: 'ADD_GRID', gridData };
-}
```
axelboc, line 243: Unused

ui/src/actions/sampleview.js
```diff
@@ -336,92 +310,47 @@ export function sendGoToBeam(x, y) {
   };
 }
 
-export function sendStartAutoCentring() {
```
axelboc, line 339: Doesn't seem to be used. The endpoint still exist, though, so not sure :shrug:

ui/src/actions/sampleview.js
```diff
@@ -336,92 +310,47 @@ export function sendGoToBeam(x, y) {
...
-        'Content-type': 'application/json',
-      },
-    }).then((response) => {
-      if (response.status >= 400) {
-        throw new Error('Server refused to start autocentring');
-      }
+export function addShape(shapeData = {}, successCb = null) {
+  return async (dispatch) => {
+    try {
+      const json = await sendAddOrUpdateShapes([shapeData]);
```
axelboc, line 316: Note the use of the same fetch function `sendAddOrUpdateShapes` in both `addShape` and `updateShapes`.

ui/src/actions/sampleview.js
```diff
@@ -492,73 +421,6 @@ export function sendStartClickCentring() {
...
-        'Content-type': 'application/json',
-      },
-    }).then((response) => {
-      if (response.status >= 400) {
-        throw new Error('Server refused to stop motor');
-      }
-    });
-  };
-}
-
```
axelboc, line 561: None of these seem to be used. Some of the endpoints still seem to exist in `samplecentring.py`, though.

ui/src/actions/sampleview.js
```diff
@@ -626,81 +488,6 @@ export function changeAperture(size) {
...
-          throw new Error('Server refused to return points position');
-        }
-        return response.json();
-      })
-      .then((json) => {
-        dispatch(updateShapes(json));
-      });
-  };
-}
-
```
axelboc, line 703: Again, none of these seem to be used.

axelboc:
There's something weird going on with three-point centring. When I try to add a new point, it moves the previous one I had saved. I've taken the screen recording below on `develop` to make sure it's not related to this PR:

Peek 2024-07-04 11-54

At some point, there seems to be multiple labels on top of each other too.

ui/src/actions/sampleview.js
marcus-oscarsson: Great !

marcus-oscarsson:
Wonderful, very nice cleanup @axelboc well done.

The thing you noticed for the centering above is because the simulation centering always reports the same position for all
positions so they are moved to the same point. We have been talking about changing that so that its atleast not the same position.

PR #1288, opened 2024-07-04
Improved yaml linting and tightened config. Set YAML 1.2 everywhere

rhfogh:
Companion pull request to mxcubecore PR #957

There may be problems with poetry.lock and/or with mxcubeweb PR #1285.


poetry.lock
fabcor-maxiv, line 1: This file (`poetry.lock`) should not have changed. I do not understand where this change comes from.

Maybe you could restore the file to its previous state and re-run `poetry install` or something like this.

elmjag:
Same comment as for this PR: https://github.com/mxcube/mxcubecore/pull/957

Let's merge this into `develop`. I don't see a need to intermingle these changes with our XML to YAML conversion work.

marcus-oscarsson:
@rhfogh Did you notice that there is a conflict ?

rhfogh:
@marcus-oscarsson  The conflict should be fixed now. Thanks for pointing it out.

PR #1289, opened 2024-07-05
Manually transferred yaml linting improvements from rhfogh_xml_yaml

rhfogh:
Adds # %YAML 1.2 to top of yaml files and changes some Booleans to lower case. Manual recreation of (most of) a commit from xml_yaml brannch


fabcor-maxiv:
@marcus-oscarsson I am quite sure this was not ready to be merge. This breaks the github actions workflows. See https://github.com/mxcube/mxcubecore/issues/960

marcus-oscarsson:
No, it was really not ! I must have missed that it was a draft there are actually several issues with this. Ill revert this

rhfogh:
@fabcor @marcus-oscarsson You could argue that this is actually correct, and the problem is on the testing side, in that the current github workflows cannot deal with correct YAML 1.2. But OK, we had better decide how we want to solve that before we merge it in.

PR #1290, opened 2024-07-05
Separate out sample changer API from Redux

axelboc:
Only the Queue and LIMS APIs left.

### DIFF
--- a/ui/src/actions/sampleChanger.js
+++ b/ui/src/actions/sampleChanger.js
@@ -1,10 +1,18 @@
 /* eslint-disable promise/catch-or-return */
-/* eslint-disable promise/no-nesting */
 /* eslint-disable promise/prefer-await-to-then */
-/* eslint-disable sonarjs/no-duplicate-string */
-import fetch from 'isomorphic-fetch';
 import { showErrorPanel } from './general';
 import { clearCurrentSample } from './queue'; // eslint-disable-line import/no-cycle
+import {
+  fetchLoadedSample,
+  fetchSampleChangerContents,
+  sendAbortSampleChanger,
+  sendMountSample,
+  sendSampleChangerCommand,
+  sendScanSampleChanger,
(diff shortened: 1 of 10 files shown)


ui/src/actions/sampleChanger.js
```diff
@@ -1,10 +1,18 @@
 /* eslint-disable promise/catch-or-return */
-/* eslint-disable promise/no-nesting */
 /* eslint-disable promise/prefer-await-to-then */
-/* eslint-disable sonarjs/no-duplicate-string */
-import fetch from 'isomorphic-fetch';
 import { showErrorPanel } from './general';
 import { clearCurrentSample } from './queue'; // eslint-disable-line import/no-cycle
+import {
+  fetchLoadedSample,
```
axelboc, line 6: There's a naming inconsistency in the back-end between "load" and "mount". I've tried to match the back-end naming - i.e. `fetchLoadedSample` but `sendMountSample` and `sendUnmountSample`.

ui/src/actions/sampleChanger.js
```diff
@@ -58,175 +66,85 @@ export function selectDrop(drop_index) {
...
-    fetch('mxcube/api/v0.1/sample_changer/loaded_sample', {
-      method: 'GET',
-      headers: {
-        Accept: 'application/json',
-        'Content-type': 'application/json',
-      },
-      credentials: 'include',
-    }).then((response) => {
-      if (response.status >= 400) {
-        throw new Error('Error refreshing sample changer contents');
```
axelboc, line 87: In case this hasn't been clear so far, when the error is just thrown like this (instead of shown in the UI via `showErrorPanel`), I've tended to simply remove the `throw` statement, since `wretch` already throws out of the box when it receives an error response.

Obviously we'll have to implement more robust error management at some point.

marcus-oscarsson:
Thanks !

PR #1294, opened 2024-07-08
[Revert 1288] 

marcus-oscarsson:
I must have merged PR1288 a bit too early, So I'm reverting. 

 - It seems like there are #hashes missing before the %YAML 1.2 
 - There a few linting errors that are "non trivial" (there is no automatic fix for them)
 - yamllint does not seem to auto format files


marcus-oscarsson:
There is possible this tool that can help but it does not seem to fix line length issues and so on:
   https://github.com/lyz-code/yamlfix

rhfogh:
@marcus-oscarsson 
yamlfix does not look good to me. Basically, if we have yamllint we should be able to do the rest ourselves, rather than rely on an opinionated formatter.
Some items:

- Automatically converting truthy strings to 'false'/'true' risks converting 'On' to 'true'. Not what we want. Yaml 1.2 plus 'truthy' warnings looks safer

- The only long lines that were not easy to fix by hand were those that showed command line commands (and that were longer than 88 characters). I do not like the idea of having those automatically split, just to fit within a style guide.

- Yaml does have some tricky corner cases (regrettably)  to do with the automatic type conversion. It seems more reliable to me if we know what we are doing, rather than leave it to a black box.

marcus-oscarsson:
Sorry, my mistake it should be pr #1289 not #1288

@rhfogh I agree it does not seem to be a good choice. It was what I found that could help in an initial conversion phase but its not a good choice for everyday use.

PR #1295, opened 2024-07-09
Updated yamllint fixes and avoided some rule exceptions

rhfogh:
Fixed yamllint error and replaced 'truthy' exception with use of "on":

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -2,9 +2,9 @@
 
 name: Python build and test
 
-on:  # yamllint disable-line rule:truthy
-  pull_request:
-  push:
+"on":
+  pull_request: null
+  push: null
 
 jobs:
   build-linux:
--- a/.github/workflows/bump_version_and_tag.yml
+++ b/.github/workflows/bump_version_and_tag.yml
@@ -4,7 +4,7 @@
 
 name: Create/Update Tag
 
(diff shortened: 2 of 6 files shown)


fabcor-maxiv (commented):
@rhfogh Should we also add all of the following in `.yamllint.yaml` so that it matches what we (will) have in mxcubecore?

```yaml

  # octal values must start with 0o, numbers may not start with e.g. 01
  # It is simpler, and costs nothing, to disallow values like `010`
  # except as quoted strings
  octal-values:
    forbid-implicit-octal: true

  # This bans implicit nulls, nulls must be explicit.
  empty-values: enable

  # With these settings, and `%YAML 1.2` in the file header, we have :
  #
  #  Booleans: true, false
  #  None: null, Null, NULL
  #  not-a-number: .NaN, .NAN, .nan
  #  Infinity:  .Inf, .INF, .inf
  #  Strings: On, ON, Off, OFF, YES, Yes, NO, No
  #  Disallowed if not quoted: True, TRUE, False, FALSE
  #
  # Setting `truthy: disabled` would allow the disabled values, as Booleans
  #
  # If desired, you could ban unquoted infinity and not-a-number by
  # float-values: forbind-int: true
  # float-values: forbind-nan: true
```

rhfogh:
@fabcor-maxiv Indeed. Done.

marcus-oscarsson:
Thanks, very nice !

PR #1296, opened 2024-07-09
Rhfogh xml yaml

rhfogh:
Fixed yaml syntax and yaml linting, conf. recent changes in develop branch


rhfogh:
The remaining test failures have nothing to do with recent changes and are nowhere near my code. Could someone have a look?

test/test_authn.py
```diff
@@ -51,7 +51,7 @@ def server(request, login_type):
 
     hw_repo = mxcubecore.HardwareRepository.get_hardware_repository()
     lims = hw_repo.get_hardware_object("lims")
-    lims.set_property("loginType", login_type)
+    lims._set_property("loginType", login_type)
```
elmjag, line 55: This does not look correct to me.

I did this change: https://github.com/elmjag/mxcubeweb/commit/14b5bbd35853fc6e09705570fe70ea755ad1abe3
With that change, at least locally, all the `pytest` tests pass for me.

elmjag:
If these changes are coming directly from recent `develop` branch, I don't see a big point on merging it here. It's better that we rebase `xml_yaml_conversion` on top of the recent `develop`. Then we will "automatically" get all the changes above.

In fact, I was planning to do a rebase of `xml_yaml_conversion`, basically as soon as https://github.com/mxcube/mxcubeweb/issues/1372 and https://github.com/mxcube/mxcubeweb/issues/1370 issues are resolved.

test/test_authn.py
fabcor-maxiv: Yes this PR seems to include a bunch of older changes that I am quite sure were already merged but I could be wrong. Maybe some rebase gone bad?

rhfogh:
@elmjag Indeed. This was more a housekeeping  thing, I wanted that old PR to get out of the way, but you are quite right this should simply be left for the later merge. PR withdrawn

PR #1297, opened 2024-07-09
Use async/await in Redux thunks

axelboc:
There are still a few `then` chains in some components here and there. I focused on removing them from Redux thunks.

### DIFF
--- a/ui/src/actions/queue.js
+++ b/ui/src/actions/queue.js
@@ -1,6 +1,3 @@
-/* eslint-disable promise/prefer-await-to-callbacks */
-/* eslint-disable promise/catch-or-return */
-/* eslint-disable promise/prefer-await-to-then */
 import { showErrorPanel } from './general';
 import { mountSample } from './sampleChanger'; // eslint-disable-line import/no-cycle
 import { abortCentring, updateShapes } from './sampleview';
@@ -37,18 +34,12 @@ export function clearAll() {
 }
 
 export function setQueueAction(queue) {
-  return {
-    type: 'SET_QUEUE',
-    sampleOrder: queue.sampleOrder,
-    sampleList: queue.sampleList,
-  };
+  const { sampleOrder, sampleList } = queue;
+  return { type: 'SET_QUEUE', sampleOrder, sampleList };
 }
 
 export function setCurrentSample(sampleID) {
-  return {
-    type: 'SET_CURRENT_SAMPLE',
-    sampleID,
-  };
+  return { type: 'SET_CURRENT_SAMPLE', sampleID };
 }
 
 export function setQueue(queue) {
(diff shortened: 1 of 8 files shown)


ui/src/actions/remoteAccess.js
```diff
@@ -41,57 +36,51 @@ export function requestControl(
...
   };
 }
 
 export function updateTimeoutGivesControl(timeoutGivesControl) {
-  return (dispatch) => {
-    sendUpdateTimeoutGivesControl(timeoutGivesControl);
+  return async (dispatch) => {
+    await sendUpdateTimeoutGivesControl(timeoutGivesControl);
     dispatch({ type: 'SET_TIMEOUT_GIVES_CONTROL', timeoutGivesControl });
   };
```
axelboc, line 85: Notice that there were no `then` chains, so the store was updated "optimistically". This was risky since there is no error handling, so the store could get out of sync. I'm now waiting for the request to succeed before updating the store. This may make the checkboxes less responsive, but at least it is safer for now.

ui/src/actions/remoteAccess.js
```diff
@@ -41,57 +36,51 @@ export function requestControl(
   name = '',
   userInfo = {},
 ) {
-  return () => {
-    sendRequestControl(control, message, name, userInfo);
-  };
+  return () => sendRequestControl(control, message, name, userInfo);
```
axelboc, line 39: By returning the promise here, I turn the arrow function into an asynchronous arrow function (there's a linting rule to prevent adding `async` without `await`, so it's not super obvious). Returning promises from thunks allows awaiting `dispatch` calls - e.g. 

```js
await dispatch(requestControl(...))
// do something after the `sendRequestControl` promise has fullfilled
```

ui/src/actions/queue.js
```diff
@@ -126,18 +110,20 @@ export function addSamplesToQueue(sampleDataList) {
...
+          const json = await sendAddQueueItem([sampleData]);
+          dispatch(setQueue(json));
+          dispatch(selectSamplesAction([sampleData.sampleID]));
+        } catch (error) {
+          if (error.status >= 400) {
+            dispatch(showErrorPanel(true, 'Server refused to add sample'));
+          }
+        }
+
+        dispatch(queueLoading(false));
```
axelboc, line 126: Note that we don't need a `finally` block since the `try` block doesn't return and the `catch` block doesn't re-throw.

axelboc (commented):
Big thanks to VS Code's [`async` refactoring](https://code.visualstudio.com/docs/typescript/typescript-refactoring#_code-suggestions) feature. :pray:

marcus-oscarsson:
> Big thanks to VS Code's [`async` refactoring](https://code.visualstudio.com/docs/typescript/typescript-refactoring#_code-suggestions) feature. 🙏

Nice one !

ui/src/actions/remoteAccess.js
marcus-oscarsson: Well done !

marcus-oscarsson:
Ahh, feels much cleaner already :)

PR #1298, opened 2024-07-09
Remove obsolete 4xx checks + add missing try/catch blocks

axelboc:
Reasoning for removing all the `error.status >= 400` checks inside the try/catch blocks:

- Wretch throws errors when `response.ok` is `false`, which happens when the status code is **outside the 2xx range**.
- If they are well-formed, 1xx and 3xx responses are typically **"interim" response codes**; the browser deals with them transparently, which means that in our `catch` blocks, we are likely to see only 4xx or 5xx errors (so the `error.status >= 400` checks are redundant).
- If we do get a 1xx or 3xx error, it's likely a programming error and we need to report it, so we really don't want to exclude them with `error.status >= 400`.

### DIFF
--- a/ui/src/actions/harvester.js
+++ b/ui/src/actions/harvester.js
@@ -32,127 +32,110 @@ export function setHarvesterCommandResponse(response) {
 
 export function refresh() {
   return async (dispatch) => {
-    const response = await sendRefresh();
-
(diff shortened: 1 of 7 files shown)


marcus-oscarsson:
Excellent !

PR #1315, opened 2024-07-11
Fix spinner sizes

axelboc:
Apologies, looks like I forgot to push this after I took the screenshots. :sweat:

`10rem` is not a valid value for the HTML `width` attribute.


marcus-oscarsson:
Ok, no worries

PR #1320, opened 2024-07-12
Refactor navigation bar

### LINKED ISSUE #1306: Nav link not highlighted when first logging in
The `/` route renders the _Data Collection_ page, but the corresponding nav link is not highlighted:

image

Either the link must be highlighted on `/`, or the `/` route should redirect to `/datacollection` (my preference).

### LINKED ISSUE #1307: Nav links are not focusable with keyboard
The nav links are actually rendered as `div` elements with `href` attributes, which is obviously not valid HTML and means they aren't focusable with the keyboard:

image

axelboc:
- Fix #1306 — after logging in, the user is now redirected to `/datacollection`, and I've added a fallback route so that `/` and any other unknown route gets redirected to `/datacollection`.
- Fix #1307 — the navigation links are now real links that can be focused and selected with the keyboard (and they have visible focus outlines).

I took the opportunity to completely refactor the `MXNavbar` component. Here is a summary of the changes:

- Remove link to `/remoteaccess` from "MXCuBE-Web" logo
- Improve responsive design, especially on small desktop
- Improve visual design, notably hover and active states
- Control expanded state of mobile nav to fix collapsing when navigating
- Convert `MXNavbar` to functional component
- Replace global styles with CSS modules
- Remove `MXNavbarContainer` and use React Redux' `useSelector` and `useDispatch` instead

Peek 2024-07-12 14-59


axelboc:
@walesch-yan the E2E tests seem to be failing. I think Cypress can't find the "Remote" navigation link, because the nav bar is now collapsed (for mobile) at a higher breakpoint than before. I've added `hidden: true` to the `findByRole` query in `takeControl`, but it still doesn't work. Since I'm off for two weeks, do you think you could try to find a solution to make the tests pass and mark the PR as ready when they are? Maybe just configure Cypress with a higher screen resolution?

walesch-yan:
@axelboc yes, I can do that next week. Enjoy your time off!

ui/src/components/MXNavbar/MXNavbar.module.css
jbflo, line 1: Great job on making this responsive :ok_hand:

walesch-yan:
I set a higher screen resolution in cypress to pass the tests, for the time being, this seems like the best approach. In the future, we could add additional tests for smaller screens if necessary

marcus-oscarsson:
Wonderful nice work guys :), I like the team effort !

PR #1323, opened 2024-07-16
[bug] FIX potential  TypeError :

jbflo:
the parameter was forgotten when adding the logger. 

To avoid having a 
Logger.exception() missing 1 required positional argument: 'msg' error

### DIFF
--- a/mxcubeweb/routes/lims.py
+++ b/mxcubeweb/routes/lims.py
@@ -28,7 +28,9 @@ def init_route(app, server, url_prefix):  # noqa: C901
         try:
             res = jsonify(app.lims.synch_with_lims())
         except Exception as ex:
-            logging.getLogger("MX3.HWR").exception()
+            logging.getLogger("MX3.HWR").exception(
+                "Could not synchronize with LIMS %s" % str(ex)
+            )
             res = (
                 "Could not synchronize with LIMS",
                 409,


mxcubeweb/routes/lims.py
```diff
@@ -28,7 +28,9 @@ def proposal_samples():
         try:
             res = jsonify(app.lims.synch_with_lims())
         except Exception as ex:
-            logging.getLogger("MX3.HWR").exception()
+            logging.getLogger("MX3.HWR").exception(
+                "Could not synchronize with LIMS %s" % str(ex)
+            )
```
fabcor-maxiv, line 33: The `exception` method should add the exception and its traceback automatically. Unless this is unwanted, it should be okay to simplify to:

```suggestion
            logging.getLogger("MX3.HWR").exception(
                "Could not synchronize with LIMS"
            )
```

https://docs.python.org/3/library/logging.html#logging.exception

jbflo: I expected a comment like that from you @fabcor-maxiv  :hugs:
but when reading the logs i prefer having the message and the exception together in a line. 
it's like error msg follow by reason in one line.  but you right, my point is just aesthetic.

fabcor-maxiv: Then it's fine : )

The logging is all over the place in MXCuBE. I still have hope that one day we can clean up all the logging.

fabcor-maxiv (approved):

PR #1330, opened 2024-07-31
support python 3.11

elmjag:
- mark python 3.11.* as supported for this package
- require a version of mxcubecore that supports python 3.11
- require a version of mxcube-video-streamer that supports python 3.11

### DIFF
--- a/conda-environment.yml
+++ b/conda-environment.yml
@@ -9,7 +9,7 @@ dependencies:
   # Runtime dependencies
   # ====================
 
-  - python >=3.8,<3.11  # Make sure it matches `pyproject.toml`
+  - python >=3.8,<3.12  # Make sure it matches `pyproject.toml`
 
   # We install `python-ldap` from conda because it is "hard" to install.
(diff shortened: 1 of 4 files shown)


marcus-oscarsson:
@elmjag Should be fine now PR for the video streamer is merged

elmjag:
So, Cypress tests are failing for some reason. I have tested running MxCUBE-Web in python 3.11, it seems to work fine. I presume that these cypress failures are not related to python 3.11.

fabcor-maxiv:
> So, Cypress tests are failing for some reason. I have tested running MxCUBE-Web in python 3.11, it seems to work fine. I presume that these cypress failures are not related to python 3.11.

Seems like transient issue. I recommend re-triggering (via the web-UI) until it passes. I just did re-trigger the failing job.

marcus-oscarsson:
Everything looks good now, I;m sorry but I cant rebase from here do you think you can do it (and then merge) ?

PR #1331, opened 2024-08-02
Remove undefined elements in ContextMenu

### LINKED ISSUE #1300: Placeholder in dropdown dialog
The dropdown lists that open by right clicking on the canvas item (resp. placed 2d-points) contain clickable placeholders. In the screenshots below I highlighted the clickable area:

Placeholder1
Placeholder2

walesch-yan:
This PR closes #1300.
It prevents undefined dropdown elements to be shown as placeholders in the dropdown menu.

### DIFF
--- a/ui/src/components/SampleView/ContextMenu.js
+++ b/ui/src/components/SampleView/ContextMenu.js
@@ -512,6 +512,10 @@ export default class ContextMenu extends React.Component {
   }
 
   listOptions(type) {
+    if (type.text === undefined) {
+      return undefined;
+    }
+
     let el = (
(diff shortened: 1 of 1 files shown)


axelboc (approved):

fabcor-maxiv:
Just a drive-by comment...

Do we know why we have some `undefined` in the list? Would that make sense to work on preventing `undefined` from entering the list in the first place instead?

walesch-yan:
I assumed that these `undefined` only exist in the mockup due to some components not being available resp. only placeholders being available.

PR #1332, opened 2024-08-02
queue: read precision setting via PathTemplate class

elmjag:
Accessing session precision config via 'session["file_info"]' syntax is not supported when YAML configuration file is used. Use the PathTemplate class instead.

Without this change, it's not possible to add vanilla data collection tasks to the queue, when Session HWOBJ is configured with a YAML file.


elmjag:
I guess is objecting to this change. I'll merge it later today. @rhfogh

PR #1333, opened 2024-08-05
Prevent error from holding ctrl on 2D points

### LINKED ISSUE #1301: Runtime error on multiple point selection in mockup version
When you select multiple points by holding the ctrl button and left click on each 2D-point you want to select, a runtime error is thrown if you click on one point twice. While the intended behavior should be to unselect the point.

multiple_click_error

This behavior cannot be observed if only a single point is selected, in that case the point becomes unselected and no error is thrown (keep in mind that the ctrl button is hold during the process)

multiple_click_no_error

walesch-yan:
This PR closes #1301.
By catching the error that is thrown from an undefined function (undefined because of no previously selected points), we can ensure the correct behavior for selecting multiple points on the canvas element. See result below:
multiple_click_fix

### DIFF
--- a/ui/src/components/SampleView/SampleImage.jsx
+++ b/ui/src/components/SampleView/SampleImage.jsx
@@ -511,23 +511,6 @@ export default class SampleImage extends React.Component {
 
     this.drawGridPlugin.clearMouseOverGridLabel(this.canvas);
 
-    if (option.target && option.target.type === 'activeSelection') {
-      const group = this.canvas.getActiveObject();
-      const clickPoint = new fabric.Point(option.e.offsetX, option.e.offsetY);
-
-      // Important to call this for containsPoint to work properly
-      // this.canvas.discardActiveObject();
-
-      group.getObjects().forEach((obj) => {
-        if (!objectFound && obj.containsPoint(clickPoint) && obj.selectable) {
-          objectFound = obj;
-        } else {
-          // this.deSelectShape([obj], option.e.ctrlKey);
-        }
-      });
-    } else if (option.target) {
-      objectFound = option.target;
-    }
     const {
       sampleViewActions,
       clickCentring,
@@ -553,6 +536,19 @@ export default class SampleImage extends React.Component {
     } else if (this.props.drawGrid) {
       this.drawGridPlugin.startDrawing(option, this.canvas, imageRatio);
     } else if (option.target && !(option.e.shiftKey || option.e.ctrlKey)) {
+      if (option.target.type === 'activeSelection') {
+        const group = this.canvas.getActiveObject();
+        const clickPoint = new fabric.Point(option.e.offsetX, option.e.offsetY);
+
+        group.getObjects().forEach((obj) => {
+          if (!objectFound && obj.containsPoint(clickPoint) && obj.selectable) {
+            objectFound = obj;
+          }
+        });
+      } else {
+        objectFound = option.target;
+      }
+
       const shapeData = this.drawGridPlugin.setPixelsPerMM(
         this.props.pixelsPerMm,
         this.props.shapes[objectFound.id],


ui/src/components/SampleView/SampleImage.jsx
```diff
@@ -518,13 +518,21 @@ export default class SampleImage extends React.Component {
       // Important to call this for containsPoint to work properly
       // this.canvas.discardActiveObject();
 
-      group.getObjects().forEach((obj) => {
```
axelboc, line 521: The error says that `group` is `undefined`, so I feel like it would be better to investigate why that is. Somehow there's no active object on the canvas when first selecting a point while pressing `Ctrl` (and only while pressing `Ctrl`).

walesch-yan: I tested different scenarios and observed the change of the `ActiveObject` in the `canvas` element. Important to note (Unfortunately, I did not mention this in #1301 ) is that in the previous code, there was no error thrown if more than two points were selected, it only happens at exactly two selected points. Also when you use single click (without pressing `ctrl`) on two selected points, no error is thrown.

From my understanding, you can not enter the code section that I changed, when only 1 point is selected, since the type must be set to `activeSelection`. Now, we can assume that in this part of the code we should have two points that are selected at least. So when using `canvas.getActiveObject()` the returned object should contain an array of all the selected points which we access by using `getObjects` method. 

However, in the case where we hold `ctrl` while clicking on one of exactly two selected points, this section of the code is actually entered because the type is still set to `activeSelection`, but the `ActiveObject` has already been updated and contains only a single point, hence `getObjects` is undefined (important group is **not** undefined!).

I used the try catch block here, since it only happens in this specific scenario and actually we only assign `objectFound` in this section, which is never used in the remaining part of the function if `ctrl` is pressed. Hence, in our case, we could completely skip the if by checking for the key. That might be a better solution, and I mainly used the try catch to see if another error might pop up at some point.

walesch-yan: Also the type was correctly changed after the function was called. Hence if you clicked away the error message, everything worked as expected

walesch-yan: I hope this explanation makes sense and please correct me if I'm wrong :)

axelboc: Thanks for the details. I'd be interested to understand what this part of the code does feature-wise, though.

It seems that `objectFound` is only used in the last `else if` block, which seems to relate to the grid feature and more specifically to sending over the clicked "cell" index of a grid over to Braggy so it can display the corresponding image.

So the first problem is perhaps that this whole `objectFound` logic should live inside the last `else if` block?

Second red flag for me is that the last `else if` condition is fairly generic and may be triggered more often than needed; for instance, there's no check that a grid object is indeed clicked.

And third, I don't get the relevance of this "activeSelection" type. With a point and a grid on the canvas, if I press `Ctrl` and select the point, `option.target.type` is "SAVED", then if I select the grid, the type is "GridGroup". It's only if I deselect either the point or the grid that the type becomes "activeSelection". So it seems to represent a "remaining selection". If the remaining selection holds more than one object, then it's a group and the `getObjects()` method is available; otherwise if only a single object remains selected, then it's not a group but the object itself, so the `getObjects()` method is not available. So a straightforward fix would be to check if the target is a group before going through the problematic code path. However, my question is rather: how does deselecting an object relate to displaying images in Braggy? Shouldn't this "activeSelection" case be skipped altogether?

All in all, I think it'd be worth getting Marcus' insight on this, since he's the one who changed it last when upgrading `fabric`: https://github.com/mxcube/mxcubeweb/commit/be4446905ecd2bfe898b4df4c4b9095072487e2d

walesch-yan: For the first and second problem: I agree, we can actually trigger the final `else if` condition with any canvas object, while it seems that only grid objects should trigger it, maybe it would be wise to keep the `objectFound` logic from the first `if` block, but change it, so that it checks if the object is actually a grid and then use this to trigger the last `else if` condition.

Regarding the third problem, you are right that this is the behavior of the type, when we use `ctrl` and deselect objects, which is in the end the reason, why the error is triggered. However, there is a second scenario, where the type is "activeSelection". Namely when selecting an area that contains multiple canvas objects, when we now click on any selected object, the type of the target is "activeSelection" and the target itself is a group (when only one object is in the area, the type will be the type of that object, i.e. "Saved" or "GridGroup"). I assume that this case was the intention behind the "activeSelection" case logic and it would make sense to just entirely skip this part when pressing `ctrl`. From my opinion the expected behavior should not be to send anything to Braggy when pressing `ctrl`, since we do not expect this with area selection either (only when clicking on an object afterwards). 

Also, when we click on an object after selecting multiple objects with `ctrl`, the target is the object itself, and the type is therefore not "activeSelection". While this seems to solve the issue with the left click, I wonder if there are any other unexpected behaviors coming from this "remaining selection" problem, when using right click for example (I did not investigate on this). 

In general, I think you are right to wait for Marcus' insight on this, as we might miss something here.  For the time being I will change this to a draft and thanks a lot for your comments, I think they helped me also to understand the logic here better. :)

marcus-oscarsson: We have some custom way of handling the selection that is not really standard within `fabric`. We are using `ctrl` as modifier to add and remove items from a selection while `fabric` uses `shift`. When an item within a group of items are clicked in `fabric` we get the a group as target (the `activeSelection`). There might be some better ways of dealing with this in more recent versions of fabric but this is to my understanding how it used to work and this logic has remained the same.

In some version update of `fabric` the type `group` disappeared and was replaced with `activeSelection` which is something I overlooked at the time and unfortunate because it can be both a simple fabric object and a group having the `getObjects` method. 

The logic of the code is the following:

  - If target is a group then find the clicked object within that group and select it
  - If target was not a group then the target is the object we want to select

There is perhaps another way of checking if target is a group that would work. On the other hand reading the comment from @walesch-yan, perhaps this function is called twice in more recent versions of fabric one for the selection and one for the object, that would make the first piece of logic unnecessary ?

walesch-yan:
Moved the whole logic of the `activeSelection` condition into the final `else if` that way, this part of the code is protected from being entered when holding `ctrl` while clicking an object and ensures the correct object selection when using area selection before clicking.

ui/src/components/SampleView/SampleImage.jsx
walesch-yan: As clarification after a discussion with @marcus-oscarsson : the function is not called twice, however there is a scenario where the `activeSelection` can be a simple fabric object but the old code checked for the `getObjects` method. Since this scenario involves using `ctrl` for deselection. We decided to move everything inside the last `else if` as @axelboc already proposed as solution, since the deselection of an object should not call this part of the code.

marcus-oscarsson:
:+1: Nice :)

ui/src/components/SampleView/SampleImage.jsx
axelboc: Nice, looks good. For future reference, I was able to replicate the `activeSelection` scenario mentioned by selecting a point and a grid with <kbd>Ctrl</kbd>, and then clicking on the grid without holding <kbd>Ctrl</kbd>. It is disappointing that fabric doesn't seem expose all the hit targets — only the top-most group. :disappointed:

PR #1335, opened 2024-08-09
Replace Create React App with Vite

### LINKED ISSUE #1273: Replace Create React App with Vite
[Create React App](https://create-react-app.dev/docs/getting-started), through its `react-scripts` package, has not received updates since [April 2022](https://github.com/facebook/create-react-app/releases). It is time to move to another build tool and I've had good experiences so far with [Vite](https://vitejs.dev/) ... and so has the [web community](https://2023.stateofjs.com/en-US/libraries/build_tools/#build_tools_ratios).

axelboc:
Fix #1273

See [Vite's documentation](https://vitejs.dev/) to learn about this modern front-end bundler.

I've kept the build folder the same (`ui/build`) but I did change the front-end port to `5173`, which is Vite's default port. You'll note that both the dev and build are much, much faster thanks to [esbuild](https://esbuild.github.io/) — the "UI" CI workflow is more than one minute faster.


ui/package.json
```diff
@@ -50,7 +51,6 @@
     "react-zoom-pan-pinch": "3.3.0",
     "redux": "4.1.2",
     "redux-form": "^8.3.8",
-    "redux-logger": "^3.0.6",
```
axelboc, line 53: I've removed `redux-logger` (cf my later comment).

ui/src/components/App.jsx
```diff
@@ -16,7 +16,7 @@ import LoggerContainer from '../containers/LoggerContainer';
 import RemoteAccessContainer from '../containers/RemoteAccessContainer';
 import HelpContainer from '../containers/HelpContainer';
 import Main from './Main';
-import LoadingScreen from '../components/LoadingScreen/LoadingScreen';
+import LoadingScreen from './LoadingScreen/LoadingScreen';
```
axelboc, line 19: I guess this is thanks to VS Code's "organise imports" feature that ran when I renamed `App.js` to `App.jsx`.

ui/src/components/App.jsx
axelboc, line 1: With Vite, any file containing JSX must have the `.jsx` extension. I had to rename quite a few files.

ui/src/index.html
axelboc, line 1: An old `index.html` file it seems?

ui/src/index.jsx
```diff
@@ -10,14 +10,6 @@ import DefaultErrorBoundary from './containers/DefaultErrorBoundary';
...
 import '@fortawesome/fontawesome-free/css/all.min.css';
 
-if (module.hot) {
-  // Enable Webpack hot module replacement for reducers
-  module.hot.accept('./reducers', () => {
-    const nextReducer = require('./reducers');
-    store.replaceReducer(nextReducer);
-  });
-}
-
```
axelboc, line 20: Moved to `store.js`.

ui/src/setupProxy.js
axelboc, line 1: Now configured in `vite.config.js`.

ui/src/setupProxy.js
```diff
@@ -1,18 +0,0 @@
...
-    createProxyMiddleware({
-      target: 'http://127.0.0.1:8081',
-      ws: false,
-    }),
-  );
-  app.use(
-    '/socket.io/*',
-    createProxyMiddleware({
-      target: 'http://127.0.0.1:8081',
-      ws: false,
```
axelboc, line 15: Not quite sure why this was configured like this, since I used to get socket.io errors in development. I switched to `ws: true` and `target: 'ws://127.0.0.1:8081` and the error is now gone.

ui/package.json
```diff
@@ -19,6 +20,7 @@
   },
   "dependencies": {
     "@fortawesome/fontawesome-free": "^5.15.4",
+    "@redux-devtools/extension": "3.3.0",
```
axelboc, line 23: Simpler way to set up Redux devtools: https://github.com/reduxjs/redux-devtools/tree/main/extension#13-use-redux-devtoolsextension-package-from-npm

ui/src/store.js
```diff
@@ -1,26 +1,17 @@
...
 import thunk from 'redux-thunk';
 import rootReducer from './reducers';
+import { composeWithDevTools } from '@redux-devtools/extension';
 
-function initStore() {
-  // Logger MUST BE the last middleware
-  const middleware = [thunk, createLogger()];
+export const store = createStore(
+  rootReducer,
+  composeWithDevTools(applyMiddleware(thunk)),
```
axelboc: The `@redux-devtools/extension` package makes things much simpler: https://github.com/reduxjs/redux-devtools/tree/main/extension#13-use-redux-devtoolsextension-package-from-npm

ui/src/store.js
```diff
@@ -1,26 +1,17 @@
-import { createStore, applyMiddleware, compose } from 'redux';
-import { createLogger } from 'redux-logger';
+import { createStore, applyMiddleware } from 'redux';
 import thunk from 'redux-thunk';
 import rootReducer from './reducers';
+import { composeWithDevTools } from '@redux-devtools/extension';
 
-function initStore() {
-  // Logger MUST BE the last middleware
-  const middleware = [thunk, createLogger()];
```
axelboc, line 8: I've decided to remove `redux-logger`, which clutters the console (potentially hiding important warnings/errors), and to instead enable `redux-devtools` in production (i.e. not just in development).

So if you need to debug MXCuBE on a beamline, just install the Redux devtools extension: https://github.com/reduxjs/redux-devtools/tree/main/extension#installation. It's much easier to debug the Redux state with this extension than by trying to read the `redux-logger` logs.

Note that I've also enabled sourcemaps generation in `vite.config.js`, which should also make production debugging easier.

ui/vite.config.js
```diff
@@ -0,0 +1,23 @@
...
+  plugins: [
+    react(),
+    { ...eslintPlugin(), apply: 'serve' }, // dev only to reduce build time
+  ],
+  server: {
+    open: true, // open default browser on start
+    strictPort: true, // fail if port already in use (must be 5173 to match server's CORS config)
+    proxy: {
+      '/mxcube/api': 'http://127.0.0.1:8081',
+      '/socket.io/': { target: 'ws://127.0.0.1:8081', ws: true },
```
axelboc, line 15: The new proxy settings.

ui/cypress.config.js
axelboc, line 1: I've added `"type": "module"` to `package.json`, which means that every `.js` file is now treated as [ESM instead of CommonJS](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules). In ESM, modules must be `import`ed instead of `require`d, which is why I'm changing the module syntax in this file (and others). The alternative would have been to rename the file with the `.cjs` extension.

ui/index.html
axelboc, line 1: With Vite, `index.html` becomes the [entry point](https://vitejs.dev/guide/#index-html-and-project-root) (instead of `index.js`). This is why it's at the root of the `ui` folder. There's also no need for `%PUBLIC_URL%` placeholders.

ui/less-watcher.config.json
axelboc, line 1: Obsolete config file for the LESS styles preprocessor.

ui/public/favicon.ico
axelboc, line 1: Removing images of the React logo from the CRA boilerplate.

axelboc:
The three-click centring E2E test still seems to be failing after mounting the sample (the "Current" tab doesn't change to "Sample: test - test"). I have a feeling it's linked to the fact that socket.io switches to `polling` transport mode during Cypress tests for some reason I can't figure out...

I've skipped it for now, since it seems to work outside of Cypress.

ui/cypress/e2e/sampleControls.cy.js
```diff
@@ -15,8 +15,11 @@ describe('3-click centring', () => {
     );
   });
 
-  it('Each click is rotating the sample by 90 degrees', () => {
+  it.skip('Each click is rotating the sample by 90 degrees', () => {
```
axelboc, line 18: Skipping this test for now, as explained, but if you have any clue @walesch-yan, please let me know :pray:

ui/src/store.js
fabcor-maxiv: I am not sure we can install browser extensions at the beamline. I mean there is always ways around, but I doubt it is a case of "just install". Maybe we will never even need this, so not sure it is worth reconsidering this now.

ui/index.html
fabcor-maxiv: This is surprising to me that it is not `ui/src/index.html`. Is that really `ui/index.html`?

fabcor-maxiv (commented):
I always give thumbs up for cleaning up, removing things, and modernizing : )

ui/cypress/e2e/sampleControls.cy.js
walesch-yan: Unfortunately, I have no direct answer to this behavior. My guess would be that some configuration mismatch on cypress' side leads to networking issues. This definitely needs to be investigated in more detail.

ui/index.html
axelboc: yep :D

ui/src/store.js
axelboc: I could also keep `redux-logger` but make it an opt in via an environment variable for instance.

fabcor-maxiv: Personally, I trust your judgement and I rarely used redux-logger anyway.

I don't know about the others, maybe they rely on redux-logger, maybe it is worth asking (on the mailing list?) if they mind the migration to redux-devtools or not.

meguiraun: I usually rely on `redux-logger` for debugging, this removal will make my life a bit worse 😔. We cannot install extensions on the beamline as we don't have internet access

elmjag: I vote for making `redux-logger` an opt in via an environment variable. It _is_ crazy spammy, would be nice not to look at that spam all the time. But, an easy way to enable it would be certainly very useful for us at MAXIV.

If you add such a variable, please add documentation for it at some good place here: https://mxcubeweb.readthedocs.io/en/latest/dev/index.html

marcus-oscarsson:
Very nice thanks for the excellent work :)

I don't mind using a browser extension for the redux store debug. I guess its possible to add a variable for it if its absolutely necessary, @meguiraun and @elmjag ? Maybe in that case it would be nice to have the option to have no logging at all. But a question out of curiosity, how do you do to update browser version and so on if you don't have access to internet ?

I'm happy to merge this :), there is a conflict from a previously merged PR, maybe you could have a look @axelboc

fabcor-maxiv:
> how do you do to update browser version and so on if you don't have access to internet ?

As far as I understand we have our own proxies/mirrors for the system package repositories of the Linux distribution.

marcus-oscarsson:
> > how do you do to update browser version and so on if you don't have access to internet ?
> 
> As far as I understand we have our own proxies/mirrors for the system package repositories of the Linux distribution.

I see, 

I'll go a head and merge this later today and we can sort out any remaining issues with how to handle the redux logging in a future PR, let me know if you think this approach is too progressive or if you have other concerns.

axelboc:
I've added an environment variable called `VITE_REDUX_LOGGER_ENABLED` and documented how to go about overriding it: 6f678d7e5162fc57d4c6508d73ec022a45dae30b

ui/.env
```diff
@@ -0,0 +1,8 @@
+#####
+# UI ENVIRONMENT VARIABLEs
+# Copy and configure the variables below in your own `.env.local` or `.env.[mode].local` file
+# https://vitejs.dev/guide/env-and-mode.html#env-files
+#####
+
+# Redux logger ('true' to enable)
+VITE_REDUX_LOGGER_ENABLED=
```
marcus-oscarsson, line 8: Would this mean that its undefined by default or what does leaving out the right hand side of the `=` result in ?

marcus-oscarsson:
Thanks !

fabcor-maxiv:
I am thinking maybe this needs some documentation...

If things have changed for developer experience in a non-backwards-compatible way -- for example the change of port number, removal of redux-logger (assuming we go through with this part) -- then it might be worth an announcement on the mailing list.

And then it might be worth documenting (in `/docs`) that we have redux-devtools and that it needs a browser extension (again assuming we go through with this).

marcus-oscarsson:
@fabcor-maxiv: The last commit also added the possibility to use `redux-logger` and it was documented as well

https://github.com/mxcube/mxcubeweb/pull/1335/files#diff-425da2324e83ab4adea4825089d26a6b2b0d0dc15cbd54a452499c1f4445909f

The missing part is perhaps the port change from 3000 to 5173 ?

fabcor-maxiv:
> The last commit also added the possibility to use `redux-logger` and it was documented as well

Oh, good, I had not seen this had been added : )

~I am not sure, though, can I use this `VITE_REDUX_LOGGER_ENABLED` environment variable at run-time or is it only for build-time (and hot-reload server)?~ Well, it can't possibly be at run-time, can it?

ui/.env
axelboc: It ends up as an empty string. Environment variables are just strings, so in the code, I test for the string `"true"`.

axelboc:
> Well, it can't possibly be at run-time, can it?

Environment variables are read when the dev server starts (`pnpm --prefix ui start`) and the project is built (`pnpm --prefix ui build`).

axelboc:
> If things have changed for developer experience in a non-backwards-compatible way -- for example the change of port number, removal of redux-logger (assuming we go through with this part) -- then it might be worth an announcement on the mailing list.

I can revert the port change if it's easier?

axelboc:
> And then it might be worth documenting (in /docs) that we have redux-devtools and that it needs a browser extension (again assuming we go through with this).

Good idea, I can add a section about the devtools.

fabcor-maxiv:
> > If things have changed for developer experience in a non-backwards-compatible way -- for example the change of port number, removal of redux-logger (assuming we go through with this part) -- then it might be worth an announcement on the mailing list.
> 
> I can revert the port change if it's easier?

I'd rather we leave things to their default values. But maybe some people will be surprised by the change. That is why I think an announcement on the dev mailing list would make sense.

And/or there could be a line in our documentation explaining how to set the port to 3000. But I assume it would be no more than a copy-paste of Vite's documentation and a link back to Vite's documentation, which I find slightly pointless to have (and maintain) in our own documentation pages. So that would not be my preferred solution.

---

Independent from all that, can we have a paragraph in <https://mxcubeweb.readthedocs.io/en/latest/dev/front-end.html> that says that we have Vite as the builder (with link to their home page)?

axelboc:
> Independent from all that, can we have a paragraph in https://mxcubeweb.readthedocs.io/en/latest/dev/front-end.html that says that we have Vite as the builder (with link to their home page)?

Yep, working on it :+1:

axelboc:
I took the opportunity to expand the front-end documentation: https://github.com/mxcube/mxcubeweb/pull/1335/files#diff-425da2324e83ab4adea4825089d26a6b2b0d0dc15cbd54a452499c1f4445909f

marcus-oscarsson:
Nice, also good with the best practices for functional components and hooks connecting to the store :+1:

docs/source/dev/front-end.md
```diff
@@ -1,9 +1,76 @@
...
+- `pnpm preview` - serve the production build at http://localhost:5173
+
+The codebase is formatted with Prettier, linted with ESLint and tested with Cypress:
+
+- `pnpm prettier` - check formatting
+- `pnpm format` - fix formatting
+- `pnpm eslint` - check for linting errors
+- `pnpm fix` - fix linting errors
+- `pnpm cypress` - open Cypress testing UI
+- `pnpm e2e` - run Cypress E2E tests
```
fabcor-maxiv, line 31: @axelboc Just noticed... Shouldn't these be `pnpm run <something>`?

fabcor-maxiv: OK, apparently not... I am surprised

axelboc: Yeah, very convenient. Works with any custom script that is not a known `pnpm` command.

PR #1337, opened 2024-08-13
Fix two React errors

axelboc:
Fixing two React errors that became more visible with `redux-logger` disabled:

1. Missing `key` in `SampleChanger`
2. Checkbox going from uncontrolled to controlled in `SampleListViewContainer` when syncing with LIMS (`checked` prop going from `undefined` to defined due to a mismatch between the component code and the Redux store).

### DIFF
--- a/ui/src/reducers/sampleGrid.js
+++ b/ui/src/reducers/sampleGrid.js
@@ -32,8 +32,7 @@ const INITIAL_STATE = {
     notCollected: false,
     cellFilter: '',
     puckFilter: '',
-    limsFilter: false,
-    useFilter: false,
+    limsSamples: false,
   },
 };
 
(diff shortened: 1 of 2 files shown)


ui/src/reducers/sampleGrid.js
```diff
@@ -32,8 +32,7 @@ const INITIAL_STATE = {
     notCollected: false,
     cellFilter: '',
     puckFilter: '',
-    limsFilter: false,
-    useFilter: false,
+    limsSamples: false,
```
axelboc, line 35: `useFilter` seems obsolete. I couldn't find a reference to it in the codebase.

marcus-oscarsson:
:+1:

PR #1338, opened 2024-08-13
Make checkboxes clickable

### LINKED ISSUE #1279: Checkbox labels should be clickable in RA Options
image

Also, random question: what does "Enable remote access" actually do? Because I can still give/take control even when it's unchecked. Am I missing something? (I run the mock beamline.)

axelboc:
Fix #1279 

Bootstrap's `<Form.Check>` with a `label` prop requires an `id` for the label to be properly associated with the checkbox and become clickable. This is now fixed for checkboxes in the remote access options and queue settings .

<details>
<summary>Reverted changes</summary>

On `localhost` or `127.0.0.1`, I match the server's behaviour (cf. `usermanager.py`, [line 390](https://github.com/mxcube/mxcubeweb/blob/develop/mxcubeweb/core/components/user/usermanager.py#L390)) by forcing the checkbox to be ticked and by disabling it so it can't be unticked:

image

For completeness, I also change the `/ra/allow_remote` endpoint to respond with a 400 when the client is on localhost (whether trying to enable to disable remote access; either way it doesn't make sense to call the endpoint since the setting is ignored).

</details>

### DIFF
--- a/ui/src/containers/RemoteAccessContainer.jsx
+++ b/ui/src/containers/RemoteAccessContainer.jsx
@@ -1,81 +1,57 @@
 import React from 'react';
-import { connect } from 'react-redux';
-import { bindActionCreators } from 'redux';
-
 import { Container, Card, Form, Row, Col } from 'react-bootstrap';
-
-import RequestControlForm from '../components/RemoteAccess/RequestControlForm';
-import UserList from '../components/RemoteAccess/UserList';
+import { useDispatch, useSelector } from 'react-redux';
 
 import {
   updateAllowRemote,
   updateTimeoutGivesControl,
 } from '../actions/remoteAccess';
+import RequestControlForm from '../components/RemoteAccess/RequestControlForm';
+import UserList from '../components/RemoteAccess/UserList';
 
-export class RemoteAccessContainer extends React.Component {
-  getRAOptions() {
-    return (
-      <Col sm={4}>
-        <Card className="mb-3">
-          <Card.Header>RA Options</Card.Header>
-          <Card.Body>
-            <Form.Check
-              type="checkbox"
-              onChange={(e) => this.props.updateAllowRemote(e.target.checked)}
-              checked={this.props.remoteAccess.allowRemote}
-              label="Enable remote access"
-            />
-            <Form.Check
-              type="checkbox"
-              onChange={(e) =>
-                this.props.updateTimeoutGivesControl(e.target.checked)
-              }
-              checked={this.props.remoteAccess.timeoutGivesControl}
-              label="Timeout gives control"
-            />
-          </Card.Body>
-        </Card>
-      </Col>
-    );
-  }
+function RemoteAccessContainer() {
+  const dispatch = useDispatch();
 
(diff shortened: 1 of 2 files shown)


ui/src/containers/RemoteAccessContainer.jsx
```diff
@@ -1,81 +1,60 @@
...
+        <Col sm={4}>
+          <Card className="mb-3">
+            <Card.Header>Options</Card.Header>
+            <Card.Body>
+              <Form.Check
+                type="checkbox"
+                onChange={(e) => dispatch(updateAllowRemote(e.target.checked))}
+                checked={remoteAccess.allowRemote || isLocalhost}
+                label="Enable remote access"
+                id="allow-remote"
```
axelboc, line 38: Passing an `id` is what fixed the label interactivity issue.

ui/src/containers/RemoteAccessContainer.jsx
```diff
@@ -1,81 +1,60 @@
...
-              }
-              checked={this.props.remoteAccess.timeoutGivesControl}
-              label="Timeout gives control"
-            />
-          </Card.Body>
-        </Card>
-      </Col>
-    );
-  }
+function RemoteAccessContainer() {
```
axelboc, line 12: I've converted the component to a function, so I recommend hiding whitespace changes to review the diff.

marcus-oscarsson:
However, we should not disable the checkbox, it should still be possible to deactivate the remote login feature when logged in from localhost.  [line 390](https://github.com/mxcube/mxcubeweb/blob/develop/mxcubeweb/core/components/user/usermanager.py#L390) makes sure that login can only be done from localhost if remote access is disabled.

axelboc:
If the remote access checkbox is disabled, I don't expect to be able to log in from another browser as an observer, even on `localhost`. Perhaps `localhost` shouldn't be a special case then, and one should always enable remote access manually?

I'll revert this change for now, as it was outside of the scope of the issue anyway.

axelboc:
I've reverted the localhost-related changes and while I was at it, I noticed that the checkboxes in the queue settings were also missing IDs, so I've added them.

marcus-oscarsson:
Thanks, I understand that the remote access behavior is a bit confusing. 

Its always possible to login, any number of times, as a local user (from localhost) regardless of the RA (remote access) option. Remote access is the possibility for users from other hosts to connect to the application. The server can be started with the option on or off but a local user (localhost) can always decide to toggle it if needed.

However you are right that it does not make any sense for a remote user to be able to to toggle this option on or off. Ive not checked recently but the RA options should not be shown for remote users.

axelboc:
> Its always possible to login, any number of times, as a local user (from localhost) regardless of the RA (remote access) option. 

Riiight, thanks, it finally clicked in my brain. 😅 For some reason, I made an equivalence between remote users and observers, but this is not the case.

marcus-oscarsson:
Oh yeah, the observers also makes things even more confusing its somewhat a very unconventional web-application in this sense

PR #1339, opened 2024-08-13
Fix  clear Filter for lims Samples

### LINKED ISSUE #1334: [Sample List] Clear Button / Filter  does not work on ISPY FIlter
When Apply Filter on ISPYB Samples the Clear Filter Action does not work.

Filter

I guess it's because, in the pass the ISPYB Samples was not part of the Filtered List

jbflo:
This fix https://github.com/mxcube/mxcubeweb/issues/1334


axelboc (approved):
Nice, sorry I missed that in #1337

PR #1340, opened 2024-08-14
fix wrong json format for automount and autoadddiffplan

meguiraun:
I was getting:

```
415 Unsupported Media Type: Did not attempt to load JSON data because the request Content-Type was not 'application/json'.
```

When setting "Automount next sample" and "Auto add diff plan" in the queue settings

### DIFF
--- a/mxcubeweb/routes/queue.py
+++ b/mxcubeweb/routes/queue.py
@@ -325,11 +325,10 @@ def init_route(app, server, url_prefix):  # noqa: C901
     @server.require_control
     @server.restrict
     def set_autmount():
-        automount = request.get_json()
-        app.queue.set_auto_mount_sample(automount)
-        resp = jsonify({"automount": automount})
+        data = request.get_json()
+        app.queue.set_auto_mount_sample(data.get("automount", False))
+        resp = jsonify(data)
         resp.status_code = 200
-
         return resp
 
     @bp.route("/num_snapshots", methods=["PUT"])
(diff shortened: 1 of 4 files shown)


marcus-oscarsson:
Ok, the tests are failing, it looks like a simple update

meguiraun:
Fixed py-tests!
but I still get: "Cypress failed to make a connection to Firefox.", I have re-run the job several with same result, no idea what else to try

fabcor-maxiv:
> but I still get: "Cypress failed to make a connection to Firefox.", I have re-run the job several with same result, no idea what else to try

It will pass eventually. I do not know if there are ways to improve this situation. Maybe @walesch-yan, @axelboc have some suggestions.

axelboc:
Cypress doesn't actually install Firefox, so I guess my first stupid question would be: is Firefox available in the CI? Does it get installed by Mamba? Perhaps somehow it gets excluded from the Conda cache so it's not available when the cache is restored?

If Firefox gets properly installed/cached, then perhaps it's a bug with a specific version of Cypress and/or Firefox and upgrading might fix it. I can take a look on Friday maybe or next week.

axelboc:
In the meantime, maybe we should remove the `--firefox` argument when running `cypress run` so that the tests run in Electron (which comes bundled with Cypress).

fabcor-maxiv:
@axelboc @walesch-yan I opened a dedicated ticket to discuss this: <https://github.com/mxcube/mxcubeweb/issues/1341>.

marcus-oscarsson:
I noticed that the last time the E2E tests worked (about 3h ago) the Firefox version was 128 now its 129. Its perhaps some update that is causing this issue.

marcus-oscarsson:
@meguiraun I think the E2E tests will pass if you rebase on develop now ?

walesch-yan:
Update: There seems to be an issue with firefox 129 and our old version of cypress, the version upgrade has been merged and the tests here should pass after a rebasing the branch

marcus-oscarsson:
@meguiraun, its a bank holiday in France tomorrow. But as always when a PR is reviewed and approved feel free to rebase and merge

marcus-oscarsson (approved):

PR #1342, opened 2024-08-14
Using import.meta instead of process.env

marcus-oscarsson:
The recent change to Vite change the way the environment variables are accessed, so using `process.env` will result in an error.

We are not using the `helpLinks` variable, I'm not sure who is but I think its perhaps you guys at MAX-IV (@meguiraun, @fabcor-maxiv, @elmjag) ?

### DIFF
--- a/ui/src/containers/HelpContainer.jsx
+++ b/ui/src/containers/HelpContainer.jsx
@@ -57,8 +57,8 @@ export class HelpContainer extends React.Component {
   render() {
     let links = [];
 
-    if (process.env.helpLinks) {
-      links = process.env.helpLinks.map((link) => (
+    if (import.meta.env.VITE_HELP_LINKS) {
+      links = import.meta.env.VITE_HELP_LINKS.map((link) => (
         // eslint-disable-next-line react/jsx-key
         <div>
           <a target="_blank" href={link.url} rel="noreferrer">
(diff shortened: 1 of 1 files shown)


axelboc:
Oh yeah, I noticed it yesterday. I have no idea how `process.env` could end up with a `helpLinks` key. I would have expected `process.env.REACT_APP_HELP_LINKS`, which would then convert to `import.meta.env.VITE_HELP_LINKS` with Vite.

Either way, I think it's worth rethinking whether this should be moved to an optional server-side configuration option.

marcus-oscarsson:
@axelboc Yes you are right that I should have changed it to `process.env.REACT_APP_HELP_LINKS`. This was introduced before the REACT_APP prefix was introduced. If its still used I would also vote for moving it to server side configuration.

marcus-oscarsson:
Ill change it to `process.env.REACT_APP_HELP_LINKS` for now and then we will see exactly what to do

marcus-oscarsson:
Yeah, so I changed it to `process.env.VITE_HELP_LINKS` and not `process.env.REACT_APP_HELP_LINKS`

marcus-oscarsson:
Ill merge this even if it probably will be change within shortly, It at least maintains comparability with the old behavior

PR #1344, opened 2024-08-14
upgrade cypress to version 13.13.2

walesch-yan:
This version upgrade should solve issues between cypress and firefox 129


marcus-oscarsson:
Thanks !

PR #1346, opened 2024-08-16
Improve wait dialog

### LINKED ISSUE #1281: "Asking for control" modal should not be dismissable
The backdrop shouldn't be clickable — the modal should be dismissed only by clicking _Cancel_. Otherwise we end up in a weird state as shown below and the person we're asking to hand over control is not notified.

Peek 2024-07-02 16-48

axelboc:
I've made separate commits to fix and refactor multiple aspects of the `PleaseWaitDialog` component and corresponding global state:

1. I fix #1281 by preventing "blocking" modals from being dismissed when their backdrop is clicked. This is done by passing `backdrop="static"` to the Bootstrap modal. In this commit, I also fix keyboard accessibility by notably allowing the modal to be dismissed with <kbd>Escape</kbd> (unless `backdrop="static"` is set). 8720b00c2312db4725fe0f109e22bf3840cfb886
2. I install a few dev dependencies to fix auto-imports, linting, etc., notably in VS Code. 01312af1d678cb418e5a0ec9e0762a099f533987
3. I turn the `PleaseWaitDialog` component into a function component. 12ce793d5793f9da8d09c01ab3d0d67c26daad3f
4. I inline the loading state of the `Login` component (which controls the visibility of the little spinner in the "Sign in" button) so that the `general.loading` state in the Redux store is only used by the wait dialog (in preparation for my fifth commit). 38aaa4c60f73bd98feb7a8cb92945d9397a2a3f7
5. I isolate the wait dialog's state into its own Redux state slice, along with its own reducer and actions. This ensures that I can select the whole `waitDialog` slice without triggering unrelated re-renders. 98077c4dbe4a1405d820981080673168f033bb2e In this commit, I also:
    - replace the generic `setLoading` action with two more intuitive actions: `showWaitDialog` and `hideWaitDialog`;
    - make sure not to reset the `title` and `message` in `hideWaitDialog` to avoid a "Please wait" flash while the dialog is being animated out;
    - hide the entire body of the dialog if there's no message or progress bar to show.

### DIFF
--- a/ui/src/reducers/waitDialog.js
+++ b/ui/src/reducers/waitDialog.js
@@ -1,0 +1,35 @@
+const INITIAL_STATE = {
+  show: false,
+  title: undefined,
+  message: undefined,
+  blocking: false,
+  abortFun: undefined, // not serializable! https://redux.js.org/style-guide/#do-not-put-non-serializable-values-in-state-or-actions
+};
+
+function waitDialogReducer(state = INITIAL_STATE, action = {}) {
+  switch (action.type) {
+    case 'SHOW_WAIT_DIALOG': {
+      return {
+        ...state,
+        show: true,
+        title: action.title,
+        message: action.message,
+        blocking: action.blocking,
+        abortFun: action.abortFun,
+      };
+    }
+    case 'HIDE_WAIT_DIALOG': {
+      return {
+        ...state, // keep title and message while dialog is hiding
+        show: false,
+        blocking: false,
+        abortFun: undefined,
+      };
+    }
+    default: {
+      return state;
+    }
+  }
+}
+
+export default waitDialogReducer;
--- a/ui/src/serverIO.js
+++ b/ui/src/serverIO.js
@@ -37,7 +37,7 @@ import {
   getQueue,
 } from './actions/queue';
 import { collapseItem, showResumeQueueDialog } from './actions/queueGUI';
-import { setLoading, showConnectionLostDialog } from './actions/general';
+import { showConnectionLostDialog } from './actions/general';
 
 import {
   showWorkflowParametersDialog,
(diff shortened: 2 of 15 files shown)


ui/src/reducers/waitDialog.js
```diff
@@ -0,0 +1,35 @@
+const INITIAL_STATE = {
+  show: false,
+  title: undefined,
+  message: undefined,
+  blocking: false,
+  abortFun: undefined, // not serializable! https://redux.js.org/style-guide/#do-not-put-non-serializable-values-in-state-or-actions
```
axelboc, line 6: In theory, the Redux state should always be [serialisable](https://redux.js.org/style-guide/#do-not-put-non-serializable-values-in-state-or-actions), so keeping a function in there isn't great. A problem for another day, though.

ui/src/serverIO.js
```diff
@@ -329,22 +327,12 @@ class ServerIO {
         }
 
         case 'loadReady': {
-          this.dispatch(
-            setLoading(false, 'SC Ready', record.message, true, () =>
-              this.dispatch(stopQueue()),
-            ),
-          );
-
+          this.dispatch(hideWaitDialog());
```
axelboc, line 330: Not sure about this to be honest. The purpose of the previous code is not clear: it hides the wait modal while still giving it a title, message and abort function. I traced it back to [here](https://github.com/mxcube/mxcubeweb/pull/597/files#diff-3ce702edf4adc9b09256a43ff92b1001d01e8ca66e6d547a52a587f4a5ffb580R175).

ui/src/serverIO.js
```diff
@@ -329,22 +327,12 @@ class ServerIO {
...
         }
 
         case 'inSafeArea': {
-          this.dispatch(
-            setLoading(false, 'SC Safe', record.message, true, () =>
-              this.dispatch(stopQueue()),
-            ),
-          );
-
+          this.dispatch(hideWaitDialog());
```
axelboc, line 335: Same same, but this line was introduced [here](https://github.com/mxcube/mxcubeweb/pull/807/files#diff-3ce702edf4adc9b09256a43ff92b1001d01e8ca66e6d547a52a587f4a5ffb580R218).

marcus-oscarsson:
Thanks :+1:

PR #1347, opened 2024-08-16
Fix login redirect

axelboc:
The login redirect was broken - e.g. when logging out while on `/samplegrid` and then logging back in, one expects to land back on `/samplegrid` instead of the default `/datacollection` screen.

We were still setting the location to redirect to in `PrivateOutlet`: `<Navigate to="/login" state={{ from: location }} replace />` — but for some reason we were not actually redirecting to it in `LoginContainer`.

As always, while I'm at it, I do a bit of refactoring on the components I interact with. In this case, I move `PrivateOutlet` into a dedicated file, I rename `Login` to `LoginForm`, I refactor `App`, `LoginContainer`, `LoginForm` and `SelectProposalContainer` to use the Redux and React Router hooks APIs (instead of `connect` and `withRouter`).


marcus-oscarsson:
Really nice, wonderful ! :100:

PR #1351, opened 2024-08-19
Draft: automatically load&sync samples upon login

meguiraun:
And now a bit of controversy ;)

We would like to populate the samples list and sync with lims automatically (save the user two clicks). Also, if a sample is already loaded in the md3, the todo queue would also be populated.

opinions?

### DIFF
--- a/ui/src/actions/sampleGrid.js
+++ b/ui/src/actions/sampleGrid.js
@@ -103,6 +103,20 @@ export function syncSamples() {
   };
 }
 
+/**
+ * Sends a request to get the sample list and then synchronizes the samples with lims.
+ */
+export function sendGetandSyncSamples() {
+  return async (dispatch) => {
+    try {
+      await dispatch(sendGetSampleList());
+      await dispatch(syncSamples());
+    } catch {
+      dispatch(showErrorPanel(true, 'Could not get and sync samples list'));
+    }
+  };
+}
+
 // update list crystal from crims
 export function updateCrystalList(crystalList) {
   return { type: 'UPDATE_CRYSTAL_LIST', crystalList };
--- a/ui/src/actions/login.js
+++ b/ui/src/actions/login.js
@@ -14,6 +14,7 @@ import { fetchHarvesterInitialState } from '../api/harvester';
 import { fetchImageData, fetchShapes } from '../api/sampleview';
 import { fetchRemoteAccessState } from '../api/remoteAccess';
 import { sendSelectProposal } from '../api/lims';
+import { sendGetandSyncSamples } from './sampleGrid';
 
 export function setLoginInfo(loginInfo) {
   return {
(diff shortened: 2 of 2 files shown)


ui/src/actions/sampleGrid.js
```diff
@@ -103,6 +103,20 @@ export function syncSamples() {
   };
 }
 
+/**
+ * Sends a request to get the sample list and then synchronizes the samples with lims.
+ */
+export function sendGetandSyncSamples() {
```
axelboc, line 109: `send` prefix is now [reserved](https://mxcubeweb.readthedocs.io/en/latest/dev/front-end.html#fetching-layer) for API functions (in the `api` folder).

Since `sendGetandSyncSamples` is an action creator, you can just call it `getAndSyncSamples`.

Actually, it looks like that there are still a few action creators with the `send` prefix (including `sendGetSampleList`). I'll take care of it asap.

marcus-oscarsson:
I think synchronizing the sample changer with LIMS can be done entirely on the backend and I think there already something for it in place, that can be extended: 

https://github.com/mxcube/mxcubeweb/blob/d22f4f9905fb69a20222aca2ed29cd1ad3432744/mxcubeweb/core/components/user/usermanager.py#L168

I have a vague memory of that there used to be some logic that also did the synchronization of LIMS and SC at login, but I'm not entirely certain. 

I would not be in favor of populating the "TODO" queue unless its an option that the user selects somewhere

It would be neat with "at_login" signal or callback so that the logic below can be handled  in some other place
https://github.com/mxcube/mxcubeweb/blob/d22f4f9905fb69a20222aca2ed29cd1ad3432744/mxcubeweb/core/components/user/usermanager.py#L168

For example instead of the above do something like:
```
 self.handle_at_login()
```

Like this it would also be easier to customize this behavior as the which `UserManager` class to use is configurable

meguiraun:
mmm, I will look at doing this in the backend. Definitely easier if we want to customize this behavior.

Note that populating the "TODO" is happening already if one of the samples is loaded. Or maybe we are talking abut different things (?)

marcus-oscarsson:
>Note that populating the "TODO" is happening already if one of the samples is loaded. Or maybe we are talking abut different things (?)

Oh yes, It would be the expected behavior that the currently mounted sample (or loaded) is also the current sample. However we don't know what the user intends to do or how they intend to work so also adding/enqueuing the other samples would be a bit strange from a usability/workflow point of view. 

I can imagine that one would like some sort of "automatic" mode where the users synchronizes with LIMS and sample changer and then enqueue the samples that have a collection plan/strategy associated with them. That however should be a option somehow

meguiraun:
Ah, I see what you meant now. I was only referring to the currently loaded sample, no additional samples

meguiraun:
I have a "MAXIVUserManager*" class that does what we need (only changing `login` method), but I want to investigate a bit more and more importantly let our staff play with it. Closing this MR for now.

*I hate naming it like that... AutoLimsSyncUserManager???

PR #1348, opened 2024-08-16
Notify requester when denying request for control

axelboc:
Draft until #1346 is merged.

Fix #1280. I also put a bit of colour on the buttons to approve/deny the request.

Peek 2024-08-16 16-17

PR #1352, opened 2024-08-19
Refactor remote access components

axelboc:
Before doing what I mentioned in https://github.com/mxcube/mxcubeweb/pull/1348#issuecomment-2296489048, I start by: refactoring all the remote access components to function components; adding an endpoint to cancel control requests (instead of handling that in `/request_control`); and improving a couple more things in the UI as follows:

1. Show _Not provided_ as a fallback while an observer is entering their name:

| Before | After |
| --- | --- |
| image | image |

2. Fix `UserList` layout when in control

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/ui/src/actions/remoteAccess.js
+++ b/ui/src/actions/remoteAccess.js
@@ -1,5 +1,6 @@
 import {
   fetchRemoteAccessState,
+  sendCancelControlRequest,
   sendGiveControl,
   sendLogoutUser,
   sendRequestControl,
@@ -30,13 +31,12 @@ export function updateNickname(name) {
   };
 }
 
-export function requestControl(
-  control = true,
-  message = '',
-  name = '',
-  userInfo = {},
-) {
-  return () => sendRequestControl(control, message, name, userInfo);
+export function requestControl(message) {
+  return () => sendRequestControl(message);
+}
+
+export function cancelControlRequest() {
+  return () => sendCancelControlRequest();
 }
 
 export function takeControl() {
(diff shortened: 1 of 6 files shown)


ui/src/actions/remoteAccess.js
```diff
@@ -30,13 +31,12 @@ export function updateNickname(name) {
...
 
-export function requestControl(
-  control = true,
-  message = '',
-  name = '',
-  userInfo = {},
-) {
-  return () => sendRequestControl(control, message, name, userInfo);
+export function requestControl(message) {
+  return () => sendRequestControl(message);
```
axelboc, line 35: `message`, `name`, `userInfo` are not actually used in the back-end. I keep `message` for now, since the goal is to eventually show the message to the user in control.

ui/src/components/RemoteAccess/PassControlDialog.jsx
```diff
@@ -1,102 +1,62 @@
...
             defaultValue="Here you go !"
             type="textarea"
             placeholder="Message"
             rows="3"
           />
           <br />
-          <Button size="sm" variant="outline-secondary" onClick={this.accept}>
-            {' '}
-            Give control to "{observer.nickname}"{' '}
+          <Button type="submit" name="allow" variant="success">
```
axelboc, line 50: Note that I've added a `Form` element with an `onSubmit` handler. The trick here is that there are two submit buttons that both need access to the form's content. I'm assigning a `name` to each button so I can find which one is used with `evt.nativeElement.submitter.name` in the form's submit handler.

marcus-oscarsson:
Nice

PR #1353, opened 2024-08-19
fixed: motors do not correctly abort

walesch-yan:
These changes ensure that motor stop is correctly called by clicking the abort button in the UI.


marcus-oscarsson:
Thanks ;)

PR #1354, opened 2024-08-20
Refactor remote access messaging

axelboc:
The `observersChanged` websocket message had two responsibilities; it used to let all logged in clients know that:

1. the `observers` state has changed (e.g. because of a change of user in control; because an observer is currently requesting control; ...)
2. the `user` state _may_ have changed (e.g. because _this_ user gained/lost control or is requesting control)

Turns out that notifying all clients that their `user` state may have changed is unnecessary—we only need to notify the client(s) whose `user` state actually changed.

So in this PR, I follow the single responsibility principle by splitting the `observersChanged` message in two and rethinking when the resulting messages, `observersChanged` and `userChanged`, actually need to be emitted (and to which clients).

I also refactor how those messages are handled in the front-end:

- Now each message forces a refetch of its corresponding slice of state independently (`remoteAccess` and `login`).
- The `userChanged` message is now the one responsible for showing the correct "in control" dialog ("You were given control", or "You lost control") and this logic is now based solely on the toggling of the `login.user.inControl` state (instead of relying on the `userChanged` message payload).

### DIFF
--- a/mxcubeweb/core/components/user/usermanager.py
+++ b/mxcubeweb/core/components/user/usermanager.py
@@ -72,17 +72,6 @@ class BaseUserManager(ComponentBase):
 
         return user
 
-    def emit_observers_changed(self, message=""):
-        operator = self.app.usermanager.get_operator()
-
-        data = {
-            "observers": [_u.todict() for _u in self.app.usermanager.get_observers()],
-            "message": message,
-            "operator": operator.todict() if operator else {},
-        }
-
-        self.app.server.emit("observersChanged", data, namespace="/hwr")
-
     def update_active_users(self):
         for _u in User.query.all():
             if (
@@ -95,8 +84,11 @@ class BaseUserManager(ComponentBase):
                     f"Logged out inactive user {_u.username}"
                 )
                 self.app.server.user_datastore.deactivate_user(_u)
+                self.app.server.emit(
+                    "userChanged", room=_u.socketio_session_id, namespace="/hwr"
+                )
 
-        self.emit_observers_changed()
+        self.app.server.emit("observersChanged", namespace="/hwr")
 
     def update_operator(self, new_login=False):
         active_in_control = False
@@ -202,7 +194,7 @@ class BaseUserManager(ComponentBase):
         self.app.server.user_datastore.deactivate_user(user)
         flask_security.logout_user()
 
-        self.emit_observers_changed()
+        self.app.server.emit("observersChanged", namespace="/hwr")
 
     def is_authenticated(self):
         return current_user.is_authenticated()
--- a/mxcubeweb/routes/ra.py
+++ b/mxcubeweb/routes/ra.py
@@ -43,15 +43,15 @@ def init_route(app, server, url_prefix):  # noqa: C901
                 return make_response(msg, 409)
 
         current_user.requests_control = True
-        server.user_datastore.commit()
+        app.usermanager.update_user(current_user)
+
+        server.emit("observersChanged", namespace="/hwr")
 
         gevent.spawn(
             handle_timeout_gives_control,
             current_user.username,
             timeout=10,
         )
-
-        app.usermanager.emit_observers_changed()
 
         return make_response("", 200)
 
@@ -60,8 +60,10 @@ def init_route(app, server, url_prefix):  # noqa: C901
     def cancel_request():
         """Cancel request for control"""
         current_user.requests_control = False
-        server.user_datastore.commit()
-        app.usermanager.emit_observers_changed()
+        app.usermanager.update_user(current_user)
+
+        server.emit("observersChanged", namespace="/hwr")
+
         return make_response("", 200)
 
     @bp.route("/take_control", methods=["POST"])
@@ -92,7 +94,11 @@ def init_route(app, server, url_prefix):  # noqa: C901
         name = request.get_json().get("name")
         current_user.nickname = name
         app.usermanager.update_user(current_user)
-        app.usermanager.emit_observers_changed()
+
+        server.emit(
+            "userChanged", room=current_user.socketio_session_id, namespace="/hwr"
+        )
+        server.emit("observersChanged", namespace="/hwr")
 
         return make_response("", 200)
 
@@ -113,7 +119,11 @@ def init_route(app, server, url_prefix):  # noqa: C901
         app.usermanager.update_user(oldop)
         app.usermanager.update_user(newop)
 
-        app.usermanager.emit_observers_changed(message)
+        server.emit("userChanged", room=oldop.socketio_session_id, namespace="/hwr")
+        server.emit(
+            "userChanged", message, room=newop.socketio_session_id, namespace="/hwr"
+        )
+        server.emit("observersChanged", namespace="/hwr")
 
     def remain_observer(user, message):
         observer = user.todict()
--- a/ui/src/actions/remoteAccess.js
+++ b/ui/src/actions/remoteAccess.js
@@ -32,11 +32,17 @@ export function updateNickname(name) {
 }
 
 export function requestControl(message) {
-  return () => sendRequestControl(message);
+  return async (dispatch) => {
+    await sendRequestControl(message);
+    dispatch(getLoginInfo());
+  };
 }
 
 export function cancelControlRequest() {
-  return () => sendCancelControlRequest();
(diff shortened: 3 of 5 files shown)


mxcubeweb/core/components/user/usermanager.py
```diff
@@ -72,17 +72,6 @@ def set_operator(self, username):
...
-    def emit_observers_changed(self, message=""):
-        operator = self.app.usermanager.get_operator()
-
-        data = {
-            "observers": [_u.todict() for _u in self.app.usermanager.get_observers()],
-            "message": message,
-            "operator": operator.todict() if operator else {},
-        }
-
-        self.app.server.emit("observersChanged", data, namespace="/hwr")
```
axelboc, line 84: So effectively, every call to `emit_observers_changed()` is now replaced with:

```py
# zero, one or two times depending on the situation, optionally with a `message`
self.app.server.emit("userChanged", room=some_socket_id, namespace="/hwr")

self.app.server.emit("observersChanged", namespace="/hwr")
```

mxcubeweb/core/components/user/usermanager.py
```diff
@@ -95,8 +84,11 @@ def update_active_users(self):
                     f"Logged out inactive user {_u.username}"
                 )
                 self.app.server.user_datastore.deactivate_user(_u)
+                self.app.server.emit(
+                    "userChanged", room=_u.socketio_session_id, namespace="/hwr"
+                )
```
axelboc, line 89: Send `userChanged` only to users who have been "deactivated".

mxcubeweb/core/components/user/usermanager.py
```diff
@@ -202,7 +194,10 @@ def signout(self):
         self.app.server.user_datastore.deactivate_user(user)
         flask_security.logout_user()
 
-        self.emit_observers_changed()
+        self.app.server.emit(
+            "userChanged", room=user.socketio_session_id, namespace="/hwr"
```
axelboc: Send `userChanged` only to user who as been signed out.

mxcubeweb/routes/ra.py
```diff
@@ -43,25 +43,27 @@ def handle_timeout_gives_control(sid, timeout=30):
                 return make_response(msg, 409)
 
         current_user.requests_control = True
-        server.user_datastore.commit()
+        app.usermanager.update_user(current_user)
```
axelboc, line 46: `server.user_datastore.commit()` was inconsistent with other user updates in this file.

mxcubeweb/routes/ra.py
```diff
@@ -43,25 +43,27 @@ def handle_timeout_gives_control(sid, timeout=30):
...
+
+        server.emit("observersChanged", namespace="/hwr")
```
axelboc, line 48: `userChanged` doesn't need to be sent here. Instead, the `user` state is now refetched right after calling the endpoint in the front-end, in the `requestControl` thunk.

mxcubeweb/routes/ra.py
```diff
@@ -43,25 +43,27 @@ def handle_timeout_gives_control(sid, timeout=30):
...
     @bp.route("/cancel_request", methods=["POST"])
     @server.restrict
     def cancel_request():
         """Cancel request for control"""
         current_user.requests_control = False
-        server.user_datastore.commit()
-        app.usermanager.emit_observers_changed()
+        app.usermanager.update_user(current_user)
+
+        server.emit("observersChanged", namespace="/hwr")
```
axelboc, line 65: Same, the user state is now refetched right after calling the endpoint in the front-end, in the `cancelControlRequest` thunk.

mxcubeweb/routes/ra.py
```diff
@@ -92,7 +94,11 @@ def update_user_nickname():
         name = request.get_json().get("name")
         current_user.nickname = name
         app.usermanager.update_user(current_user)
-        app.usermanager.emit_observers_changed()
+
+        server.emit(
+            "userChanged", room=current_user.socketio_session_id, namespace="/hwr"
```
axelboc, line 99: Sending `userChanged` only to the user who took control.

mxcubeweb/routes/ra.py
```diff
@@ -113,7 +119,11 @@ def toggle_operator(username, message):
         app.usermanager.update_user(oldop)
         app.usermanager.update_user(newop)
 
-        app.usermanager.emit_observers_changed(message)
+        server.emit("userChanged", room=oldop.socketio_session_id, namespace="/hwr")
+        server.emit(
+            "userChanged", message, room=newop.socketio_session_id, namespace="/hwr"
+        )
```
axelboc, line 125: Sending `userChanged` only to the old and new operators.

ui/src/actions/remoteAccess.js
```diff
@@ -32,11 +32,17 @@ export function updateNickname(name) {
 }
 
 export function requestControl(message) {
-  return () => sendRequestControl(message);
+  return async (dispatch) => {
+    await sendRequestControl(message);
+    dispatch(getLoginInfo());
```
axelboc, line 37: As explained, we no longer rely on a socket even to update the `user` state; we refetch it right away.

ui/src/serverIO.js
```diff
@@ -376,27 +376,24 @@ class ServerIO {
...
-      ) {
-        this.dispatch(showWaitDialog('You were given control', data.message));
-      } else if (
-        data.observers.length > 0 &&
-        state.login.user.inControl &&
-        data.observers
-          .map((el) => el.username)
-          .includes(state.login.user.username)
-      ) {
+      await this.dispatch(getLoginInfo());
```
axelboc, line 383: So each message now refecthes its own slice of state:

- `userChanged` dispatches `getLoginIngo()`
- `observersChanged` dispatches `getRaState()`

ui/src/serverIO.js
```diff
@@ -376,27 +376,24 @@ class ServerIO {
...
+
+      const newState = store.getState();
+      const { inControl } = newState.login.user;
+
+      if (!wasInControl && inControl) {
+        this.dispatch(showWaitDialog('You were given control', message));
+      } else if (wasInControl && !inControl) {
         this.dispatch(showWaitDialog('You lost control'));
       }
```
axelboc, line 392: This is the new logic to show the "You were given control" and "You lost control" dialog. It looks only for a change in the `login.user.inControl` state — the server no longer needs to send a payload with the `userChanged` message (apart from an optional `message` to display in the dialog).

I've done my best to test this and can't seem to find a case where the new logic behaves differently from the old logic...

The key was to `await` the refetch of the `login` state and to compare the values before and after. I will use the same pattern to show/hide the "asking for control" dialog and the like in future PRs.

ui/src/actions/remoteAccess.js
marcus-oscarsson: Ok, nice, easier to understand like this

ui/src/serverIO.js
marcus-oscarsson: Very nice, thanks alot. We will have to make some thorough testing of this during startup but its well worth it :)

mxcubeweb/core/components/user/usermanager.py
axelboc: Oh FYI, I tried to send `observersChanged` here only when at least one user had been deactivated, but then existing users weren't getting notified when a new observer logged in. This is not great, but didn't want to risk breaking things, so I didn't go any further.

axelboc:
A couple of tests are breaking, I'm investigating :wink:

mxcubeweb/core/components/user/usermanager.py
axelboc: This was the culprit. Notifying the user who just signed out doesn't make sense, since the user is no longer listening for socket messages. Also, the UI already refetches the `login` state, obviously.

mxcubeweb/core/components/user/usermanager.py
marcus-oscarsson: I see, that is a bit odd

mxcubeweb/core/components/user/usermanager.py
marcus-oscarsson: Ok

marcus-oscarsson:
We have to remember to put some extra effort into testing but it looks good

PR #1356, opened 2024-08-20
Improve UX of requesting/approving/denying control

### LINKED ISSUE #1280: Denying request for control does not notify requester
The _Asking for control_ modal doesn't close itself and the requester doesn't know he's been denied control.

Peek 2024-07-02 16-42

axelboc:
Multiple fixes and improvements:

1. At last, I fix #1280 by making sure that the "Asking for control" modal closes after the request is denied by the user in control.
2. I also address [@fabcor-maxiv's comment](https://github.com/mxcube/mxcubeweb/pull/1348#issuecomment-2293614365) by showing the requester's message to the user in control. Note that this required adding a field to the `user` database table, so you'll need to **delete your local database** with `rm /tmp/mxcube-user.db` so it gets recreated the next time you start the server.
3. Finally, I improve the design and UX of the `PassControlDialog` ("[observer] is asking for control"), notably to ensure that the user in control changes the default message ("Here you go!") when denying a request.

Here is a screen recording of the entire flow:

Peek 2024-08-20 15-37

### DIFF
--- a/mxcubeweb/routes/ra.py
+++ b/mxcubeweb/routes/ra.py
@@ -44,7 +44,7 @@ def init_route(app, server, url_prefix):  # noqa: C901
 
         data = request.get_json()
         current_user.requests_control = True
-        current_user.requests_control_msg = data['message']
+        current_user.requests_control_msg = data["message"]
         app.usermanager.update_user(current_user)
 
         server.emit("observersChanged", namespace="/hwr")
@@ -189,7 +189,12 @@ def init_route(app, server, url_prefix):  # noqa: C901
             new_op.requests_control = False
             new_op.requests_control_msg = None
             app.usermanager.update_user(new_op)
-            server.emit("userChanged", data["message"], room=new_op.socketio_session_id, namespace="/hwr")
+            server.emit(
+                "userChanged",
+                data["message"],
+                room=new_op.socketio_session_id,
+                namespace="/hwr",
+            )
             server.emit("observersChanged", namespace="/hwr")
 
         return make_response("", 200)
--- a/mxcubeweb/routes/ra.py
+++ b/mxcubeweb/routes/ra.py
@@ -125,17 +125,6 @@ def init_route(app, server, url_prefix):  # noqa: C901
         )
         server.emit("observersChanged", namespace="/hwr")
 
-    def remain_observer(user, message):
-        observer = user.todict()
-        observer["message"] = message
-
-        server.emit(
-            "setObserver",
-            observer,
-            room=user.socketio_session_id,
-            namespace="/hwr",
-        )
-
     @bp.route("/", methods=["GET"])
     @server.restrict
     def rasettings():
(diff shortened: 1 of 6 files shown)


mxcubeweb/routes/ra.py
```diff
@@ -186,14 +181,21 @@ def request_control_response():
...
+            # Request denied
+            new_op.requests_control = False
+            new_op.requests_control_msg = None
+            app.usermanager.update_user(new_op)
+            server.emit(
+                "userChanged",
+                data["message"],
+                room=new_op.socketio_session_id,
+                namespace="/hwr",
+            )
```
axelboc, line 197: The fix for the "asking for control" dialog not hiding after being denied, together with the new condition in the `userChanged` message handler in `serverIO`.

ui/src/actions/remoteAccess.js
```diff
@@ -34,7 +35,16 @@ export function updateNickname(name) {
...
+
     dispatch(getLoginInfo());
+    dispatch(
+      showWaitDialog(
+        'Asking for control',
+        'Please wait while asking for control',
+        true,
+        () => dispatch(cancelControlRequest()),
+      ),
+    );
```
axelboc, line 47: Just moved this out of `RequestControlForm`

mxcubeweb/routes/ra.py
```diff
@@ -42,7 +42,9 @@ def handle_timeout_gives_control(sid, timeout=30):
                 msg = "Another user is already asking for control"
                 return make_response(msg, 409)
 
+        data = request.get_json()
         current_user.requests_control = True
+        current_user.requests_control_msg = data["message"]
```
axelboc, line 47: Storing the requester's message in the corresponding user object in the DB so that it gets sent to the user in control (inside the requester's `observer` object)

ui/src/components/RemoteAccess/PassControlDialog.jsx
```diff
@@ -11,41 +12,72 @@ function PassControlDialog() {
...
           </Modal.Title>
         </Modal.Header>
+
         <Modal.Body>
-          User "{requestingObs?.nickname}" is asking for control:
-        </Modal.Body>
-        <Modal.Footer>
+          {requestingObs?.requestsControlMsg && (
+            <Alert>{requestingObs.requestsControlMsg}</Alert>
+          )}
```
axelboc, line 65: Displaying the requester's message here.

ui/src/components/RemoteAccess/PassControlDialog.jsx
```diff
@@ -11,41 +12,69 @@ function PassControlDialog() {
...
+        { message: "Please explain why you're denying this request" },
+        { shouldFocus: true },
+      );
+      return;
+    }
+
+    await dispatch(respondToControlRequest(isAllow, data.message));
   }
 
+  useLayoutEffect(() => {
```
marcus-oscarsson: Hm, what does this do ?

marcus-oscarsson:
Hey !, thats nice :)

ui/src/components/RemoteAccess/PassControlDialog.jsx
axelboc: The component is always rendered, even when hidden (so it can be animated out, I guess), so we must ensure that the form's state is properly reset if the modal is shown multiple times. React Hook Form takes care of resetting on submit, but this effect takes care of the case where the modal closes because the requester cancelled their request.

`useLayoutEffect` was supposed to reset the form before React committed the changes to the DOM, but after double checking, I realise it is no longer needed. I also added a condition to reset the state only when the modal closes.

marcus-oscarsson: I see, tricky :)

PR #1357, opened 2024-08-20
Fix forced sign-out

### LINKED ISSUE #1282: Logging out remote users doesn't log them out from the UI
They're only logged out if they refresh the page:

Peek 2024-07-02 16-52

axelboc:
Fix #1282

Easy fix! :tada: 

... too easy? :joy_cat:

### DIFF
--- a/ui/src/serverIO.js
+++ b/ui/src/serverIO.js
@@ -47,7 +47,7 @@ import {
 
 import { getRaState, incChatMessageCount } from './actions/remoteAccess';
 
-import { forcedSignout, getLoginInfo } from './actions/login';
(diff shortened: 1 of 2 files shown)


ui/src/serverIO.js
```diff
@@ -413,8 +413,7 @@ class ServerIO {
     });
 
     this.hwrSocket.on('forceSignout', () => {
-      this.dispatch(forcedSignout());
-      serverIO.disconnect();
```
axelboc, line 419: Don't think we want to disconnect the WS connection, since we connect _before_ logging in, in `App`.

marcus-oscarsson:
:+1:

Issue #1350, opened 2024-08-19
Rename directory `/test` to `/demo`


marcus-oscarsson:
Rename directory `test` (https://github.com/mxcube/mxcubeweb/tree/develop/test) to `demo`.

fabcor-maxiv:
Related to this? https://github.com/mxcube/mxcubecore/issues/944

marcus-oscarsson:
Yes a subset :)

fabcor-maxiv:
If we rename `/test` to `/demo`, then *pytest* will not be able to find the tests anymore (not by default, not without telling pytest explicitly where the test files are).

Or am I missing something? I have not spent much time thinking about this.

elmjag:
IMHO the `https://github.com/mxcube/mxcubeweb/tree/develop/test/HardwareObjectsMockup.xml` should be moved up out from `test` and be renamed to `demo`.

The rest of the `test` folder should remain as-is. There _are_ actual tests there as well.

Another detail, the tests _probably_ needs to be updated, I think they reply on files in `HardwareObjectsMockup.xml`.

marcus-oscarsson:
Yes, then I guess we have to discuss exactly how we want to configure the tests.  We still need a set of configuration files that configure the test hardware objects. So, `https://github.com/mxcube/mxcubeweb/tree/develop/test/HardwareObjectsMockup.xml` should perhaps be copied to a folder called `demo` or even `demo-configuration` (one level up as you mentioned @elmjag). 

How do you think we should configure the tests, should we have one folder with all configuration files that we can override for specific tests, should each test have their set of configuration files (quite tedious IMO) or some other solution ?

elmjag:
I think we can start with changing tests to use the new `demo` folder. Not sure if it's possible, but I think it would be good that there are at least some tests that exercise the `demo` folder.

Future test may use their own config files, if they need to test some configuration not covered by `demo`. They should go inside the `test` folder.

I can take a look at this. I have an itch since long time to rename the `HardwareObjectsMockup.xm` folder :)

marcus-oscarsson:
Ok

elmjag:
I started look into this, but got blocked by this: https://github.com/mxcube/mxcubeweb/issues/1370

elmjag:
It turned out to be pretty straight forward to move `HardwareObjectsMockup.xml `to `demo`. Just changing the default path for beamline configuration files is enough for the tests to find them:

Basically this change does the trick:
[...]

PR #1358, opened 2024-08-21
adds mockup beamline configured with YAML files

elmjag:
Adding an example of the mockup beamline, where all HWOBJs are configured with YAML files.

To start MXCuBE with this version of beamline, use '--repository' command line argument. For example:

`   ./mxcubeweb-server --repository demo.yaml`

This provides an example and a convenient way to test YAML configuration files code.


fabcor-maxiv:
There was this discussion: https://github.com/mxcube/mxcubecore/issues/944

So is it really used for test purposes? Is it a mockup? Is it used for demo purposes? All of the above?

elmjag:
> There was this discussion: [mxcube/mxcubecore#944](https://github.com/mxcube/mxcubecore/issues/944)
> 
> So is it really used for test purposes? Is it a mockup? Is it used for demo purposes? All of the above?

This is not used for (automatic) tests. This is a demo beamline only.

The folder structure should be re-arranged, my suggestion is something like:

```
  demo       <--- current test/HardwareObjectsMockup.yml/aperture.yml
  demo.xml   <--- current test/HardwareObjectsMockup.yml/aperture.xml
```
  
 With the plan to remove `demo.xml` folder once we drop support for XML files.
 
 But I think this should be done as a separate activity, not as a port of the _XML to YAML_ work.

fabcor-maxiv:
See also: <https://github.com/mxcube/mxcubeweb/issues/1350>.

elmjag:
Marking it as draft until https://github.com/mxcube/mxcubeweb/issues/1350 is resolved.

elmjag:
So, I decided to put all these files in `demo.yaml` folder for now. I don't want to put it in `demo` folder yet, as that means switching over mxcubeweb tests to use yaml files. 

I think that should be done after the `xml_yaml_conversion` is merged and YAML stuff is somewhat settled.

fabcor-maxiv:
Do we need the `test_samples` directory and all of its content? @rhfogh, maybe?

_[I feel like I have already asked this before. Sorry for the repeat question if it is the case.]_

rhfogh:
@fabcor-maxiv 
Not sure we have discussed, this but it is an understandable question. 

This is needed to test the GPhL workflow in mock mode. some of the files (the *.hkli) files are only needed for those who have the GPhL diffraction simulator (GPhL, basically) - this allows you to test the entire run, including processing, with simulated images. Other parts (SPOT.XDS) allows you to run the UI part of the global phasing workflow without a diffraction simulator, which could be run by others as well. I certainly need this, and I need it where MXCuBE can find it, as part of the configuration tree. If people think it should not b checked in that is something we can discuss.

marcus-oscarsson:
Thanks @elmjag

elmjag:
I noticed that 'Beam size' drop-down is now messed up, when running MXCuBE with yaml configs. (It works fine with XML configs.)

I think I just need to update some beam definer related configs to make it work. I'll take a look at this, before merging.

rhfogh:
@elmjag Updating the beam definer configs might also solve a couple of errors that you seem to get when you start up the program.
Again, thanks for looking.

demo.yaml/beam.yaml
```diff
@@ -0,0 +1,9 @@
+%YAML 1.2
+---
+class: BeamMockup.BeamMockup
+configuration:
+  defaultBeamDivergence: (12.0, 5.0)
```
beteva: I think this should rather be
```
   beam_divergence_vertical: 12.0
   beam_divergence_horizontal: 5.0
```
And eventually:
  `     definer_type: aperture`

demo.yaml/aperture.yaml
```diff
@@ -0,0 +1,5 @@
+%YAML 1.2
+---
+class: ApertureMockup.ApertureMockup
+configuration:
+  diameter_size_list: '[10, 20, 50, 100, 150]'
```
beteva, line 5: configuration not needed - hard coded in the ApertureMockup

rhfogh:
@beteva 
Good you are looking at this.
On general principles would it be better to configure the diameter_size_list for the mock aperture, rather than hardcode it in the HardwareObject?

demo.yaml/aperture.yaml
beteva: In principle it is configurable, so something could be put in the configuration file instead. The hard coding is more for the test - to be sure that we are using well defined values.
This is how it would look if you have an xml file configuration:
```
 <object class="ApertureMockup">
   <username>aperture</username>
   <values>{"A5": (5, 0.11), A10": (10, 0.15), "A20": (20, 0.3), "A30": (30, 0.63), "A50": (50, 0.9), "A100": (100, 1)}</values>
 </object>
```
I guess we could try to put this in the configuration file as the ApertureMockup inherits from AbstractNState, but we should be sure to put the above values in order to satisfy the test.
I guess the yml file woul be something like this:

```
class: ApertureMockup.ApertureMockup
configuration:
  username: aperture
  values: {"A5": (5, 0.11), A10": (10, 0.15), "A20": (20, 0.3), "A30": (30, 0.63), "A50": (50, 0.9), "A100": (100, 1)}
```
  It would be great if you could tell me if using a configuration file works for you.

demo.yaml/beam.yaml
elmjag: Yes, that was it, thanks!

elmjag:
As far as I can tell, the YAML configured demo beamline works as good as the XML configured. I'm merging this.

PR #1359, opened 2024-08-21
Remove `send` prefix from action creators

axelboc:
There are a couple of exceptions where the endpoint itself includes the word `send`, like `send_command`. In this case, I've tried to use a more specific name for the fetch function (e.g. `sendHarvesterCommand`). This is what I already had done in other domains, like `sampleChanger`.

### DIFF
--- a/ui/src/actions/harvester.js
+++ b/ui/src/actions/harvester.js
@@ -4,9 +4,9 @@ import {
   sendHarvestAndLoadCrystal,
   sendCalibratePin,
   sendValidateCalibration,
-  sendAbort,
-  sendDataCollectionToCrims as sendDCToCrims,
-  sendCommand as sendCmd,
+  sendAbortHarvester,
+  sendDataCollectionInfoToCrims,
+  sendHarvesterCommand,
 } from '../api/harvester';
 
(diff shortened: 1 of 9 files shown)


ui/src/actions/harvester.js
```diff
@@ -68,22 +68,22 @@ export function harvestAndLoadCrystal(xtalUUID, successCb = null) {
...
     } catch (error) {
       dispatch(showErrorPanel(true, error.response.headers.get('message')));
-      throw new Error('Server refused to Harveste or Load Crystal');
+      throw new Error('Server refused to harvest or load crystal');
     }
   };
 }
 
-export function sendDataCollectionToCrims() {
+export function sendDCToCrims() {
```
fabcor-maxiv:
```suggestion
export function sendDcToCrims() {
```

I prefer like this, this seems more correct to me (and it hurts my eyes less). Also because we would not write `send_d_c_to_crims`.

But then above we have `sendDataCollectionInfoToCrims`. So we could also go with:

```suggestion
  sendDataCollectionToCrims,
```

fabcor-maxiv (approved):
Seems fine to me

PR #1360, opened 2024-08-21
Change currentAperture type to str

beteva:
currentAperture is a string as it represents beam size label, not just the size any more. Corresponds to any beam definer.


meguiraun (approved):

marcus-oscarsson:
Thanks !

PR #1362, opened 2024-08-22
Refactor `Main` as function component

axelboc:
I also extract the `ChatWidget` into its own component.

### DIFF
--- a/ui/src/components/ChatWidget.jsx
+++ b/ui/src/components/ChatWidget.jsx
@@ -1,0 +1,75 @@
+import React, { useEffect } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import {
+  incChatMessageCount,
+  resetChatMessageCount,
+} from '../actions/remoteAccess';
+import { fetchChatMessages, sendChatMessage } from '../api/remoteAccess';
+import {
+  addResponseMessage,
+  addUserMessage,
+  setBadgeCount,
+  Widget,
+} from 'react-chat-widget';
+import Draggable from 'react-draggable';
+import { store } from '../store';
+
+function ChatWidget() {
+  const dispatch = useDispatch();
+
+  const username = useSelector((state) => state.login.user.username);
+  const observers = useSelector((state) => state.remoteAccess.observers);
+  const chatMessageCount = useSelector(
+    (state) => state.remoteAccess.chatMessageCount,
+  );
+
+  useEffect(() => {
+    const { user } = store.getState().login; // non-reactive so effect runs only on mount
+
+    (async function loadMessages() {
+      const { messages } = await fetchChatMessages();
+      let unread = 0;
+
+      messages.forEach((entry) => {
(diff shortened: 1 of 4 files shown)


ui/src/components/ChatWidget.jsx
```diff
@@ -0,0 +1,55 @@
...
+      messages.forEach((entry) => {
+        if (entry.username === username) {
+          addUserMessage(`${entry.date} **You:** \n\n ${entry.message} \n\n`);
+        } else {
+          addResponseMessage(
+            `${entry.date} **${entry.nickname}:** \n\n ${entry.message}`,
+          );
+        }
+      });
+    })();
```
axelboc, line 47: The [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) is because `useEffect` doesn't directly accept `async` functions.

ui/src/components/Main.jsx
```diff
@@ -19,144 +16,70 @@ import WorkflowParametersDialog from '../containers/WorkflowParametersDialog';
...
-class Main extends React.Component {
-  constructor(props) {
-    super(props);
-    this.handleClick = this.handleClick.bind(this);
-    this.handleNewUserMessage = this.handleNewUserMessage.bind(this);
-    this.onChatContainerClick = this.onChatContainerClick.bind(this);
-  }
-
-  componentDidMount() {
-    if (!this.props.applicationFetched) {
```
axelboc, line 53: I removed this condition, since when this component renders, `applicationFetched` is always `false`. If the user logs out, it gets unmounted before `applicationFetched` is set back to `false`.

ui/src/components/Main.jsx
```diff
@@ -19,144 +16,70 @@ import WorkflowParametersDialog from '../containers/WorkflowParametersDialog';
...
-  handleNewUserMessage(message) {
-    sendChatMessage(message, this.props.login.user.username);
-  }
+  const showReadOnlyDiv =
+    !inControl && pathname !== '/remoteaccess' && pathname !== '/help';
 
-  handleClick(e) {
-    e.preventDefault();
-    e.stopPropagation();
-    e.nativeEvent.stopImmediatePropagation();
```
axelboc, line 89: I've removed this handler from the read-only `div`. I don't think it's needed, since the read-only `div` is on top of the main part of the page and it's a stand-alone element with no children, so it already prevents pointer events from reaching any of the elements it covers. It would be needed if the backdrop was implemented as a CSS pseudo-element on `div.main`, for instance, since pointer events would then bubble down to the `div`'s children.

ui/src/components/ChatWidget.jsx
marcus-oscarsson: Oh, ok

axelboc:
Okay, I think it's ready. I've changed the base branch to `chat-fix` so it can be merged into it.

marcus-oscarsson:
Thanks, alot have a nice arvo ;)

PR #1364, opened 2024-08-23
Chat - Saving information about the user who wrote a message

### LINKED ISSUE #1361: Chat messages not restored on refresh
`GET /remoteaccess/chat` throws a silent error like in #1355, so the messages aren't retrieved.

```
  File "/home/bocciare/tools/micromamba/envs/mxcubeweb/lib/python3.10/site-packages/flask/app.py", line 880, in full_dispatch_request rv = self.dispatch_request()
  File "/home/bocciare/tools/micromamba/envs/mxcubeweb/lib/python3.10/site-packages/flask/app.py", line 865, in dispatch_request return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args) # type: ignore[no-any-return]
  File "/home/bocciare/tools/micromamba/envs/mxcubeweb/lib/python3.10/site-packages/flask_security/decorators.py", line 432, in decorated_view return current_app.ensure_sync(fn)(*args, **kwargs)
  File "/home/bocciare/workspace/mxcubeweb/mxcubeweb/routes/ra.py", line 215, in get_all_mesages return jsonify({"messages": app.chat.get_all_messages()})
  File "/home/bocciare/workspace/mxcubeweb/mxcubeweb/core/components/chat.py", line 40, in get_all_messages "username": _m.username,

AttributeError: 'Message' object has no attribute 'username' 
```

marcus-oscarsson:
The database schema for our chat is a bit limited in the relation messages to users. There is currently not really a way to know who authored a message. From a database model perspective one should probably have a author column with the author id in the `message` table or some other design of the mapping message to users. However I'm not sure its needed for our quite simple chat. 

Currently missing in this PR is a field `read` to signify if the message was read by the receiver or not.

This PR Closes #1361.

### DIFF
--- a/mxcubeweb/core/models/usermodels.py
+++ b/mxcubeweb/core/models/usermodels.py
@@ -43,6 +43,10 @@ class Message(Base):
     id = Column(Integer(), primary_key=True)
     at = Column(DateTime())
     message = Column(Text())
+    read = Column(Boolean(False))
+    from_username = Column(String(255), unique=False)
+    from_nickname = Column(String(255), unique=False)
+    from_host = Column(String(255), unique=False)
 
 
 class User(Base, UserMixin):
--- a/ui/src/api/remoteAccess.js
+++ b/ui/src/api/remoteAccess.js
@@ -51,3 +51,7 @@ export function fetchChatMessages() {
 export function sendChatMessage(message, username) {
   return endpoint.post({ message, username }, '/chat').res();
 }
+
+export function sendSetAllMessagesRead(message, username) {
+  return endpoint.post({ message, username }, '/chat/set_all_read').res();
+}
--- a/ui/src/components/Main.jsx
+++ b/ui/src/components/Main.jsx
@@ -19,8 +19,16 @@ import WorkflowParametersDialog from '../containers/WorkflowParametersDialog';
 import GphlWorkflowParametersDialog from '../containers/GphlWorkflowParametersDialog';
 import SelectProposalContainer from '../containers/SelectProposalContainer';
 import diagonalNoise from '../img/diagonal-noise.png';
-import { resetChatMessageCount } from '../actions/remoteAccess';
-import { Widget, addResponseMessage, addUserMessage } from 'react-chat-widget';
+import {
+  resetChatMessageCount,
+  incChatMessageCount,
+} from '../actions/remoteAccess';
+import {
+  Widget,
+  addResponseMessage,
+  addUserMessage,
+  setBadgeCount,
+} from 'react-chat-widget';
 import { showDialog } from '../actions/general';
 import { LimsResultDialog } from './Lims/LimsResultDialog';
 import LoadingScreen from './LoadingScreen/LoadingScreen';
@@ -48,6 +56,7 @@ class Main extends React.Component {
 
     // eslint-disable-next-line promise/prefer-await-to-then, promise/catch-or-return
     fetchChatMessages().then((json) => {
+      let unread = 0;
       json.messages.forEach((entry) => {
         if (entry.username === this.props.login.user.username) {
           addUserMessage(`${entry.date} **You:** \n\n ${entry.message} \n\n`);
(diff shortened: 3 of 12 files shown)


mxcubeweb/core/models/usermodels.py
```diff
@@ -43,6 +43,10 @@ class Message(Base):
     id = Column(Integer(), primary_key=True)
     at = Column(DateTime())
     message = Column(Text())
+    read = Column(Boolean(False))
```
marcus-oscarsson, line 46: Added the following columns to keep track of the author and if the message was read

ui/src/components/Main.jsx
```diff
@@ -75,6 +90,7 @@ class Main extends React.Component {
   }
 
   render() {
+    setBadgeCount(this.props.remoteAccess.chatMessageCount);
```
marcus-oscarsson: the chat widget (`react-chat-widget`) now apparently handles the message count internally (see setBadgeCount, https://github.com/Wolox/react-chat-widget) so we need to override that behavior

ui/src/reducers/remoteAccess.js
```diff
@@ -36,8 +36,11 @@ function remoteAccessReducer(state = INITIAL_STATE, action = {}) {
     case 'RESET_CHAT_MESSAGE_COUNT': {
       return { ...state, chatMessageCount: 0 };
     }
-    case 'INC_CHAT_MESSAGE_COUNT': {
-      return { ...state, chatMessageCount: state.chatMessageCount + 1 };
+    case 'SET_CHAT_MESSAGE_COUNT': {
```
marcus-oscarsson: Simply made it possible to pass the number of unread messages

ui/src/components/Main.jsx
```diff
@@ -48,13 +56,20 @@ class Main extends React.Component {
...
         } else {
           addResponseMessage(
             `${entry.date} **${entry.nickname}:** \n\n ${entry.message}`,
           );
+
+          if (!entry.read) {
+            unread++;
+          }
+
+          this.props.setChatMessageCount(unread);
```
axelboc: Probably should be called after the loop?

ui/src/reducers/remoteAccess.js
axelboc: The naming doesn't make it clear that the message count is being incremented. I think the `INC` prefix was fine before; "to increment" doesn't necessarily mean by 1.

ui/src/components/Main.jsx
axelboc: This should be called in `componentDidUpdate` (or `useEffect` if it was a function component).

If you like, I can open a sub-PR to refactor `Main` into a function component (I had already done so yesterday actually, but hadn't pushed it yet).

ui/src/api/remoteAccess.js
```diff
@@ -51,3 +51,7 @@ export function fetchChatMessages() {
 export function sendChatMessage(message, username) {
   return endpoint.post({ message, username }, '/chat').res();
 }
+
+export function setAllMessagesRead(message, username) {
```
axelboc: `sendSetAllMessagesRead`

ui/src/components/Main.jsx
marcus-oscarsson: Actually, that for some reason did not work. I had to move `setBadeCount` to render to make it work. Wonderful for the functional component, I was hesitating of doing it as well. Ill let you continue if you already started :)

marcus-oscarsson: Ill quickly address some of your comments below and then you can go for it

ui/src/api/remoteAccess.js
marcus-oscarsson: Oh yes

ui/src/components/Main.jsx
marcus-oscarsson: Yes, of course

ui/src/reducers/remoteAccess.js
marcus-oscarsson: I see,
 
> "to increment" doesn't necessarily mean by 1.

True, I somehow found it nicer with SET but INC is just as fine

ui/src/components/Main.jsx
marcus-oscarsson: Should be ok now :)

axelboc: Thanks, I'll take over.

Issue #1355, opened 2024-08-20
Errors sent as 200 reponse payloads


axelboc:
When a back-end error occurs inside some endpoints, the server still responds with 200, with the error trace as body.

I last experienced this in #1354: an error was thrown in the [`signout` function](https://github.com/mxcube/mxcubeweb/blob/develop/mxcubeweb/core/components/user/usermanager.py#L180) and the front-end wasn't aware of it. This was the body of the response:

![image](https://github.com/user-attachments/assets/1823e12d-723f-44ca-919d-27e5c458dca9)

This kind of error is detected only when the front-end needs to read the response body; otherwise it's just ignored and nothing gets logged to the console since it's a 200 response.

---

This kind of server error should result in `500 - Internal Server Error` responses.

elmjag:
Another details here is that front-end probably does not check the Content-Type header of the responses. It just tries to parse this response as JSON.

For each response, the front-end should check that:

 - status code is 200
 - the Content-Type is `application/json`

and only _then_ try to parse the reply as JSON.

Right now we just get these cryptic errors in the console, when it tries to parse the traceback text as JSON.

axelboc:
You are very much right. The status code is checked obviously, since anything else than 200 throws a fetch error, but the `Content-Type` doesn't seem to be checked before `JSON.parse()` is called.

EDIT: fixed in #1365

PR #1365, opened 2024-08-23
Check MIME type before parsing responses as JSON

axelboc:
Addresses https://github.com/mxcube/mxcubeweb/issues/1355#issuecomment-2306379987

I was able to write a wretch add-on for this, which is pretty awesome. :tada: The add-on adds a method called `expectJson` to wretch's chainable API (more specifically to the response chain, so after `get`, `post`, etc.)

So you can now do: `.get(...).expectJson().json()` and you'll see something like this:

image

You have to go down the stack trace a bit to find the location of the problematic fetch call, but I think that's okay.

### DIFF
--- a/ui/src/api/addons/safeJson.js
+++ b/ui/src/api/addons/safeJson.js
@@ -1,0 +1,28 @@
+/* eslint-disable promise/prefer-await-to-then */
+
+function safeJsonMiddleware(state) {
+  return (next) => (url, opts) => {
+    return next(url, opts).then((response) => {
+      const contentType = response.headers.get('content-type');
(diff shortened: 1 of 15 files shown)


elmjag:
Is it possible to include call do the json parsing inside the `expectJson` function?

It would be nice if we could just do:

`return endpoint.get('/beam/info').expectJson();`

axelboc:
Even better, thanks :+1: 

I called it `safeJson`, though.

elmjag (approved):

marcus-oscarsson:
Nice

marcus-oscarsson:
Thanks @elmjag and @axelboc

PR #1366, opened 2024-08-23
Clean-up of unmount methods

### LINKED ISSUE #1363: 'NoneType' object has no attribute '_set_loaded' error when unmounting a sample
When running with mocked beamline, you get the error below after clicking the 'Unmount' button.

image

In the log, there is this error:

```
2024-08-23 08:40:30,137 |MX3.HWR|ERROR  | [SC] sample could not be mounted
  File "/home/elmjag/mxcube/web/mxcubeweb/core/components/samplechanger.py", line 273, in unmount_sample_clean_up
    HWR.beamline.sample_changer.unload(sample["location"], wait=False)
  File "/home/elmjag/mxcube/core/mxcubecore/HardwareObjects/mockup/SampleChangerMockup.py", line 92, in unload
    sample._set_loaded(False, True)
    ^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute '_set_loaded'
```

How to reproduce:

- goto `Queued Samples` tab
- click `Create new sample` button
- write sample name and protein acronym, click `Mount`
- click `Unmount` button

marcus-oscarsson:
Simplified the un-mount logic slightly. I guess someone took a tumble in the logical garden at some point (probably me ;). There was a method that unmounts a sample with a given address/position. However that obviously does not  make any sense since we only can have one sample mounted at a time :)

Removed the function `unmountSample` that took the sample address and replaced it with the method that unmounts the currently mounted sample ;)

Closes #1363

### DIFF
--- a/ui/src/api/sampleChanger.js
+++ b/ui/src/api/sampleChanger.js
@@ -30,10 +30,6 @@ export function sendMountSample(sampleData) {
   return endpoint.post(sampleData, '/mount').res();
 }
 
-export function sendUnmountSample(sample) {
-  return endpoint.post({ sample: { location: sample } }, '/unmount').res();
-}
-
 export function sendUnmountCurrentSample() {
   return endpoint.post(undefined, '/unmount_current').res();
 }
--- a/mxcubeweb/routes/samplechanger.py
+++ b/mxcubeweb/routes/samplechanger.py
@@ -89,25 +89,6 @@ def init_route(app, server, url_prefix):  # noqa: C901
 
         return resp
 
-    @bp.route("/unmount", methods=["POST"])
-    @server.require_control
-    @server.restrict
-    def unmount_sample():
-        try:
-            resp = jsonify(
-                app.sample_changer.unmount_sample(request.get_json()["sample"])
-            )
-        except Exception as ex:
-            return (
-                "Cannot unload sample",
-                409,
-                {
-                    "Content-Type": "application/json",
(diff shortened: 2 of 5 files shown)


ui/src/api/sampleChanger.js
```diff
@@ -30,10 +30,6 @@ export function sendMountSample(sampleData) {
   return endpoint.post(sampleData, '/mount').res();
 }
 
-export function sendUnmountSample(sample) {
-  return endpoint.post({ sample: { location: sample } }, '/unmount').res();
```
marcus-oscarsson, line 34: So the issue you had @elmjag was here, `sample` is an object that contains `location` so this logic simply wrapped that object in a object with the key `location` . On the server side that object gets passed as a `dict` instead of the location `str`

axelboc (approved):
Looks like I missed [this line](https://github.com/mxcube/mxcubeweb/blame/1bdb6b8db673a05c59071f6b23079f8ed91b00c6/ui/src/actions/sampleChanger.js#L176) while refactoring... My apologies!

marcus-oscarsson:
> Looks like I missed [this line](https://github.com/mxcube/mxcubeweb/blame/1bdb6b8db673a05c59071f6b23079f8ed91b00c6/ui/src/actions/sampleChanger.js#L176) while refactoring... My apologies!

:), not at all a problem, the logic was so odd that it would have been difficult to anticipate that I think

PR #1367, opened 2024-08-23
Show error when another observer is already asking for control

axelboc:
Screenshot from 2024-08-22 16-54-59

I'm also:

- refactoring the `ErrorNotificationPanel` and reducing the width of the error modal;
- fixing a couple of conflicts between the "pass-control dialog" and the "wait dialog" — this happened for instance when taking control while a request for control is in progress (the _[user] is asking for control_ dialog was shown on top of the _You were given control_ dialog visually but not in the DOM, which caused issues);
- removing the name of the user from the "Give control" button:
    
    Screenshot from 2024-08-22 16-39-58

### DIFF
--- a/ui/src/components/RemoteAccess/PassControlDialog.jsx
+++ b/ui/src/components/RemoteAccess/PassControlDialog.jsx
@@ -3,6 +3,7 @@ import { useDispatch, useSelector } from 'react-redux';
 import { Modal, Button, Form, Alert } from 'react-bootstrap';
 import { respondToControlRequest } from '../../actions/remoteAccess';
 import { useForm } from 'react-hook-form';
+import { hideWaitDialog } from '../../actions/waitDialog';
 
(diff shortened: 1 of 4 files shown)


ui/src/components/RemoteAccess/PassControlDialog.jsx
```diff
@@ -40,10 +41,12 @@ function PassControlDialog() {
   }
 
   useEffect(() => {
-    if (!showModal) {
+    if (showModal) {
+      dispatch(hideWaitDialog()); // avoid conflict with any yet-to-be-dismissed dialog
```
axelboc, line 45: Obviously less than ideal. It would be better if the Redux state held a single dialog visibility state but this isn't trivial.

marcus-oscarsson:
Oh, yes very nice :)

PR #1368, opened 2024-08-23
Fix initial position of chat widget

axelboc:
The chat widget is now a bit more tucked into the bottom-right corner, as expected (I think):

image

Took the opportunity to replace all occurrences I could find of `0px` in the CSS, which can be written simply as `0`.

### DIFF
--- a/ui/src/components/rachat.css
+++ b/ui/src/components/rachat.css
@@ -1,9 +1,7 @@
 .chat-widget-dragable {
   position: absolute;
-  height: 100px;
-  width: 100px;
-  bottom: 0px;
-  right: 0px;
+  bottom: 5px;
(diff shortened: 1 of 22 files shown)


marcus-oscarsson:
:+1:

ui/src/components/rachat.css
```diff
@@ -1,9 +1,7 @@
 .chat-widget-dragable {
   position: absolute;
-  height: 100px;
-  width: 100px;
```
axelboc, line 4: This `height` was the cause of the positioning issue.

marcus-oscarsson: Ok, I see thanks !

PR #1369, opened 2024-08-23
Refactor and improve `ObserverDialog`

axelboc:
- Remove unused `showObserverDialog` action and state flag.
- Convert to function component.
- Use React Hook Form to manage form (there's no validation whatsoever since the `name` is optional, but it provides a cleaner way to retrieve the value of the `name` field in the `submit` handler, and to provide a reactive `defaultValue` for the field).
- Clean up design slightly (size of modal, button colour, etc.)
- Autofocus `name` field when dialog appears so users can type and press Enter right away.
- Label the field and make sure the field receives focus when the label is clicked (and do the same in `PassModalDialog`).

image


marcus-oscarsson:
Thanks :), more and more functional components :+1:

PR #1373, opened 2024-08-27
Tame Cypress tests

axelboc:
Our E2E tests are currently limited by the fact that **WebSockets don't seem to work on the CI**. The front-end tries to connect but the connection is never established. It is a very hard issue to debug. I've tried with Chrome and Electron browsers instead of Firefox; I've played with the WebSocket URL; I've refactored the `ServerIO` class ; ... nothing worked.

I did find [this issue](https://github.com/cypress-io/cypress/issues/22556), which is still open and suggests a workaround, but before I dig into this, I thought I'd first try to **stabilise the tests** with the knowledge that WebSockets don't work on the CI.

---

### 1) Run tests locally with WebSockets disabled

I add an environment variable called `VITE_DISABLE_WEBSOCKETS` that I set to `true` before launching the Cypress UI locally (in the `cypress` script in `package.json`). This allows me to run the E2E tests locally more like how they are run on the CI.

### 2) Make tests easier to debug on the CI

Debugging tests that fail on the CI is challenging because we can't inspect the DOM or see `console.log()` messages, and Cypress' output is generally quite sparse. I fix this by installing a Cypress plugin called [cypress-terminal-report](https://github.com/archfz/cypress-terminal-report), which logs all the commands and `console` messages for failing tests.

### 3) Refetch local state only when needed

Basically, without WebSockets, any UI change that relies on a server-sent update must be expected **after refetching the Redux state**. This can be done in three ways: by reloading the page (`cy.reload()`), by performing a full browser navigation (`cy.visit(...)`), or by signing out and logging back in.

I noticed that not all calls to `cy.reload()` or `cy.visit(...)` were necessary in our current tests. For instance, after taking control, we can navigate back to "Data collection" instantly by simply clicking on the corresponding navigation link in the top bar. So I replace all the unnecessary calls, which should speed up the tests a bit.

Two tests remain with calls to `cy.reload()`. Ideally, I would prefer to log out and back in, but this has a number of side effects that are incompatible with our tests, including unmounting the current sample. So I've left them as is.

### 4) Stabilise flaky tests

The two tests calling `cy.reload()` were particularly flaky, to the point that we had to skip one of them. This happened because `cy.reload()` was called before the browser had time to send the preceding command to the server.

The solution consists in awaiting the API call and expecting a UI change to confirm that the call went through successfully before reloading.

### DIFF
--- a/ui/cypress/e2e/sampleControls.cy.js
+++ b/ui/cypress/e2e/sampleControls.cy.js
@@ -15,12 +15,18 @@ describe('3-click centring', () => {
     );
   });
 
-  it.skip('Each click is rotating the sample by 90 degrees', () => {
+  it('Each click is rotating the sample by 90 degrees', () => {
     cy.mountSample();
     cy.findByRole('button', { name: 'Sample: test - test' }).should(
       'be.visible',
     );
+
     cy.findByRole('button', { name: '3-click centring' }).click();
+    cy.findByRole('button', { name: '3-click centring' }).should(
+      'have.class',
+      'active',
+    );
+
     cy.get('.form-control[name="diffractometer.phi"]')
       .invoke('val')
       .then((value) => {
--- a/ui/cypress/e2e/sampleControls.cy.js
+++ b/ui/cypress/e2e/sampleControls.cy.js
@@ -2,13 +2,13 @@
 
 describe('3-click centring', () => {
   beforeEach(() => {
-    cy.login();
-    cy.findByRole('heading', { name: 'MXCuBE-Web (OSC)' }).should('be.visible');
-    cy.takeControl();
+    cy.loginWithControl();
   });
 
   it('3-click centring should not work without sample', () => {
-    cy.clearSamples();
+    cy.clearSamples(); // another test may have mounted a sample
+    cy.findByRole('link', { name: /Data collection/u, hidden: true }).click();
+
     cy.findByRole('button', { name: '3-click centring' }).click();
     cy.findByRole('alert', 'Error: There is no sample mounted').should(
       'be.visible',
--- a/ui/cypress/e2e/sampleControls.cy.js
+++ b/ui/cypress/e2e/sampleControls.cy.js
@@ -1,11 +1,11 @@
-/* global cy, it, describe, beforeEach, Cypress */
+/* global cy, it, describe, beforeEach */
 
 describe('3-click centring', () => {
   beforeEach(() => {
     cy.loginWithControl();
   });
 
-  it('3-click centring should not work without sample', () => {
+  it('displays error when no sample is mounted', () => {
     cy.clearSamples(); // another test may have mounted a sample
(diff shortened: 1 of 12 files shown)


ui/cypress/e2e/sampleControls.cy.js
```diff
@@ -1,42 +1,43 @@
...
     );
+
     cy.findByRole('button', { name: '3-click centring' }).click();
+    cy.findByText(/Clicks left: 3/u, { hidden: true }).should('exist');
+
     cy.get('.form-control[name="diffractometer.phi"]')
       .invoke('val')
-      .then((value) => {
-        let omegaValue = Number.parseFloat(value);
-        Cypress._.times(2, () => {
```
axelboc, line 28: Note that I also simplify this test by performing only one rotation.

ui/src/actions/queue.js
```diff
@@ -106,22 +106,18 @@ export function addSamplesToQueue(sampleDataList) {
 }
 
 export function addSampleAndMount(sampleData) {
-  return (dispatch) => {
-    dispatch(
-      mountSample(sampleData, async () => {
```
axelboc, line 111: Replacing the callback syntax so it's easier to follow.

ui/src/components/Tasks/AddSample.jsx
```diff
@@ -45,12 +45,13 @@ function AddSample(props) {
...
   }, [setFocus, show]);
 
-  function addAndMount(params) {
+  async function addAndMount(params) {
     const sampleData = getSampleData(params);
     addSamplesToList([sampleData]);
-    addSampleAndMount(sampleData);
     hide();
 
+    await addSampleAndMount(sampleData);
```
axelboc, line 53: Calling after `hide` so the modal doesn't wait for the call to succeed before disappearing (unlike `showList`, which is the important part, since we wait for the "Queued samples" tab to no longer be active in the test as an indication that the call has succeeded)

ui/cypress/e2e/sampleControls.cy.js
walesch-yan: It might not be a "3"-click centring test then, but for the sake of performance increase, I'm happy to reduce the test to a single rotation.

walesch-yan:
The changes look very good, thanks a lot!

ui/cypress/e2e/sampleControls.cy.js
marcus-oscarsson: I guess a more complete test would be to perform the 3-click centering and check that the point gets added. But  that could be a separate test I guess.

marcus-oscarsson:
Wonderful, thanks for all the care full investigation (it takes some time) and the insight :)

ui/cypress/e2e/sampleControls.cy.js
axelboc: Yeah exactly, but without WebSockets, it's too complicated a feature to test for now, in my opinion.

PR #1375, opened 2024-08-27
Clean-up `ServerIO`

axelboc:
- Move "refresh interval" initialisation and clean-up to `App.jsx`.
- Remove unused `ServerIO` class members.
- Log to the console when sockets are disconnected.
- Fix `connect` listener registered on wrong socket.

### DIFF
--- a/ui/src/serverIO.js
+++ b/ui/src/serverIO.js
@@ -64,7 +64,6 @@ import {
 import { setEnergyScanResult } from './actions/taskResults';
 
 import { CLICK_CENTRING } from './constants';
-import { sendRefreshSession } from './api/login';
 import { store } from './store';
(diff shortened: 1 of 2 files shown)


ui/src/serverIO.js
```diff
@@ -154,6 +120,8 @@ class ServerIO {
     });
 
     this.loggingSocket.on('disconnect', (reason) => {
+      console.log('loggingSocket disconnected!'); // eslint-disable-line no-console
```
marcus-oscarsson, line 123: For debugging then I guess ?

marcus-oscarsson: Oh, yes you even mention that above :)

axelboc: Yeah, I figured there's one for `connect`, might as well have one for `disconnect`.

PR #1378, opened 2024-08-28
Mesh results

meguiraun:
Fixing mesh result display when sending via mxcubeweb API. 

Also fixing the call to `getGridCellCenter`, this method no longer exists.

Format of the data: (to be documented in the documentation, not just here in an MR...)
```
{"result": {<cell_number>:  [R, G, B, alpha], ....}, "data_file": <str>}
```
for example:
```
{"result": 
   {
    '44': [255, 255, 255, 0.0], 
    '61': [255, 255, 255, 0.0], 
    '56': [255, 255, 255, 0.0],
    '10': [255, 255, 255, 0.0]},
"data_file": "a/path/to/file"}
```

Anyone else using this functionality?

For PNG format, who is using it, @marcus-oscarsson? Can you give me an example data and I test with this? Do you use this API or you go via workflow?

Related mxcubecore https://github.com/mxcube/mxcubecore/pull/1004

Working with mockups:

change the following: mesh_result_format: RGB in beamline_config.yml

Mock results, assuming grid G1 exists. Update this code with number of cells (the 266 number below)
```
import requests, json, random

url = "http://localhost:8081/mxcube/api/v0.1/sampleview/shapes/G1"
from random import randrange
MAX = 100


SENT = []

def send():
    data = {}
    cells = []
    for i in range(5):
        cells.append(random.randint(1, 266))

    for i in cells:
        if i not in SENT:
            data.update({str(i): 
            [random.randrange(0, 255, 1), 
                random.randrange(0, 255, 1), 
                random.randrange(0, 255, 1), 0.0]
                # [255, 255, 255, 0.0]
            }
            )
            SENT.append(i)
    requests.post(url, json={"result": data, "data_file": "/data/visitors/2132"})


while(True):
    send()
    time.sleep(1)
```

mesh_display

### DIFF
--- a/ui/src/components/SampleView/DrawGridPlugin.jsx
+++ b/ui/src/components/SampleView/DrawGridPlugin.jsx
@@ -329,7 +329,7 @@ export default class DrawGridPlugin {
       for (let nh = 0; nh < row; nh++) {
         for (let nw = 0; nw < col; nw++) {
           const index = nw + nh * col + 1;
-          if (index < Object.keys(result).length) {
+          if (result[index] != undefined) {
             fillingMatrix[nw][nh] = this.heatMapColorForValue(
               gd,
               result[index],
--- a/ui/src/components/SampleView/DrawGridPlugin.jsx
+++ b/ui/src/components/SampleView/DrawGridPlugin.jsx
@@ -329,7 +329,7 @@ export default class DrawGridPlugin {
       for (let nh = 0; nh < row; nh++) {
         for (let nw = 0; nw < col; nw++) {
           const index = nw + nh * col + 1;
-          if (result[index] != undefined) {
+          if (result[index] !== undefined) {
             fillingMatrix[nw][nh] = this.heatMapColorForValue(
               gd,
               result[index],
--- a/ui/src/components/SampleView/DrawGridPlugin.jsx
+++ b/ui/src/components/SampleView/DrawGridPlugin.jsx
@@ -325,10 +325,12 @@ export default class DrawGridPlugin {
       for (let nh = 0; nh < row; nh++) {
         for (let nw = 0; nw < col; nw++) {
           const index = nw + nh * col + 1;
-          fillingMatrix[nw][nh] = this.heatMapColorForValue(
-            gd,
-            result[index][1],
-          );
+          if (index < Object.keys(result).length) {
+            fillingMatrix[nw][nh] = this.heatMapColorForValue(
+              gd,
+              result[index],
+            );
+          }
         }
       }
     }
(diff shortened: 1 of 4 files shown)


ui/src/components/SampleView/DrawGridPlugin.jsx
```diff
@@ -325,10 +329,12 @@ export default class DrawGridPlugin {
       for (let nh = 0; nh < row; nh++) {
         for (let nw = 0; nw < col; nw++) {
           const index = nw + nh * col + 1;
-          fillingMatrix[nw][nh] = this.heatMapColorForValue(
-            gd,
-            result[index][1],
-          );
+          if (result[index] != undefined) {
```
axelboc:
```suggestion
          if (result[index] !== undefined) {
```

marcus-oscarsson:
@meguiraun We are using the XMLRPC interface, so we are not using the REST interface of mxcubeweb to add mesh/grid results.  We have some longer term plans to replace the XMLRPC interface with for instance gRPC or other technology and integrate that with the `AbstractProcessing` concept that we have been speaking bout in the past. However, we are for now using XMLRPC.

The XMLRPC method is using `SampleView.set_grid_data(self, sid, result_data, data_file_path)`  and the `result_data` argument is a base64 encoded PNG, as far as I know the argument `data_file_path` (and we can probably remove it).

We would not recommend using the mxcubeweb REST interface to communicate from other processes to mxcube. The first reason would be for security concerns the other would be that logic is better situated in `mxubecore`. The idea would be to share the same interface for integrating processing results.

Also be aware, that rendering the mesh like above also does not perform very well when the meshes gets bigger, i.e 100k cells. Perhaps you already tested this and where happy with the results ?

meguiraun:
We are using this already in Biomax and we are happy with how it works. We will have bigger mesh-es in micromax and although I agree about the performance issues, we need to reach that bridge first and then we might reconsider and change to PNG. What does this png represent for you? Is it a heatmap overlay image?

Regarding security, we are doing this communication in internal networks so we don't have same concerns. My experience with gRPC has been discouraging, a big effort is needed to maintain the interface (as opposed to secure http).

I would not mind securing this rest api, but we are bit stretched now...

marcus-oscarsson:
Sure, no problem to keep it like this for the moment so that its compatible with your current solution. We can discuss exactly how to best handle this at a later moment.

marcus-oscarsson:
Yes the PNG image is heatmap that is overlay-ed

elmjag:
> Also be aware, that rendering the mesh like above also does not perform very well when the meshes gets bigger, i.e 100k cells. Perhaps you already tested this and where happy with the results ?

You are correct. The 100k sized meshes absolutely kills UI performance in RGB mode. 

However here is PR that addresses that: https://github.com/mxcube/mxcubeweb/pull/1416

I have tested the change above this with ~ 130k sized meshed, it works ok. I did not measure the actual performance, but the overall latency feels ok.

PR #1381, opened 2024-08-30
move mocked beamline configs to 'demo'

elmjag:
Move out and rename the mocked beamline configuration files folder out from the test folder. This to make it more clear that it's not only used by the tests, but also as a 'demo' beamline.

Solves: https://github.com/mxcube/mxcubeweb/issues/1350


fabcor-maxiv (approved):
Seems fine to me.

marcus-oscarsson:
:) but is it more clear that the demo directory is also used for tests :p ?

marcus-oscarsson:
Looks good

PR #1385, opened 2024-09-02
working with sessions

meguiraun:
here at MAXIV we dont necessarily have sessions in LIMS at the time the user logins, so we create on demand, hence [this change](https://github.com/mxcube/mxcubeweb/blob/d2a4464f6ef54972ca89c86fed2b34fc5e4b8b26/mxcubeweb/core/components/lims.py#L325).

However, the current implementation checks (and creates) session for all user's proposals, which feels an unnecessary step (see deleted code in this MR).

does this breaks other peoples' login procedures?

### DIFF
--- a/mxcubeweb/core/components/lims.py
+++ b/mxcubeweb/core/components/lims.py
@@ -226,10 +226,6 @@ class Lims(ComponentBase):
                 )
                 return ERROR_CODE
 
-            for prop in session["proposal_list"]:
-                todays_session = HWR.beamline.lims.get_todays_session(prop)
-                prop["Session"] = [todays_session["session"]]
-
             if hasattr(
                 HWR.beamline.session, "commissioning_fake_proposal"
             ) and HWR.beamline.session.is_inhouse(loginID, None):
@@ -326,7 +322,7 @@ class Lims(ComponentBase):
             )
 
             todays_session = HWR.beamline.lims.get_todays_session(
-                proposal_info, create_session=False
+                proposal_info, create_session=True
             )
             HWR.beamline.session.session_id = todays_session.get("session").get(
                 "sessionId"
--- a/mxcubeweb/core/components/user/usermanager.py
+++ b/mxcubeweb/core/components/user/usermanager.py
@@ -383,7 +383,11 @@ class UserManager(BaseUserManager):
             raise Exception("Remote access disabled")
 
         # Only allow remote logins with existing sessions
-        if self.app.lims.lims_valid_login(login_res)
+        if (
+            self.app.lims.lims_valid_login(login_res)
+            and is_local_host()
+            and HWR.beamline.lims.loginType.lower() != "user"
+        ):
         and is_local_host()
         and HWR.beamline.lims.loginType.lower() != "user":
             msg = "[LOGIN] Valid login from local host (%s)" % str(info)
--- a/mxcubeweb/core/components/user/usermanager.py
+++ b/mxcubeweb/core/components/user/usermanager.py
@@ -388,7 +388,6 @@ class UserManager(BaseUserManager):
             and is_local_host()
             and HWR.beamline.lims.loginType.lower() != "user"
         ):
-        and is_local_host()
         and HWR.beamline.lims.loginType.lower() != "user":
             msg = "[LOGIN] Valid login from local host (%s)" % str(info)
             logging.getLogger("MX3.HWR").info(msg)
--- a/mxcubeweb/core/components/user/usermanager.py
+++ b/mxcubeweb/core/components/user/usermanager.py
@@ -383,7 +383,9 @@ class UserManager(BaseUserManager):
             raise Exception("Remote access disabled")
 
         # Only allow remote logins with existing sessions
-        if self.app.lims.lims_valid_login(login_res) and is_local_host():
+        if self.app.lims.lims_valid_login(login_res)
+        and is_local_host()
+        and HWR.beamline.lims.loginType.lower() != "user":
             msg = "[LOGIN] Valid login from local host (%s)" % str(info)
             logging.getLogger("MX3.HWR").info(msg)
         elif self.app.lims.lims_valid_login(
--- a/mxcubeweb/core/components/user/usermanager.py
+++ b/mxcubeweb/core/components/user/usermanager.py
@@ -384,9 +384,6 @@ class UserManager(BaseUserManager):
 
         # Only allow remote logins with existing sessions
         if self.app.lims.lims_valid_login(login_res) and is_local_host():
-            if not self.app.lims.lims_existing_session(login_res):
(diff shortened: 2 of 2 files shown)


mxcubeweb/core/components/lims.py
```diff
@@ -326,7 +322,7 @@ def select_proposal(self, proposal):
             )
 
             todays_session = HWR.beamline.lims.get_todays_session(
-                proposal_info, create_session=False
+                proposal_info, create_session=True
```
marcus-oscarsson, line 325: Hm, the sessions should already exist when its being selected, at least with the current logic. `select_proposal` will for instance be called when the application is fetched so creating a session here would break that logic. A session otherwise created above: 

https://github.com/mxcube/mxcubeweb/pull/1385/files#diff-3c54a0ef3e4a4efd2c578368ac1933aa853c8f8553112067cc4c75058a25b476R242

The problem is maybe in your case that you also call this logic when a user selects a proposal when login-in as user ? In that case, would it be possible to add an argument `create_proposal` from that route. I imagine its post to `/porposal` ?

mxcubeweb/core/components/user/usermanager.py
```diff
@@ -384,9 +384,6 @@ def _login(self, login_id: str, password: str):
 
         # Only allow remote logins with existing sessions
         if self.app.lims.lims_valid_login(login_res) and is_local_host():
-            if not self.app.lims.lims_existing_session(login_res):
```
marcus-oscarsson, line 387: This would break the idea of that a session needs to exist to be able to login with proposal based logins. We need to check the login type in that case

mxcubeweb/core/components/user/usermanager.py
```diff
@@ -384,9 +384,6 @@ def _login(self, login_id: str, password: str):
 
         # Only allow remote logins with existing sessions
         if self.app.lims.lims_valid_login(login_res) and is_local_host():
```
marcus-oscarsson:
```suggestion
        if self.app.lims.lims_valid_login(login_res)
        and is_local_host()
        and HWR.beamline.lims.loginType.lower() != "user":
```

mxcubeweb/core/components/lims.py
meguiraun: `get_todays_session` only creates a session if it does not exist. So in my view, having a True here is safe in any situation (at this point the user has made a decision, the proposal to be used).

the https://github.com/mxcube/mxcubeweb/blob/develop/mxcubeweb/core/components/user/usermanager.py#L331 has the flag False always, so in our case the session is not going to be created, plus it would have been created for potentially a proposal that is not the one to be selected just after

sure we can add an argument to the call, but I still dont get why... 

Are you using user account based login? Maybe easier if we discuss this during tomorrow's meeting

mxcubeweb/core/components/user/usermanager.py
```diff
@@ -383,10 +383,9 @@ def _login(self, login_id: str, password: str):
             raise Exception("Remote access disabled")
 
         # Only allow remote logins with existing sessions
-        if self.app.lims.lims_valid_login(login_res) and is_local_host():
-            if not self.app.lims.lims_existing_session(login_res):
-                login_res = self.app.lims.create_lims_session(login_res)
-
+        if self.app.lims.lims_valid_login(login_res)
```
marcus-oscarsson:
```suggestion
        if (
            self.app.lims.lims_valid_login(login_res)
            and is_local_host()
            and HWR.beamline.lims.loginType.lower() != "user"
        ):
```

marcus-oscarsson: Sorry I got the indentation wrong

mxcubeweb/core/components/lims.py
marcus-oscarsson: We are starting to use the user based logins and we are currently evaluating the SSO integration. There will soon be PR with our draft implementation, and we will see exactly how to continue from there. 

Regarding the above, Its not entirely clear to me and it might need some testing to verify. I think one issue is that the call to get_proposal_info, above does not check with LIMS for created proposals after login, https://github.com/mxcube/mxcubeweb/blob/a8e938ab56958d3a5f29e1c3b0c3e0425b5538d9/mxcubeweb/core/components/lims.py#L304

So that `proposal_info` is not updated and a new session is created.

Meaning that, I guess, any call to `get_todays_session`, would create a new session. The session I guess would be created on the LIMS side but would not be seen by mxcube until a new login is made. I think that is why the flag is `false`.

The logic here is a bit fragile and we've not reworked this as we are also working on the SSO integration.

I believe you if you say that youv'e tested and that it works

mxcubeweb/core/components/user/usermanager.py
```diff
@@ -383,10 +383,13 @@ def _login(self, login_id: str, password: str):
...
-        if self.app.lims.lims_valid_login(login_res) and is_local_host():
-            if not self.app.lims.lims_existing_session(login_res):
-                login_res = self.app.lims.create_lims_session(login_res)
-
+        if (
+            self.app.lims.lims_valid_login(login_res)
+            and is_local_host()
+            and HWR.beamline.lims.loginType.lower() != "user"
+        ):
+        and is_local_host()
```
marcus-oscarsson:
```suggestion
```

marcus-oscarsson: Sorry, missed this extra line

mxcubeweb/core/components/user/usermanager.py
```diff
@@ -383,10 +383,12 @@ def _login(self, login_id: str, password: str):
...
-        if self.app.lims.lims_valid_login(login_res) and is_local_host():
-            if not self.app.lims.lims_existing_session(login_res):
-                login_res = self.app.lims.create_lims_session(login_res)
-
+        if (
+            self.app.lims.lims_valid_login(login_res)
+            and is_local_host()
+            and HWR.beamline.lims.loginType.lower() != "user"
+        ):
+        and HWR.beamline.lims.loginType.lower() != "user":
```
marcus-oscarsson, line 391:
```suggestion
            if not self.app.lims.lims_existing_session(login_res):
                login_res = self.app.lims.create_lims_session(login_res)
```

marcus-oscarsson: Ok, final attempt, with these suggestions for now, You get the idea :)

marcus-oscarsson:
Where are you with this PR at this moment ?

marcus-oscarsson:
Closing this for the time being, feel free re-open.

elmjag:
> Closing this for the time being, feel free re-open.

We don't need to merge this anymore. We solved this differently now.

marcus-oscarsson:
:), I sort of understood that :+1:

PR #1390, opened 2024-09-09
Fixes  issue #1389 (Saving grid crashes front end)

### LINKED ISSUE #1389: Saving grid crashes front end
Creating a grid and then saving it crashes the front end.

image

marcus-oscarsson:
Closes #1389 

Returning `false` when result is `null` using optional chaining operator.

@meguiraun, would that work for you ?

### DIFF
--- a/ui/src/components/SampleView/DrawGridPlugin.jsx
+++ b/ui/src/components/SampleView/DrawGridPlugin.jsx
@@ -375,7 +375,7 @@ export default class DrawGridPlugin {
     let color = gridData.selected
       ? 'rgba(136, 255, 91, 1)'
       : 'rgba(228, 255, 9, 0.5)';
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
@meguiraun Let me know if you also had an issue with this and if the fix works for you ?

meguiraun:
sorry for my late reply, I have been away. 

I did not see this, maybe because I did the development at another commit... but looking at the fix that would be strange.

marcus-oscarsson:
:), No problem thanks for your reply. The problem was actually due to a change in `mxcubecore`. As you know `result` used to be an emtpy list and since:

https://github.com/mxcube/mxcubecore/pull/1004/files#diff-9dcb8055597b7dac5aff21ed156197f1a115664df4df6f0b14e8fcd24cc03325R547

Its no longer an empty list but `None`, meaning `null` in Javascript world. So we needed to handle that `null` result somehow.

PR #1391, opened 2024-09-09
Mesh grids not treated as mesh grids

dominikatrojanowska:


dominikatrojanowska:
Issue solved for now: 
context menu after right clicking on a new grid that overlaps with the old ones seems to behave incorrectly (like for a selected group of shapes)
- clicking on a grid: 
2024-09-06 09_51_17-MxCuBE-Web – idtest0
- clicking on an overlapped area: 
2024-09-06 09_50_55-MxCuBE-Web – idtest0


Fixed - when drawing, centering, measuring etc, canvas selection option is disabled:
- clicking on a grid: 
image
- clicking on an overlapped area: 
image
- clicking out of the grid: 
image

While it's still possible to select many shapes with left clicks and ctrl/shift pressed or dragging the mouse with left button down, when not drawing, centering, measuring etc.: 
image

PR #1392, opened 2024-09-10
set content-type to json for beamline get routes

elmjag:
For beamline 'get attributes' API routes, make sure Content-Type header is `application/json`. The front-end is now checking the content-type header for replies, and assumes an error unless it is set to `application/json`.

This fixes the issue where changing data collection parameters in the Task Dialog breaks the validation, making it impossible to run the job with non-default parameters.

### DIFF
--- a/mxcubeweb/routes/beamline.py
+++ b/mxcubeweb/routes/beamline.py
@@ -26,7 +26,7 @@ def create_get_route(app, server, bp, adapter, attr):
             **{"return": getattr(app.mxcubecore.get_adapter(name), attr)(**args)}
         )
 
-        return make_response(result.json(), 200)
+        return make_response(result.dict(), 200)
 
     route_url = f"{atype}/<string:name>/{attr}"
     endpoint = f"{atype}_{attr}"


fabcor-maxiv:
Naive question:

Should it be done systematically on all endpoints? Do we have endpoints that should not return JSON?

mxcubeweb/routes/beamline.py
```diff
@@ -26,7 +26,7 @@ def get_func(name):
             **{"return": getattr(app.mxcubecore.get_adapter(name), attr)(**args)}
         )
 
-        return make_response(result.json(), 200)
+        return make_response(result.dict(), 200)
```
axelboc, line 29: How this change leads to the `Content-Type` header being set to `application/json` is just hilarious :joy_cat:

axelboc:
> Should it be done systematically on all endpoints? Do we have endpoints that should not return JSON?

Ideally yes, similarly to how exceptions should always lead to 500 errors (#1355). Needs to be done with care, though. If I recall, some error responses (sometimes with the wrong 200 response code) include a message as raw text instead of JSON, for instance.

mxcubeweb/routes/beamline.py
marcus-oscarsson: Indeed, one would have expected `.json()` to also set `Content-Type` to `application/json`

marcus-oscarsson: Well, actually, on second thought it somehow makes sense. I guess `json()` returns a string so the content is interpreted as text, while a dictionary always can be interpreted as json when returned as a response

fabcor-maxiv (approved):
I am fine with this.

I think we should look into checking that all endpoints that send output should send it as JSON, always, even for errors (5xx, 4xx). And that the JavaScript UI frontend is ready to handle JSON in all cases.

https://github.com/mxcube/mxcubeweb/issues/1393

mxcubeweb/routes/beamline.py
elmjag: This is Flask logic. By default, if you give it a string, it creates a text response. If you give it a dictionary, it makes it a `application/json` response.

And yes, `result` is a pydantic model (I think), so it's `json()` method returns a string.

marcus-oscarsson: Yeah, sure it does make sense.

PR #1397, opened 2024-09-13
Refactoring GridForm

walesch-yan:
This PR includes some refactoring and clean-up on the `GridForm` and `SampleImage` components

Changes include:
  - Some button adujstments: `+` and `-` now have the same size, radio buttons use `onChange` instead of `onClick`
  -  Changed `GridForm` from a class component to a functional component
  -  Previously right-clicking on the `GridForm` would open the `SampleImage` context menu which also happened when the form was moved away from the canvas element
Before:  
   Broken_ContextMenu
Now:
   Fixed_contextMenu
  - Previously upon closing and opening the form, it would always cover parts of the video view even when previously dragged 
away. Now it remembers its old position.
Before:
Broken_position
Now:
Fixed_position
  - removed the `hideGridForm` function from `SampleImage`, which was never used, since this is handled by the GridForm's `show` variable.
  - `GridForm` had some advanced settings, that were not used. These allowed the setting of the Cell space, I removed code from `SampleImage` and `GridForm` related to this.

### DIFF
--- a/ui/src/components/SampleView/GridForm.jsx
+++ b/ui/src/components/SampleView/GridForm.jsx
@@ -13,14 +13,12 @@ export default function GridForm(props) {
   const {
     getGridOverlayOpacity,
     gridList,
-    gridResultType,
     removeGrid,
     rotateTo,
     saveGrid,
     selectedGrids,
     selectGrid,
     setGridOverlayOpacity,
-    setGridResultType,
     show,
     toggleVisibility,
   } = props;
@@ -182,36 +180,6 @@ export default function GridForm(props) {
                   defaultValue={getGridOverlayOpacity()}
                   onChange={setGridOverlayOpacity}
                   name="overlaySlider"
-                />
-              </Col>
-            </Form.Group>
-            <Form.Group className="mb-2" as={Row}>
-              <Col sm="4">
-                {' '}
-                <Form.Label>Heat map</Form.Label>{' '}
-              </Col>
-              <Col sm="1"> : </Col>
-              <Col sm="7">
-                <Form.Check
-                  name="resultType"
-                  type="radio"
-                  onChange={() => setGridResultType('heatmap')}
-                  checked={gridResultType === 'heatmap'}
-                />
-              </Col>
-            </Form.Group>
-            <Form.Group as={Row}>
-              <Col sm="4">
-                {' '}
-                <Form.Label>Crystal map</Form.Label>{' '}
-              </Col>
-              <Col sm="1"> : </Col>
-              <Col sm="7">
-                <Form.Check
-                  name="resultType"
-                  type="radio"
-                  onChange={() => setGridResultType('crystalmap')}
-                  checked={gridResultType === 'crystalmap'}
                 />
               </Col>
             </Form.Group>
--- a/ui/src/components/SampleView/GridForm.jsx
+++ b/ui/src/components/SampleView/GridForm.jsx
@@ -1,6 +1,6 @@
 /* eslint-disable react/jsx-key */
 /* eslint-disable jsx-a11y/control-has-associated-label */
-import React, { useEffect, useRef } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { Row, Col, Form, Button, Table } from 'react-bootstrap';
 import Draggable from 'react-draggable';
 
@@ -30,6 +30,7 @@ export default function GridForm(props) {
   } = props;
 
   const draggableRef = useRef(null);
+  const [position, setPosition] = useState({ x: 20, y: 64 });
 
   useEffect(() => {
     const draggableElement = draggableRef.current;
--- a/ui/src/components/SampleView/GridForm.jsx
+++ b/ui/src/components/SampleView/GridForm.jsx
@@ -72,6 +72,7 @@ export default class GridForm extends React.Component {
           <td>
             <Button
               size="sm"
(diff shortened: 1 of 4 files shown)


walesch-yan:
Can it be confirmed that nobody uses the advanced settings for changing the cell space, from the code it looked like the flag is never set to true.

walesch-yan:
Also I was wondering if the crystal map feature is used or if it also could be removed?

ui/src/components/SampleView/GridForm.jsx
```diff
@@ -135,93 +134,86 @@ export default class GridForm extends React.Component {
...
-              </Form.Group>
-            </Form>
-          </Col>
-        </Row>
-      </Draggable>
-    );
-
-    return this.props.show ? gridForm : null;
-  }
+  return show ? (
```
axelboc: You can use an early return here to reduce nesting:

```tsx
if (!show) {
  return null;
}

return ...
```

ui/src/components/SampleView/GridForm.jsx
```diff
@@ -135,93 +134,86 @@ export default class GridForm extends React.Component {
...
+    <Draggable
+      position={position}
```
axelboc, line 144: If you can keep the component [uncontrolled](https://react.dev/learn/sharing-state-between-components#controlled-and-uncontrolled-components) using `defaultPosition`, it will save a lot of re-renders when dragging:

```tsx
defaultPosition={position}
```

axelboc: Actually, never mind. Everything will re-render anyway because of the local state. You could store the position in a ref to really avoid re-renders, but it's not worth it.

ui/src/components/SampleView/GridForm.jsx
```diff
@@ -1,42 +1,63 @@
...
+    if (draggableElement) {
+      draggableElement.addEventListener('contextmenu', handleContextMenu);
+    }
+
+    return () => {
+      if (draggableElement) {
+        draggableElement.removeEventListener('contextmenu', handleContextMenu);
+      }
+    };
+  }, [show]);
```
axelboc, line 43: Did you try simply passing `onContextMenu={evt => evt.stopPropagation()}` to the `Row` element?

https://stackoverflow.com/a/56328162/758806

walesch-yan: Yes I actually did, but it didn't change the behavior because the parent element (in `SampleImage`) doesn't use onContextMenu either, but addEventlistener, but now that you mention it, maybe changing the behaviour of the parent element here would have been a better solution

ui/src/components/SampleView/GridForm.jsx
walesch-yan: Well spotted thanks!

ui/src/components/SampleView/GridForm.jsx
walesch-yan: Upon this comment I tried changing the event call from the parent element from addEventListener('contextmenu') to onContextMenu, however this changed the offset position of where the event happened, which is relatively crucial for the behavior of the event. I am however not too sure about why this change happens, but it seems that changing the parent event call is not an option in this case

ui/src/components/SampleView/GridForm.jsx
walesch-yan: Changed this in the latest push

walesch-yan:
As mentioned previously I also removed the radio buttons and every action and mention to changing the `gridResultType`, as from my understanding only the heatmap is used

ui/src/components/SampleView/GridForm.jsx
axelboc: Hmm, so `stopPropagation` prevents triggering an event handler on a parent element. Maybe you could try with `stopImmediatePropagation` as well to prevent triggering handlers on the same element (I think React actually registers the event handlers on the root element of the application):

```ts
onContextMenu={evt => {
  evt.stopImmediatePropagation();
  evt.stopPropagation();
}
```

(If it works then try with `stopImmediatePropagation` alone.)

walesch-yan: Tried that as well earlier, unfortunately this never prevented the unwanted behavior. From my understanding the problem is that React's synthetic event system does not prevent calls to events called on the native event systems. The event call of the parent element looks like this
```
    const imageOverlay = document.querySelector('#insideWrapper');
    imageOverlay.addEventListener('contextmenu', this.rightClick, false);
```
When I tried to change it to use `onContextMenu` (i.e. `onContextMenu={this.rightClick}`) instead, it actually worked with your proposed solution, but unfortunately that added another issue of the canvas context menu being misplaced (always on the top left corner of the canvas element).

axelboc: Haha okay, thanks for trying all that anyway. No worries, use what you have for now and we'll revisit this `querySelector` at a later time. Maybe just put a link to this thread in a comment above your `useEffect` so we can remember this discussion.

marcus-oscarsson:
Thanks @walesch-yan, nicely done !

elmjag:
> Can it be confirmed that nobody uses the advanced settings for changing the cell space, from the code it looked like the flag is never set to true.

Sorry being late to the discussion. We are actually using V-Space and H-Space controls at MAX IV.

I have made this issue: https://github.com/mxcube/mxcubeweb/issues/1443

PR #1398, opened 2024-09-17
Pydantic version 1 and 2 compatability

marcus-oscarsson:
Adding compatibility between Pydantic v1 and v2. We are using [spectree](https://github.com/0b01001001/spectree?tab=readme-ov-file) which for the moment only supports version 1.x of Pydantic.

```
Validate query, JSON data, response data with [pydantic](https://github.com/samuelcolvin/pydantic/) 😉

    If you're using Pydantic V2, you will need to import the BaseModel from pydantic.v1 to make it compatible
```

Pydantic provides a `v1` namespace to make the migration to v2 easier and that we have to use to be able to continue to use `spectree`. The `v1` namespace is available from version `v1.10.17`

I don't see an issue with it for the moment but it would of course be nice if `spectree` eventually could support Pydantic v2

An alternative to `spectree` could be f[lask-openapi3](https://github.com/luolingchun/flask-openapi3?tab=readme-ov-file)

This PR is dependent on https://github.com/mxcube/mxcubecore/pull/1020


pyproject.toml
```diff
@@ -37,11 +37,11 @@ mock = "^4.0.3"
 scandir = "^1.10.0"
 redis = "^4.3.5"
 jsonschema = "^4.17.1"
-pydantic = "^1.10.2"
+pydantic = "^1.10.18"
 PyDispatcher = "^2.0.6"
 pytz = "^2022.6"
 tzlocal = "^4.2"
-mxcubecore = ">=1.145.0"
+mxcubecore = ">=1.153.0"
```
fabcor-maxiv: I guess this needs the version that will be released after the corresponding change in mxcubecore is merged, right?

fabcor-maxiv (changes requested):

pyproject.toml
marcus-oscarsson: Yes, but the tests were failing with the older version of `mxcubecore` because of an update to the `AbstractBeam` interface. That is also why all other PR's having failing tests (I think)

fabcor-maxiv: Ah!?

This? <https://github.com/mxcube/mxcubeweb/actions/runs/10904178365/job/30259973161#step:8:65>

```none
>       assert ap == actual_aperture
E       AssertionError: assert 'A100' == 'slits'
E         - slits
E         + A100
```

Should be addressed separately: <https://github.com/mxcube/mxcubeweb/issues/1401>.

elmjag: This is an intermittent issue. I just run tests on latest mxcubeweb with mxcubecore 1.145.0 locally. All tests passed.

elmjag: I should add I did manage to make it fail as well, with the above assertion, after running the test many times.

beteva: The problem comes from the evaluation of the size of the slits. If the slits are smaller than the aperture, than the beam size is taken from the slits and the test_set_aperture fails systematically with 'A100' == 'slits'.  There are three solutions in order the tst to pass. Either: 
 1. Remove the slits from the beam.xml
 or
2. Set the slits bigger (or at least equal) of the biggest aperture (A100). The values are hardcoded in  the SlitsMockup, so we need to change the mockup to take into account the values, set in the configuration.
or
3. Remove the biggest aperture from the tests.

While waiting for the SlitMockup to be changed and to make it take the values from the configuration file, I vote for 1.

fabcor-maxiv: Feels like it would make sense to discuss further on the dedicated ticket: https://github.com/mxcube/mxcubeweb/issues/1401

beteva: > This is an intermittent issue. I just run tests on latest mxcubeweb with mxcubecore 1.145.0 locally. All tests passed.

The problem is intermittent, because the test are using random aperture size, so when A100 is not used, the tests pass

fabcor-maxiv: @beteva Please, discuss this on <https://github.com/mxcube/mxcubeweb/issues/1401>.

fabcor-maxiv (approved):

PR #1408, opened 2024-09-19
Refactor misc components

axelboc:
Just a whole bunch of refactorings, mainly converting class-based components to function components and replacing `connect()` with `useSelector` + `useDispatch`. I did notice and fixed two bugs though (see comments).


ui/src/components/Equipment/GenericEquipment.jsx
```diff
@@ -1,57 +1,28 @@
...
-              {renderEquipmentState()}{' '}
-              {renderCollapsibleHeaderClose(styles.generic_equipment_arrow_p)}
-            </div>
-          }
-          triggerWhenOpen={
-            <div>
-              {' '}
-              {renderEquipmentState()}{' '}
-              {renderCollapsibleHeaderOpen(styles.generic_equipment_arrow_p)}
-            </div>
```
axelboc, line 49: Instead of rendering two different triggers wen open/closed, I now use the `aria-expanded` attribute to rotate the chevron icon.

ui/src/containers/EquipmentContainer.jsx
```diff
@@ -46,142 +46,136 @@ import HarvesterMaintenance from '../components/Equipment/HarvesterMaintenance';
...
+          <GenericEquipment
+            state={props.sampleChangerState}
+            name={props.contents?.name}
+            initialOpen
+          >
+            {props.contents.name === 'PlateManipulator' ? (
+              <Row className="row">
+                <Col sm={6}>
+                  <PlateManipulator
+                    contents={props.contents}
```
axelboc, line 63: Eventually, all these props coming from the Redux state should go away. Each component should retrieve its own state information with `useSelector`.

ui/src/containers/ResumeQueueDialog.jsx
```diff
@@ -1,54 +1,33 @@
...
-          <Button onClick={this.reject}> OK </Button>
-        </Modal.Footer>
-      </Modal>
-    );
-  }
-}
-
-function mapStateToProps(state) {
-  return {
-    show: state.queue.showResumeQueueDialog,
```
axelboc, line 44: Bug here: this should be `queueGUI`. The `ResumeQueueDialog` was likely never showing up.

ui/src/containers/TaskContainer.jsx
```diff
@@ -1,29 +1,42 @@
...
 
-class TaskContainer extends React.Component {
-  constructor(props) {
-    super(props);
-    this.addTask = this.addTask.bind(this);
-  }
+function TaskContainer() {
+  const dispatch = useDispatch();
+
+  const sampleList = useSelector((state) => state.sampleGrid.sampleList);
```
axelboc, line 21: Same as before, all these will eventually move down into child components once they're all converted to function components.

ui/src/containers/EquipmentContainer.jsx
```diff
@@ -46,142 +46,136 @@ import HarvesterMaintenance from '../components/Equipment/HarvesterMaintenance';
...
-                      contents={this.props.haContents}
-                      handleRefresh={this.props.haRefresh}
-                    />
-                  </Col>
-                  <Col sm={3}>
-                    <HarvesterMaintenance
-                      contents={this.props.haContents}
-                      commands={this.props.haCommands}
-                      global_state={this.props.haGlobal_state}
-                      commands_state={this.props.commands_state}
```
axelboc, line 145: Bug here: `commands_state` refers to `state.sampleChangerMaintenance.commands_state`. I think it's supposed to be `haCommands_state` which refers to `state.harvesterMaintenance.commands_state`.

ui/src/components/Equipment/GenericEquipment.jsx
marcus-oscarsson: Ah, neat !,  and `aria-expanded` is handed by `bootstrap` then I assume, so the only thing you need to do is to add a style ?

ui/src/containers/EquipmentContainer.jsx
marcus-oscarsson: Probably yes, well spotted it seems very likely indeed. Maybe @jbflo can comment when he is back from vacation if its not the case.

marcus-oscarsson:
Thanks @axelboc

ui/src/components/Equipment/GenericEquipment.jsx
axelboc: Indeed. Could have used other solutions too, like a different class when opened, one of the default global classes provided by `react-collapsible`, etc. But yeah the idea is to rotate the icon with CSS instead of rendering two different icons.

PR #1413, opened 2024-09-23
Drop support for fake commissioning proposals in LIMS

elmjag:
We here at MAXIV no longer use this feature with fake commissioning proposals.

As far as we understand, we were the only users of this. Let's get rid of unneeded code.


marcus-oscarsson:
Perfect !

PR #1416, opened 2024-09-25
UI: make mesh grid drawing more performant

elmjag:
Replace per-cell result ellipse fabric objects with one single custom fabric object. When the cells are handful pixels large,    switch to drawing the results as a rectangle. It does not make a visual difference, but is faster.
    
These changes reduces UI latency to usable levels, when drawing meshes with 100k+ cells in `RGB` mode.
    
Also, this PR skips drawing the dashed inner lines, when cells are small. This does not affect performance in any noticeable way, but IMHO makes the display of results easier to comprehend.


marcus-oscarsson:
Pages build takes forever but I guess we can skip that one for now

PR #1418, opened 2024-09-25
Refactor equipment maintenance components

axelboc:
This refers to the `SampleChangerMaintenance`, `HarvesterMaintenance` and `PlateManipulatorMaintenance` components rendered on the _Equipment_ page.

- Move `ActionGroup`, `ActionButton`, `ActionField` into separate files and refactor them slightly (notably, `ActionButton` no longer takes care of sending the maintenance command; it just exposes an `onClick` prop).
- Let `PlateManipulatorMaintenance` use those "Action" components (it was using its own duplicated "Action" components).
- Refactor the "Maintenance" components to function components, refactor them to remove unnecessary render methods (`renderActionGroup`, `renderActionButton`, etc.)
- Let the "Maintenance" retrieve the state they need from the Redux store instead of passing it down from `EquipmentContainer`.

### DIFF
--- a/ui/src/components/Equipment/ActionButton.jsx
+++ b/ui/src/components/Equipment/ActionButton.jsx
@@ -1,0 +1,19 @@
+import React from 'react';
+import { Button } from 'react-bootstrap';
+
+function ActionButton(props) {
+  const { label, disabled, onSend } = props;
+  return (
+    <Button
+      className="me-2"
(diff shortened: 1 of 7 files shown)


marcus-oscarsson:
Ahh, it always feels nice with a good cleanup :) Thanks !

PR #1419, opened 2024-09-26
Extract and refactor clear queue dialog

axelboc:
Still continuing to convert class components. Here I'm mainly focusing on the dialog that appears when clicking on the "Clear sample list" button to confirm the action. I'm creating a dedicated component for this dialog so it is self-contained and can be extracted from the `SampleListViewContainer` component.

I'm also tweaking the UX of the dialog slightly:

- Narrower dialog
- Dialog can be dismissed by clicking outside
- Title changed to match trigger button wording
- "Ok" button renamed "Clear"
- "Clear" button now uses the `warning` theme variant
- "Clear" button now comes before "Cancel" button (more common pattern, and means it's focusable first)
- I've reworded the text to highlight the fact that the action also clears the queue ~and unmounts the current sample, if any~.

The last point makes me wonder whether the naming of the "Clear sample list" button is right... The main action is in fact the clearing of the queue... I'll open an issue to discuss this.

image

image

### DIFF
--- a/ui/src/components/GenericDialog/ConfirmActionDialog.jsx
+++ b/ui/src/components/GenericDialog/ConfirmActionDialog.jsx
@@ -1,56 +1,46 @@
-/* eslint-disable react/jsx-handler-names */
 import React from 'react';
 import { Modal, Button } from 'react-bootstrap';
 
-export default class ConfirmActionDialog extends React.Component {
-  constructor(props) {
-    super(props);
-    this.onOkClick = this.onOkClick.bind(this);
-    this.onCancelClick = this.onCancelClick.bind(this);
-  }
+function ConfirmActionDialog(props) {
+  const {
+    title,
+    okBtnLabel = 'OK',
+    cancelBtnLabel = 'Cancel',
+    show = false,
+    children,
+    onHide,
+    onOk,
+    onCancel,
+  } = props;
 
-  onOkClick() {
-    if (this.props.onOk) {
-      this.props.onOk();
-    }
-
-    this.props.hide();
-  }
-
-  onCancelClick() {
-    if (this.props.onCancel) {
-      this.props.onCancel();
-    }
-
-    this.props.hide();
-  }
-
-  render() {
-    return (
-      <Modal show={this.props.show}>
-        <Modal.Header>
-          <Modal.Title>{this.props.title}</Modal.Title>
-        </Modal.Header>
-        <Modal.Body>{this.props.message}</Modal.Body>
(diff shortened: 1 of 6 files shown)


ui/src/components/SampleGrid/ClearQueueDialog.jsx
```diff
@@ -0,0 +1,27 @@
...
+  return (
+    <ConfirmActionDialog
+      title="Clear sample list?"
+      okBtnLabel="Clear"
+      show={show}
+      onOk={() => dispatch(clearQueue())}
+      onHide={() => dispatch(showConfirmClearQueueDialog(false))}
+    >
+      This will also <strong>clear the queue</strong> and{' '}
+      <strong>unmount</strong> the current sample.
```
meguiraun: do we actually umount the sample? or it simply dissapears from the UI?

ui/src/components/GenericDialog/ConfirmActionDialog.jsx
axelboc, line 1: I kept this generic `ConfirmActionDialog` component even though it's only used by the `ClearQueueDialog`. I think it's a good abstraction so it seemed worth keeping.

ui/src/components/Main.jsx
```diff
@@ -54,6 +55,7 @@ function Main() {
       )}
 
       <SelectProposalContainer />
+      <ClearQueueDialog />
```
axelboc, line 58: Sure, this dialog only makes sense in the context of the _Samples_ page, but it doesn't really matter since its state is managed globally. I prefer to have all the dialogs in one place. It might uncover opportunities to simplify the global state and combine a few dialogs together into more general abstractions.

ui/src/components/SampleGrid/ClearQueueDialog.jsx
axelboc: Excellent question :smile:

axelboc: Right, it doesn't. If you sync again, the sample reappears.

axelboc: I've reworded. I think clearing the mounted sample from the UI is less than ideal but this can be discussed in #1420.

ui/src/components/Main.jsx
marcus-oscarsson: Yes, it would be nice to only have, if possible, one main dialog component

marcus-oscarsson:
Thanks !

PR #1421, opened 2024-09-26
Abstract away tooltip overlay triggers

axelboc:
React Bootstrap's abstractions for tooltips is quite verbose. I'm adding a small component to make this a bit more palatable.

I'm also mitigating a very annoying issue, which is that when you click for instance on the "Clear sample list" button and then dismiss the dialog, the button receives focus again, which shows the tooltip.

That's because React Bootstrap's tooltips appear both on hover and on focus for accessibility; this is good but the way it's implemented is far from perfect, since when you close the dialog, you want to give focus back to the button but not show the tooltip... I couldn't find a workaround for this, unfortunately, but a way to mitigate it is to at least close the tooltip when clicking anywhere else on the screen. This is done with the `rootClose` prop, which was previously set on some tooltips but not all.

Peek 2024-09-26 14-40

### DIFF
--- a/ui/src/components/Equipment/PlateManipulator.jsx
+++ b/ui/src/components/Equipment/PlateManipulator.jsx
@@ -6,12 +6,12 @@ import {
   ButtonToolbar,
   OverlayTrigger,
   Popover,
-  Tooltip,
 } from 'react-bootstrap';
 import { contextMenu, Menu, Item, Separator } from 'react-contexify';
 
 import { MdSync } from 'react-icons/md';
 import styles from './equipment.module.css';
+import TooltipTrigger from '../TooltipTrigger';
 
 const strokeColor = 'rgb(136, 136, 136)';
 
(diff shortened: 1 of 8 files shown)


ui/src/components/Equipment/PlateManipulator.jsx
```diff
@@ -634,33 +634,23 @@ export default function PlateManipulator(props) {
...
       <Col className="ms-3">
         <ButtonToolbar className="ms-4">
-          {!inPopover ? (
+          {!inPopover && (
             <div className="me-4">
               <b>{cplate_label}</b>
             </div>
-          ) : null}
-          <OverlayTrigger
-            variant="outline-success"
```
axelboc, line 643: I don't think this was a valid prop and had any effect.

ui/src/components/Equipment/PlateManipulator.jsx
```diff
@@ -634,33 +634,23 @@ export default function PlateManipulator(props) {
...
-            placement="bottom"
-            overlay={
-              <Tooltip id="select-samples">
-                Refresh if Plate Location not Updated
-              </Tooltip>
-            }
+          )}
+          <TooltipTrigger
+            id="refresh-tooltip"
```
axelboc, line 643: I changed all the IDs as well, since some of them had been copy-pasted and did not match the purpose of the tooltip.

axelboc: When we make the switch to React 18, we can generate random IDs with [`useId()`](https://react.dev/reference/react/useId)

marcus-oscarsson:
Thanks !

PR #1422, opened 2024-09-26
Use `&&` operator to render JSX conditionally where possible

axelboc:
I see a lot of ternaries in the JSX where the false branch renders `null`. This can often be simplified with the [`&&` operator](https://react.dev/learn/conditional-rendering#logical-and-operator-). The only pitfall is that if the condition resolves to `0`, then React  happily renders that — otherwise, if it's an any other falsy value (empty string, `false`, `null`, etc.), it renders nothing.

So in this PR, I'm going over and refactoring all the cases that are 100% safe, and only those — that is, where the condition resolves necessarily to a boolean (`===`, `>`, `!`, etc.) For the rest, it will have to be done on a case-by-base basis (or after we migrate to TypeScript).

### DIFF
--- a/ui/src/components/Equipment/PlateManipulator.jsx
+++ b/ui/src/components/Equipment/PlateManipulator.jsx
@@ -278,7 +278,7 @@ export default function PlateManipulator(props) {
           >
             Move to drop {selectedDrop}
           </Item>
-          {crystalForSelectedWell?.shelf === selectedDrop ? (
+          {crystalForSelectedWell?.shelf === selectedDrop && (
             <>
               <Separator />
               <Item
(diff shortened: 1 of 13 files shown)


marcus-oscarsson:
That's nice, good initiative !

ui/src/components/Equipment/PlateManipulator.jsx
```diff
@@ -638,14 +638,9 @@ export default function PlateManipulator(props) {
...
           )}
-          <OverlayTrigger
-            variant="outline-success"
-            placement="bottom"
-            overlay={
-              <Tooltip id="select-samples">
-                Refresh if Plate Location not Updated
-              </Tooltip>
-            }
+          <TooltipTrigger
```
marcus-oscarsson: Did I get this right @axelboc ?

marcus-oscarsson:
I can't seem to rebase this, from GitHub, can you do it @axelboc ?

PR #1425, opened 2024-09-27
Pin micromamba version in GitHub Actions workflows

### LINKED ISSUE #1424: CI is sloooow
You've probably noticed it. Not sure if it's a temporary slow down or if it's something to do with the recent change to test all supported Python versions. https://github.com/mxcube/mxcubeweb/actions/runs/11069955541

image

fabcor-maxiv:
Pin version of micromamba to `1.5.10-0` in GitHub Actions workflows because of following issue:
https://github.com/mamba-org/setup-micromamba/issues/225

GitHub: fix https://github.com/mxcube/mxcubeweb/issues/1424

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -41,12 +41,14 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      # Pin micromamba version because of following issue:
+      # https://github.com/mamba-org/setup-micromamba/issues/225
       - name: Set up Conda environment
         uses: mamba-org/setup-micromamba@v1
         with:
           create-args: >-
             python=${{ matrix.python-version }}
-          micromamba-version: "latest"
+          micromamba-version: 1.5.10-0
           environment-file: conda-environment.yml
           cache-environment: true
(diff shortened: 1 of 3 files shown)


fabcor-maxiv:
As said in <https://github.com/mxcube/mxcubeweb/issues/1424#issuecomment-2379418620>, I am a bit reluctant to pin versions like this. If I knew for sure that Dependabot would help us keep this pin up-to-date, then it would not be an issue, but I doubt Dependabot can do that.

We can still merge this, and hopefully we will think about updating this pinned version once in a while.

fabcor-maxiv:
It was necessary to clear the GitHub Actions caches and retrigger the workflows (I amended the commit and force pushed).

fabcor-maxiv:
The `pages-build` still took a very long time before starting, not clear why. Maybe unrelated issue. Maybe transient issue.

marcus-oscarsson:
I agree but I guess we don't have lots of options for the moment, we just have to make sure to remember to unpin later. I think dependabot can handle parts of the actions but I'm not sure to what extent.

Thanks, well done !

PR #1428, opened 2024-09-30
[react-contexify]  update to 6.0.0

### LINKED ISSUE #1426: Seems like [react-contexify] Context Menu stopped working
[react-contexify] Context Menu stopped working on components/Equipment/Harvester.jsx && components/Equipment/PlateManipulator.jsx . 

Maybe in there also components/SampleQueue/CurrentTree.jsx

@marcus-oscarsson  can you check if this the case for the render of  components/SSXChip/SSXChip.jsx

it's like the Context are are not showing at all, or show from a unknown position.  

*** Edited add error Message from the library  ***
image

jbflo:
This fix [Issue](https://github.com/mxcube/mxcubeweb/issues/1426)

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -37,7 +37,7 @@
     "react-bootstrap": "2.8.0",
     "react-chat-widget": "^3.1.4",
     "react-collapsible": "^2.10.0",
-    "react-contexify": "^5.0.0",
+    "react-contexify": "6.0.0",
     "react-copy-to-clipboard": "^5.1.0",
     "react-dom": "^17.0.2",
     "react-draggable": "^4.4.4",
--- a/ui/pnpm-lock.yaml
+++ b/ui/pnpm-lock.yaml
@@ -63,8 +63,8 @@ importers:
         specifier: ^2.10.0
         version: 2.10.0(react-dom@17.0.2(react@17.0.2))(react@17.0.2)
       react-contexify:
-        specifier: ^5.0.0
-        version: 5.0.0(react-dom@17.0.2(react@17.0.2))(react@17.0.2)
+        specifier: 6.0.0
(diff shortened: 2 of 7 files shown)


axelboc:
There are a bunch of changes in v6, including breaking changes (CSS import path, CSS classes, etc.), might be worth doubling checking those. There may also be opportunities to improve/simplify the code with some of the new features, maybe? :shrug:  https://github.com/fkhadra/react-contexify/releases/tag/v6.0.0

marcus-oscarsson:
@axelboc, yes that would probably be worth doing. What do you think @jbflo ? Ill merge this for now though as it seems to fix the issue, and then we can follow up with some improvements.

marcus-oscarsson:
@jbflo, @axelboc, is it perhaps worth double checking the impact of the backward incompatible changes before merging ?

jbflo:
Yes. We can even remove the use of this library since we had some issues with the positioning in the pass for menu with very big Height.


We have a Generic ContexMenu Component  already we can make the necessary modifications to make it work everywhere.

marcus-oscarsson:
@jbflo, yes I remember that we already discussed. Sounds like a good idea then

jbflo:
> @jbflo, yes I remember that we already discussed. Sounds like a good idea then

I had a look at the Release notes [I should had check before sorry]
I made the change for the css  Breaking changes.  but it's weird that it was still working with the old css path. 

You can Merge this as it is. 
Or 
we can close this one and go on  [all in one] improve/simplify the code on another PR.  
both fine by me.

axelboc:
It's worth merging I think, in case the refactor/migration to the other solution takes longer than expected.

marcus-oscarsson:
Thanks guys !

PR #1429, opened 2024-10-01
Simplify UI for changing step of motor input

### LINKED ISSUE #1311: Step input popover is a bit cumbersome to use
The step input popover has a few flaws:

- It requires too many user interactions to change the step.
- The up/down arrow buttons to adjust the step are apparently rarely used because it's often quicker to enter a value manually (especially since some of the increments are not well configured by default - e.g. 0.1 for Omega is way too small).
- The checkmark button feels unnecessary since one is already typing a value, so pressing <kbd>Enter</kbd> is faster.
- When the step changes, the motor input becomes "dirty" for no apparent reason (i.e. orange background, indicating that the value of the input has changed but hasn't yet been sent to the server).

Peek 2024-07-11 11-14

**Proposal:** Turn the step value into an input field that can be edited directly and committed with <kbd>Enter</kbd> (no up/down arrow buttons). The step input should be marked as dirty when edited but not yet committed (but not the motor input).

This solution would also greatly simplify the logic of the step component, since it currently deals with disabling the popover behaviour when the motor input is already displayed inside a popover:

image

axelboc:
Fix #1311 

I replace the cumbersome popover UI with a simple number input.

Peek 2024-10-01 10-28

Since the step is a front-end state, which is updated synchronously in the Redux store, we can simply update it on every `change` event of the step input. No need to require the user to press <kbd>Enter</kbd> as I initially proposed in #1311.

---

Now, slight problem (which was already present before this change): in the mock environment, I can't seem to get the step inputs of the following "motors" to work:

- Sample Vertical
- Sample Horizontal
- Focus

If I recall, those are actually abstract motors that control multiple physical motors, so perhaps it is a limitation of the mock environment?

Peek 2024-10-01 10-29

Interestingly, if you pay close attention, when I press the arrows, the motor input sometimes receives a value that does not make sense whatsoever for a split millisecond (starting with `3` when pressing the up arrow, or `-0.` when pressing the down arrow) — it seems to overshoot before going back to the initial value +/- 0.1.

### DIFF
--- a/ui/src/components/MotorInput/MotorInput.jsx
+++ b/ui/src/components/MotorInput/MotorInput.jsx
@@ -3,8 +3,8 @@
 import React from 'react';
 import cx from 'classnames';
 import { Button } from 'react-bootstrap';
-import PopInput from '../PopInput/PopInput';
 import { MOTOR_STATE } from '../../constants';
+import styles from './MotorInput.module.css';
 import './motor.css';
 import '../input.css';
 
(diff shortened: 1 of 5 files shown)


ui/src/components/MotorInput/MotorInput.jsx
```diff
@@ -96,114 +96,76 @@ export default class MotorInput extends React.Component {
     return (
       <div className="motor-input-container">
         <p className="motor-name">{this.props.label}</p>
-        <div className="d-flex">
-          <form className="d-flex" onSubmit={this.handleKey} noValidate>
-            <div style={{ display: 'flex', width: '160px' }}>
```
axelboc, line 101: Removing some unnecessary wrappers to keep only the `form`. You may want to hide whitespace changes when reviewing.

marcus-oscarsson:
Thanks, I don't really know why you have that strange intermediate value. I think you are right that its maybe something odd with the mockups, we have to investigate closer.

PR #1432, opened 2024-10-01
Improve beamline attribute controls

axelboc:
My initial goal was to simplify the `PopInput` component, since the only place it was still used after #1429 was in `BeamlineSetupContainer` (for editing beamline attributes: energy, wavelength, etc.) ... but I ended up going quite a bit further.

First, I'm removing three abstract components: `PopIntput`, `LabeledValue` and `NumericInput`, and replacing them with simpler, more specific components for rendering and editing beamline attributes:

- `BeamlineAttribute`: the root component that renders both read-only and editable attribute values;
- `BeamlineAttributeForm`: the input + button that appear when clicking on an editable attribute value.

Second, I'm using a plain, uncontrolled `number` input in `BeamlineAttributeForm` instead of `RcInputNumber` from the `rc-input-number` package (which is now removed). It's one fewer dependency and I honestly couldn't see the added value of `RcInputNumber`.

Third, I'm improving the accessibility and overall UX of the beamline attribute editing UI:

- `button` instead of `a` elements
- focusable with the keyboard
- dotted underline on hover/focus
- increased clickable area
- visible outline when focused with the keyboard
- select input text on open so user can type a new value right away
- re-select input text after a value has been applied
- give focus back to button on close so user can keep using the keyboard to quickly edit another attribute 

Peek 2024-10-01 16-48


ui/src/components/SampleGrid/SampleGridTable.css
```diff
@@ -247,7 +247,7 @@ button[disabled] {
   top: 0;
 }
 
-.btn-link {
+.btn-link:not([data-default-styles]) {
```
axelboc, line 250: This `color` override, which I assume was for link buttons inside the sample grid, is in fact a global style, so it applies to all `<Button variant="link" />`. To avoid breaking anything for now, I'm providing a way to opt-out of this style declaration, like I did with modals.

axelboc (commented):
Big clean-up :grin: 

image

ui/src/components/SampleGrid/SampleGridTable.css
marcus-oscarsson: Nice

marcus-oscarsson:
Very nice cleanup, good idea to `PopIntput`, `LabeledValue` and `NumericInput` with something more specific and in the end simpler :)

PR #1434, opened 2024-10-02
Fix size of step inputs in Chrome

### LINKED ISSUE #1433: Motor value does not show in ui
After the latest changes the UI does not show the current values of some motors. Only the step input is visible.
Screenshot from 2024-10-02 15-14-36Screenshot from 2024-10-02 15-14-54

One exception are the motor fields hidden behind the setting button of sample alignment
Screenshot from 2024-10-02 15-15-09

axelboc:
Fix #1433 

@walesch-yan can you please double check? I'm remote and can't run MXCuBE-Web on my Windows machine. Otherwise, I can investigate tomorrow.

### DIFF
--- a/ui/src/components/BeamlineAttribute/BeamlineAttribute.module.css
+++ b/ui/src/components/BeamlineAttribute/BeamlineAttribute.module.css
@@ -21,6 +21,7 @@
 }
 
 .input {
+  width: 9em;
   height: 36px;
   appearance: auto !important;
 }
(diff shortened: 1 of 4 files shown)


walesch-yan:
This did not solve it

axelboc:
Ok, strange. I'll look at it tomorrow, then. The styles of the `MotorInput` need a deeper clean-up, I think.

axelboc:
Turns out Chrome just ignores the `size` attribute on the `input`...

image

fabcor-maxiv:
But what kind of `<input>` is that? Not all of them have the `size` attribute. Or am I misunderstanding things?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#size

axelboc:
Wow you're right, it's [not valid for `number` inputs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#controlling_input_size). Why it isn't, I don't understand. Anyways, well spotted! :fishing_pole_and_fish: 

I've removed the `size` attribute and used an `em` width in the CSS. I've done the same in `BeamlineAttributeForm`.

PR #1438, opened 2024-10-04
Refactor styling of hardware states

axelboc:
The state-based styling for motor inputs and beamline attribues was previously based on `input-bg-<state>` classes, located in a global stylesheet called `input.css`, which was imported multiple times. This approach led to complicated `className` logic like in `MotorInput.jsx`.

Instead, I now use `data-` attributes, CSS modules, and CSS custom properties (e.g. `--hw-ready--bg`).

Additionally, I noticed a few references to non-existent `MOTOR_STATE` keys, like `ALARM`, `OFFLINE`, etc. so I clean this up. The `MOTOR_STATE` object becomes `HW_STATE` (since it is used by other kinds of hardware) and it now contains only valid [base hardware object states](https://github.com/mxcube/mxcubecore/blob/03c89f2eef8af604b211f5788813df3ad4216138/mxcubecore/BaseHardwareObjects.py#L61). Note that the [abstract motor states](https://github.com/mxcube/mxcubecore/blob/03c89f2eef8af604b211f5788813df3ad4216138/mxcubecore/HardwareObjects/abstract/AbstractMotor.py#L40) actually re-use the base hardware states, so from what I can tell, we never end up with `MOVING` or `LOWLIMIT` in the front-end.

While I'm at it, I improve the `ApertureInput` drop-down so it turns yellow when the aperture is being changed:

Peek 2024-10-04 12-04


ui/src/components/SampleView/ApertureInput.jsx
```diff
@@ -1,38 +1,33 @@
...
 
-  render() {
-    const inputCSS = cx('form-control input-sm', {
-      'input-bg-moving': this.props.aperture === 'Unknown',
-      'input-bg-ready': this.props.aperture !== 'Unknown',
-    });
+  const value = useSelector((state) => state.sampleview.currentAperture);
+  const options = useSelector((state) => state.sampleview.apertureList);
+  const state = useSelector(
+    (state) => state.beamline.hardwareObjects['beam.aperture']?.state,
```
axelboc, line 14: Is the aperture HWO guaranteed to be called `beam.aperture`? I used optional chaining just in case it's not.

ui/src/components/MotorInput/OneAxisTranslationControl.module.css
```diff
@@ -0,0 +1,21 @@
+.input {
+  background-color: var(--hw-ready--bg) !important;
+  transition: background-color 100ms ease-in;
```
axelboc, line 3: Each component redefines its own state-based styles (but using the global CSS custom properties). This gives more flexibility (which HWO states to style, which styles to apply, etc.)

ui/src/components/MotorInput/OneAxisTranslationControl.jsx
```diff
@@ -101,47 +97,26 @@ export default class OneAxisTranslationControl extends React.Component {
...
-          disabled={
-            this.props.state !== MOTOR_STATE.READY || this.props.disabled
-          }
+          disabled={this.props.state !== HW_STATE.READY || this.props.disabled}
           onClick={() => this.stepChange(motorName, step, -1)}
         >
           <i className="fas fa-angle-left" />
         </Button>
-        <input
+        <Form.Control
```
axelboc, line 119: This avoids declaring the `form-control` class manually.

ui/src/constants.js
```diff
@@ -117,34 +117,15 @@ export const SPACE_GROUPS = [
...
   UNKNOWN: 'UNKNOWN',
   WARNING: 'WARNING',
-  MOVING: 'MOVING',
   BUSY: 'BUSY',
   READY: 'READY',
   FAULT: 'FAULT',
   OFF: 'OFF',
 };
-
-export const MOTOR_STATE_DESC = {};
```
axelboc, line 130: Seems unused.

ui/src/containers/BeamlineSetupContainer.jsx
```diff
@@ -151,7 +151,7 @@ function BeamlineSetupContainer(props) {
         <td
           key={`bs-val-${uiprop.attribute}`}
           style={{
-            padding: '0 0.625rem 0 0',
+            padding: '0.125rem 0.625rem 0.125rem 0',
```
axelboc, line 154: Adding a little vertical gap between beamline attribute values so the yellow backgrounds (while values are changing) don't merge into one.

ui/src/components/SampleView/ApertureInput.jsx
marcus-oscarsson: It should always be called beam.aperture, but this fine :)

ui/src/components/BeamlineAttribute/BeamlineAttribute.module.css
```diff
@@ -10,6 +10,11 @@
   text-decoration: none;
 }
 
+.valueBtn[data-busy] {
```
marcus-oscarsson, line 13: Nice with the `data-`

marcus-oscarsson:
Thanks alot, really nice !

ui/src/components/SampleView/ApertureInput.jsx
beteva: > It should always be called beam.aperture, but this fine :)

Is this true, @marcus-oscarsson. For me here it should get the beam size, not the aperture. Maybe we are missing a method in the Beam HO to return the definer state.

marcus-oscarsson: We are displaying the state of the aperture object, which at the moment always is accessed via `beam.aperture`. It should ideally be `beam.state` but as you mentioned there is no such thing at the moment, in the same way `currentAperture` and `appertureList` should  be simple be `currentSize` and `sizeList` or similair.

Issue #1437, opened 2024-10-04
retrieving samples from SC and LIMS Sync


meguiraun:
halo,

for these buttons

<img width="244" alt="image" src="https://github.com/user-attachments/assets/d2b636ef-d33b-437b-8908-799c76c1c413">

- the first one displays all the samples present in the sample changer, no matter if the user should see them or not
- the second one displays only the ones the user is supposed to see

Here at MAX IV we dont want the first one, I was thinking on replacing the first one by the second, and have only one button. The equipment tab still displays all the samples.

Any opinion? (I am sure it works like this since a long time ago)

axelboc:
We've also received user feedback recently saying that the "Get samples from SC" button was maybe obsolete.

meguiraun:
https://github.com/mxcube/mxcubeweb/pull/1439

marcus-oscarsson:
Closing :)

PR #1439, opened 2024-10-04
remove "get samples from SC" button

meguiraun:
Dealing with issue https://github.com/mxcube/mxcubeweb/issues/1437

from 
Screenshot 2024-10-04 at 11 01 34

to:
Screenshot 2024-10-04 at 12 39 28

### DIFF
--- a/ui/src/containers/SampleListViewContainer.jsx
+++ b/ui/src/containers/SampleListViewContainer.jsx
@@ -13,7 +13,6 @@ import {
   Col,
   Form,
   Button,
-  SplitButton,
   DropdownButton,
   InputGroup,
   Dropdown,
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
Seems reasonable to me. I think I would prefer the `Get samples` button before the `Create new sample` button. It somehow feels like `Get samples` will be the one that is used more often ?

meguiraun:
> Seems reasonable to me. I think I would prefer the `Get samples` button before the `Create new sample` button. It somehow feels like `Get samples` will be the one that is used more often ?

sure, makes sense for ux point of view

marcus-oscarsson:
Looks good, to me. You can rebase and merge :)

jbflo:
What about having the get samples trigger automatically .on login for example ? I remember this was asked before .

PR #1441, opened 2024-10-08
Refactor `MotorControls` and `TwoAxisTranslationControl`

axelboc:
As usual, converting to function components and introducing `useSelector`.

One change worth mentioning is that the decision to show the `TwoAxisTranslationControl` (the buttons in a cross in the sidebar) is now based only on the existence of the `uiproperties` objects (with roles `sample_vertical` and `sample_horizontal`) instead of the corresponding hardware objects. This simplifies the logic significantly.

### DIFF
--- a/ui/src/components/MotorInput/TwoAxisTranslationControl.jsx
+++ b/ui/src/components/MotorInput/TwoAxisTranslationControl.jsx
@@ -1,137 +1,153 @@
 import React from 'react';
 import { Button, OverlayTrigger, Popover } from 'react-bootstrap';
-import { HW_STATE } from '../../constants';
+import { HW_STATE, QUEUE_RUNNING } from '../../constants';
 import MotorInput from './MotorInput';
 import './motor.css';
+import { useDispatch, useSelector } from 'react-redux';
+import { setAttribute } from '../../actions/beamline';
+import { setStepSize } from '../../actions/sampleview';
+import { stopBeamlineAction } from '../../actions/beamlineActions';
 
-export default class TwoAxisTranslationControl extends React.Component {
-  constructor(props) {
-    super(props);
-    this.stepChange = this.stepChange.bind(this);
-  }
+function TwoAxisTranslationControl(props) {
+  const { verticalMotorProps, horizontalMotorProps } = props;
 
-  stepChange(name, step, operator) {
-    const { value } = this.props.motors[name];
-    const newValue = value + step * operator;
-    this.props.save(this.props.motors[name].attribute, newValue);
-  }
+  const dispatch = useDispatch();
 
-  renderMotorSettings() {
-    return (
-      <Popover style={{ maxWidth: 'fit-content' }} title="">
-        <Popover.Header as="h3">Sample alignment motors</Popover.Header>
-        <Popover.Body>
-          <MotorInput
-            save={this.props.save}
-            value={this.props.motors.sample_vertical.value}
-            saveStep={this.props.saveStep}
-            step={this.props.motors.sample_vertical.step}
-            motorName={this.props.motors.sample_vertical.attribute}
-            label={this.props.motors.sample_vertical.label}
-            suffix={this.props.motors.sample_vertical.suffix}
-            decimalPoints={this.props.motors.sample_vertical.precision}
-            state={this.props.motors.sample_vertical.state}
-            stop={this.props.stop}
-            disabled={this.props.motorsDisabled}
-            inplace
-          />
-          <MotorInput
-            save={this.props.save}
-            value={this.props.motors.sample_horizontal.value}
-            saveStep={this.props.saveStep}
-            step={this.props.motors.sample_horizontal.step}
-            motorName={this.props.motors.sample_horizontal.attribute}
-            label={this.props.motors.sample_horizontal.label}
-            suffix={this.props.motors.sample_horizontal.suffix}
-            decimalPoints={this.props.motors.sample_horizontal.precision}
-            state={this.props.motors.sample_horizontal.state}
-            stop={this.props.stop}
-            disabled={this.props.motorsDisabled}
-            inplace
-          />
-        </Popover.Body>
-      </Popover>
-    );
-  }
+  const verticalMotor = useSelector(
+    (state) => state.beamline.hardwareObjects[verticalMotorProps.attribute],
+  );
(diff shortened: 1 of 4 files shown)


ui/src/containers/SampleViewContainer.jsx
```diff
@@ -178,20 +168,8 @@ class SampleViewContainer extends Component {
...
-                uiproperties={motorUiProperties}
-                hardwareObjects={motorhardwareObjects}
-                motorsDisabled={
-                  this.props.motorInputDisable ||
-                  this.props.queueState === QUEUE_RUNNING
-                }
-                steps={motorSteps}
-                stop={this.props.stopBeamlineAction}
-                sampleViewActions={this.props.sampleViewActions}
-                sampleViewState={this.props.sampleViewState}
```
axelboc, line 193: Accessing state lower down, where it's actually needed, helps to avoid unused props like these two.

ui/src/containers/SampleViewContainer.jsx
```diff
@@ -57,22 +56,13 @@ class SampleViewContainer extends Component {
...
       return null;
     }
 
-    const motorUiProperties = uiproperties.sample_view.components.filter(
-      ({ value_type }) => value_type === 'MOTOR',
-    );
-
-    const motorhardwareObjects = Object.values(hardwareObjects).filter(
-      ({ type }) => type === 'MOTOR',
-    );
```
axelboc, line 72: `motorhardwareObjects` was then used in `MotorControls` to find the `sample_vertical` and `sample_horiztonal` hardware objects, which were then spread with the UI properties and passed down to `TwoAxisTranslationControl`. This was overly complex, especially because of `Object.values`.

Now I simply retrieve the two hardware objects in `TwoAxisTranslationControl` by their keys:

```ts
  const verticalMotor = useSelector(
    (state) => state.beamline.hardwareObjects[verticalMotorProps.attribute],
  );
  const horizontalMotor = useSelector(
    (state) => state.beamline.hardwareObjects[horizontalMotorProps.attribute],
  );
```

ui/src/components/SampleView/MotorControls.jsx
```diff
@@ -1,113 +1,69 @@
...
+      </Button>
 
-            {this.state.showAll && (
-              <div style={{ marginTop: '0.5rem' }}>
-                {this.renderMotorInputs(3, numel)}
-              </div>
-            )}
-          </div>
-        </div>
-      </Row>
```
axelboc, line 110: Removing a lot of unnecessary wrappers.

ui/src/components/SampleView/MotorControls.jsx
```diff
@@ -1,113 +1,69 @@
...
-    });
-
-    const numel = this.props.uiproperties.length;
-
-    if (!sample_vertical || !sample_horizontal) {
-      return <Row className="row">{this.renderMotorInputs(0, numel)}</Row>;
-    }
+  if (!verticalMotorProps || !horizontalMotorProps) {
+    return motorsProps.map(({ attribute: k, role }) => (
+      <MotorInputContainer key={k} component="sample_view" role={role} />
```
axelboc, line 29: Note that I haven't refactored `MotorInputContainer` and `MotorInput`, so things might evolve further.

ui/src/components/MotorInput/TwoAxisTranslationControl.jsx
```diff
@@ -1,137 +1,153 @@
...
+        placement="right"
+        overlay={
+          <Popover>
+            <Popover.Header as="h3">Sample alignment motors</Popover.Header>
+            <Popover.Body>
+              <MotorInput
+                save={(name, val) => dispatch(setAttribute(name, val))}
+                value={verticalMotor.value}
+                saveStep={(name, val) => dispatch(setStepSize(name, val))}
+                step={verticalMotorProps.step}
```
axelboc, line 81: I think I found why changing the steps of the sample vertical/horizontal motors doesn't work: the value in the step input is read from `state.uiproperties`, but the up/down/left/right buttons read the step from `state.sampleview.motorSteps`.

The `SET_STEP_SIZE` action does save the step in these two places, but when it's read from `state.sampleview.motorSteps`, we use hardcoded keys instead of the motor names:

```js
// Before refactoring
const { sample_verticalStep, sample_horizontalStep } = this.props.steps;

// After refactoring
const { sample_verticalStep, sample_horizontalStep } = useSelector(
  (state) => state.sampleview.motorSteps,
);
```

The `sampleview` reducer stores the updated step in `motorSteps` against the name of the motor (e.g. `diffractometer.phiy`), not against `sample_verticalStep` and `sample_horizontalStep`, so the cross never gets the updated values.

Should I strive to remove this `sampleview.motorSteps` state in my next PR? Or on the contrary, should the step always be read from `sampleview.motorSteps` instead of `uiproperties`?

- One could argue that `uiproperties` is a server state that shouldn't be modified locally, in which case managing the local steps state via `motorSteps` makes sense.
- On the other hand, perhaps the steps state shouldn't be a local state but a server state, in which case accessing it via `uiproperites` makes sense (as long as it's updated by refetching the latest `uiproperties` state from the server).

marcus-oscarsson: Thanks very nice as always :), I think removing `sampleview.motorSteps` seems like a good idea as well as regarding`uiproperites` as a server state. Unless I'm overlooking something.

ui/src/components/SampleView/MotorControls.jsx
marcus-oscarsson: :)

marcus-oscarsson:
Nice cleanup thanks !

ui/src/components/MotorInput/TwoAxisTranslationControl.jsx
axelboc: Ideally, `uiproperties` should not be modified locally – it should always match the state of `uiproperties` on the server, so if we want to keep it as the source of truth for the step size, there should be a route to update the step size in `uiproperties`.

Since this decision has other implications (observer mode, persistence between users, etc.), and since `motorSteps` seems to be the legacy implementation, I reckon I'll remove `motorSteps` for now and keep modifying the step size locally in `uiproperties` for now. I'll add a comment in the code so we remember to revisit this later.

PR #1442, opened 2024-10-09
ensure shape key is the upper-cased id

meguiraun:
fixing an issue when the shape id reported by the server is lower cased (p1), but the UI expects upper cased (P1)

how to reproduce:
1 - Create a new point in the ui, check the state and you will see the ui state as  "P1"
2 - (you can run data collections on this)
3 - Refresh the page, check the state and you will see the ui state as  "p1"
4 - Now you cannot run tasks as the addTask method will try to access the shape with "P1" key

If you enable and the. disable the 3click centring, you would end up with both keys ("p1" and "P1") in the state (???)

I tried to hack my way into the `to_camel`method for achieving this but I got no luck.

### DIFF
--- a/mxcubeweb/core/components/sampleview.py
+++ b/mxcubeweb/core/components/sampleview.py
@@ -266,8 +266,10 @@ class SampleView(ComponentBase):
         for shape in HWR.beamline.sample_view.get_shapes():
             s = shape.as_dict()
             shape_dict.update({shape.id: s})
-
-        return {"shapes": to_camel(shape_dict)}
+            # shape key comes case lowered from the to_camel (2dp1), this breaks ui
+            # lest ensure it upper case by only came casing the dict data
+            _shape = {shape.id: to_camel(shape_dict[shape.id])}
+        return {"shapes": _shape}
 
     def get_shape_width_sid(self, sid):
         shape = HWR.beamline.sample_view.get_shape(sid)


marcus-oscarsson:
Thanks, @meguiraun. Seems like a viable fix. How about looking into why its uppercase to begin with. Would it make sense to make sure that its lower case to start with and then simply rely on `to_camel` completely, just a thought ?

meguiraun:
I assumed uppercase was what it should be, as set by the UI... I can have a look at the reducer...

meguiraun:
we have already a similar approach here: https://github.com/mxcube/mxcubeweb/blob/develop/mxcubeweb/core/components/sampleview.py#L147

and the tasks are added based on the shape ID (capitalized), see https://github.com/mxcube/mxcubeweb/blob/develop/ui/src/components/Tasks/DataCollection.jsx#L53

I feel this MR is the less intrusive approach

marcus-oscarsson:
Ok, I think you are right. It seems like this is the approach that is used elsewhere in the code base. It is a bit error prone as we need to remember to extract the ID, I guess thats why we have this problem ;) . Lets keep it as this for the moment then.

mxcubeweb/core/components/sampleview.py
```diff
@@ -266,8 +266,10 @@ def get_shapes(self):
         for shape in HWR.beamline.sample_view.get_shapes():
             s = shape.as_dict()
             shape_dict.update({shape.id: s})
-
-        return {"shapes": to_camel(shape_dict)}
+            # shape key comes case lowered from the to_camel (2dp1), this breaks ui
+            # lest ensure it upper case by only came casing the dict data
+            _shape = {shape.id: to_camel(shape_dict[shape.id])}
+        return {"shapes": _shape}
```
walesch-yan, line 272: Hey @meguiraun, we tested your changes and run into the following issue when there are no shapes in the `sample_view` hwr:
Screenshot from 2024-10-14 10-23-07
Another issue is, when we there are multiple shapes in `sample_view`, it seems to return a dict that only contains one element, I would suggest the following change:
`_shape[shape.id] = to_camel(shape_dict[shape.id])` 
and initializing `_shape` before the for-loop, if this sounds correct to you i can make a new PR for this, as this one has already been accepted?

walesch-yan: Or maybe a more elegant way would be to just change the update function of `shape_dict` so that the function would look like this:
```    
    def get_shapes(self):
        shape_dict = {}

        for shape in HWR.beamline.sample_view.get_shapes():
            s = shape.as_dict()
            # shape key comes case lowered from the to_camel (2dp1), this breaks ui
            # lest ensure it upper case by only came casing the dict data
            shape_dict.update({shape.id: to_camel(s)})

        return {"shapes": shape_dict}
```
Which would remove the necessity of an additional dict object (`_shape`)?

meguiraun: sure, please go ahead with a PR

PR #1444, opened 2024-10-10
Refactor `MotorInputContainer` and `MotorInput`

axelboc:
I'll probably refactor them further, but since the conversion of `MotorInput` to a functional component wasn't trivial, I think it's best that I split things up.

Note that I also fix the styling of "dirty"/"edited", which I broke in #1438.

All the interactions still work:

- edit value manually then press <kbd>Enter</kbd>
- focus field then press <kbd>ArrowUp/Down</kbd>
- click on up/down arrow buttons
- change step
- abort/retry

Peek 2024-10-10 11-50

### DIFF
--- a/docs/source/dev/tests.md
+++ b/docs/source/dev/tests.md
@@ -58,6 +58,8 @@ pnpm --prefix ui build
 CYPRESS_BASE_URL="http://127.0.0.1:8081" pnpm --prefix ui e2e
 ```
 
+> Currently, WebSockets don't work when running Cypress in the CI. In order to replicate the CI environement more accurately, WebSockets can be disabled locally via environment variable `VITE_DISABLE_WEBSOCKETS`. For instance, you can start the development server with sockets disabled with: `VITE_DISABLE_WEBSOCKETS=true pnpm --prefix ui start`.
+
 ### Best Practices
 
 The best use for development would be to test the components you changed first individually by using the cypress app. This allows for better error tracking and time efficiency as it prevents you from running tests on components that are not effected by your changes multiple times during development.
--- a/ui/src/components/SampleView/SampleControls.module.css
+++ b/ui/src/components/SampleView/SampleControls.module.css
@@ -108,7 +108,6 @@
   margin-top: 0.5rem;
   padding: 0.625rem 0.75rem;
   background-color: rgba(0, 0, 0, 0.6);
-  color: rgb(255, 255, 255);
   border-radius: 4px;
 }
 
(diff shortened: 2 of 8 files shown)


docs/source/dev/tests.md
```diff
@@ -58,6 +58,8 @@ pnpm --prefix ui build
 CYPRESS_BASE_URL="http://127.0.0.1:8081" pnpm --prefix ui e2e
 ```
 
+> Currently, WebSockets don't work when running Cypress in the CI. In order to replicate the CI environement more accurately, WebSockets can be disabled locally via environment variable `VITE_DISABLE_WEBSOCKETS`. For instance, you can start the development server with sockets disabled with: `VITE_DISABLE_WEBSOCKETS=true pnpm --prefix ui start`.
```
axelboc, line 61: Running Cypress locally was still broken. I introduced that env var `VITE_DISABLE_WEBSOCKETS` a few weeks back, but I was setting it in the `cypress` script in `ui/package.json`. This had no effect since the env var is used by the front-end code to decide not to open socket connections -- not by Cypress.

Obviously we don't want to always start the dev server with sockets disabled, so I decided the simplest was to document the command `VITE_DISABLE_WEBSOCKETS=true pnpm --prefix ui start` for now. 

To summarize, to run E2E tests locally against the front-end development server, run the following three commands in separate terminals:

- `mxcubeweb-server -r $(pwd)/demo/ --static-folder $(pwd)/ui/build/ -L debug`
- `VITE_DISABLE_WEBSOCKETS=true pnpm --prefix ui start`
- `pnpm --prefix ui cypress`

ui/src/components/SampleView/SampleControls.module.css
```diff
@@ -108,7 +108,6 @@
   margin-top: 0.5rem;
   padding: 0.625rem 0.75rem;
   background-color: rgba(0, 0, 0, 0.6);
-  color: rgb(255, 255, 255);
```
axelboc, line 111: Slightly unrelated: this fixes the text color of the input field of the "Focus" overlay on the sample view (which uses `MotorInput` via `OneAxisTranslationControl`).

axelboc:
Sorry, switching back to draft, as I'd like to refactor `OneAxisTranslationControl` while I'm at it, since it basically should use `MotorInput` as well.

axelboc:
Nope, rabbit hole... I'll do it separately :sweat_smile:

docs/source/dev/tests.md
marcus-oscarsson: I think we should add `127.0.0.1:8081` to he CORS hosts as well since we are doing `CYPRESS_BASE_URL: http://127.0.0.1:8081` in the CI actions.

marcus-oscarsson: Just a thought, even if its not directly related :)

marcus-oscarsson:
Thanks !,

docs/source/dev/tests.md
axelboc: You're a genius!

image

marcus-oscarsson: ;), not sure but I sometimes manage to get things right

PR #1445, opened 2024-10-14
Refactor `OneAxisTranslationControl`

axelboc:
`OneAxisTranslationControl` has an input field that behaves the same way as the input field in `MotorInput`: you can type a number and press enter, or use up/down arrow to increment decrement, and it updates the value and shows the state of the underlying motor.

So in this PR, I extract this shared input from `MotorInput` into a `BaseMotorInput` component, and use it in `OneAxisTranslationControl`. This resolves a few minor inconsistencies between the two components and removes a bunch of duplicated code.

Peek 2024-10-14 09-30

Still more refactoring to go — notably, I'd like to reduce the number of props that are passed down to some of the components.

### DIFF
--- a/ui/src/components/MotorInput/OneAxisTranslationControl.jsx
+++ b/ui/src/components/MotorInput/OneAxisTranslationControl.jsx
@@ -1,169 +1,74 @@
 import React from 'react';
-import { Button, Form, Popover } from 'react-bootstrap';
+import { Button } from 'react-bootstrap';
+
+import BaseMotorInput from './BaseMotorInput';
 import { HW_STATE } from '../../constants';
-import MotorInput from './MotorInput';
 import './motor.css';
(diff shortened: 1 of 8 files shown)


ui/src/components/MotorInput/OneAxisTranslationControl.jsx
```diff
@@ -1,169 +1,74 @@
...
-            decimalPoints="3"
-            state={this.props.motors.sample_horizontal.state}
-            stop={this.props.stop}
-            disabled={this.props.motorsDisabled}
-            inplace
-          />
-        </div>
-      </Popover>
-    );
-  }
```
axelboc, line 94: This was dead code — maybe from copy-pasting `TwoAxisTranslationControl`.

marcus-oscarsson:
Thanks !

PR #1446, opened 2024-10-14
fix minor issues in get_shapes method

walesch-yan:
The changes introduced in #1442 fixed an issue related to the camel casing of the shape ids. However the introduction of a new `dict` object came with issues (see #1442 for a more in detail review) related to:
- no shapes: the new dict object was not correctly initialized
- more than one shape: the dict object would only contain a single object

The changes proposed in this PR should fix the issues and remove the necessity of an additional `dict` object


meguiraun (approved):
thanks!

PR #1447, opened 2024-10-14
Fix beamstop and capillary not shown

axelboc:
This fixes the beamstop and capillary beamline attributes in the mock environment ... and probably the beamstop distance overlay in production environments too??

image

image

image

### DIFF
--- a/demo/beamstop.xml
+++ b/demo/beamstop.xml
@@ -1,5 +1,5 @@
 <object class="ExporterNStateMockup">
   <username>beamstop</username>
-  <values>{"OUT": "OFF", "IN": "BEAM"}</values>
+  <values>{"IN": "BEAM", "OUT": "OFF"}</values>
   <use_hwstate>True</use_hwstate>
 </object>
--- a/demo/mxcube-web/ui.yaml
+++ b/demo/mxcube-web/ui.yaml
@@ -100,9 +100,9 @@ beamline_setup:
   id: beamline_setup
   components:
     - label: Beamstop
-      attribute: beamstop
+      attribute: diffractometer.beamstop
     - label: Capillary
-      attribute: capillary
+      attribute: diffractometer.capillary
     - label: Fast Shutter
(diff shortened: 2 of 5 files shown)


demo/beamstop.xml
```diff
@@ -1,5 +1,5 @@
 <object class="ExporterNStateMockup">
   <username>beamstop</username>
-  <values>{"OUT": "OFF", "IN": "BEAM"}</values>
+  <values>{"IN": "BEAM", "OUT": "OFF"}</values>
```
axelboc, line 3: The front-end expects the "good" state (green) to come first.

demo/mxcube-web/ui.yaml
```diff
@@ -100,9 +100,9 @@ beamline_setup:
   id: beamline_setup
   components:
     - label: Beamstop
-      attribute: beamstop
+      attribute: diffractometer.beamstop
     - label: Capillary
-      attribute: capillary
+      attribute: diffractometer.capillary
```
axelboc, line 105: This makes the two attributes appear in the UI.

ui/src/containers/BeamlineSetupContainer.jsx
```diff
@@ -23,32 +22,27 @@ function BeamlineSetupContainer(props) {
...
-  function handleSetAttribute(name, value) {
-    setAttribute(name, value);
-  }
-
   function renderBeamstopAlignmentOverlay() {
     const { hardwareObjects } = beamline;
     const motorInputList = [];
     let popover = null;
 
-    const motor = hardwareObjects.beamstop_alignemnt_x;
```
axelboc, line 42: There was a typo, but the hardware object is no longer called `beamstop_alignment_x` anyway but `diffractometer.beamstop_distance`.

This and the changes below fix the beamstop distance overlay as shown in the third screenshot (the cog icon + overlay with `OneAxisTranslationControl`).

PR #1448, opened 2024-10-15
Fix changing motor step of some motors

axelboc:
Notably the _Sample Horizontal_ and _Sample Vertical_ motors:

image

The bug was a combination of two factors:

- The steps were saved in two places in the store: `uiproperties.sample_view.components[index].step` and `sampleview.motorSteps`. Some components read the step from `uiproperties`, while others read it from `sampleview`. This was discussed in https://github.com/mxcube/mxcubeweb/pull/1441#discussion_r1791900401
- Some motors' `uiproperties` point to the same underlying hardware object (e.g. _Y_ and _Sample Vertical_ both point to `diffractometer.phiy`) and we were updating the step only in the first `uiproperties` object that we would find for that hardware object.

The fix consists in:

- Removing `sampleview.motorSteps` from the store and updating the steps only in `uiproperties`.
- Looking for the `uiproperties` object to update based on the `role` (e.g. `sample_vertical`) rather than the `attribute` (e.g. `diffractometer.phiy`), as `role` is unique.

Additionally, I remove a few remaining uses of an old API endpoint to update motor positions: `/sampleview/<motor_id>/<new_pos>` and related code. This includes other similar unused endpoints (`/sampleview/zoom`, `/sampleview/backlighton`, etc.), the `sendUpdateMotorPosition` API function, and the `updateMotorPosition` action creator (replaced with `setAttribute`).


ui/src/actions/sampleview.js
```diff
@@ -41,8 +40,8 @@ export function setAperture(size) {
   return { type: 'SET_APERTURE', size };
 }
 
-export function setStepSize(name, value) {
-  return { type: 'SET_STEP_SIZE', componentName: 'sample_view', name, value };
+export function setMotorStep(role, value) {
+  return { type: 'SET_MOTOR_STEP', role, value };
```
axelboc, line 44: Making the action creator more specific, since we now use the `role` to find the `uiproperties` object to update, and this makes sense only for sample view motors.

ui/src/components/SampleView/SampleImage.jsx
```diff
@@ -528,58 +528,36 @@ export default class SampleImage extends React.Component {
...
+    const { components } = uiproperties.sample_view;
     const keyPressed = this._keyPressed;
 
-    const phi = hardwareObjects['diffractometer.phi'];
-    const focus = hardwareObjects['diffractometer.focus'];
-    const zoom = hardwareObjects['diffractometer.zoom'];
-
-    if (keyPressed === 'r' && phi.state === HW_STATE.READY) {
-      // then we rotate phi axis by the step size defined in its box
-      if (e.deltaX > 0 || e.deltaY > 0) {
```
axelboc, line 541: I'm no longer checking for `deltaX` here, as this assumes a mouse/pad that supports horizontal scroll, and I feel one would not instinctively scroll horizontally to perform this interaction anyway.

ui/src/containers/BeamlineSetupContainer.jsx
```diff
@@ -45,7 +43,7 @@ function BeamlineSetupContainer(props) {
             value={motor.value}
             min={motor.limits[0]}
             max={motor.limits[1]}
-            step={step}
```
axelboc, line 48: I've hit a bit of a snag here.

The `diffractometer.beamstop_distance` has no corresponding `uiproperties` object (which is where the step is stored). If I add one by declaring a _Beamstop distance_ component in `ui.yaml`, under `beamline_setup`, then it shows up as a beamline attribute in the UI (next to _Energy_, _Resolution_, etc.)

Any thoughts?

marcus-oscarsson: Hm, I don't have a good solution for it right away, one way would be to add property called display that can be either true or false and that would simply be displayed in the beamline setup. Id prefer another solution though because that is rather counter intuitive.

Perhaps by adding the possibility to have sub-components, which will make things rather complicated. Would it make sense to add another section called `beamstop` like ?:
```
beamstop:
  id: beamstop
  components:
    - label: Beamstop distance
      attribute: diffractometer.beamstop_distance
      role: beamstop_distance
      step: 0.01
      precision: 2
      suffix: mm
```

axelboc: Right, not a bad idea! :thinking: 

I can't help but think that the UI config needs a bigger overhaul, though. Is it okay if we keep the hardcoded step for now and if I add add a link to this thread in a comment so we can revisit it later?

marcus-oscarsson: Yes, your probably right, the idea was to keep it very simple. But it of-course need to solve our main problems :)

PR #1449, opened 2024-10-15
Merge `MotorInputContainer` and `MotorInput`

axelboc:
Getting there... I'm now able to merge `MotorInputContainer` and `MotorInput`. The new `MotorInput` component now accepts a single `role` prop. It looks for the `uiproperties` object with that role and renders the corresponding motor input. This removes a bunch of code from `TwoAxisTranslationControl`.

Additionally, I make the props of `OneAxisTranslationControl` and `TwoAxisTranslationControl` consistent: they now accept one or two `uiproperties` objects respectively. It is then up to `<One|Two>AxisTranslationControl` to retrieve the corresponding motor hardware object.

Finally, a very small cosmetic change: I remove the colons from the motor labels to lighten things up a bit:

image


ui/src/components/MotorInput/MotorInput.jsx
```diff
@@ -55,32 +69,34 @@ function MotorInput(props) {
...
             precision={precision}
             step={step}
-            testId={`MotorInput_value_${motorName}`}
+            testId={`MotorInput_value_${attribute}`}
             disabled={disabled}
-            onChange={(val) => save(motorName, val)}
+            onChange={(val) => dispatch(setAttribute(attribute, val))}
           />
         </div>
-        {saveStep &&
```
axelboc, line 63: Worth pointing out that I'm able to remove this condition because `OneAxisTranslationControl` no longer uses `MotorInput` since #1445 (it uses the lower-level component `BasicMotorInput`).

axelboc (commented):
Next up, removing the global styles in `motor.css`.

marcus-oscarsson:
Awesome, thanks !

PR #1451, opened 2024-10-16
Modularize `MotorInput` styles

axelboc:
In other words, I get rid of the global styles in `motors.css` (`rw-widget`, `.arrow-control`, etc.) and rewrite them with CSS Modules.

Hopefully there won't be any styling regression. On the contrary, I think I actually fix a few subtle things like the double border on motor input fields, the lack of hover/focus styles on the up/down arrow buttons, etc.

image

I also change the design of the beamstop distance overlay slightly by making better use of Bootstrap:

image


ui/src/components/MotorInput/BaseMotorInput.jsx
```diff
@@ -63,9 +61,8 @@ function BaseMotorInput(props) {
 
   return (
     <form className="d-flex" noValidate onSubmit={handleSubmit}>
-      <Form.Control
```
axelboc, line 66: Bootstrap's styles kind of get in the way of our customisation needs, so I prefer to remove them by switching to a bare `input`.

ui/src/components/SampleGrid/SampleGridTable.css
```diff
@@ -151,10 +151,6 @@
   max-width: 160px !important;
 }
 
-button[disabled] {
-  opacity: 0.2;
-}
```
axelboc, line 156: Moved to `main.css`, as this is a good thing to have, and added `pointer-events: none`.

ui/src/components/SampleView/MotorControls.jsx
```diff
@@ -25,21 +25,22 @@ function MotorControls() {
   );
 
   if (!verticalMotorProps || !horizontalMotorProps) {
-    return motorsProps.map(({ attribute: k, role }) => (
-      <MotorInput key={k} role={role} />
-    ));
+    return motorsProps.map(({ role }) => <MotorInput key={role} role={role} />);
```
axelboc, line 28: Unrelated: using the `role` as key instead of the attribute to fix React warning about duplicate keys.

ui/src/components/SampleView/MotorControls.jsx
```diff
@@ -25,21 +25,22 @@ function MotorControls() {
...
+      {motorsProps.slice(0, 3).map(({ role }) => (
+        <MotorInput key={role} role={role} />
       ))}
 
-      <TwoAxisTranslationControl
-        verticalMotorProps={verticalMotorProps}
-        horizontalMotorProps={horizontalMotorProps}
-      />
+      <div className={motorInputStyles.container}>
+        <p className={`${motorInputStyles.label} mb-1`}>Sample alignment</p>
```
axelboc, line 38: I've moved the container and label out of `TwoAxisTranslationControl` for consistency with `OneAxisTranslationControl`.

marcus-oscarsson:
Nice, thanks :)

PR #1453, opened 2024-10-17
Test main navigation and motor inputs

axelboc:
I've made separate commits but basically, I:

- increase the viewport when testing so it's a bit nicer to work with;
- merge the two 3-click centring tests to save time;
- add a smoke test for the main navigation to make sure that all pages render as expected (expect "System logs" since it is likely to be removed);
- test motor inputs in the sidebar (I think I cover most possible user interactions)
- label motor inputs for accessibility (and testing);
- add `data-testid` attributes on some elements that are a bit difficult to label properly for now (so I can find them in the tests);
- fix a subtle bug when reapplying the same value in a motor input (e.g. if the value is `0.1` and the precision is `1`, typing `0.09` and <kbd>Enter</kbd> would keep `0.09` in the field instead of re-applying `0.1`)

Peek 2024-10-17 11-19

### DIFF
--- a/ui/cypress/e2e/sampleControls.cy.js
+++ b/ui/cypress/e2e/sampleControls.cy.js
@@ -1,41 +1,42 @@
 /* global cy, it, describe, beforeEach */
 
-describe('3-click centring', () => {
+describe('Sample controls', () => {
   beforeEach(() => {
     cy.loginWithControl();
   });
 
-  it('displays error when no sample is mounted', () => {
+  it('3-click centring', () => {
     cy.clearSamples(); // another test may have mounted a sample
     cy.findByRole('link', { name: /Data collection/u, hidden: true }).click();
 
+    // Start 3-click centring
     cy.findByRole('button', { name: '3-click centring' }).click();
-    cy.findByRole('alert', 'Error: There is no sample mounted').should(
-      'be.visible',
-    );
-  });
 
-  it('rotates sample by 90 degrees on click', () => {
+    // Error because no sample is mounted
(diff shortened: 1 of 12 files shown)


ui/cypress/e2e/sampleControls.cy.js
```diff
@@ -1,41 +1,42 @@
...
-    cy.findByTestId('MotorInput_value_diffractometer.phi')
+    // Click on sample view to rotate sample once
+    cy.findByLabelText('Omega')
       .invoke('val')
       .then((initialValue) => {
         cy.get('.canvas-container').click();
         cy.findByText(/Clicks left: 2/u, { hidden: true }).should('exist');
 
-        // Wait for omega motor to finish moving
-        cy.wait(1000);
```
axelboc, line 34: This sleep timer is no longer needed now that Web Sockets are working.

ui/cypress/e2e/sampleControls.cy.js
```diff
@@ -1,41 +1,42 @@
...
       'be.visible',
     );
 
+    // Start 3-click centring again
     cy.findByRole('button', { name: '3-click centring' }).click();
     cy.findByText(/Clicks left: 3/u, { hidden: true }).should('exist');
 
-    cy.findByTestId('MotorInput_value_diffractometer.phi')
+    // Click on sample view to rotate sample once
+    cy.findByLabelText('Omega')
```
axelboc, line 30: Labeling motor inputs makes them easier to find.

PR #1454, opened 2024-10-17
Extract `SnapshotControl` out of `SampleControls`

axelboc:
Still with the aim of passing fewer props down from container components. I'll extract the other controls in the `SampleControls` toolbar in my next PRs.


ui/src/components/SampleControls/SampleControls.jsx
```diff
@@ -1,24 +1,19 @@
 /* eslint-disable jsx-a11y/control-has-associated-label */
 /* eslint-disable react/jsx-handler-names */
 /* eslint-disable sonarjs/no-duplicate-string */
-import './SampleView.css';
```
axelboc, line 4: Import moved to `SampleViewContainer`

ui/src/components/SampleControls/SampleControls.jsx
```diff
@@ -32,10 +27,6 @@ export default class SampleControls extends React.Component {
     this.availableVideoSizes = this.availableVideoSizes.bind(this);
   }
 
-  componentDidMount() {
-    window.takeSnapshot = this.doTakeSnapshot;
```
axelboc, line 36: Refactored as a `useEffect` in `SnapshotControl`.

ui/src/components/SampleControls/SampleControls.jsx
```diff
@@ -49,39 +40,6 @@ export default class SampleControls extends React.Component {
...
-    imgDataURI = this.props.canvas.toDataURL({ format: 'jpeg' });
-    this.props.canvas.setBackgroundImage(0);
-    this.props.canvas.renderAll();
-    return { data: imgDataURI.slice(23), mime: imgDataURI.slice(0, 23) };
-  }
-
-  takeSnapShot(evt) {
-    /* eslint-disable unicorn/consistent-function-scoping */
-    function imageEpolog(props) {
-      const { sampleID } = props.currentSampleID;
```
axelboc, line 68: Note that there was a bug here: `currentSampleID` is a string.

ui/src/components/SampleControls/SampleControls.jsx
```diff
@@ -49,39 +40,6 @@ export default class SampleControls extends React.Component {
...
-    const img = document.querySelector('#sample-img');
-    const fimg = new fabric.Image(img);
-    fimg.scale(this.props.imageRatio);
-    let imgDataURI = '';
-    this.props.canvas.setBackgroundImage(fimg);
-    this.props.canvas.renderAll();
-    imgDataURI = this.props.canvas.toDataURL({ format: 'jpeg' });
-    this.props.canvas.setBackgroundImage(0);
-    this.props.canvas.renderAll();
-    return { data: imgDataURI.slice(23), mime: imgDataURI.slice(0, 23) };
```
axelboc, line 62: Splitting the URI string is unnecessary, since we then do `evt.currentTarget.href = img.mime + img.data` later on... :grin:

ui/src/components/SampleControls/SampleControls.jsx
```diff
@@ -160,19 +118,7 @@ export default class SampleControls extends React.Component {
     return (
       <div className={styles.controls}>
         {this.props.getControlAvailability('snapshot') && (
-          <Button
-            as="a"
```
axelboc, line 164: Since the `href` is generated on click, the anchor element adds no value. I convert it to a `button` element, and instead of setting `evt.currentTarget.href`, I call a utility function called `download`, which creates a temporary anchor element to trigger the download.

ui/src/components/SampleControls/SnapshotControl.jsx
```diff
@@ -0,0 +1,60 @@
...
+    const filename = `${proposal}-${currentSampleName}.jpeg`;
+    download(filename, canvas.toDataURL({ format: 'jpeg' }));
+
+    canvas.setBackgroundImage(0);
+    canvas.renderAll();
+  }, [canvas, currentSampleName, imageRatio, proposal]);
+
+  useEffect(() => {
+    // Allow server to trigger snapshots (cf. `serverIO`)
+    window.takeSnapshot = takeSnapShot;
```
axelboc: In `serverIO.js`:

```js
    this.hwrSocket.on('take_xtal_snapshot', (cb) => {
      cb(window.takeSnapshot());
    });
```

Just to check: is the `take_xtal_snapshot` socket event still used? It feels a bit brittle, since the canvas has to be displayed for the snapshot to work, so if the event is received when the user is on the _Sample_ screen, it won't work.

marcus-oscarsson: Its not really, used and we are preparing a PR together with @walesch-yan that will remove it :)

marcus-oscarsson:
Could you rebase this branch I don't seem to be able to do it from the GitHub UI

PR #1455, opened 2024-10-17
Draft: User cannot login if previous user did not logout

dominikatrojanowska:
Depends on: https://github.com/mxcube/mxcubecore/pull/1066

KeyError regarding [Proposal](https://github.com/mxcube/mxcubeweb/blob/develop/mxcubeweb/core/components/lims.py#L306) was blocking effective log in in case of:

* another user was logged
* and an active proposal was not accessible by the current user

When, so called, another user logged out, the login (by pressing Sign in button on login page) of current user was still impossible because the current user already appeared as active (https://github.com/mxcube/mxcubeweb/blob/develop/mxcubeweb/core/components/user/usermanager.py#L343)
Log in was possible when the another user logs out and instead of clicking Sign in button the session was refreshed.


test/test_authn.py
```diff
@@ -137,26 +183,25 @@ def test_authn_same_proposal(make_client):
...
-#     client_1 = make_client()
-#     resp = client_1.post(URL_SIGNIN, json=CREDENTIALS_1)
-#     assert resp.status_code == 200
-#     import pdb
-#     pdb.set_trace()
-#     assert resp.json["msg"] == "Could not authenticate"
+# Test against proposal-based authentication only
+@pytest.mark.parametrize("login_type", ["proposal"], indirect=True)
+def test_authn_different_proposal(make_client):
+    """Test two users for different proposals.
```
marcus-oscarsson: So this is not actually testing for different proposals, its testing two users with the same proposal

mxcubeweb/core/components/lims.py
```diff
@@ -298,24 +297,19 @@ def get_proposal(self, user):
 
     def select_proposal(self, proposal):
         proposal_info = self.get_proposal_info(proposal)
-
-        if (
-            HWR.beamline.lims.loginType.lower() == "user"
-            and "Commissioning" in proposal_info["Proposal"]["title"]
-        ):
-            if hasattr(HWR.beamline.session, "set_in_commissioning"):
+        if proposal_info:
```
marcus-oscarsson, line 300: Was it these lines that caused the issue ?

marcus-oscarsson:
Hi @dominikatrojanowska and @fabcor-maxiv  Where in the PR is the actual issue solved ?

mxcubeweb/core/components/lims.py
dominikatrojanowska: the issue was TypeError (if proposal_info was None) or KeyError raised here:
```
and Commissioning" in proposal_info["Proposal"]["title"]
```
instead of returning False at the end, and finally it was carried and re-raised here: https://github.com/mxcube/mxcubeweb/blob/develop/mxcubeweb/routes/login.py#L43

marcus-oscarsson: > ...user already appeared as active (https://github.com/mxcube/mxcubeweb/blob/develop/mxcubeweb/core/components/user/usermanager.py#L343)

I see, the first comment above threw me off a bit. So I imagine that this is with user based login ? Out of curiosity how come that `proposal_info` was None ?

marcus-oscarsson:
So If understand this correctly the issue you are facing is that you can't login with two different users belonging to the same proposal/session ?

test/test_authn.py
dominikatrojanowska: ok, I will cherry pick from develop for this piece

mxcubeweb/core/components/lims.py
dominikatrojanowska: sorry for my mistake - it was some time ago I was debugging that, the error raised here was KeyError caused by 1st or 2nd keyword, and of course proposal_info may be an empty dict, not None

marcus-oscarsson:
@dominikatrojanowska: As we mentioned before, I suggest that you try this use case with the `AbstractLims` PR's and perhaps your issue is solved, let us know how it goes ?

PR #1456, opened 2024-10-17
Extract `GridControl` and `CentringControl` from `SampleControls`

axelboc:
I'm moving a bunch of the `onClick` logics into Redux thunks to avoid subscribing components to too much global state.

### DIFF
--- a/ui/src/actions/sampleview.js
+++ b/ui/src/actions/sampleview.js
@@ -143,7 +143,21 @@ export function setGridOverlay(level) {
 }
 
 export function toggleDrawGrid() {
-  return { type: 'DRAW_GRID' };
+  return async (dispatch, getState) => {
+    const { sampleview, queue } = getState();
+
+    // Stop three-click centring if active
+    if (sampleview.clickCentring) {
+      await dispatch(abortCentring());
+    }
+
+    if (!sampleview.drawGrid && !queue.currentSampleID) {
+      dispatch(showErrorPanel(true, 'There is no sample mounted'));
+      return;
+    }
+
+    dispatch({ type: 'DRAW_GRID' });
+  };
 }
 
 export function centringClicksLeft(clicksLeft) {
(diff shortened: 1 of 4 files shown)


ui/src/actions/sampleview.js
```diff
@@ -143,7 +143,21 @@ export function setGridOverlay(level) {
...
+
+    // Stop three-click centring if active
+    if (sampleview.clickCentring) {
+      await dispatch(abortCentring());
+    }
+
+    if (!sampleview.drawGrid && !queue.currentSampleID) {
+      dispatch(showErrorPanel(true, 'There is no sample mounted'));
+      return;
+    }
```
axelboc, line 157: Note that it is now allowed to toggle off `drawGrid` when there is no sample mounted.

(I mean it was allowed so far but only because of the bug mentioned in my other comment, which is now fixed).

ui/src/actions/sampleview.js
```diff
@@ -247,33 +261,48 @@ export function unselectShapes(shapes) {
...
-    const { shapes } = getState();
+    dispatch(clearSelectedShapes());
     dispatch(unselectShapes(shapes));
 
-    if (queue.currentSampleID) {
-      const json = await sendStartClickCentring();
-      const { clicksLeft } = json;
+    if (!queue.currentSampleID) {
+      dispatch(showErrorPanel(true, 'There is no sample mounted'));
+      return;
```
axelboc, line 290: Just turning the `else` branch into an early return.

ui/src/components/SampleControls/SampleControls.jsx
```diff
@@ -22,40 +24,9 @@ class SampleControls extends React.Component {
...
       'diffractometer.backlight',
     );
-    this.toggleCentring = this.toggleCentring.bind(this);
-    this.toggleDrawGrid = this.toggleDrawGrid.bind(this);
     this.availableVideoSizes = this.availableVideoSizes.bind(this);
   }
 
-  toggleDrawGrid() {
-    // Cancel click centering before draw grid is started
-    if (this.props.currentSampleID === '') {
```
axelboc, line 32: There was a bug here: `currentSampleID` is `null` when no sample is mounted.

PR #1457, opened 2024-10-17
Extract `FocusControl` and `ZoomControl` from `SampleControls`

axelboc:
No difficulty, but note that I've removed the "optimistic" update when interacting with the zoom slider (more info in comment):

#### BEFORE

Peek 2024-10-17 15-57

#### AFTER

Peek 2024-10-17 15-56

### DIFF
--- a/ui/src/components/SampleControls/SampleControls.jsx
+++ b/ui/src/components/SampleControls/SampleControls.jsx
@@ -4,13 +4,14 @@
 import React from 'react';
 import { OverlayTrigger, Button, Dropdown } from 'react-bootstrap';
 
-import OneAxisTranslationControl from '../MotorInput/OneAxisTranslationControl';
 import SnapshotControl from './SnapshotControl';
 import { HW_STATE } from '../../constants';
 
 import styles from './SampleControls.module.css';
 import GridControl from './GridControl';
 import CentringControl from './CentringControl';
+import FocusControl from './FocusControl';
+import ZoomControl from './ZoomControl';
 
 class SampleControls extends React.Component {
   constructor(props) {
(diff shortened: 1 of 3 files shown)


ui/src/components/SampleControls/ZoomControl.jsx
```diff
@@ -0,0 +1,62 @@
...
+import styles from './SampleControls.module.css';
+
+const ZOOM_HWO_ID = 'diffractometer.zoom';
+
+function ZoomControl() {
+  const dispatch = useDispatch();
+
+  const { state, value, commands } = useSelector(
+    (state) => state.beamline.hardwareObjects[ZOOM_HWO_ID],
+  );
```
axelboc, line 17: I don't need the `uiproperties` object, so I access the hardware object directly by its key.

ui/src/components/SampleControls/SampleControls.jsx
```diff
@@ -76,96 +77,15 @@ class SampleControls extends React.Component {
...
-                      'diffractometer.zoom',
-                      zoomMotor.commands[Number.parseFloat(e.target.value)],
-                    );
-                  }}
-                  onChange={(e) => {
-                    this.props.setBeamlineAttribute(
-                      'diffractometer.zoom',
-                      zoomMotor.commands[Number.parseFloat(e.target.value)],
-                    );
-                  }}
```
axelboc, line 143: We would make a local update to the store `onChange` and a server update `onMouseUp`. This led to a "jump" of the slider thumb when interacting with the mouse. More importantly, server updates were missed when interacting with the keyboard!

I kept only `onChange` and replaced `setBeamlineAttribute` with `setAttribute` to always (and only) perform server updates. The local state is still updated via websockets.

marcus-oscarsson:
On a streak ! :1st_place_medal:

Issue #1443, opened 2024-10-10
bring back the Draw grid V-Space and H-Space


elmjag:
The` V-Space` and `H-Space` boxes in the *Draw grid* widget have been removed in this PR: https://github.com/mxcube/mxcubeweb/pull/1397

![image](https://github.com/user-attachments/assets/921c7230-2890-4f5c-98d6-7f40cb06d09f)

We are actually using these controls. We did tests doing mesh grid scans with H-Space separation last week, and it worked for us.

Can we bring them back?

As I understand, these controls are seldom used, so let's make it configurable if they are shown. I suggest to add these two UI configs:

```
show_mesh_grid_vspace   true/false
show_mesh_grid_hspace   true/false

```
With default values set to false for both of them.

As a side note, at MicroMAX currently we only want to show H-Space, but in the future we want to show both. Thus it would be nice to make each box individually configurable.

marcus-oscarsson:
We wont have anything against re-introducing the h and v-space controls if you are using them.  They where added because we thought there might be a use for h/v space but in the end no one actually used them. Many users also found the controls strange, for some reason, and where mostly just wondering why we had those. The decision was thus to remove them and possibly replace them with something more "user friendly" when actually requested. 

It might be necessary to have a second look at these controls when we have some more feedback from the scientists but adding them back as they where would be fine.

elmjag:
@mockoocy will work on this.

PR #1458, opened 2024-10-17
brings back v/h-space controls (#1443)

mockoocy:
This PR brings back controls removed in this PR https://github.com/mxcube/mxcubeweb/pull/1397 
As this feature is rarely used, it can be optionally enabled via `ui.yaml` config file.
Here's part of configuration which enables these:
```yaml
sample_view_video_controls:
  id: sample_view_video_controls
  grid_settings:
    show_mesh_grid_vspace: true
    show_mesh_grid_hspace: true
```
The `grid_settings` key can be omitted and these new keys will have default value of `false` (added at the time of `pydantic` validation.
Additionally changed the way add / remove grid buttons' sizes are being calculated, as adding these h-space/v-space controls may cause them to become awkwardly small ( at least it did so on my 1920x1080 screen :))

### DIFF
--- a/docs/source/config/ui.rst
+++ b/docs/source/config/ui.rst
@@ -102,6 +102,9 @@ The section has the following syntax:
 
     sample_view_video_controls:
       id: sample_view_video_controls
+      grid_settings:
+        show_mesh_grid_vspace: false
+        show_mesh_grid_hspace: false
       components:
         - id: snapshot
           show: <show>
@@ -120,8 +123,12 @@ The section has the following syntax:
         - id: video_size
           show: <show>
 
-It is a list of all supported sample video widgets.
-``<show>`` flag is a boolean, when it's *true* the widget is included, when *false* it is omitted.
+It consists of two keys:
+- `grid_settings` - settings related to the mesh grid controls in the `draw_grid` component.
+- `components` - a list of all supported sample video widgets, in the format:
+  - id: widget id
+  - show: <show> - boolean flag, when it's *true* the widget is included, when *false* it is omitted.
+
 
 .. image:: video_controls.png
 
--- a/docs/source/config/ui.rst
+++ b/docs/source/config/ui.rst
@@ -102,14 +102,13 @@ The section has the following syntax:
 
     sample_view_video_controls:
       id: sample_view_video_controls
-      grid_settings:
-        show_mesh_grid_vspace: false
-        show_mesh_grid_hspace: false
       components:
         - id: snapshot
           show: <show>
         - id: draw_grid
           show: <show>
+          show_hspace: <show_hspace>
+          show_vspace: <show_vspace>
         - id: 3_click_centring
           show: <show>
         - id: focus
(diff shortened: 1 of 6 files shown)


mockoocy:
Ran `pre-commit` and fixed the trailing whitespace 😅. CI should work now

docs/source/config/ui.rst
```diff
@@ -120,8 +123,12 @@ The section has the following syntax:
...
           show: <show>
 
-It is a list of all supported sample video widgets.
-``<show>`` flag is a boolean, when it's *true* the widget is included, when *false* it is omitted.
+It consists of two keys:
+- `grid_settings` - settings related to the mesh grid controls in the `draw_grid` component.
+- `components` - a list of all supported sample video widgets, in the format:
+  - id: widget id
+  - show: <show> - boolean flag, when it's *true* the widget is included, when *false* it is omitted.
+
```
elmjag: The rendered document look a bit odd:

image

You can see the result here: https://mxcubeweb--1458.org.readthedocs.build/en/1458/config/ui.html#sample-view-video-controls

elmjag:
I suggest we make configs less wordy. I have two suggestions:

```
sample_view_video_controls:
  id: sample_view_video_controls
  draw_grid:
    show_vspace: true
    show_hspace: true
```
The `draw_grid` is more in-line with existing configs, as it's the key name to enabled the *Draw grid* button.

Another suggestion is to put v/h space keys in the same place where *Draw grid* button is enabled. Like this:

```
sample_view_video_controls:
  id: sample_view_video_controls
  components:
    - id: snapshot
      show: true
    - id: draw_grid
      show: true
      show_vspace: true
      show_hspace: true
```
I don't know if this is possible to achieve in the current code that deals with UI configs. If it's very complicated, I'm fine with the first option.

elmjag:
Overall, this looks good. I did some light testing.

This PR will do the trick for as at MAXIV for out spaced mesh scan needs.

marcus-oscarsson:
Thanks.

Id also prefer as @elmjag suggests, its more concise and consistent. I don't think it would make things more complicated.

```
sample_view_video_controls:
  id: sample_view_video_controls
  components:
    - id: snapshot
      show: true
    - id: draw_grid
      show: true
      show_vspace: true
      show_hspace: true
``

**Comment by mockoocy** [comment]
Changed yml schema to match @elmjag's proposal

File: ui/src/containers/SampleViewContainer.jsx [hunk 2]
```diff
@@ -46,13 +46,10 @@ class SampleViewContainer extends Component {
   }
 
   getControlAvailability(name) {
-    const available = find(
```
elmjag, line 48: Remove `find` import at line 36 above. This will fix the eslint warning.

elmjag:
Looks good, just need a small fix to pass JS-Lint: https://github.com/mxcube/mxcubeweb/pull/1458/files#r1810054097

Otherwise, I'm happy!

mockoocy:
> Looks good, just need a small fix to pass JS-Lint: https://github.com/mxcube/mxcubeweb/pull/1458/files#r1810054097
> 
> Otherwise, I'm happy!

should be ok now :)

elmjag (approved):
LGTM, ship it!

marcus-oscarsson:
Wonderful, thanks !

PR #1459, opened 2024-10-17
Extract `LightControl` from `SampleControls`

axelboc:
The new control is used for both the frontlight and backlight controls, which reduces duplication


marcus-oscarsson:
Looks good to me :)

PR #1460, opened 2024-10-18
Adds a way to preserve user-preferred mesh grid state

mockoocy:
Fixes a bug / bad UX in which caused mesh grid to became unhidden after certain operations like DataCollection or Phase change.

The steps to reproduce the bugs were as following:
- mount a sample
- open `draw_grid` control
- draw a few grids
- make some of them hidden
- right click on canvas, press `Data Collection` button
- run the task
After the task was done, each of the grids was shown again.
It's also worth adding that in the demo beamline many of the motors seem to not be mocked, so the behavior can't be observed. 

The reason for this behavior was further described in this Pull request https://github.com/mxcube/mxcubecore/pull/1044 for mxcubecore. This PR will be marked as draft until the mxcubecore one is merged.

The fix writes to state set by user to `Shape.user_state` attribute on the mxcubecore side.
When the data collection or phase change is finished, then the user-preferred state is restored by setting `Shape.state` to be equal to `Shape.user_state`.

### DIFF
--- a/ui/src/components/SampleView/GridForm.jsx
+++ b/ui/src/components/SampleView/GridForm.jsx
@@ -58,12 +58,13 @@ export default function GridForm(props) {
 
     for (const grid of Object.values(gridList)) {
       const selectedStyle = selectedGrids.includes(grid.id) ? 'selected' : '';
+      const hiddenStyle = grid.userState === 'HIDDEN' ? 'hid-by-user' : '';
       const vdim = grid.numRows * (grid.cellHeight + grid.cellVSpace);
       const hdim = grid.numCols * (grid.cellWidth + grid.cellHSpace);
 
       gridControlList.push(
         <tr
-          className={selectedStyle}
+          className={`${selectedStyle} ${hiddenStyle}`}
           key={grid.name}
           onClick={(e) => selectGrid([grid], e.ctrlKey)}
         >
--- a/ui/src/components/SampleView/GridForm.jsx
+++ b/ui/src/components/SampleView/GridForm.jsx
@@ -5,9 +5,26 @@ import Draggable from 'react-draggable';
 
 import './SampleView.css';
 import { useSelector } from 'react-redux';
+import TooltipTrigger from '../TooltipTrigger';
 
 function handleContextMenu(e) {
   e.stopPropagation();
+}
+/**
+ * Returns the tooltip content for the hide button
+ *
+ * @param {boolean} userHidden - if the grid is hidden by the user
+ * @param {boolean} autoHidden - if the grid is hidden automatically (out of view)
+ * @returns {string} - the tooltip content
+ */
+function getHideButtonTooltip(userHidden, autoHidden) {
+  if (userHidden) {
+    return 'Grid is hidden by user';
+  }
+  if (autoHidden) {
+    return 'Grid is out of view';
+  }
+  return 'Hide grid';
 }
 
 export default function GridForm(props) {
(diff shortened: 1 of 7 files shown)


mockoocy:
Turns out the same issue happens for the zoom and focus controls. Will also try to fix it for these in this PR.

fabcor-maxiv:
> Turns out the same issue happens for the zoom and focus controls. Will also try to fix it for these in this PR.

@mockoocy Note that there was work on focus and zoom controls very recently: <https://github.com/mxcube/mxcubeweb/pull/1457>. I do not know if it is related or not.

mockoocy:
added some changes; 
now the UI has some indicator that specific grid is being hidden (it appears with a darker background)
I have also changed `HWR.beamline.sample_view.add_shape_from_mpos` function, also when updating shapes there are optionally more arguments passed to it, so that `shape` and `user_shape` are not changed when grids is being redrawn sometimes.

marcus-oscarsson:
Thanks @mockoocy, do you happen to have a screenshot of the changes ?

mockoocy:
> Thanks @mockoocy, do you happen to have a screenshot of the changes ?

image

ui/src/components/SampleView/SampleView.css
```diff
@@ -176,3 +176,7 @@
   color: #000;
   font-weight: bold;
 }
+
+.hid-by-user {
+  background-color: #aaa !important;
```
fabcor-maxiv: Isn't it usually bad practice to use `!important` in CSS?

elmjag:
@marcus-oscarsson I think we have different ideas how this mesh hidding/un-hidding is suppose to work.

Like what is the actual desired behavior. Perhaps we can discuss this in a video meeting? I'm at least a bit confused right now.

ui/src/components/SampleView/SampleView.css
mockoocy: You're right. I switched to using data attributes to override bootstrap's default css.

marcus-oscarsson:
@elmjag sure, we can have a quick chat if you like. How about tomorrow ? Otherwise I think the current proposal is fine maybe the color could be a bit more fun (other than dark gray :) ) and what do you think about a text perhaps under the hide button (saying "hidden by user" and perhaps "out of view" when hidden automatically) ?

mockoocy:
Undrafting.
As per @marcus-oscarsson's suggestion I changed the background color for table with grid info. 
Additionally I added a tooltip that shows the reason why a grid is hidden. I think it's nicer and informative enough :) 

image
withTooltip

marcus-oscarsson:
Would it be possible to get some more distinct color difference, between "hidden by user" and "out of range", like light blue and light green or something ?

marcus-oscarsson:
Or maybe no change for "our of range" and light blue or green for "hidden by user" ?

mockoocy:
> Or maybe no change for "our of range" and light blue or green for "hidden by user" ?

Light green's already reserved for selected grid, so I guess we'll go with light blue  :smile_cat:   
image

marcus-oscarsson:
Ok, yeah that looks better I guess, does it not ?. Would it make sense to make the text of the hidden meshes italic ?

mockoocy:
> Ok, yeah that looks better I guess, does it not ?. Would it make sense to make the text of the hidden meshes italic ?

Yeah, that's a good idea - done :)

marcus-oscarsson:
Cool, thanks alot very nice !

ui/src/components/SampleView/GridForm.jsx
```diff
@@ -5,10 +5,27 @@ import Draggable from 'react-draggable';
...
 import { useSelector } from 'react-redux';
+import TooltipTrigger from '../TooltipTrigger';
 
 function handleContextMenu(e) {
   e.stopPropagation();
 }
+/**
+ * Returns the tooltip content for the hide button
+ *
+ * @param {boolean} userHidden - if the grid is hidden by the user
```
marcus-oscarsson, line 16: :+1: Nice, thanks

ui/src/components/SampleView/SampleView.css
```diff
@@ -164,8 +164,13 @@
...
 
-.selected {
-  background-color: #dff0d8 !important;
+table.table-striped > tbody > tr[data-selected='true'] {
+  background-color: #dff0d8;
   color: #000;
   font-weight: bold;
 }
+
+table.table-striped > tbody > tr[data-user-hidden='true'] {
```
marcus-oscarsson, line 173: :+1:

marcus-oscarsson:
I guess this is good to go then ?, Ill merge  https://github.com/mxcube/mxcubecore/pull/1044 then you can update `pyrproject.toml`

marcus-oscarsson:
My apologies it took sometime to get PR on `mxcubecore` merged, its done now :)

PR #1462, opened 2024-10-21
Extract `VideoSizeControl` from `SampleControls`

axelboc:
Since it was the last component to be extracted, I also refactor `SampleControls` and clean-up the props that are passed down to it from `SampleViewContainer` through `SampleImage`.

Finally, I review all the sample controls to make sure they have a consistent design (notably when hovered/active/expanded) and are more accessible (`id`, outline on `:focus-visible`, etc.) This includes using Bootstrap's `Popover` component to have consistent overlay styles, using Bootstrap's CSS variables to remove a few custom styles, and using `:not(data-default-styles)` to remove some of our own global styles that make things less accessible (e.g. by removing focus outlines).

Here's a screen recording that demonstrates the result:

Peek 2024-10-21 11-24

... and another one to demonstrate the focus outlines when navigating with the keyboard:

Peek 2024-10-21 11-28

### DIFF
--- a/ui/src/actions/beamline.js
+++ b/ui/src/actions/beamline.js
@@ -45,10 +45,6 @@ export function updateBeamlineHardwareObjectStateAction(data) {
   return { type: BL_UPDATE_HARDWARE_OBJECT_STATE, data };
 }
 
-export function setBeamlineAttribute(name, value) {
-  return updateBeamlineHardwareObjectAction({ name, value });
-}
-
 export function setAttribute(name, value) {
   return (_, getState) => {
     const state = getState();
--- a/ui/src/components/SampleControls/SampleControls.jsx
+++ b/ui/src/components/SampleControls/SampleControls.jsx
@@ -22,10 +22,10 @@ function SampleControls(props) {
       {useShowControl('focus') && <FocusControl />}
       {useShowControl('zoom') && <ZoomControl />}
       {useShowControl('backlight') && (
-        <LightControl label="Frontlight" hwoId="diffractometer.backlight" />
+        <LightControl label="Backlight" hwoId="diffractometer.backlight" />
       )}
       {useShowControl('frontlight') && (
-        <LightControl label="Backlight" hwoId="diffractometer.frontlight" />
+        <LightControl label="Frontlight" hwoId="diffractometer.frontlight" />
       )}
       {useShowControl('video_size') && <VideoSizeControl />}
     </div>
--- a/ui/src/components/SampleControls/SampleControls.jsx
+++ b/ui/src/components/SampleControls/SampleControls.jsx
@@ -8,31 +8,28 @@ import ZoomControl from './ZoomControl';
 import LightControl from './LightControl';
 import VideoSizeControl from './VideoSizeControl';
 
+import { useShowControl } from './utils';
 import styles from './SampleControls.module.css';
 
-class SampleControls extends React.Component {
-  render() {
-    const { getControlAvailability } = this.props;
+function SampleControls(props) {
+  const { canvas } = props;
 
-    return (
-      <div className={styles.controls}>
-        {getControlAvailability('snapshot') && (
-          <SnapshotControl canvas={this.props.canvas} />
-        )}
-        {getControlAvailability('draw_grid') && <GridControl />}
-        {getControlAvailability('3_click_centring') && <CentringControl />}
-        {getControlAvailability('focus') && <FocusControl />}
-        {getControlAvailability('zoom') && <ZoomControl />}
-        {getControlAvailability('backlight') && (
-          <LightControl label="Frontlight" hwoId="diffractometer.backlight" />
-        )}
-        {getControlAvailability('frontlight') && (
-          <LightControl label="Backlight" hwoId="diffractometer.frontlight" />
-        )}
-        {getControlAvailability('video_size') && <VideoSizeControl />}
-      </div>
-    );
-  }
+  return (
+    <div className={styles.controls}>
+      {useShowControl('snapshot') && <SnapshotControl canvas={canvas} />}
+      {useShowControl('draw_grid') && <GridControl />}
+      {useShowControl('3_click_centring') && <CentringControl />}
+      {useShowControl('focus') && <FocusControl />}
+      {useShowControl('zoom') && <ZoomControl />}
+      {useShowControl('backlight') && (
+        <LightControl label="Frontlight" hwoId="diffractometer.backlight" />
+      )}
+      {useShowControl('frontlight') && (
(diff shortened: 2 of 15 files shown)


ui/src/actions/beamline.js
```diff
@@ -45,10 +45,6 @@ export function updateBeamlineHardwareObjectStateAction(data) {
   return { type: BL_UPDATE_HARDWARE_OBJECT_STATE, data };
 }
 
-export function setBeamlineAttribute(name, value) {
-  return updateBeamlineHardwareObjectAction({ name, value });
-}
-
```
axelboc, line 51: Obsolete — this was modifying only the local state. Updating a beamline attribute should always be done via `setAttribute` and corresponding API call.

ui/src/components/SampleControls/SampleControls.jsx
```diff
@@ -1,95 +1,35 @@
...
-        )}
-      </div>
-    );
-  }
+function SampleControls(props) {
+  const { canvas } = props;
+
+  return (
+    <div className={styles.controls}>
+      {useShowControl('snapshot') && <SnapshotControl canvas={canvas} />}
```
axelboc, line 19: Now that it's a function component, I can use a hook to check whether a control is to be shown.

ui/src/components/SampleControls/SampleControls.module.css
```diff
@@ -101,16 +87,36 @@
...
 }
 
-.overlay {
-  position: absolute;
-  z-index: 1001;
-  display: flex;
-  margin-top: 0.5rem;
-  padding: 0.625rem 0.75rem;
-  background-color: rgba(0, 0, 0, 0.6);
-  border-radius: 4px;
```
axelboc, line 111: This all goes away except the custom background colour thanks to `Popover` (cf. `.popover`). The background colour is now defined via Bootstrap's `--bs-popover-bg` and matches the background colour of the sample controls bar (slightly grey instead of black so that when the sample video is black, the overlays still stand out).

ui/src/components/SampleControls/SampleControls.jsx
```diff
@@ -1,95 +1,35 @@
...
-        key="reset"
-        onClick={() => {
-          this.props.sampleViewActions.setVideoSize(
-            this.props.width,
-            this.props.height,
-          );
-        }}
-      >
-        <span className="far fa-redo" /> Reset
-      </Dropdown.Item>,
```
axelboc, line 51: I've removed the _Reset_ item from the _Video size_ dropdown, as I don't think it did anything (it just reapplies the current width/height...) Did it ever work? If so what was its purpose?

ui/src/components/SampleControls/VideoSizeControl.jsx
```diff
@@ -0,0 +1,46 @@
...
+    >
+      <Dropdown.Toggle className={styles.dropdownBtn} data-default-styles>
+        <i className={`${styles.controlIcon} fas fa-video`} />
+        <i className={`${styles.dropdownIcon} fas fa-sort-down`} />
+        <span className={styles.controlLabel}>Video size</span>
+      </Dropdown.Toggle>
+
+      <Dropdown.Menu className={styles.dropdownMenu}>
+        {videoSizes.map(([w, h], i) => {
+          const isActive = w.toString() === width;
```
axelboc, line 27: The sizes are defined as numbers in `sampleview.videoSizes`, but as strings in `sampleview.width/height`. It's not ideal. I traced back the latter to `HWR.beamline.sample_view.camera.get_width/height()`. Not sure what the feasibility would be of changing these methods to return numbers.

ui/src/components/SampleView/SampleImage.jsx
```diff
@@ -890,13 +891,10 @@ export default class SampleImage extends React.Component {
               selectedGrids={this.props.selectedGrids.map((grid) => grid.id)}
             />
             {this.createVideoPlayerContainer(this.props.videoFormat)}
-            <SampleControls
-              showErrorPanel={this.props.showErrorPanel}
-              {...this.props}
```
axelboc, line 895: This line was giving me nightmares :jack_o_lantern: :joy_cat:

ui/src/components/SampleView/SampleView.css
```diff
@@ -89,14 +89,6 @@
   position: relative;
 }
 
-.btn:focus {
-  outline: none !important;
```
axelboc, line 93: This was a duplicate global style. In the remaining style declaration, I add `:not([data-default-styles])` to allow opting out (see further down).

ui/src/components/SampleView/SampleView.css
```diff
@@ -89,14 +89,6 @@
...
-}
-
-.btn:active {
-  color: grey !important;
```
axelboc, line 97: This style would apply when pressing (and remaining pressed) on any button in the UI. It was quite ugly in my opinion, as it would make button labels/icons unreadable on click. I couldn't find a button where this style made the active state look better, so I think it's safe to remove it. If there's a regression somewhere, it should be barely noticeable. But do let me know if you notice one.

ui/src/components/SampleControls/utils.js
```diff
@@ -1,3 +1,13 @@
+import { useSelector } from 'react-redux';
+
+export function useShowControl(name) {
```
marcus-oscarsson, line 3: Nice !

ui/src/components/SampleControls/VideoSizeControl.jsx
marcus-oscarsson: I think that can be done quite easily :)

ui/src/components/SampleControls/SampleControls.jsx
marcus-oscarsson: I think its a regression from the past, It used to change to set the video size to the current size, meaning that it would simply restart the streaming with the same size

axelboc: It was working as expected then, but this brings up two questions: 
- Why would one want to restart de streaming ? Can it ever stop? 
- Would a dedicated button called "restart streaming" somewhere be clearer?

axelboc: As discussed, re-selecting the current video size has the same effect as the now-removed "Reset" button, so I won't re-add the "Reset" button.

axelboc:
FYI, I've fixed the mislabeling of the Backlight and Frontlight buttons introduced in #1459

ui/src/components/SampleControls/SampleControls.jsx
marcus-oscarsson: Thanks !

PR #1464, opened 2024-10-22
Enforce import order with `isort`

fabcor-maxiv:


fabcor-maxiv:
The critical import + monkey patching from gevent seems to be preserved:

https://github.com/mxcube/mxcubeweb/pull/1464/files#diff-e9f205861e6eeb2baea071733449d0d817ad8b35c7f1f24c90ad7dc4a82139b3R1

axelboc:
FYI, I'll do the same in the front-end once we've removed most global styles imports.

elmjag (approved):
Looks nice!

PR #1466, opened 2024-10-23
makes app read collect HWO config when started

mockoocy:
This PR intends to fix the bug in which when some values in the app config change, even though they shouldn't.

For example
- the app is started
- `NUM_SNAPSHOTS` is set to 4 here: https://github.com/mxcube/mxcubeweb/blob/c6c520b220f5140317a71713991fc2fb5b08f8c8/mxcubeweb/app.py#L241
- users logs out 
- `NUM_SNAPSHOTS` is set to value from `HWR.beamline.collect` here https://github.com/mxcube/mxcubeweb/blob/c6c520b220f5140317a71713991fc2fb5b08f8c8/mxcubeweb/core/components/queue.py#L2125-L2133

The expected outcome should be that these settings should be loaded when the app starts instead of when user logs out. It also messes up with default setting for `NUM_SNAPSHOTS` we want at MicroMAX, as we want to set it 0.

### DIFF
--- a/mxcubeweb/app.py
+++ b/mxcubeweb/app.py
@@ -238,7 +238,7 @@ class MXCUBEApplication:
     AUTO_ADD_DIFFPLAN = False
 
     # Number of sample snapshots taken before collect
-    NUM_SNAPSHOTS = 4
+    DEFAULT_NUM_SNAPSHOTS = 4
 
     # Remember collection paramters between samples
     # or reset (defualt) between samples.
@@ -520,7 +520,9 @@ class MXCUBEApplication:
             "VIDEO_FORMAT": MXCUBEApplication.VIDEO_FORMAT,
             "AUTO_MOUNT_SAMPLE": MXCUBEApplication.AUTO_MOUNT_SAMPLE,
             "AUTO_ADD_DIFFPLAN": MXCUBEApplication.AUTO_ADD_DIFFPLAN,
-            "NUM_SNAPSHOTS": MXCUBEApplication.NUM_SNAPSHOTS,
+            "NUM_SNAPSHOTS": HWR.beamline.collect.get_property(
+                "num_snapshots", MXCUBEApplication.DEFAULT_NUM_SNAPSHOTS
+            ),
             "UI_STATE": MXCUBEApplication.UI_STATE,
         }
 
--- a/mxcubeweb/core/components/queue.py
+++ b/mxcubeweb/core/components/queue.py
@@ -41,6 +41,7 @@ ORIGIN_MX3 = "MX3"
 class Queue(ComponentBase):
     def __init__(self, app, config):
         super().__init__(app, config)
+        self.init_queue_settings()
 
     def build_prefix_path_dict(self, path_list):
         prefix_path_dict = {}
--- a/mxcubeweb/core/components/queue.py
+++ b/mxcubeweb/core/components/queue.py
@@ -265,7 +265,6 @@ class Queue(ComponentBase):
         settings = {}
 
         for setting_name in [
-            "NUM_SNAPSHOTS",
             "REMEMBER_PARAMETERS_BETWEEN_SAMPLES",
             "CENTRING_METHOD",
             "AUTO_ADD_DIFFPLAN",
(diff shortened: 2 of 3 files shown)


marcus-oscarsson:
Nice, I guess it means that we can remove:

https://github.com/mxcube/mxcubeweb/blob/c6c520b220f5140317a71713991fc2fb5b08f8c8/mxcubeweb/app.py#L241

And perhaps remove the reference to `app.NUM_SNAPSHOTS` completely and replace it with  `HWR.beamline.collect.number_of_snapshots`

elmjag:
I have tested this, it works. However, looking that the code I agree with @marcus-oscarsson, let's take a chance and clean-up this code a bit.

As Marcus suggests, let's remove the [NUM_SNAPSHOTS](https://github.com/mxcube/mxcubeweb/blob/develop/mxcubeweb/app.py#L241) variable. I suggest that we replace `NUM_SNAPSHOTS` references with something like this:

`HWR.beamline.collect.get_property("num_snapshots", DEFAULT_NUM_SNAPSHOTS)`

This way the `num_snapshots` will still be an optional property for the collect hardware object.

marcus-oscarsson:
> `HWR.beamline.collect.get_property("num_snapshots", DEFAULT_NUM_SNAPSHOTS)`
> 
> This way the `num_snapshots` will still be an optional property for the collect hardware object.

:+1: seems reasonable :)

mxcubeweb/app.py
```diff
@@ -520,7 +520,9 @@ def save_settings():
             "VIDEO_FORMAT": MXCUBEApplication.VIDEO_FORMAT,
             "AUTO_MOUNT_SAMPLE": MXCUBEApplication.AUTO_MOUNT_SAMPLE,
             "AUTO_ADD_DIFFPLAN": MXCUBEApplication.AUTO_ADD_DIFFPLAN,
-            "NUM_SNAPSHOTS": MXCUBEApplication.NUM_SNAPSHOTS,
+            "NUM_SNAPSHOTS": HWR.beamline.collect.get_property(
+                "num_snapshots", MXCUBEApplication.DEFAULT_NUM_SNAPSHOTS
+            ),
```
mockoocy, line 525: Haven't seen any references to the `save_settings` function these lines are part of, nor the file it creates. 
I wonder if it is even used or is some kind of leftover code? I haven't seen a way to load it.

mockoocy:
Removed the `NUM_SNAPSHOTS` variable.

mxcubeweb/app.py
elmjag: I _think_ this is dead code. Some kind remains of the code for saving MXCuBE state to disk!?

Feel free to make a PR which removes the `save_settings()`.

mxcubeweb/core/components/queue.py
```diff
@@ -2481,3 +2487,10 @@ def set_setting(self, name_value: SimpleNameValue) -> tuple:
             result = ()
 
         return result
+
+    def set_num_snapshots(self, num_snapshots: int):
+        """Sets the number of snapshots to take during data collection
+        Args:
```
elmjag, line 2493: Change `Args` to `Parameters`.

This is the mark-up style our sphix doc generator expects here.

fabcor-maxiv: The list of sections is here: <https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#docstring-sections>.

`Args`, and `Arguments` are aliases for `Parameters`.

I do not know that we have a policy of picking one name over the others.

mockoocy: Changed in the latest push, but I am a bit confused.
https://mxcubeweb.readthedocs.io/en/latest/dev/contributing.html#style-guidelines
here it says that `Args` should be used and searching for string `'Parameters:'` in the codebase did not yield any results in python files. 

Edit: seems like both will work as @fabcor-maxiv  says, personally I think that Args would be better (more consistent with what's used more commonly in codebase)

fabcor-maxiv: Indeed, we agreed to use Google style, and now that I check it again the Google style seems to be recommending `Args`: <https://google.github.io/styleguide/pyguide.html#doc-function-args>. So let's go with `Args`, but really any alias is fine for now. Hopefully at some point we will introduce a linter that checks and enforces the docstrings format for us (because we keep forgetting).

elmjag: I did not know that `Args` and `Arguments` also work, TIL!

Then *I* was doing this wrong. `Args` is fine then, doh.

elmjag:
FYI, there seems to be issues with actually using the  `Crystal snapshots` widget: https://github.com/mxcube/mxcubeweb/issues/1479

However, as it's already buggy on the `develop` branch, so that's a topic for a different PR.

elmjag (approved):
LGTM!

marcus-oscarsson:
Thanks !

Issue #1405, opened 2024-09-18
Remove `System log` from UI


marcus-oscarsson:
We would like to propose to remove the `System log` tab/view completely from the UI. It does not really add any useful functionality and we don't want to propagate this sort of information to the UI (at least not here at ESRF). What do the rest of you think of removing this from the UI.

![image](https://github.com/user-attachments/assets/32682c4d-1871-4189-a16b-77aa3e0391b5)

marcus-oscarsson:
@elmjag, @meguiraun @fabcor-maxiv, what do you think ?

elmjag:
I have no idea if this is somehow used by the users. We in software group don't need that page, as we can see the log in other places. I think it's probably fine to remove it.

A more cautious approach is to make it configurable if the tab is visible or not. Say put a new option into `mxcube-web/ui.yaml` `system_log_visible`, with default setting false.

Wait X amount of months. If nobody asks about the missing system log tab, just delete it. If somebody is missing the tab, we add documentation on how to make it visible.

marcus-oscarsson:
I agree that would be a more cautious approach. However we are sure that nobody is using it here, and we actually don't want to propagate this sort of information to the UI. If you guys have a doubt we can make it configurable otherwise Ill suggest to simply remove it

meguiraun:
I rarely use this tab, I usually go and check the log file or a kibana dashboard. I checked with our scientist and they don't use it much.

I am ok with removing it.

fabcor-maxiv:
I do not think I have ever used this.

marcus-oscarsson:
:), Ok I think that settles it. Lets remove it. You can send all complains of it missing to me :)

elmjag:
Yes, remove it!

Try to do it in one commit, then in case we'll need to bring it back, it would be easier :)

Issue #1393, opened 2024-09-10
Ensure all REST endpoints respond with JSON


fabcor-maxiv:
Follow-up to <https://github.com/mxcube/mxcubeweb/pull/1392>. In this case we had an endpoint that was not returning JSON. We should investigate if it is possible to make all our REST endpoints return JSON content (if they return content at all), even in case of error (`4xx` and `5xx`).

Unless we have a good reason not to return JSON for some specific cases?

Maybe it can be enforced in a systematic manner, instead of making it for each REST route individually.

And of course we should check that the JavaScript UI frontend expects JSON from the server.

elmjag:
For `500 Internal Server Error` replies, I I don't really see a big point of getting a JSON reply.

Normally 500 reply means that there was some unexpected exception in the back-end python code. In that case you want to see the traceback of that exception. It is very helpful when that traceback is included into the reply body, as text with correct line breaks.

If the traceback, together with request method and URL ended up in the javascript console, that would be very nice, and save time when debugging these issues.

elmjag:
> And of course we should check that the JavaScript UI frontend expects JSON from the server.

Frontend does expect  `application/json` replies:
https://github.com/mxcube/mxcubeweb/blob/develop/ui/src/api/addons/safeJson.js#L7

axelboc:
I would generally think that sending a backend stack trace is a significant security flaw as it reveals the intervals of the application... but since the codebase is public it probably doesn't matter.

Putting this stack trace in a JSON response body still makes sense for consistency. We don't want to have to detect and handle 500 errors specifically in every try/catch block on the front-end. We want our error handling code to be simple and generic. Stringified `\n` line breaks can easily be dealt with.

fabcor-maxiv:
> For `500 Internal Server Error` replies, I I don't really see a big point of getting a JSON reply.

Whenever something goes wrong on the Python backend the JavaScript UI frontend should be able to show meaningful error messages to the user, so the frontend needs to get the error message from the backend, parse it and translate it into a human readable message. That is for the parsing step that JSON makes sense. Typically there would be some JSON like this:

```json
{
  "error": "NothingWorksError",
  "details": "Nothing works, everything is broken",
  "traceback": "..."
}
```

[...]

PR #1468, opened 2024-10-24
removing system log

walesch-yan:
This PR removes the System log from the UI as discussed in #1405 including actions and state variables related to it.

While at it I modified the `MX3LoggingHandler`, so that only records from the `user-level-log` are kept in the front-end part.


mxcubeweb/logging_handler.py
```diff
@@ -26,14 +21,10 @@ def _record_to_json(self, record):
             "message": record.getMessage(),
             "severity": record.levelname,
             "timestamp": record.asctime,
-            "logger": record.name,
-            "stack_trace": stack_trace,
```
walesch-yan, line 30: I removed the `logger` and `stack_trace` from the LoggingHandler, as we only need the user_level_log to be shown in the UI and due to some concerns about the stack_trace being sent to the front-end in #1393. Could this removal be confirmed as accepted or is the general opinion to keep it?

marcus-oscarsson:
Wonderful :)

PR #1469, opened 2024-10-24
[Queue] - Removed queue_mount_sample

marcus-oscarsson:
This is the PR corresponding to https://github.com/mxcube/mxcubecore/pull/971 and https://github.com/mxcube/mxcubecore/issues/970

Replaced `queue_mount_sample` with the corresponding method in `mxubcecore` as it should have been from the beginning ;): https://github.com/mxcube/mxcubecore/blob/14fa9ddfca44dbed45785651ca85e1f0ef928afd/mxcubecore/queue_entry/base_queue_entry.py#L768

Drafting until  https://github.com/mxcube/mxcubecore/pull/971 is merged


beteva:
Autumn clean. Very good!

mxcubeweb/core/components/samplechanger.py
```diff
@@ -413,181 +408,3 @@ def _gripper_changed(self):
...
-            mxcube.sample_changer.mount_sample_clean_up(sample)
-        else:
-            sample = {
-                "location": data_model.loc_str,
-                "sampleID": data_model.loc_str,
-            }
-
-            # in this case the sample changer takes the sample from an Harvester
-            # We Harvest the sample and make it ready to load
-            if mount_from_harvester:
```
jbflo, line 465: this harvester  logic need to be ported here 
https://github.com/mxcube/mxcubecore/pull/971  right ?

mxcubeweb/core/components/samplechanger.py
```diff
@@ -413,181 +408,3 @@ def _gripper_changed(self):
...
-                # This is to preserve backward compatibility (load_sample was
-                # supposed to return None); if sample could not be loaded, but
-                # no exception is raised, let's skip the sample
-
-                raise queue_entry.QueueSkipEntryException(
-                    "Sample changer could not load sample", ""
-                )
-
-    # Harvest Next sample while loading current
-    if mount_from_harvester and not HWR.beamline.harvester.get_room_temperature_mode():
```
jbflo, line 488: Here also,  to be ported here on this PR
https://github.com/mxcube/mxcubecore/pull/971

mxcubeweb/core/components/samplechanger.py
```diff
@@ -413,181 +408,3 @@ def _gripper_changed(self):
...
-        dm = HWR.beamline.diffractometer
-
-        if mount_from_harvester:
-            try:
-                logging.getLogger("user_level_log").info(
-                    "Start Auto Harvesting Centring"
-                )
-
-                computed_offset = (
-                    HWR.beamline.harvester.get_offsets_for_sample_centering()
```
jbflo, line 516: Here also,  to be ported here on this PR
https://github.com/mxcube/mxcubecore/pull/971

jbflo (commented):
Some logic were removed here, that will cause issues if not ported to mxcubecore 
this PR i Guess : https://github.com/mxcube/mxcubecore/pull/971

mxcubeweb/core/components/samplechanger.py
marcus-oscarsson: Actually, it would perhaps be better if this was integrated in the harvester or the sample changer so that we don't need to add the extra logic for handling the harvester in this method.

Maybe int this case it would be better to "proxy" the sample changer via the harvester and like this the harvester is the sample changer. It would make the entire configuration easier

jbflo: Agreed, and i think it would be better to do it right away in the mxcubecore related PR.
this one :  https://github.com/mxcube/mxcubecore/pull/971

marcus-oscarsson: We would need to add `queue_harvest_sample` and `queue_harvest_next_sample` to https://github.com/mxcube/mxcubecore/blob/14fa9ddfca44dbed45785651ca85e1f0ef928afd/mxcubecore/HardwareObjects/EMBLFlexHarvester.py#L141. I could do something but it would probably be better if you could help me out as you know how the harvester works a bit better.

jbflo: OK, 
good  I remove them from this pr.

marcus-oscarsson: Thanks @jbflo

elmjag:
I was not able to test this against latest develop branch due to merge conflicts. 

However, I suspect that this PR needs to bump mxcubecore version to at least  1.173.0. So that this https://github.com/mxcube/mxcubecore/commit/cb742731c8669149bc5741681f006bf1d0d58823 commit is included.

Issue #1473, opened 2024-10-28
CI failing on poetry-export 


marcus-oscarsson:
The CI is failing on poetry-export, it seems that we now need to install this plugin explicitly ?
```
poetry-export............................................................Failed
- hook id: poetry-export
- files were modified by this hook
```

Warning: poetry-plugin-export will not be installed by default in a future version of Poetry.

fabcor-maxiv:
> it seems that we now need to install this plugin explicitly ?

We already do: <https://github.com/mxcube/mxcubeweb/pull/1269/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9R38>.

Where does it fail? This one <https://github.com/mxcube/mxcubeweb/actions/runs/11549893904/job/32143790326?pr=1471>?

fabcor-maxiv:
So, it should not show the warning. The issue has been reported: <https://github.com/python-poetry/poetry-plugin-export/issues/249>.

marcus-oscarsson:
Here:
   https://github.com/mxcube/mxcubeweb/actions/runs/11527304882/job/32092840490

On second thought I don't think its the plugin install that fails but that this will always fail on `dependabot` updates right ?

walesch-yan:
Yes, this one https://github.com/mxcube/mxcubeweb/actions/runs/11549893904/job/32143790326?pr=1471 failed due to poetry-export, but only after rebasing it to the newest changes in mxcubeweb

fabcor-maxiv:
> On second thought I don't think its the plugin install that fails but that this will always fail on `dependabot` updates right ?

Oh... Yeah... That might be the issue...

I need to think about it.

walesch-yan:
> Here: https://github.com/mxcube/mxcubeweb/actions/runs/11527304882/job/32092840490
> 
> On second thought I don't think its the plugin install that fails but that this will always fail on `dependabot` updates right ?

I think this was actually the issue, since we did not bump the version of `werkzeug` in the `requirements.txt` file. 



> Yes, this one https://github.com/mxcube/mxcubeweb/actions/runs/11549893904/job/32143790326?pr=1471 failed due to poetry-export, but only after rebasing it to the newest changes in mxcubeweb

This one then failed because I did not re-run the pre-commits after the rebase, I tried it again with the needed changes for `requirements.txt` and passed the tests.

walesch-yan:
[...]

PR #1475, opened 2024-10-28
fix CI failing on poetry-export

walesch-yan:
This PR should fix the current CI failures mentioned in #1473.


marcus-oscarsson:
Thanks !

PR #1478, opened 2024-10-29
Refactor visibility and positioning logic of context menu in sample view

axelboc:
I'm starting to refactor `SampleView/ContextMenu.jsx`, which is the component that renders the context menu of the sample view. It's a bit class component with a lot of logic for all the various actions that the context menu allows, and therefore a lot of props. The logic for rendering the menu options is difficult to follow, and the component uses `UNSAFE_componentWillReceiveProps`, which is obviously not great. I'd like to extract a lot of the action-related logic to reduce the reliance on global state, render the options more declaratively in JSX if possible, and then of course convert the component to a function and replace all the props with `useSelector`.

But first, I do a bit of a clean-up of the visibility and positioning logic of the context menu:

- I make better use of Bootstrap's `Dropdown` component and all its features, which avoids having to manually hide the menu ourselves after every action or when clicking outside of it (`autoClose` + `onToggle`).
- I show/hide the menu declaratively rather than imperatively with the `show` prop, which removes the need for `UNSAFE_componentWillReceiveProps`.
- I render the menu as a direct child of `document.body` via a React Portal, which simplifies the positioning logic: I can now save the position of the right-click event relative to the page (`pageX` and `pageY`) and use that directly as the menu's `left` and `top` positions.

Note that the top left-hand corner of the context menu is now exactly under the cursor, which matches the behaviour of the native context menu:

[Screencast from 2024-10-29 09-33-11.webm](https://github.com/user-attachments/assets/47c8e9b7-eeed-44bb-b493-80a605001e1c)

### DIFF
--- a/ui/src/components/SampleView/ContextMenu.jsx
+++ b/ui/src/components/SampleView/ContextMenu.jsx
@@ -1,48 +1,32 @@
 /* eslint-disable react/jsx-handler-names */
 import React from 'react';
+import { createPortal } from 'react-dom';
 import { Dropdown } from 'react-bootstrap';
 import { getLastUsedParameters } from '../Tasks/fields';
 
-// eslint-disable-next-line react/no-unsafe
+const BESPOKE_TASK_NAMES = new Set([
+  'datacollection',
+  'characterisation',
+  'xrf_spectrum',
+  'energy_scan',
+  'mesh',
+  'helical',
+  'workflow',
+  'interleaved',
+]);
+
 export default class ContextMenu extends React.Component {
   constructor(props) {
     super(props);
     this.toggleDrawGrid = this.toggleDrawGrid.bind(this);
     this.menuOptions = this.menuOptions.bind(this);
-    this.hideContextMenu = this.hideContextMenu.bind(this);
-  }
-
-  componentDidMount() {
-    document.addEventListener('click', this.hideContextMenu);
-  }
-
-  UNSAFE_componentWillReceiveProps(nextProps) {
-    if (nextProps.show) {
-      this.showContextMenu(nextProps.x, nextProps.y);
-    } else {
-      this.hideContextMenu();
-    }
-  }
-
-  componentWillUnmount() {
-    document.removeEventListener('click', this.hideContextMenu);
   }
 
   // eslint-disable-next-line sonarjs/cognitive-complexity
   menuOptions() {
-    const bespokeTaskNames = new Set([
-      'datacollection',
-      'characterisation',
-      'xrf_spectrum',
-      'energy_scan',
-      'mesh',
-      'helical',
-      'workflow',
-      'interleaved',
-    ]);
     const generalTaskNames = Object.keys(
       this.props.taskForm.defaultParameters,
-    ).filter((tname) => !bespokeTaskNames.has(tname));
+    ).filter((tname) => !BESPOKE_TASK_NAMES.has(tname));
 
     const genericTasks = {
       point: [],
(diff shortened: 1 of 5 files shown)


ui/src/components/SampleView/ContextMenu.jsx
```diff
@@ -1,48 +1,32 @@
...
+const BESPOKE_TASK_NAMES = new Set([
+  'datacollection',
+  'characterisation',
+  'xrf_spectrum',
+  'energy_scan',
+  'mesh',
+  'helical',
+  'workflow',
+  'interleaved',
+]);
```
axelboc, line 16: Unrelated, just extracting this as a constant.

ui/src/components/SampleView/ContextMenu.jsx
```diff
@@ -1,48 +1,32 @@
...
   constructor(props) {
     super(props);
     this.toggleDrawGrid = this.toggleDrawGrid.bind(this);
     this.menuOptions = this.menuOptions.bind(this);
-    this.hideContextMenu = this.hideContextMenu.bind(this);
-  }
-
-  componentDidMount() {
-    document.addEventListener('click', this.hideContextMenu);
-  }
```
axelboc, line 17: This is already taken care of by Bootstrap's `Dropdown`.

ui/src/components/SampleView/ContextMenu.jsx
```diff
@@ -1,48 +1,32 @@
...
-
-  UNSAFE_componentWillReceiveProps(nextProps) {
-    if (nextProps.show) {
-      this.showContextMenu(nextProps.x, nextProps.y);
-    } else {
-      this.hideContextMenu();
-    }
-  }
```
axelboc, line 25: Replaced with declarative visibility and positioning in the JSX.

ui/src/components/SampleView/ContextMenu.jsx
```diff
@@ -430,18 +402,14 @@ export default class ContextMenu extends React.Component {
         }
       });
     }
-
-    this.props.sampleViewActions.showContextMenu(false);
```
axelboc, line 434: All of these are replaced with Bootstrap's Dropdown's `autoClose` feature.

ui/src/components/SampleView/ContextMenu.jsx
```diff
@@ -538,10 +488,28 @@ export default class ContextMenu extends React.Component {
     } else {
       optionList = menuOptions.NONE.map(this.listOptions);
     }
-    return (
-      <Dropdown.Menu show id="contextMenu" role="menu">
-        {optionList}
-      </Dropdown.Menu>
```
axelboc, line 544: Note that we weren't actually using the root `Dropdown` component, which is where all the interesting features are. Of course, what matters is that we don't render a `Dropdown.Trigger` since this is not an actual "dropdown" UI element.

ui/src/components/SampleView/ContextMenu.jsx
```diff
@@ -538,10 +488,28 @@ export default class ContextMenu extends React.Component {
...
+
+    return createPortal(
```
axelboc, line 495: Note the use of [`createPortal(..., document.body)`](https://react.dev/reference/react-dom/createPortal) as mentioned.

ui/src/components/SampleView/SampleImage.jsx
```diff
@@ -500,7 +500,7 @@ export default class SampleImage extends React.Component {
       this.canvas.discardActiveObject();
     }
 
-    showContextMenu(true, ctxMenuObj, e.offsetX, e.offsetY);
```
axelboc, line 503: `offsetX/Y` is relative to the container (here the sample view container). By rendering the menu under `document.body`, and using `pageX/Y` instead, we make the positioning/stacking/... more robust. For instance, we could now put `overflow: hidden` on the sample view container if we ever need to.

axelboc:
I should point out that other context menus in the app use [`react-contextify`](https://github.com/fkhadra/react-contexify). Since this library has not been updated in 2 years, I think it's safer to not rely on it more than we already do.

marcus-oscarsson:
Whohoo :)

axelboc:
Please don't merge! Just noticed a regression: there are a couple of features that rely on the `x/y` position of the click relative to the canvas and not the page (like "Go to beam"). Will think of a solution.

marcus-oscarsson:
:), hihi not merging

axelboc:
Okay, should be fixed now. I simply store both the page-relative and image-relative coordinates in the `contextMenu` state.

axelboc:
FYI, in my next PR, I will store the sample stage coordinates (`evt.offsetX / imageRatio` and `evt.offsetY / imageRatio`) in the context menu state instead of just `evt.offsetX` and `evt.offsetY`. This will save us from having to compute them in `ContextMenu`, both when creating points and "going to beam".

ui/src/components/SampleView/ContextMenu.jsx
marcus-oscarsson: nice !

marcus-oscarsson:
Nice !

PR #1482, opened 2024-10-31
fix queue_mount_sample() signature

elmjag:
The queue_mount_sample() is now called with 3 arguments, `view` argument is no longer provided.

The invocation of queue_mount_sample() have been changed in this commit: https://github.com/mxcube/mxcubecore/commit/cb742731c8669149bc5741681f006bf1d0d58823

This change makes mxcubeweb incompatible with older versions of mxcubecore, thus we need to bump mxcubecore version.

This change fixed the pytest tests, when running against latest mxcubecore.

### DIFF
--- a/mxcubeweb/core/components/samplechanger.py
+++ b/mxcubeweb/core/components/samplechanger.py
@@ -416,7 +416,7 @@ class SampleChanger(ComponentBase):
 
 
 # Disabling C901 function is too complex (19)
-def queue_mount_sample(view, data_model, centring_done_cb, async_result):  # noqa: C901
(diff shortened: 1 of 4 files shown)


elmjag:
After I made this PR, I noticed this one: https://github.com/mxcube/mxcubeweb/pull/1469

I'm guessing that https://github.com/mxcube/mxcubeweb/pull/1469 also solves the problem with failing tests. However, I was not able to test it, as there is some merge conflict now with latests develop branch.

We should probably merge #1469 rather then this PR. But I'll leave this PR open for now, until we sort out the failing tests.

marcus-oscarsson:
Thanks @elmjag, its indeed meant to be taken care of in #1469. Ill close this for now

PR #1483, opened 2024-10-31
Upgrade mxcubecore and remove obsolete conversion of sampleview size in `VideoSizeControl`

axelboc:
The check for the currently selected video size is now a tiny bit simpler.

I also found another instance in `SampleImage` where the sampleview width/height was being implicitly coerced from a string to a number. No change needed, though — the coercion just no longer happens, which is always good (TypeScript would have complained about it, for instance).

Thanks @walesch-yan for the fix in the back-end :wink:


axelboc:
There's a test failure, but I don't think it's related :shrug:

axelboc:
Nevermind, thanks @elmjag #1482

walesch-yan:
@axelboc, thanks for doing the necessary changes in the front-end. From the log message in the failing test, it seems to be coming from a mistake that I made in the fix. I used the built-in python types for type hints, however these are only supported from `Python 3.9` onwards, hence the tests failed on version 3.8. I made a new PR mxcube/mxcubecore#1070 for this.

marcus-oscarsson:
@axelboc I tried to fix the conflicts related to the poetry lock file, I think I got right, finally. Sorry for the mess

PR #1484, opened 2024-10-31
fix mesh grid drawn in wrong positions sometimes

mockoocy:
## Mesh grid being drawn in wrong place
This PR fixes the following bug:
1. Open draw grid control
2. Draw some mesh grid
3. Move mouse outside of the unsaved grid
4. drag mouse somewhere (in other direction that bottom-left)
5. right click on previously drawn grid
6. Press 'Save Grid'
The grid would be drawn in some other place on the canvas

This bug was caused, because the top-left coordinates of the grid were set on `SampleImage`'s `onMouseDown` handler. So if we first drew the grid, and then dragged the mouse the following would happen:
- if the direction of the drag is towards bottom-left corner of the screen, then new grid would be drawn - this is ok
- otherwise, no new grid would appear, but the old grid's top-left corner coordinates would be changed - not ok.

I fixed it by only allowing the position of drawn grid's top-left corner to be changed if a grid should be really be redrawn.

Also there are 2 other minor changes:
### Typedef for `GridData`
I decided to add JSDoc [@typedef](https://jsdoc.app/tags-typedef) annotation to provide some kind of type hinting and better autocomplete in code editors. Regarding the type hint, I am not 100% sure if the `result` property is typed properly, and if it used at all? I haven't noticed a way to make it take another value than `''` or `null`, so I guess it could be removed in some other PR?
### small refactor
Removed 2 functions that were unused.

### DIFF
--- a/ui/src/components/SampleView/DrawGridPlugin.jsx
+++ b/ui/src/components/SampleView/DrawGridPlugin.jsx
@@ -99,6 +99,26 @@ const MeshGridResult = fabric.util.createClass(fabric.Object, {
     }
   },
 });
+/**
+ * @typedef {Object} GridData
+ * @property {[number, number]} screenCoord
+ * @property {number} top
+ * @property {number} left
+ * @property {number} width
+ * @property {number} height
+ * @property {number} cellWidth
+ * @property {number} cellHeight
+ * @property {number} cellVSpace
+ * @property {number} cellHSpace
+ * @property {number} numCols
+ * @property {number} numRows
+ * @property {string|null} cellCountFun - name of the function
+ * @property {boolean} selected
+ * @property {string|null} id
+ * @property {'' | null | number[]} result
+ * @property {number} pixelsPerMMX
+ * @property {number} pixelsPerMMY
+ */
 
 /**
  * GridData object defines a grid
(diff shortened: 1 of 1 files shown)


mockoocy:
Bug on video:
https://github.com/user-attachments/assets/00ff11de-a20a-4371-80de-577fce337d2d

fhernandezvivanco:
Hi @mockoocy , we're also encountering a similar issue, but in our case the the grid shifts after rendering the heatmap. Did you also experience the same behavior?

mockoocy:
> Hi @mockoocy , we're also encountering a similar issue, but in our case the the grid shifts after rendering the heatmap. Did you also experience the same behavior?

Hi, 
I did not notice anything weird, but there's a chance I just can't reproduce it :)

mockoocy:
Drafted, because I've found that there is another, similar bug

1. Draw grid
2. Right click on it
3. Click somewhere else
4. Right click again
5. Save Grid
It will also appear in some unexpected spot

marcus-oscarsson:
Nice, well done. Thanks for having a look at this. Very nice with the video showing the issue as well.

mockoocy:
Fixed another similar bug (the one that was actually visible on the video :D)
Before, pressing right click on unsaved grid caused it's top-left coordinates (`gridData.screenCoord`) to be scaled.
The issue was that when pressing right click multiple times, these coordinates were scaled multiple times, causing the grid to migrate towards corner coordinates. This was because the spread operator `...` performs a shallow copy of an object, so I used `structuredClone` instead.

elmjag:
> Fixed another similar bug (the one that was actually visible on the video :D) 

I'll test these changes next week.

marcus-oscarsson:
Thanks !

PR #1486, opened 2024-10-31
Added GPhL WF configuration option (suppress log output)

rhfogh:


demo/gphl/gphl-setup.yml
```diff
@@ -44,7 +44,7 @@ gphl_persistname: persistence
 
 software_paths:
   # Mandatory. Directory with workflow config input, e.g. instrumentation.nml
-  gphl_beamline_config: gphl_beamline_config
+  gphl_beamline_config: gphl/config_esrf_id30b
```
marcus-oscarsson, line 47: Are you sure about this ` gphl/config_esrf_id30b`

rhfogh: @marcus-oscarsson  Yes - as it stands - but I admit one might consider changing something.

gphl_beamline_config defines a subdirectory that must contain (among other things) instrumentation.nml, the file that specifies detector type, axis direction, etc. etc. This is by definition beamline-specific, and in real life lives in the beamline configuration set-up. 'gphl_beamline_config' is wrong, there is no such subdirectory. 'gphl/config_esrf_id30b' exists in the check-in, and contains the requisite configuration files, so you can run the workflow in mock mode. A mechanism for selecting the configs for various beamlines would be better, but who would use or maintain it? We could choose another beamline instead of ID30B, but is massif-1 or P14 better? If we want something generic, what detector and goniostat should we configure?

I can see this does not look ideal (though it works) so I am open to advice on how to do it better.

marcus-oscarsson: Its alright as it is, I just wanted to make sure that its what you meant :). You could have a `gphl/demo` or `/examples/` I guess. 

> A mechanism for selecting the configs for various beamlines would be better, but who would use or maintain it? 

I guess that's doable somehow, you mean apart from specifying the configuration directory on the command line ?

PR #1487, opened 2024-11-04
Fix connection lost dialog appearing after saving a file in development

axelboc:
When the entire app reloaded after saving a file such as `serverIO.js` or a reducer file, the "connection lost" dialog would appear and remain until dismissed even though the sockets had successfully reconnected.

### DIFF
--- a/ui/src/serverIO.js
+++ b/ui/src/serverIO.js
@@ -83,10 +83,10 @@ class ServerIO {
   }
 
   disconnect() {
-    this.hwrSocket?.close();
+    this.hwrSocket?.disconnect();
     this.hwrSocket = null;
 
-    this.loggingSocket?.close();
+    this.loggingSocket?.disconnect();
     this.loggingSocket = null;
   }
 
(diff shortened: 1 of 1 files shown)


ui/src/serverIO.js
```diff
@@ -83,10 +83,10 @@ class ServerIO {
   }
 
   disconnect() {
-    this.hwrSocket?.close();
```
axelboc, line 86: `close` is actually just an alias for `disconnect`.

ui/src/serverIO.js
```diff
@@ -104,16 +104,12 @@ class ServerIO {
 
     this.hwrSocket.on('disconnect', (reason) => {
       console.log('hwrSocket disconnected!'); // eslint-disable-line no-console
-      const socket = this.hwrSocket;
```
axelboc, line 107: This was the bug: we're saving a reference to the old socket that was disconnected. Since we recreate a socket every time we call `listen()`, `socket` would keep referring to the previous, disconnected socket, and `socket.connected` would remain `false`.

ui/src/serverIO.js
```diff
@@ -104,16 +104,12 @@ class ServerIO {
...
-        setTimeout(() => {
-          socket.connect(); // try reconnecting
-        }, 500);
-      }
 
       setTimeout(() => {
-        dispatch(showConnectionLostDialog(!socket.connected));
+        dispatch(
+          // Show message if socket still hasn't reconnected (and wasn't manually disconnected in the first place)
+          showConnectionLostDialog(this.hwrSocket && !this.hwrSocket.connected),
```
axelboc, line 111: I now check that the socket is still defined. If it's not, it means the `disconnect()` method was called (and therefore that the entire app reloaded).

ui/src/serverIO.js
```diff
@@ -104,16 +104,12 @@ class ServerIO {
 
     this.hwrSocket.on('disconnect', (reason) => {
       console.log('hwrSocket disconnected!'); // eslint-disable-line no-console
-      const socket = this.hwrSocket;
-
-      if (reason === 'io server disconnect') {
-        setTimeout(() => {
-          socket.connect(); // try reconnecting
-        }, 500);
-      }
```
axelboc, line 113: This is not needed since socket IO already has its own reconnection mechanism.

marcus-oscarsson:
Nice, thanks !

PR #1488, opened 2024-11-04
Disabling test until mxcubecore issues `mxcubecore/#1071` is fixed

marcus-oscarsson:
There was some nice work done to improve the tests in https://github.com/mxcube/mxcubecore/pull/1066, however there is a smaller issue with the current solution, outlined here. https://github.com/mxcube/mxcubecore/issues/1071.

Disabling that specific tests until that issue is fixed.

### DIFF
--- a/test/test_authn.py
+++ b/test/test_authn.py
@@ -137,24 +137,26 @@ def test_authn_same_proposal(make_client):
     assert resp.json["user"]["inControl"] == False
 
 
-# Test against proposal-based authentication only
-@pytest.mark.parametrize("login_type", ["proposal"], indirect=True)
-def test_authn_different_proposals(make_client):
-    """Test two users for different proposals.
+# # Test against proposal-based authentication only
+# @pytest.mark.parametrize("login_type", ["proposal"], indirect=True)
+# def test_authn_different_proposals(make_client):
+#     """Test two users for different proposals.
 
-    If a user signs in for a different proposal than an already signed in user,
-    this user should not be allowed to sign in.
-    """
-    client_0 = make_client()
-    resp = client_0.post(URL_SIGNIN, json=CREDENTIALS_0)
-    assert resp.status_code == 200
-    resp = client_0.get(URL_INFO)
-    assert resp.status_code == 200
+#     If a user signs in for a different proposal than an already signed in user,
+#     this user should not be allowed to sign in.
+#     """
+#     client_0 = make_client()
+#     resp = client_0.post(URL_SIGNIN, json=CREDENTIALS_0)
+#     assert resp.status_code == 200
+#     resp = client_0.get(URL_INFO)
+#     assert resp.status_code == 200
 
-    client_1 = make_client()
-    resp = client_1.post(URL_SIGNIN, json=CREDENTIALS_1)
-    assert resp.status_code == 200
-    assert resp.json["msg"] == "Could not authenticate"
+#     client_1 = make_client()
+#     resp = client_1.post(URL_SIGNIN, json=CREDENTIALS_1)
+#     assert resp.status_code == 200
+#     import pdb
+#     pdb.set_trace()
+#     assert resp.json["msg"] == "Could not authenticate"
 
 
 def test_authn_session_timeout(client):


marcus-oscarsson:
Merging this for now, @fabcor-maxiv and @dominikatrojanowska let me know if you have some input

dominikatrojanowska:
Hi Marcus, thanks for being perceptive :)

This PR https://github.com/mxcube/mxcubecore/pull/1066 was made to allow for those changes #1455, especially passing correctly this [test case](https://github.com/mxcube/mxcubeweb/blob/2d3c486ec8862b8ba761763eb0d0d8af73a71fbf/test/test_authn.py#L142). Here is the [uodated version](https://github.com/mxcube/mxcubeweb/pull/1455/files#diff-948603f00beb75aff4ef52d02fed569c1702af735c26ee1fe13496af0679e7fcR189)

I updated also the test description to explain Max IV use case. Let me know what do you think about it. If the use case does not fits to ESRF's one I think we can keep it on our maxiv-develop branch only. Maybe let's move this part of conversation into #1455 thread.

Regarding mxcubecore: in the previous version (with `idtest0` only) the [test case](https://github.com/mxcube/mxcubeweb/blob/2d3c486ec8862b8ba761763eb0d0d8af73a71fbf/test/test_authn.py#L142) was passing just because user with `login_id = idtest1` did not exist. The credentials were incorrect thus this part was blocking login https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/HardwareObjects/mockup/ISPyBClientMockup.py#L98 Seemed it's not the case that should be tested there.

marcus-oscarsson:
I think that the fix you did for the test case as it is [here](https://github.com/mxcube/mxcubeweb/pull/1455/files#diff-948603f00beb75aff4ef52d02fed569c1702af735c26ee1fe13496af0679e7fcR189) with `CREDENTIALS_0` and `CREDENTIALS_1` is fine and tests what you are intending. 

The problem is that the fix in #1066 was only partial and allowed you to login with `idtest0` and `idtest1` but the same proposal is returned for both logins allowing login thus the test fails. The proposal is returned by the `get_proposal` method and returns the internal `__test_proposal` and this structure does not change with login

dominikatrojanowska:
Should I introduce `__test_proposal_0` and `__test_proposal_1` or just keep [the one that already exists](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/HardwareObjects/mockup/ISPyBClientMockup.py#L53) but change some relevant fields, for instance `personId`?

fabcor-maxiv:
I guess we need two different mockup proposals indeed.

Maybe there is a smart way to do it while avoiding too much copy-paste stuff, I do not know...

dominikatrojanowska:
yes, it could be a:
* method taking user id as an argument and returning relevant dictionary eg.
```
def __test_proposal(self, user: Optional[str] = "") -> dict:
        proposal = None if not user else {
            "status": {"code": "ok"},
            "Person": {
                "personId": user[-1],
                "laboratoryId": 1,
                "login": None,
                "familyName": "operator on IDTESTeh1",
            },
            "Proposal": {
                "code": "idtest",
                "title": "operator on IDTESTeh1",
                "personId": user[-1],
                "number": "0",
                "proposalId": 1,
                "type": "MX",
            },
            "Session": [
                {
                    "scheduled": 0,
                    "startDate": "2013-06-11 00:00:00",
                    "endDate": "2023-06-12 07:59:59",
                    "beamlineName": self.beamline_name,
                    "timeStamp": "2013-06-11 09:40:36",
                    "comments": "Session created by the BCM",
                    "sessionId": 34591,
                    "proposalId": 1,
                    "nbShifts": 3,
                }
            ],
            "Laboratory": {"laboratoryId": 1, "name": "TEST eh1"},
        }
        return proposal
```
but the reading way would change to `self.__test_proposal(user_id)`
* or property, overwritten when the user logs in. Reading way wouldn't change (`self.__test_proposal`)
```
    @property
    def __test_proposal(self, user: Optional[str] = "") -> dict:
        return self._test_proposal

    @__test_proposal.setter
    def __test_proposal(self, user: str):
        self._test_proposal = {
            "status": {"code": "ok"},
            "Person": {
                "personId": user[-1],
                "laboratoryId": 1,
                "login": None,
                "familyName": "operator on IDTESTeh1",
            },
        ...
```
I dont know if anything else should be changed... in my opinion I would keep it as simple and less effort as possible since those are tests only. Tell me if I'm wrong or you have different opinion. Also some explanation of what should differ between those two proposals (corresponding to `idtest0` an `idtest1`) would be appreciated :)

fabcor-maxiv:
@dominikatrojanowska I think this is the wrong place to discuss this further. Maybe we should continue over there: <https://github.com/mxcube/mxcubecore/issues/1071>.

PR #1489, opened 2024-11-04
Store sampleview coords in context menu state

axelboc:
I now store `e.offset<X|Y> / imageRatio` instead of just `e.offset<X|Y>` in the state, which saves `ContextMenu` from having to do this calculation in a couple of places.

I've made a couple of additional commits to move some logic around, inline some methods, etc., and to let the right-clicked shape be `undefined` in the state (instead of storing a fake shape object `{ type: 'NONE' }`).

### DIFF
--- a/ui/src/components/SampleView/ContextMenu.jsx
+++ b/ui/src/components/SampleView/ContextMenu.jsx
@@ -299,7 +299,21 @@ export default class ContextMenu extends React.Component {
   showModal(modalName, extraParams = {}, _shape = null) {
     const { sampleID, shape, sampleData, defaultParameters } = this.props;
 
-    const sid = _shape ? _shape.id : shape.id;
+    if (this.props.clickCentring) {
+      this.props.sampleViewActions.stopClickCentring();
+      this.props.sampleViewActions.acceptCentring();
+    }
+
+    if (!sampleData) {
+      this.props.showErrorPanel(
+        true,
+        'There is no sample mounted, cannot collect data.',
+      );
+
+      return;
+    }
+
+    const sid = _shape ? _shape.id : shape?.id;
     if (Array.isArray(sid)) {
       // we remove any line
       // in case we have selected (by drawing a box) two points
@@ -316,11 +330,6 @@ export default class ContextMenu extends React.Component {
       }
     }
 
-    if (this.props.clickCentring) {
-      this.props.sampleViewActions.stopClickCentring();
-      this.props.sampleViewActions.acceptCentring();
-    }
-
     const type =
       modalName === 'Generic' ? extraParams.type : modalName.toLowerCase();
     const name =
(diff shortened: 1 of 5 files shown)


ui/src/components/SampleView/ContextMenu.jsx
```diff
@@ -486,7 +479,7 @@ export default class ContextMenu extends React.Component {
     const menuOptions = this.menuOptions();
     let optionList = [];
 
-    if (this.props.sampleID !== undefined) {
```
axelboc, line 489: I've kept this condition for now, but I don't find it ideal. If there's no sample mounted and you right-click on a shape, you get a completely different context menu than if you right-click on the same shape and there's a sample mounted. Actions that don't require a sample to be mounted, like _Delete Line_, could very well be shown when right-clicking on a line regardless of the presence of a sample.

I would prefer for context menu actions that require a sample to be mounted to be conditionally rendered individually.

ui/src/components/SampleView/ContextMenu.jsx
```diff
@@ -34,24 +33,6 @@ export default class ContextMenu extends React.Component {
       grid: [],
       none: [],
     };
-    let twoDPoints = [];
-
-    if (this.props.enable2DPoints) {
```
axelboc, line 39: I've inlined the condition below.

ui/src/components/SampleView/ContextMenu.jsx
```diff
@@ -253,32 +252,53 @@ export default class ContextMenu extends React.Component {
...
+              },
+            ]
+          : []),
         { text: 'divider', key: 7 },
         ...genericTasks.none,
-        genericTasks.grid.none > 0 ? { text: 'divider', key: 7 } : {},
       ],
     };
 
-    if (this.props.enableNativeMesh) {
```
axelboc, line 274: Also inlined above.

ui/src/components/SampleView/ContextMenu.jsx
marcus-oscarsson: > I would prefer for context menu actions that require a sample to be mounted to be conditionally rendered individually or to show an error to the user when used.

Yes, the context menu on a shape without a sample mounted should give an error, just like we can't create a point/shape without a sample

axelboc: Sorry, I've edited my comment to clarify my thought: I think that actions that don't require a sample to be mounted, like _Delete Point_ should be shown when right-clicking on a shape regardless of the presence of a sample.

That's because when you mount a sample, perform 3-click centring, save the created point, and unmount the sample, the point remains on the sample view, so you can still right-click on it. At the moment, when you're in this situation, you basically can't delete anything via the context menu (you still can by pressing <kbd>Delete</kbd>):

image

marcus-oscarsson: Hm, yes makes sense from that "point" (no pun intended :) ) of view. But all the shapes should actually be cleared when a sample is unmounted, because they are not valid any longer. We should look at that if its not the case

marcus-oscarsson: On the beamlines, at least ours, the shapes are removed at unmount, I guess its just a event that is not registered properly in some mockup.

axelboc:
Replacing the dummy shape `{ type: 'NONE' }` with `undefined` in the context menu state seems to have caused some issues in the `showModal` method. Apologies, I'll investigate.

axelboc:
I've used optional chaining in `showModal` to handle an undefined `shape`. In the other methods, the `shape` prop is always defined.

ui/src/components/SampleView/ContextMenu.jsx
```diff
@@ -424,24 +423,20 @@ export default class ContextMenu extends React.Component {
...
-    const { imageX, imageY, imageRatio } = this.props;
-    this.createPoint(imageX / imageRatio, imageY / imageRatio, (shape) =>
-      this.showModal(name, {}, shape, extraParams),
+    const { sampleViewX, sampleViewY } = this.props;
+
+    this.props.sampleViewActions.add2DPoint(
+      sampleViewX,
+      sampleViewY,
+      'SAVED',
+      (shape) => this.showModal(name, {}, shape, extraParams),
```
walesch-yan, line 439: Maybe you already thought about it and want to do that in a different PR, but related to #1490: Couldn't we just prevent the drawing of a point when no sample is mounted by adding a check for `sampleData` (similar to how it is done in `showModal`)

axelboc: Indeed, the check for the mounted sample could just come before the creation of the point. But we could also hide or disable the context menu option.

walesch-yan: You are correct, that might be a more elegant solution to the problem. Sorry, It seems that I misunderstood the other comment, the first time I read it

PR #1492, opened 2024-11-08
 Initial work an adapting AbstractLims and integration with SSO - Part 1

marcus-oscarsson:
This is the `mxcubeweb` part of the `AbstractLims` PR on `mxcubecore`: https://github.com/mxcube/mxcubecore/pull/1079.

As with PR #1079 on `mxcubecore` this work will need to be continued with some further cleanup. As with the work on `mxcubecore` this is really a larger set of commits many also done by @antolinos and some by myself. There have been some cleanup after testing but apart from that this is code have been tested together with #1079.

This PR introduces:

   - The use of AbstractLims
   - New session selection interface 
   - Single sign on via OpenIDConnect

**Testing**
`ISPyBMockupClient` (`lims.xml` in demo environment) can be used to test both user type and proposal type login by setting the `loginType` attribute.
```
<object class="ISPyBClientMockup">
  <object hwrid="/lims_rest" role="lims_rest"/>
  <ws_root>
    https://your.lims.org
  </ws_root>
  <base_result_url>
    https://your.limsresults.org
  </base_result_url>
  <ws_username></ws_username>
  <ws_password></ws_password>
  <loginType>user</loginType>                <!-- set to user or proposal -->
  <object role="session" href="/session"/>
</object>
```

**UI updates**
proposal_selection_2
_Listing currently available sessions, and the possibility to explicitly reschedule and/or move a session from one beamline to another_

sample_list
_Possibility to select from which LIMS to retrieve sample information from, data is uploaded to both DRAC and ISPyB_


mxcubeweb/core/components/user/usermanager.py
```diff
@@ -212,39 +265,35 @@ def force_signout_user(self, username):
...
 
             res["selectedProposal"] = "%s%s" % (
                 HWR.beamline.session.proposal_code,
                 HWR.beamline.session.proposal_number,
             )
-
             res["selectedProposalID"] = HWR.beamline.session.proposal_id
         else:
-            raise Exception("Not logged in")
+            logging.getLogger("MX3.HWR").warning("Logged out")
```
axelboc: Is this warning needed? Could be an info instead. Seems to be logged twice too.

marcus-oscarsson: You are right, it should be an info :)

pyproject.toml
```diff
@@ -41,8 +41,10 @@ pydantic = ">=2.8.2,<2.9.0"
 PyDispatcher = "^2.0.6"
 pytz = "^2022.6"
 tzlocal = "^4.2"
-mxcubecore = ">=1.191.0"
+mxcubecore =  "1.194.0"
```
fabcor-maxiv, line 44: @marcus-oscarsson This might have been pinned by mistake, no? Can it be removed? `>=1.194.0` maybe?

marcus-oscarsson: Yup, It it should indeed be `>=1.194.0` now

fabcor-maxiv: Solved in <https://github.com/mxcube/mxcubeweb/pull/1520>.

PR #1496, opened 2024-11-21
Convert `SelectProposal` to function component and other fixes

axelboc:
This is based on `mo-lims` and meant to be merged into it.


ui/src/actions/login.js
```diff
@@ -54,27 +54,18 @@ export function hideProposalsForm() {
   };
 }
 
-export function selectProposalAction(prop) {
-  return {
-    type: 'SELECT_PROPOSAL',
-    proposal: prop,
-  };
-}
```
axelboc, line 62: no longer used

ui/src/actions/login.js
```diff
@@ -54,27 +54,18 @@ export function hideProposalsForm() {
...
 export function setInitialState(data) {
   return { type: 'SET_INITIAL_STATE', data };
 }
 
-export function selectProposal(number, navigate) {
+export function selectProposal(number) {
   return async (dispatch) => {
     try {
       await sendSelectProposal(number);
-      navigate('/');
```
axelboc, line 72: I don't think we need to navigate after selecting a proposal. The redirect to `/login` actually had no effect since the user was redirected right away to `/datacollection`.

ui/src/components/LoginForm/SelectProposal.jsx
```diff
@@ -1,182 +1,123 @@
...
-            disabled={
-              this.state.pNumber === null ||
-              (session && session.is_scheduled_beamline === false) ||
-              session.is_scheduled_time === false
-            }
-            onClick={this.handleSelectProposal}
+  return (
+    <Modal
+      show={show}
+      backdrop={showProposalsForm || 'static'}
```
axelboc, line 58: The dialog doesn't have to be a modal when re-opened manually (i.e. when a proposal is already selected). This line allows the dialog to be dismissed by clicking on the backdrop in this situation.

ui/src/components/LoginForm/SelectProposal.jsx
```diff
@@ -1,182 +1,123 @@
...
+      onHide={() => {
+        if (showProposalsForm) {
+          handleHide();
+        }
```
axelboc, line 62: This condition does the opposite: it makes the dialog truly modal when shown because no proposal is selected. It does so by preventing the dialog from being dismissed with `Escape` in this situation.

ui/src/components/LoginForm/SelectProposal.jsx
```diff
@@ -1,182 +1,123 @@
...
+        <ActionButton
+          selectedSession={selectedSession}
+          onClick={() => dispatch(selectProposal(selectedSessionId))}
+        />
+        <Button
+          variant="outline-secondary"
+          data-default-styles
+          onClick={() => handleHide()}
+        >
+          {showProposalsForm ? 'Cancel' : 'Sign out'}
```
axelboc, line 116: Being more explicit about what canceling does when the dialog is shown because no proposal is selected.

ui/src/components/LoginForm/SelectProposal.jsx
```diff
@@ -1,182 +1,123 @@
...
-
-    this.state = {
-      pId: 0,
-      pNumber: null,
-      session: null,
-      proposal: null,
-      filter: null,
-      sessions: props.data.proposalList,
-      filteredSessions: props.data.proposalList,
-    };
```
axelboc, line 23: Note that I've simplified the state a fair bit: I now have only a `filter` state and a `selectedSession` state. The `sessions` don't need to be stored in the state, and the rest can just be computed on every render.

ui/src/components/LoginForm/SelectProposal.jsx
```diff
@@ -1,182 +1,123 @@
...
+          {showProposalsForm ? 'Cancel' : 'Sign out'}
+        </Button>
+      </Modal.Footer>
+    </Modal>
+  );
 }
 
-const SelectProposalForm = reduxForm({
-  form: 'proposals',
-})(SelectProposal);
```
axelboc, line 178: Not sure why this was a Redux form.

ui/src/components/LoginForm/SelectProposal.jsx
```diff
@@ -1,182 +1,123 @@
...
+      show={show}
+      backdrop={showProposalsForm || 'static'}
+      onHide={() => {
+        if (showProposalsForm) {
+          handleHide();
+        }
+      }}
+    >
+      <Modal.Header closeButton>
+        <Modal.Title>Select a session</Modal.Title>
```
axelboc, line 66: There seems to be a bit of confusion in the UI with the terms "proposal" and "session". Are those the same thing?

ui/src/components/LoginForm/SessionTable.jsx
```diff
@@ -1,144 +1,134 @@
...
       <thead>
         <tr>
-          {props.params.showBeamline && <th>Beamline</th>}
-          <th>Title</th>
-          <th>Start</th>
-          <th>End</th>
-          <th>Portal</th>
-          <th>User</th>
-          <th>Logbook</th>
+          <th scope="col">ID</th>
```
axelboc, line 24: This heading cell was missing.

ui/src/components/LoginForm/SessionTable.jsx
```diff
@@ -1,144 +1,134 @@
...
- */
-const getScheduledDateComponent = (session, startDate, startTime) => {
-  if (session.is_rescheduled) {
-    return (
-      <del className={styles.time}>
-        {getDateComponent(startDate, startTime)}
-      </del>
-    );
+  if (sessions.length === 0) {
+    return <span className={styles.fallback}>No sessions</span>;
```
axelboc, line 17: Showing a fallback instead of an empty table.

ui/src/components/LoginForm/SessionTable.jsx
```diff
@@ -1,144 +1,134 @@
...
+            className={styles.row}
+            data-selected={selectedSessionId === session.session_id}
+            tabIndex={0}
+            onClick={() => onSessionSelected(session)}
+            onKeyDown={(evt) => {
+              // For keyboard accessibility; ideally add "Select" button to each row instead of making rows clickable
+              if (evt.key === 'Enter' || evt.key === ' ') {
+                onSessionSelected(session);
+              }
+            }}
```
axelboc, line 47: I'm including a bunch of accessibility fixes like this one and various `aria-label` attributes.

ui/src/serverIO.js
```diff
@@ -105,7 +107,7 @@ class ServerIO {
     this.hwrSocket.on('disconnect', (reason) => {
       console.log('hwrSocket disconnected!'); // eslint-disable-line no-console
 
-      setTimeout(() => {
+      this.connectionLostTimeout = setTimeout(() => {
```
axelboc, line 110: This timeout was triggered sometimes when logging out and back in quickly, which made the connection-lost dialog flash on the screen.

ui/src/components/LoginForm/SelectProposal.jsx
marcus-oscarsson: :), almost but not really. When using proposal to login they are considered the same but when using user based login they are not.

marcus-oscarsson: You are right that it is confusing, the form should really be called `SelectSession` or similar. This can be changed in a future PR.

ui/src/components/LoginForm/SessionTable.jsx
marcus-oscarsson: Very nice !

ui/src/serverIO.js
marcus-oscarsson: Well spotted, nice fix !

marcus-oscarsson:
Thanks, :+1:

ui/src/actions/login.js
```diff
@@ -108,16 +99,14 @@ export function ssoLogIn() {
...
-    // We make sure that user data is reseted so that websockets
-    // are keept dicconnected while logging out.
-    dispatch(resetLoginInfo());
-    await sendSignOut().finally(() =>
-      dispatch(
-        // Retreiving the user data from the backend
-        getLoginInfo(),
-      ),
-    );
+
```
marcus-oscarsson, line 104: I see, ok :)

marcus-oscarsson (approved):

PR #1497, opened 2024-11-25
Add favicon

axelboc:
Browsers still request `/favicon.ico` as a fallback regardless of the `meta` tags. Adding a `favicon.ico` file reduces the noise both in the server logs and in the browser dev tools.


fabcor-maxiv:
Finally some meaningful pull request! It does break Cypress tests, though(?).
: D

axelboc:
:sob:

PR #1499, opened 2024-11-25
Bump versions of GitHub Actions

axelboc:
There were some warnings in the GitHub UI. I went through the release notes of each action and couldn't spot any red flags: https://github.com/actions/


axelboc:
Prettier removed a few whitespaces — let me know if you prefer I add them back.

.github/workflows/pages.yaml
```diff
@@ -43,7 +37,7 @@ jobs:
       # Pin micromamba version because of following issue:
       # https://github.com/mamba-org/setup-micromamba/issues/225
       - name: "Setup conda environment with micromamba"
-        uses: "mamba-org/setup-micromamba@v1"  # tested with v1.4.4
+        uses: "mamba-org/setup-micromamba@v2"  # tested with v2.0.1
```
fabcor-maxiv: I was the one to add those comments:
```suggestion
        uses: "mamba-org/setup-micromamba@v2"
```

I do not know if they are actually helpful or not.
Maybe we should pin everything to specific versions
and wait for Dependabot to ping us.

fabcor-maxiv (approved):
> Prettier removed a few whitespaces — let me know if you prefer I add them back.

It is fine. As long as all the checks are green.

But, I am confused. For some of those I do not understand why `yamllint` did not complain earlier...

I recall that we had what looked like a conflict between yamllint and prettier not too long ago. I am left wondering: is there something that we are missing?

.github/workflows/pages.yaml
axelboc: Meh, not sure, pinning the versions and updating them via Dependabot would add a lot of noise for not much benefit IMO.

GitHub Actions follow semver very well — granted, errors always happen but the probably of a job failing because of a minor/patch update of a GitHub Action is very low.

I'd vote for removing the comment and keeping the major version tags.

axelboc:
So the VS Code settings had a typo. Also the fix-on-save was disabled for YAML files but not for GH workflow files, which have their own "language mode"...

.github/workflows/pages.yaml
fabcor-maxiv: I am OK, with putting our trust in semver and removing the comments

fabcor-maxiv (approved):

PR #1502, opened 2024-12-02
Fixes grid-scan heatmap issue

fhernandezvivanco:
The current version of the code is displaying heatmaps incorrectly (see the figure below). This is probably due to a recent refactor that is passing the wrong argument to the `heatMapColorForValue` function.

To fix this, I only changed the following line:

`this.heatMapColorForValue(
              gd,
              result[index],
              result[index],)
`

to 

`this.heatMapColorForValue(
              gd,
              result[index],
              result[index][1],)
`

The value `result[index][1]` contains the correct RGBA input values for the `heatMapColorForValue` function (which is the value that was passed as of version `4.34.0`)

heatmap_res


marcus-oscarsson:
Thanks, we are using the `PNG` option so we have not noticed this issue, but very welcome for those using the `RGB` option.

PR #1504, opened 2024-12-03
Adding documentation and dependency for missing video stream

### LINKED ISSUE #1495: Sample View image not displayed for mockup config
Hi,

Probably a simple issue/doubt: when running the mockup config (mxcubeweb/demo), the example video image `mxcube_sample_snapshot.jpeg` set in `md_camera.xml` is never displayed on the UI (the sample view widget is always black).

On the core side (`MDCameraMockup.py`), the image path is found and the image is emitted (with the `imageReceived` signal), but it's not handled by `sampleview.py on the Web side (e.g., method `_new_frame_received()` is never called apparently). No error message appears in the log regarding this.

Is this the expected behavior? Does tha camera have to be enabled manually? Would be nice to have the image showing, so we can have a base for new camera implementations (which is what I am trying to do :grin: ). Thanks!

walesch-yan:
Following the discussion on #1495. I added `ffmpeg` as a dependency in `conda-environment.yml` and added a note in the documentation for quickly solving any missing dependencies from the [video-streamer](https://github.com/mxcube/video-streamer). 

Closes #1495

### DIFF
--- a/conda-environment.yml
+++ b/conda-environment.yml
@@ -9,6 +9,7 @@ dependencies:
   # Runtime dependencies
   # ====================
 
+  - ffmpeg  # Not directly used by mxcube, but necessary for the video-streamer
   - python >=3.8,<3.12  # Make sure it matches `pyproject.toml`
 
   # We install `python-ldap` from conda because it is "hard" to install.
--- a/docs/source/dev/environment.md
+++ b/docs/source/dev/environment.md
@@ -167,6 +167,14 @@ pnpm --prefix ui start
 
 The above will automatically open a browser with the URL: <http://localhost:5173>
 
+> **Note**: In case that you decided to install MXCuBE-Web without using the `conda-environment.yml` file, the stream in the `Data Collection` tab of the UI might be missing. To resolve this:
+>
+> 1. Ensure that `ffmpeg` is properly installed on your side, to install run `sudo apt install ffmpeg`
+>    Then restart the server.
+> 1. Alternatively, you can manually install the dependencies of [MXCuBE's video-streamer](https://github.com/mxcube/video-streamer/) and restart the server afterwards.
+>
+> If the issue persists, please make sure to use the [issues](https://github.com/mxcube/mxcubeweb/issues) page and reach out for help.
+
 #### 9.3. Running the end to end (e2e) tests
 
 Keep both the backend and front-end servers running then run the following command in a third terminal, from the root directory of the project:


conda-environment.yml
```diff
@@ -24,3 +24,4 @@ dependencies:
   - nodejs ==18.*|20.*  # Make sure it matches `ui/package.json`
   - pnpm ==9.*  # Make sure it matches `ui/package.json`
   - poetry
+  - ffmpeg  # Not directly used by mxcube, but necessary for the video-streamer
```
fabcor-maxiv: 1. I recommend using alphabetical order within each section.
2. ffmpeg is a runtime dependency not a development dependency, so it should be listed in the "Runtime dependencies" section

docs/source/dev/environment.md
```diff
@@ -167,6 +167,14 @@ pnpm --prefix ui start
 
 The above will automatically open a browser with the URL: <http://localhost:5173>
 
+> **Note**: In case that you decided to install MXCuBE-Web without using the `conda-environment.yml` file, the stream in the `Data Collection` tab of the UI might be missing. To resolve this:
+>
+> 1. Ensure that `ffmpeg` is properly installed on your side, to install run `sudo apt install ffmpeg`
+>    Then restart the server.
+> 1. Alternatively, you can manually install the dependencies of [MXCuBE's video-streamer](https://github.com/mxcube/video-streamer/) and restart the server afterwards.
+>
+> If the issue persists, please make sure to use the [issues](https://github.com/mxcube/mxcubeweb/issues) page and reach out for help.
```
fabcor-maxiv, line 176: The documentation is completely written around this `conda-environment.yml` file.
I really do not wish for us to have to maintain documentation for more than one scenario.
If it were just me I would not put this note in.

Better would be to write a section about what the dependencies of MXCuBE-Web are,
besides the dependencies that are already obvious because they are written in `pyproject.toml` and `poetry.lock`.
And this section could be just as short as:

> Please, look into the `conda-environment.yml` for the list of additional dependencies.

fabcor-maxiv (changes requested):

conda-environment.yml
walesch-yan: You are right, I will change this. Thanks for pointing it out!

docs/source/dev/environment.md
walesch-yan: I think it was more meant as a troubleshooting, since a few people seemed to have the problem and the application doesn't really tell you about the missing dependency. 

Sure, a dependency list can be a very good addition to the documentation. 
Actually, I was wondering, if we should add some sort of FAQ to the documentation (maybe this would go over the scope of this PR)

marcus-oscarsson: The `video-streamer` is an integral part of `mxcubeweb`, not strictly necessary but used by default in the demo configuration. Its also the solution we recommend to use in general. One of the more common issues we've had with the video streamer is that `ffmpeg` has not been installed. They way I see it is that we want to provide a working solution that is easy to install and get to work. In that sense I think this section is very useful. Its possible that it means that we need to maintain some more lines of documentation but in this case I would say its worth it.

fabcor-maxiv: I guess it was an issue because `ffmpeg` was not listed anywhere as a dependency. If `ffmpeg` is now listed as a dependency in `conda-environment.yaml`, then it will be installed and people will not have the issue of `ffmpeg` missing. Am I misunderstanding something? :D

I see that `ffmpeg` is not listed here either:
https://github.com/mxcube/video-streamer/blob/539e65b34a8d6493945cc886957d05242cf1858b/conda-environment.yml
Is that normal?

Anyway, we can keep the note, I am not convinced it is necessary, but obviously it is not a big deal.

fabcor-maxiv (approved):

PR #1507, opened 2024-12-11
(#1500) Makes default value for REMEMBER_PARAMETERS_BETWEEN_SAMPLES and queue_manager.centringMethod configurable

### LINKED ISSUE #1500: Make 'Remember parameters between samples' default configurable
MXCuBE have a setting `Remember parameters between samples`:

image

The default value for this setting is hard-coded here:

https://github.com/mxcube/mxcubeweb/blob/542423c884ac74057f26e46b55cd055fd2ac84f1/mxcubeweb/app.py#L241

Currently, it is set to `False`, which is of cause opposite of what scientist here at MAXIV demand.

How about making this hard-coded value a configurable instead? I suggest adding a new section `settings` to `mxcube-web/ui.yaml` file. With following format:

```
settings:
 remember_parameters_between_samples:
   default: true
 some_other_setting:
   default: false
 yet_another_setting:
   default: true
```
  
What do you think?

mockoocy:
This PR makes settings mentioned in the title configurable. Also makes them behave in an expected manner, that is:
- they persist between refreshes (`centringMethod` was reset on the frontend after refresh)
- are reset properly when user logs out

The default values are set with `default_centring_method` and `default_remember_parameters_between_samples`. 

It also fixes the problem with `remember parameters between samples` checkbox. When It was changed in the settings, although boolean value was sent by client, it registered a bool-casted-to-string value, that is `"False"` or `"True"`.

Closes #1500

### DIFF
--- a/mxcubeweb/core/components/queue.py
+++ b/mxcubeweb/core/components/queue.py
@@ -12,6 +12,7 @@ from mxcubecore import queue_entry as qe
 from mxcubecore.HardwareObjects.Gphl import GphlQueueEntry
 from mxcubecore.model import queue_model_enumerables as qme
 from mxcubecore.model import queue_model_objects as qmo
+from mxcubecore.model.queue_model_enumerables import CENTRING_METHOD
 from mxcubecore.queue_entry.base_queue_entry import QUEUE_ENTRY_STATUS
 
 from mxcubeweb.core.components.component_base import ComponentBase
@@ -280,6 +281,7 @@ class Queue(ComponentBase):
             "numSnapshots": HWR.beamline.collect.get_property(
                 "num_snapshots", self.app.DEFAULT_NUM_SNAPSHOTS
             ),
+            "centringMethod": HWR.beamline.queue_manager.centring_method,
         }
 
         res.update(settings)
@@ -2076,6 +2078,21 @@ class Queue(ComponentBase):
         )
         self.app.AUTO_ADD_DIFFPLAN = HWR.beamline.collect.get_property(
             "auto_add_diff_plan", False
+        )
+        # Change value of the parameter, without changing the hardware object property.
+        # This allows to properly reset the value on logout when invoking init_queue_settings.
+        self.app.REMEMBER_PARAMETERS_BETWEEN_SAMPLES = (
+            HWR.beamline.queue_manager.get_property(
+                "remember_parameters_between_samples", False
+            )
+        )
+
+        centring_method_as_string = HWR.beamline.queue_manager.get_property(
+            "default_centring_method", "NONE"
+        )
+        HWR.beamline.queue_manager.centring_method = getattr(
+            CENTRING_METHOD,
+            centring_method_as_string,
         )
 
     def queue_start(self, sid):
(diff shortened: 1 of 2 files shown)


mxcubeweb/core/components/queue.py
```diff
@@ -2078,6 +2080,20 @@ def init_queue_settings(self):
             "auto_add_diff_plan", False
         )
 
+        self.app.REMEMBER_PARAMETERS_BETWEEN_SAMPLES = (
+            HWR.beamline.queue_manager.get_property(
+                "default_remember_parameters_between_samples", False
```
marcus-oscarsson: Looks good in general, do we need to call it `default_remember_parameters_between_samples` with the `default` prefix, would not `remember_parameters_between_samples` be sufficient ? I guess one understands that the user can change the option ?

mockoocy: Both can work. I think that providing the `default` prefix is a good idea for fields that may be modified by the user, to avoid some awkward bugs with e.g., some fields not resetting - but that's only me preference :)

fabcor-maxiv: There are two levels of "default" here, aren't there?

1. `False` is the default fall back value if `default_remember_parameters_between_samples` is not set, right?
2. and then `default_remember_parameters_between_samples` is also a "default" because it can be overridden via the checkbox in the UI, correct? It is the default value when starting MXCuBE.

marcus-oscarsson: :), yes. That can be a bit confusing. I would not make things too complicated. I don't think we need to the have the fallback/default argument configurable (`False` in this case) . I also don't think we need to keep the `default` prefix, some parameters are changeable by the users other are not, do we need to distinguish between them ? In that case a comment is perhaps more suitable.

mockoocy: Okay, maybe too many defaults there :D will remove the prefix then.

Edit: removed the prefix and included small comment on why it's done that way.

mxcubeweb/core/components/queue.py
```diff
@@ -2077,6 +2079,21 @@ def init_queue_settings(self):
...
         )
+        # Change value of the parameter, without changing the hardware object property.
+        # This allows to properly reset the value on logout when invoking init_queue_settings.
+        self.app.REMEMBER_PARAMETERS_BETWEEN_SAMPLES = (
+            HWR.beamline.queue_manager.get_property(
+                "remember_parameters_between_samples", False
+            )
+        )
+
+        default_centring_method_as_string = HWR.beamline.queue_manager.get_property(
```
marcus-oscarsson: I guess the same thing here then ?

mockoocy: and now... in this case it would not work as the `centring_method` property is being set directly as 
`HWR.beamline.queue_manager.centring_method = ...`, so if we'd want to make it behave as the other settings we need to explicitly store it.

marcus-oscarsson: I meant for the naming with the `default` prefix :)

mockoocy: Oh, changed then :)

marcus-oscarsson: :)

PR #1508, opened 2024-12-11
[Re-opened] - Convert SelectProposal to function component and other fixes #1496 

marcus-oscarsson:
I managed to remove some changes proposed by @axelboc  in a previous PR, probably due to a force push. I've "resurrected" that PR here.


axelboc (approved):
No worries!

PR #1509, opened 2024-12-16
go to login page if session is invalid

elmjag:
When trying to re-connect the socketIO connection, check if our session is still valid. If session is no longer valid, that is we are no longer logged in, go to the login page.

This takes care of the cases where session have become invalid while front-end did not have a connection to the back-end.

Currently, if session becomes invalid, the front-end will spam back-end with reconnect attempts until user reloads or closes the MXCuBE tab in the browser.

### DIFF
--- a/ui/src/serverIO.js
+++ b/ui/src/serverIO.js
@@ -107,6 +107,21 @@ class ServerIO {
     this.hwrSocket.on('disconnect', (reason) => {
       console.log('hwrSocket disconnected!'); // eslint-disable-line no-console
 
+      if (reason === 'io server disconnect') {
+        //
+        // If socket disconnects with this reason, it is possible that our
+        // session have become invalid.
+        //
+        // That is, we can establish connection to the server, but then we
+        // get a 'unauthorized' response which closes the socket.
+        //
+        // Check if that is the case by fetching the login info. If our
+        // session is invalid, getLoginInfo() will update state to
+        // 'not logged in' and the login page will be shown.
+        //
+        dispatch(getLoginInfo());
+      }
+
       this.connectionLostTimeout = setTimeout(() => {
         dispatch(
           // Show message if socket still hasn't reconnected (and wasn't manually disconnected in the first place)


axelboc (commented):
Hey @elmjag, I'm a bit confused as to how refetching the login info when the `loggingSocket` disconnects addresses the issue you describe. It definitely doesn't match with your description of the fix: "When trying to re-connect the socketIO connection, check if our session is still valid."

Could you please share steps to reproduce? I'd like to take a look if you don't mind. Especially since it seems I forgot about the `loggingSocket` in https://github.com/mxcube/mxcubeweb/pull/1487/files#r1827407890, when I removed the unnecessary `setTimeout` that tried to manually reconnect the `hwrSocket`.

elmjag:
> Could you please share steps to reproduce? I'd like to take a look if you don't mind. 

The fastest way to reproduce this is:

 * login to MXCuBE
 * stop `mxcubeweb-server`
 * delete `/tmp/mxcube-user.db` file
 * start `mxcubeweb-server` again

Another way to reproduce this issue is to instead of deleting `/tmp/mxcube-user.db` file, wait for the session timeout. That is, stop `mxcubeweb-server`, wait until the session times out, start `mxcubeweb-server` again.

The reason checking login info on `loggingSocket` disconnects works, is because currently the `loggingSocket.on('disconnect', ..` code currently works as a poller, which checks if possible to connect to the back-end. Once session becomes invalid, the `loggingSocket` will try to reconnect. As far as I understand, it will first be able to create a connection, but then it will be disconnected due to invalid credentials. Thus checking login info in that callback will be able to detect invalid session.

Initially I hoped that socket.io sockets would raise some kind of authentication error, but I was not able find a way to trap that error. Another detail, is it seems that socket.io does not try to reconnect once it gets the 'authentication error'. That the reason I put the login info check into `loggingSocket`disconnect handler. The `loggingSocket` will try reconnect, regardless of what kind of error it gets.

Perhaps this should be implemented differently. The problem I want to solve is that front-end actually detects that session is not valid anymore, instead of being permanently stuck in "lost connection" mode. Feel free to suggest a better solution.

axelboc:
I'm able to reproduce the issue and definitely agree that the user should be logged out automatically. Thanks for the detailed steps and explanations.

That being said, it bothers me to refetch the login info when one of the sockets receives a `disconnect` event:

- First, the `disconnect` event can be fired for various reasons, including when the server goes down (and fetching anything in this case doesn't make sense). Sure, we could check specifically for the ["io server disconnect"](https://socket.io/docs/v4/client-api/#event-disconnect) reason, as this seems to be the one that occurs with the reproduction steps above. However, nothing in this reason's description ("The server has forcefully disconnected the socket with `socket.disconnect()`") says why the server called `socket.disconnect()` in the first place. How do we know that we should stop trying to reconnect? Maybe the server will log the user back in at some point, we just don't know.
- Second, it's not clear which socket should listen to the `disconnect` event and refetch the login info. Should it be the logging socket or the HWR socket? Conceptually, having to make this choice is problematic.

In front-end apps, the typical way to detect when a user is no longer authenticated and should be logged out from the UI is with **401 error responses**. The UI just remains in a logged in state until the user tries to do something and a 401 is triggered.

As it turns out, we have a mechanism in place to refresh the session every 9 seconds, so the UI could remain in this stale logged in state for at most 9 seconds:

image

I'd like to investigate this approach if you don't mind.

elmjag:
Using response to `refresh_session` to detect invalid session should in principle work.

The reason I did not want to use it was because working on removing that request. To me it feels like making custom HTTP request to keep a session alive is not needed. There got to be some build-in feature in flask-login or flask-security-too package that already deals with it. But, so far I have not been able to understand how it's suppose to work.

Let's for now use `refresh_session` response for sending users to login page.

We can revisit this subject, if I manage to figure out session keep-alive features in flask packages.

In other words, feel free to implement this PR using `refresh_session` requests.

axelboc:
Yeah, not a fan of this refresh mechanism either.

Perhaps the problem is that the sockets are opened only while the user is logged in. If we opened them on page load and kept them open regardless of the user's logged in state, we would have no trouble resynching the state on reconnect (e.g. when turning the server off and on). This would also make managing the state of the "connection lost" dialog easier, and we wouldn't have to worry about disconnecting the sockets before logging out (to avoid hitting 401s).

Generally speaking, the fact that we try to keep the UI state fresh with a combination of both HTTP and WebSocket doesn't help either. We should perhaps accept that MXCuBE-Web is a real-time application and move to an architecture that's more suitable for this. Anyway, that's a much bigger discussion... :joy: 

---

Given your reply regarding the `refresh_session` mechanism, and since only fools don't change their mind, I propose that we stick to your solution for now. Could you please just:

- add a check to refetch the login info only when the disconnection reason is `"io server disconnect"`
- add a comment to explain why we refetch the login info when this happens;
- move the logic to the HWR socket, which is kind of the "main" socket in my view.

---

This doesn't remove the need to log out when we detect a 401, though. But I'm hoping we can do this with a more general wretch middleware maybe than specifically on the `refetch_session` requests. I'll try to work on that.

axelboc:
Sorry, meant to rebase my other PR :sweat:

marcus-oscarsson:
Nice discussion !, I'm all for removing the `refresh_session` request. But how do we handle the situation when a user closes the browser window, we would like to make sure that a user does not remain logged-in.

axelboc:
> But how do we handle the situation when a user closes the browser window, we would like to make sure that a user does not remain logged-in.

I assume this is where the idea of a [`Keep-Alive`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Keep-Alive) HTTP connection comes in.

Similarly, with WebSockets, the server would know right away when the socket connection closes and could then log the user out a little while later if they don't reconnect.

marcus-oscarsson:
I never tried to use `Keep-Alive` that could perhaps work. What you mean is that one would invalidate the session on timeout somehow then ? I did try to use the various websocket events but it was not working reliably at the time, but I might have overlooked or something or made some mistake somewhere.

marcus-oscarsson:
@elmjag I merged  #1512 and tried to resolve the resulting conflict, feel free to do as you like with the last commit

elmjag:
> Could you please just:
> 
> * add a check to refetch the login info only when the disconnection reason is `"io server disconnect"`
> * add a comment to explain why we refetch the login info when this happens;
> * move the logic to the HWR socket, which is kind of the "main" socket in my view.

Oh yes, I did the changes above, feels much cleaner! So the magic reason was `"io server disconnect"`, thanks for sharing 
:-)

> This doesn't remove the need to log out when we detect a 401, though. But I'm hoping we can do this with a more general wretch middleware maybe than specifically on the `refetch_session` requests. I'll try to work on that.

I agree, it would be nice if we handled all 401 replies properly.

elmjag:
> I never tried to use `Keep-Alive` that could perhaps work. What you mean is that one would invalidate the session on timeout somehow then ? I did try to use the various websocket events but it was not working reliably at the time, but I might have overlooked or something or made some mistake somewhere.

I have not managed to figure out how to implement this yet. But in principle the back-end knows exactly when user closes the browser, as it disconnects the sockets. Perhaps this can used to start the session expiration timer.

I hope there is some kind of standard solution available in the flask eco-system that we can use. MXCuBE can't be the first flask application that need this feature. So far, I have not managed to put the puzzle pieces together. I'll keep investigating this, next year :)

axelboc (approved):
:100:

marcus-oscarsson:
Very nice thanks @elmjag !

PR #1511, opened 2024-12-16
Show error boundary when development server is down

axelboc:
When the development server is down, the app gets stuck on the loading screen:

Screenshot from 2024-12-16 14-00-09

The `/login_info` request triggered by the `getLoginInfo` action (dispatched from inside the `useEffect`, in `App.jsx`) doesn't succeed since the server is down. Unfortunately, since the request is triggered from `useEffect`, the `DefaultErrorBoundary` doesn't see the error and therefore doesn't show the "Something went wrong" message below, as one would expect:

Screenshot from 2024-12-16 13-30-38

In this PR, I refactor `DefaultErrorBoundary` using an external library called [react-error-boundary](https://github.com/bvaughn/react-error-boundary), which provides more features than our basic implementation. One of these features is the `useErrorBoundary` hook, which gives us the ability to imperatively show the closest error boundary.

While I'm at it, I improve the design of the "Something went wrong" message a bit:

Screenshot from 2024-12-16 13-59-36

Granted, the error message above when the development server is down is not very clear, but for some reason Vite's proxy server triggers a 500 error, so I don't want to risk hiding a real server-side 500 error...

### DIFF
--- a/ui/src/components/App.jsx
+++ b/ui/src/components/App.jsx
@@ -19,6 +19,7 @@ import { serverIO } from '../serverIO';
 import { getLoginInfo } from '../actions/login';
 import PrivateOutlet from './PrivateOutlet';
 import { sendRefreshSession } from '../api/login';
+import { useErrorBoundary } from 'react-error-boundary';
 
 const REFRESH_INTERVAL = 9000;
 
@@ -68,10 +69,11 @@ const router = createBrowserRouter([
 
 function App() {
   const dispatch = useDispatch();
+  const { showBoundary } = useErrorBoundary();
   const loggedIn = useSelector((state) => state.login.loggedIn);
 
   useEffect(() => {
-    dispatch(getLoginInfo());
+    dispatch(getLoginInfo()).catch(showBoundary); // eslint-disable-line promise/prefer-await-to-then
 
     if (loggedIn) {
       serverIO.listen();
@@ -85,7 +87,7 @@ function App() {
 
     // no clean-up required, until we connect to serverIO
     return undefined;
-  }, [loggedIn, dispatch]);
+  }, [loggedIn, dispatch, showBoundary]);
 
   if (loggedIn === null) {
     // Fetching login info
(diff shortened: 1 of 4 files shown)


ui/src/containers/DefaultErrorBoundary.jsx
```diff
@@ -1,63 +1,37 @@
...
-            the issue.
+            Reloading the page might fix the issue. If it remains, please
+            contact support.
           </p>
-          <details style={{ whiteSpace: 'pre-wrap' }}>
-            {this.state.error && this.state.error.toString()}
-            <br />
-            {this.state.errorInfo.componentStack}
-            <br />
-            {JSON.stringify(this.state.store)}
```
axelboc, line 41: I'm no longer displaying the component stack and store snapshot. I still send both to the server via the `onError` callback and the `logFrontEndTraceBack` action, though, so they still get logged server-side.

ui/src/containers/DefaultErrorBoundary.jsx
```diff
@@ -1,63 +1,37 @@
...
+    <ErrorBoundary
+      fallbackRender={({ error }) => (
+        <div className="p-4">
+          <h3 className="mb-3">Something went wrong</h3>
+          <p className="mb-2">
+            We are terribly sorry but an error occurred:{' '}
+            <code>{error?.toString()}</code>
           </p>
           <p>
-            Simply, <a href="/datacollection">reloading</a> the page might fix
```
axelboc, line 33: The error may have been triggered from another page, so navigating to `/datacollection` may not be appropriate. I change this with a "Reload" button that calls `window.location.reload()`.

ui/src/components/App.jsx
```diff
@@ -68,10 +69,11 @@ const router = createBrowserRouter([
 
 function App() {
   const dispatch = useDispatch();
+  const { showBoundary } = useErrorBoundary();
   const loggedIn = useSelector((state) => state.login.loggedIn);
 
   useEffect(() => {
-    dispatch(getLoginInfo());
+    dispatch(getLoginInfo()).catch(showBoundary); // eslint-disable-line promise/prefer-await-to-then
```
axelboc, line 76: Obviously there are a lot more cases of uncaught async errors in the application, but at least now we have a way to deal with them if need be.

> The long term goal is to not use Redux for data fetching and take advantage of React Suspense instead (so errors are thrown during render and caught by error boundaries).

marcus-oscarsson:
Nice, thanks !

PR #1512, opened 2024-12-16
Don't manually reconnect logging websocket

axelboc:
I forgot about the logging socket when I removed the manual reconnection of the HWR socket in https://github.com/mxcube/mxcubeweb/pull/1487/files#diff-e612f081728e8f03615f1629c341ca0bc439c53d5e8d0c2139594c60a2394430L109

The reasoning is the same: socket IO already provides an automatic reconnection mechanism (and it's much better, obviously). Also, our custom reconnection mechanism would spam the server in some cases, like when restarting the server. @elmjag can you please check if this is sufficient to address the issue from #1509? If not, please do share a repro so I can investigate further.

### DIFF
--- a/ui/src/serverIO.js
+++ b/ui/src/serverIO.js
@@ -438,13 +438,6 @@ class ServerIO {
 
     this.loggingSocket.on('disconnect', (reason) => {
       console.log('loggingSocket disconnected!'); // eslint-disable-line no-console
-
-      if (reason === 'io server disconnect') {
-        const socket = this.loggingSocket;
-        setTimeout(() => {
-          socket.connect();
-        }, 500);
-      }
     });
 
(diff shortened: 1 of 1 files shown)


elmjag:
This does solve the spamming issue. MXCuBE stops trying to re-connect sockets after first 'unauthorized' response.

However, it will not go to the login page if the session is invalid. I think we should make sure it does.

Right now, if you somehow lose valid session, you end-up with MXCuBE which looks more or less normal, but is actually not usable anymore. A lot of request will be denied with '401' replies. Socket are disconnected so hardware status is no longer updated in the UI. In this situation the login page should be shown, to avoid confusion.

You can reproduce this situation as I have described it here: https://github.com/mxcube/mxcubeweb/pull/1509#issuecomment-2545897070

PR #1513, opened 2024-12-17
Hanling user "logged in" state in a more robust way in login_info

### LINKED ISSUE #1477: Timezone error on Ubutun 24 silenced by `login_info` endpoint
I've just upgraded to Ubuntu 24 and had to reinstall `mxcubeweb` from scratch. Unfortunately, I couldn't get past the login screen even though I wasn't getting any error. The `login` endpoint would just return 200 with `{ "msg": "" }` and then the subsequent `login_info` request would also return 200 but with `{ "loggedIn": False }`...

After trying to reinstall from scratch like ten times, with different versions of python, with `pip` instead of `poetry`, etc., we found that the problem came from an error in `usermanager.login_ingo()` that was getting swallowed by the `try/catch` statement in the [`login_info` endpoint](https://github.com/mxcube/mxcubeweb/blob/develop/mxcubeweb/routes/login.py#L92).

This issue is about refactoring this statement (and/or `usermanager.login_info()`) so unknown errors are not silenced in the future.

---

For information, the error that I faced was thrown when calling [`current_user.to_dict()`](https://github.com/mxcube/mxcubeweb/blob/develop/mxcubeweb/routes/login.py#L92) and had to do with timezones:

```
tzlocal.utils.ZoneInfoNotFoundError: 'Multiple conflicting time zone configurations found:\n/etc/timezone: Etc/UTC\n/etc/localtime is a symlink to: Europe/Paris\nFix the configuration, or set the time zone in a TZ environment variable.\n'
```

My `/etc/timezone` file contained the text `Etc/UTC`. I replaced this text with `Europe/Paris`, to match the target of the `/etc/localtime` symlink, and the error went away. Has anybody else tried to run `mxcubeweb` on Ubuntu 24 yet? If the problem is not isolated, then maybe we'll have to upgrade `tzlocal` or something.

marcus-oscarsson:
The request to `login_info` raised an exception and incorrectly returned a stack trace with the 200 status code.

The handling within `login_info` is perhaps still not ideal but this PR makes sure that the `is_anonymous` on `current_user` is updated correctly before the code that returns the user information is called. The "event" of "no valid sessions" should probably not be seen as an exception.

All requests raising an exceptions now returns a 501 without the traceback

Fixes https://github.com/mxcube/mxcubeweb/issues/1477

### DIFF
--- a/mxcubeweb/core/components/user/usermanager.py
+++ b/mxcubeweb/core/components/user/usermanager.py
@@ -264,10 +264,17 @@ class BaseUserManager(ComponentBase):
             self.app.server.emit("forceSignout", room=socketio_sid, namespace="/hwr")
 
     def login_info(self):
+        # update_operator will update the login status of current_user, and make
+        # sure that the is_anonymous has the correct value.
+        # Update operator calls lims.select_session that raises an exception if
+        # there are no valid LIMS sessions.
+        try:
+            self.update_operator()
+        except Exception:
+            pass
+
         if not current_user.is_anonymous:
             session_manager: LimsSessionManager = self.app.lims.get_session_manager()
-            self.update_operator()
-
             login_type = (
                 "User" if HWR.beamline.lims.is_user_login_type() else "Proposal"
             )
--- a/mxcubeweb/routes/main.py
+++ b/mxcubeweb/routes/main.py
@@ -63,11 +63,6 @@ def init_route(app, server, url_prefix):
 
     @server.flask.before_request
     def before_request():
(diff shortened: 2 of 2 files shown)


axelboc:
Fixes #1477 probably? (I've opened a ticket with our IT division about the conflicting timezone on Ubuntu 24.)

marcus-oscarsson:
:), thanks thats great because @walesch-yan also had the issue. I'm not sure it fixes the actual issue but you wont get it right away, I guess you would bump into the timezone issue at some point in anyways.

PR #1514, opened 2024-12-17
Don't call `login_info` twice on login/logout

axelboc:
Just an optimisation to avoid unnecesssary requests and re-renders. The `getLoginInfo` action dispatched from the `useEffect` in `App.jsx` would, on initial load, cause the `loggedIn` flag to change, which would cause the `useEffect` to run again, which would dispatch the `getLoginIngo` action again... A similar thing happened when logging in and out.

Splitting the `useEffect` in two in `App.jsx` ensures that `getLoginInfo` is dispatched only once in all cases (initial load, login and logout).


marcus-oscarsson:
Awesome, thanks. much appreciated :)

PR #1518, opened 2024-12-19
Log user out from UI on 401

axelboc:
Following https://github.com/mxcube/mxcubeweb/pull/1509#issuecomment-2548421307, I'm adding a global fetch error catcher to log the user out from the UI whenever they encounter a 401 response.

The challenge was dealing with import cycles... The global catcher is set on the wretch API object, which is used by all the endpoint functions, which are used by the Redux actions... I had to:

- register the 401 catcher on a separate wretch API object (in a dedicated file) specifically for authenticated endpoints;
- make sure that unauthenticated endpoint functions, including `/login_info`, made use of the base wretch API object (the one without the 401 catcher);
- make sure not to use the `getLoginInfo` action creator in the 401 catcher (otherwise, I would have had to move it into a separate file as well...)

Anyway, it was a bit of a mess, but I got there I think.

To test the behaviour, I recommend:

1. Commenting out the `forceSignout` socket event handler in `serverIO.js`
2. Logging in twice into the UI (e.g. normal and private window)
3. Logging the other user out and waiting for the `/refresh_session` or `/ra` endpoint to be called for that user

[Screencast from 2024-12-19 14-14-17.webm](https://github.com/user-attachments/assets/9fdc89ed-8cd1-4f35-a6e4-cb813e95b583)

### DIFF
--- a/ui/src/actions/beamline.js
+++ b/ui/src/actions/beamline.js
@@ -4,12 +4,6 @@ import {
   sendPrepareBeamlineForNewSample,
 } from '../api/beamline';
 import { sendLogFrontEndTraceBack } from '../api/log';
-// The different states a beamline attribute can assume.
-export const STATE = {
-  IDLE: 'READY',
-  BUSY: 'BUSY',
-  ABORT: 'UNUSABLE',
-};
 
 // Action types
 export const BL_UPDATE_HARDWARE_OBJECT = 'BL_UPDATE_HARDWARE_OBJECT';
--- a/ui/src/actions/login.js
+++ b/ui/src/actions/login.js
@@ -7,7 +7,8 @@ import { fetchAvailableWorkflows } from '../api/workflow';
 import { fetchAvailableTasks, fetchQueueState } from '../api/queue';
 
 import { showErrorPanel, applicationFetched } from './general';
-import { fetchLoginInfo, sendLogIn, sendSignOut } from '../api/login';
+import { sendSignOut } from '../api/login';
+import { fetchLoginInfo, sendLogIn } from '../api/loginBase';
 import { fetchDetectorInfo } from '../api/detector';
 import { fetchSampleChangerInitialState } from '../api/sampleChanger';
 import { fetchHarvesterInitialState } from '../api/harvester';
@@ -20,21 +21,6 @@ export function setLoginInfo(loginInfo) {
     type: 'SET_LOGIN_INFO',
     loginInfo,
   };
-}
-
-export function resetLoginInfo() {
-  return setLoginInfo({
-    beamlineName: '',
-    synchrotronName: '',
-    loginType: '',
-    user: '',
-    proposalList: [],
-    selectedProposal: '',
-    selectedProposalID: '',
-    loggedIn: false,
-    rootPath: '',
-    useSSO: false,
-  });
 }
 
 export function showProposalsForm() {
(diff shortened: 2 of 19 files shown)


ui/src/actions/beamline.js
```diff
@@ -4,12 +4,6 @@ import {
   sendPrepareBeamlineForNewSample,
 } from '../api/beamline';
 import { sendLogFrontEndTraceBack } from '../api/log';
-// The different states a beamline attribute can assume.
-export const STATE = {
-  IDLE: 'READY',
-  BUSY: 'BUSY',
-  ABORT: 'UNUSABLE',
-};
```
axelboc, line 12: Moved to `reducer/beamline` since that's the only place it's used (to reduce chance of import cycle).

ui/src/actions/login.js
```diff
@@ -27,21 +23,6 @@ export function setLoginInfo(loginInfo) {
...
-    loginType: '',
-    user: '',
-    proposalList: [],
-    selectedProposal: '',
-    selectedProposalID: '',
-    loggedIn: false,
-    rootPath: '',
-    useSSO: false,
-  });
-}
```
axelboc, line 38: All we care about is setting `loggedIn` to `false` to trigger the `useEffect` in `App.jsx` to disconnect the sockets. So I'm doing exactly that below in `signOut` by calling `setLoginInfo` directly.

ui/src/api/loginBase.js
```diff
@@ -0,0 +1,19 @@
...
+
+export function sendLogIn(proposal, password, previousUser) {
+  return endpoint.post({ proposal, password, previousUser }, '/').safeJson();
+}
+
+export function sendSSOLogIn() {
+  window.location = 'mxcube/api/v0.1/login/ssologin';
+}
+
+export function fetchLoginInfo() {
```
fabcor-maxiv: Probably very few care about this, but anyway... Can we be more consistent with casing? : D

```suggestion
export function sendLogin(proposal, password, previousUser) {
  return endpoint.post({ proposal, password, previousUser }, '/').safeJson();
}

export function sendSsoLogin() {
  window.location = 'mxcube/api/v0.1/login/ssologin';
}

export function fetchLoginInfo() {
```

My rule of thumb is to think how I would write in snake case: `send_s_s_o_log_in` or `send_sso_login`.

axelboc: - It's the verb "to log in" in this case - i.e. send + command (with some exceptions to avoid duplicating the verb "send", like `sendFeedback`).
- Regarding casing of abbreviations, that's a whole other discussion. Even the web platform is inconsistent (e.g. [`XMLHttpRequest`](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest) :rofl:) Personally, I think enforcing a specific case is unnecessary.
- Anyway, lucky for us, `sendSSOLogIn` will be removed with #1517 :joy:

ui/src/actions/beamline.js
marcus-oscarsson: Great !

ui/src/api/api.js
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable promise/prefer-await-to-callbacks */
+import baseApi from './apiBase';
+import { fetchLoginInfo } from './loginBase';
+import { store } from '../store';
+
+const api = baseApi
```
marcus-oscarsson, line 6: Ok, I see. Nice

marcus-oscarsson:
Hey @axelboc. There just a few conflicts as I already merged, #1517. It otherwise looks good to me. You can go a head and merge when the conflicts are solved.

PR #1522, opened 2025-01-08
add visual cues for development environment

mockoocy:
Add MAXIV we want to easily distinguish between development, testing and production environments.
This PR does so, by changing background color of `body` depending on the selected environment. It is opt-in behaviour, which can be enabled by setting `VITE_DEV_MODE_VISUAL_CUES_ENABLED="true"` in `.env` file. 

As a small note about the environments in vite:
- `development` is set when running with `vite`
- `production` is set when application is built with `vite build`
- `staging` is set when application is built with `vite build --mode staging`

### DIFF
--- a/docs/source/dev/front-end.md
+++ b/docs/source/dev/front-end.md
@@ -17,9 +17,9 @@ The front-end package manager is [pnpm](https://pnpm.io/):
 
 [Vite](https://vitejs.dev/) is used to build the project for production:
 
-- `pnpm start` - start the development server at http://localhost:5173
+- `pnpm start` - start the development server at <http://localhost:5173>
 - `pnpm build` - build the front-end for production into the `ui/build` folder
-- `pnpm preview` - serve the production build at http://localhost:5173
+- `pnpm preview` - serve the production build at <http://localhost:5173>
 
 The codebase is formatted with Prettier, linted with ESLint and tested with Cypress:
 
@@ -74,3 +74,4 @@ Environment variables are defined in file `ui/.env`. To override them, create yo
 The following environment variables are available:
 
 - `VITE_REDUX_LOGGER_ENABLED`: whether to log Redux actions to the browser console (disabled by default); useful if you're unable to install the [Redux devtools](https://github.com/reduxjs/redux-devtools/tree/main/extension#installation) browser extension.
+- `VITE_BG_COLOR`: background color for the `body` tag. Useful for making a distinction between various deployment environments.
--- a/ui/.env
+++ b/ui/.env
@@ -6,3 +6,5 @@
 
 # Redux logger ('true' to enable)
 VITE_REDUX_LOGGER_ENABLED=
+# Background color for the <body> element.
+VITE_BG_COLOR="transparent"
--- a/ui/index.html
+++ b/ui/index.html
@@ -15,7 +15,7 @@
     <link rel="apple-touch-icon" href="/mxcube_logo20.png" />
     <link rel="manifest" href="/manifest.json" />
   </head>
-  <body>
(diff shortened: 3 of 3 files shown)


mockoocy:
The pipeline does not seem to pass, due to: https://github.com/mxcube/mxcubeweb/pull/1521

axelboc (changes requested):
May I suggest a simpler, less intrusive implementation based on [local `.env` files](https://vite.dev/guide/env-and-mode#env-files)?

```
// .env.development.local
VITE_BG_COLOR=#ead6cc
```

```
// .env.staging.local
VITE_BG_COLOR=#b4acae
```

```html
<!-- ui/index.html -->
<body style="background-color: %VITE_BG_COLOR%;">
```

mockoocy:
> May I suggest a simpler, less intrusive implementation based on [local `.env` files](https://vite.dev/guide/env-and-mode#env-files)?
> 
> ```
> // .env.development.local
> VITE_BG_COLOR=#ead6cc
> ```
> 
> ```
> // .env.staging.local
> VITE_BG_COLOR=#b4acae
> ```
> 
> ```
> 
> <body style="background-color: %VITE_BG_COLOR%;">
> ```

Yours is much more clever! Will do it that way :)

elmjag:
How about just adding a way to override the background color on deployment?

Like adding support for optional `VITE_BG_COLOR` setting. If it's not specified, you get the current default color. So it's up to your deployment code to specify a custom background color, if desired.

This way we don't need to add hard-coded mapping between deployment type and background colors in MXCuBE itself.

elmjag:
Regardless of which implementation you want to go with, please add documentation on how to use it.

axelboc:
`.env.*.local` files are not committed, so in effect, they can be added just before building the UI in whichever environment the build runs. For documentation, I recommend adding an empty `VITE_BG_COLOR` variable in `.env` (there's already an explanation on `.env.*.local` files at the top of this file), and a mention in `front-end.md`.

marcus-oscarsson:
@axelboc Good idea with the local .env files !

mockoocy:
Rebased and changed to use the local env files.

docs/source/dev/front-end.md
```diff
@@ -17,9 +17,9 @@ The front-end package manager is [pnpm](https://pnpm.io/):
 
 [Vite](https://vitejs.dev/) is used to build the project for production:
 
-- `pnpm start` - start the development server at http://localhost:5173
+- `pnpm start` - start the development server at <http://localhost:5173>
 - `pnpm build` - build the front-end for production into the `ui/build` folder
-- `pnpm preview` - serve the production build at http://localhost:5173
+- `pnpm preview` - serve the production build at <http://localhost:5173>
```
axelboc, line 22: Oh didn't know this syntax, thanks! TIL

docs/source/dev/front-end.md
```diff
@@ -74,3 +74,4 @@ Environment variables are defined in file `ui/.env`. To override them, create yo
 The following environment variables are available:
 
 - `VITE_REDUX_LOGGER_ENABLED`: whether to log Redux actions to the browser console (disabled by default); useful if you're unable to install the [Redux devtools](https://github.com/reduxjs/redux-devtools/tree/main/extension#installation) browser extension.
+- `VITE_BG_COLOR`: background color for the `body` tag. Useful for making a distinction between various deployment environments.
```
axelboc, line 77: I would also add this var to `.env` with `transparent` as default value and a comment, both for discoverability and so the HTML/CSS in the generated `index.html` remains valid.

mockoocy: Done

axelboc (approved):
Thanks!

PR #1525, opened 2025-01-14
Adds option to create 2d point on canvas

mockoocy:
This PR just adds option to create 2d points from the sample view's context menu.


marcus-oscarsson:
Thanks

PR #1528, opened 2025-01-23
remove fixed column size for samples nav

walesch-yan:
In the `Samples` page, the column sizes of the top row including the filter and some buttons are fixed, which causes the different elements to overlap and sometimes hide features on smaller screens. 

This PR removes the fixed sizes, you can see the differences below for different screen sizes
Before:
Screenshot from 2025-01-23 15-05-48

Screenshot from 2025-01-23 15-06-10

Screenshot from 2025-01-23 15-07-06

Now:

Screenshot from 2025-01-23 15-02-46

Screenshot from 2025-01-23 15-06-33

Screenshot from 2025-01-23 15-06-49

### DIFF
--- a/ui/src/containers/SampleListViewContainer.jsx
+++ b/ui/src/containers/SampleListViewContainer.jsx
@@ -845,8 +845,8 @@ class SampleListViewContainer extends React.Component {
         ) : null}
         <Card className="samples-grid-table-card">
           <Card.Header className="samples-grid-table-card-header">
-            <Row className="samples-grid-table-row-header">
-              <Col sm={5} className="d-flex">
+            <Row className="gap-2">
+              <Col className="d-flex">
                 {this.getSynchronizationDropDownList()}
                 <span style={{ marginLeft: '1.5em' }} />
                 <Button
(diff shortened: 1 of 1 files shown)


ui/src/containers/SampleListViewContainer.jsx
```diff
@@ -845,8 +845,8 @@ class SampleListViewContainer extends React.Component {
         ) : null}
         <Card className="samples-grid-table-card">
           <Card.Header className="samples-grid-table-card-header">
-            <Row className="samples-grid-table-row-header">
```
walesch-yan, line 848: While at it, I removed this className reference, since the css class, does not seem to exist anymore

axelboc (approved):
Nice! Any chance you could add some vertical spacing when the elements wrap onto multiple lines? Maybe with `gap-2` or `style={{ rowGap: '1rem' }}` on the `Row` element probably?

walesch-yan:
> Nice! Any chance you could add some vertical spacing when the elements wrap onto multiple lines? Maybe with `gap-2` or `style={{ rowGap: '1rem' }}` on the `Row` element probably?

Thanks for pointing it out, definitely looks better with the gap!
Screenshot from 2025-01-27 16-52-25

PR #1529, opened 2025-01-27
remove duplicate LIMS call when setting active session

elmjag:
No need to call set_active_session_by_id() twice on a LIMS object.


marcus-oscarsson:
Indeed, :) thanks. We have some other minor fixes coming relatively soon.

elmjag:
> Indeed, :) thanks. We have some other minor fixes coming relatively soon.

Nice!

FYI, I also have a pile of changes related LIMS ISPyB user-login code I want to merge into mxcubecore. I'm working on putting together coherent PRs.

Issue #1519, opened 2024-12-19
ReadTheDocs build ends with error


marcus-oscarsson:
The build process on ReadTheDocs ends with the following error message:
```
...
  - Installing sphinx-last-updated-by-git (0.3.8)
  - Installing sphinxcontrib-httpdomain (1.8.1)
  - Installing tzlocal (4.3.1)

Installing the current project: mxcubeweb (4.249.0)

[rtd-command-info] start-time: 2024-12-19T13:28:40.394331Z, end-time: 2024-12-19T13:58:00.962292Z, duration: 1760, exit-code: 137
mamba run --name mxcubeweb python -m sphinx -T -E -b html -d _build/doctrees -c docs docs/source ${READTHEDOCS_OUTPUT}/html
Command killed due to timeout or excessive memory consumption
```

It seems like the conda build/install consumes too much memory

marcus-oscarsson:
The docs on http://mxcube.github.io/mxcubeweb/ is still built

fabcor-maxiv:
Still stuck today. I changed the [branch protection rules for `develop`](https://github.com/mxcube/mxcubeweb/settings/branches) so that Read The Docs build is not "required" for now, we should revert once this is solved, but we have the GitHub Pages build anyway which is more or less equivalent.

marcus-oscarsson:
Ok

marcus-oscarsson:
The `read-the-docs` builds are still failing. I think I will disable the action and if possible redirect the `read-the-docs` page to the `GitHub` pages. What do you think ?

fabcor-maxiv:
Another one of those things that is on my TO-DO list, but that keeps getting postponed... :/

I think I have an idea. I want to get rid of the "[REST API](https://mxcubeweb.readthedocs.io/en/stable/rest-api/index.html)" part of the documentation. Maybe this will help, maybe we need to get rid of more things.

fabcor-maxiv:
It seems to me like it is starting with this PR that the Readthedocs builds started failing: "[Adding documentation and dependency for missing video stream #1504](https://github.com/mxcube/mxcubeweb/pull/1504)" where we added `ffmpeg` as a dependency in `conda-environment.yml`.

fabcor-maxiv:
[...]

PR #1530, opened 2025-01-27
Remove "REST API" section of the docs

fabcor-maxiv:
Likely this section of the documentation is never looked up. On the other hand it is resource intensive to generate these pages.

GitHub: https://github.com/mxcube/mxcubeweb/issues/1519

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
           - --strict
 
   - repo: https://github.com/python-poetry/poetry
-    rev: 1.8.3
+    rev: 1.8.5
     hooks:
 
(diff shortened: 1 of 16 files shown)


marcus-oscarsson:
Ok, good idea. Lets try this :) Thanks alot for investigating this issue I know that its time consuming !

marcus-oscarsson:
I guess we don't really need this one then, or do you still want to go trough with this ?

fabcor-maxiv:
Probably not necessary to pursue this indeed, we can keep these pages.

Even though, for the record, this part of the documentation is seemingly in terribly bad shape and would need serious work.

marcus-oscarsson:
I completely agree with you on the last bit, we should really think about how to improve that part

PR #1531, opened 2025-01-28
Update pre-commit hooks and Poetry

fabcor-maxiv:
Update pre-commit hooks and Poetry. Format and lint with those updated tools.

Pin the version of Poetry in `conda-environment.yml` to match the version used in the `pre-commit` hook.

Many tools already dropped support for Python version 3.8, including Poetry, `pre-commit`, `black`, and `isort`, so we stay on slightly older versions of those tools for now.


marcus-oscarsson:
Looks good to me, please feel free to merge once rebased

PR #1532, opened 2025-01-28
Updates to UserManager related to LIMS session handling

marcus-oscarsson:
Updates to UserManager related to LIMS session handling:

  -  Not storing LIMS data in the user database
  -  Consistent handling of "force signout" and "signout" using `delete_user` in both cases
  -  Session handling; updating session list on change, automatically selecting session if there is only one available
  -  Added separate tab for sessions from other beamlines

### DIFF
--- a/mxcubeweb/core/components/user/usermanager.py
+++ b/mxcubeweb/core/components/user/usermanager.py
@@ -41,6 +41,7 @@ class BaseUserManager(ComponentBase):
 
     def handle_sessions_changed(self, sessions):
         self.app.server.emit("sessionsChanged", namespace="/hwr")
+
     def get_observers(self):
         return [
             user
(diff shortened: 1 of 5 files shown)


mxcubeweb/core/components/user/usermanager.py
```diff
@@ -433,37 +450,26 @@ def _login(self, login_id: str, password: str) -> LimsSessionManager:
...
-                raise Exception("Another user is already logged in")
-
         # Only allow local login when remote is disabled
         if not self.app.ALLOW_REMOTE and not is_local_host():
             raise Exception("Remote access disabled")
 
         return session_manager
 
-    def _signout(self):
+    def _signout(self, sso_data={}):
```
fabcor-maxiv:
```suggestion
    def _signout(self, sso_data=None):
```

Is it one of those cases where having a mutable thing as default argument leads to unexpected results?

Examples:
* <https://www.geeksforgeeks.org/use-mutable-default-value-as-an-argument-in-python/>
* <https://nikos7am.com/posts/mutable-default-arguments/>
* <https://www.pullrequest.com/blog/python-pitfalls-the-perils-of-using-lists-and-dicts-as-default-arguments/>

Looking at the code of the method, it seems like it should be fine because the dict is only read but never written. It still feels strange to me to have a dict as default argument.

PR #1533, opened 2025-01-28
Update `.readthedocs.yaml`

### LINKED ISSUE #1519: ReadTheDocs build ends with error

fabcor-maxiv:
GitHub: fix https://github.com/mxcube/mxcubeweb/issues/1519

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -10,19 +10,37 @@ version: 2
 
 build:
 
-  os: "ubuntu-22.04"
+  os: ubuntu-22.04
 
   tools:
-    python: "mambaforge-22.9"
+    python: mambaforge-22.9
 
   commands:
+
     # With shallow clone the "Last updated" timestamps can not get computed
     # See "Caveats" at <https://pypi.org/project/sphinx-last-updated-by-git/>
-    - "git fetch --unshallow || true"
-    - "mamba env create --file conda-environment.yml --force"
-    - "mamba run --name mxcubeweb poetry install --only=docs,main"
-    # yamllint disable-line rule:line-length
-    - "mamba run --name mxcubeweb python -m sphinx -T -E -b html -d _build/doctrees -c docs docs/source ${READTHEDOCS_OUTPUT}/html"
+    - git fetch --unshallow || true
+
+    # Create a Conda environment with only the dependencies necessary to build the docs:
+    # * we skip `ffmpeg` on purpose as it is not necessary to build the docs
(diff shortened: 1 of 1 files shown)


beteva:
Hi guys, thanks for the effort to sort this out. Just a side question - do we still stick to python 3.8?

fabcor-maxiv:
> do we still stick to python 3.8?

My approach is to preferably format, build, lint, and so on with the lowest supported version of Python, as I assume that since Python is mostly backwards compatible, such operations should still work with later Python versions. As of today, Python 3.8 is still the lowest version supported by MXCuBE-Web.

fabcor-maxiv:
I will merge this without approval, it is just documentation build and only on Readthedocs, not the code of MXCuBE-Web itself.

marcus-oscarsson:
> I will merge this without approval, it is just documentation build and only on Readthedocs, not the code of MXCuBE-Web itself.

Thanks again for taking time to look at this. I guess this would be a limit case in my opinion. Your solution is working and does indeed not effect the code base of mxcube itself but we can still ask for a review or ping someone. On the other hand there is nothing that prevents us from opening another PR later if needed. 

@beteva I guess we could drop `Python 3.8` as well, actually It's what I would propose. And investigate if we can start running `Python 3.13`

rhfogh:
W really ought to stop supporting Python 3.8. It is EOL. And jump as far forward as possible, to postpone the time when we have to do this again.

fabcor-maxiv:
Unless I am missing something, this change is not negatively impacted by the fact that we have to support an old EOL Python (3.8). This other change is: <https://github.com/mxcube/mxcubeweb/pull/1531>. Anyway, whatever our lowest supported version is, whether or not it is an EOL version, we will always have to write down the version string in a bunch of places and keep it in sync, normal churn.

PR #1536, opened 2025-01-31
Adding processing parameters (space group and unit) to helical scan

dominikatrojanowska:


dominikatr:
Added rows, hidden by default
image

dominikatrojanowska:
Added rows, hidden by default:
image

ui/src/components/Tasks/Helical.jsx
```diff
@@ -73,6 +75,7 @@ class Helical extends React.Component {
     resetLastUsedParameters(this);
   }
 
+  // TODO: add widgets for entering space group and unit cell parameters
```
elmjag, line 78: Remove this line.

The widgets for space group and cell params are added, thus TODO have been DONE :)

ui/src/components/Tasks/Helical.jsx
```diff
@@ -179,6 +182,45 @@ class Helical extends React.Component {
           </Form>
 
           <FieldsHeader title="Processing" />
+          <CollapsableRows>
```
marcus-oscarsson, line 185: I guess this is copied from the characterization or data-collection dialog, it could be extracted to a separate functional component, what do you think ?

elmjag: Yes, this is a copy-and-pasted code. It already appears twice in the code, in [DataCollection.jsx](https://github.com/mxcube/mxcubeweb/blob/develop/ui/src/components/Tasks/DataCollection.jsx#L295) and in [Mesh.jsx](https://github.com/mxcube/mxcubeweb/blob/develop/ui/src/components/Tasks/Mesh.jsx#L193).

I agree with @marcus-oscarsson, we should put code into one component, and re-use it in DataCollection.jsx, Mesh.jsx and Helical.jsx files.

axelboc: Hmm not sure. I'm not liking the extra `<Form>` wrapper to be honest — I wouldn't like it to end up in a reusable component without some further thoughts. And given that the markup is pretty simple, I'm not too concerned with the duplication at this time. Duplication is better than a wrong abstraction.

marcus-oscarsson: Ok, yes but then when you would like to change you have to change in three different places. How would you do to get the abstraction right so that it could possible become reusable ?

axelboc: It's just that this component is class-based and relies on `redux-form` and I'm not sure what the impact of that refactoring will be. The abstraction might actually help with the refactoring ... or it might hinder it, it's tough to know in advance. Anyway, if you feel that this code is likely to change before the refactoring of `redux-form` happens, then sure, it's probably a good idea to abstract it now.

marcus-oscarsson: Ah, good point. Actually Id like to get rid of `redux-form` as soon as we can so it would perhaps be good to think  a bit about as you mention. I guess these standard forms can be replaced by `react-form` but then we also need to see how/if `json-schema-form` can be used better for the "automatically" generated forms.

marcus-oscarsson: Seeing that this part of the form have not changed since basically the begging of the project I guess its not susceptible to change anytime soon. So lets keep it like this and take another look at it when we remove `redux-form`

PR #1537, opened 2025-01-31
[SSX] - Minor fixes to chip interface

marcus-oscarsson:
A few fixes to the chip interfaces mostly because of a `fabric.js` update


axelboc (approved):
Seems legit

PR #1540, opened 2025-01-31
[Lint] - Fixed strange if

marcus-oscarsson:
Fixed the odd if statement, the sample changer HardwareObjects should now return True and not None or True on success.

### DIFF
--- a/mxcubeweb/core/components/samplechanger.py
+++ b/mxcubeweb/core/components/samplechanger.py
@@ -213,8 +213,6 @@ class SampleChanger(ComponentBase):
                 elif sc.get_loaded_sample().get_address() != sample["location"]:
                     res = sc.load(sample["sampleID"], wait=True)
 
-                if res is None:
-                    res = True
                 if (
                     res is not False
                     and HWR.beamline.queue_manager.centring_method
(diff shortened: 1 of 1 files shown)


mxcubeweb/core/components/samplechanger.py
```diff
@@ -213,10 +213,8 @@ def mount_sample_clean_up(self, sample):
                 elif sc.get_loaded_sample().get_address() != sample["location"]:
                     res = sc.load(sample["sampleID"], wait=True)
 
-                if res is None:
-                    res = True
                 if (
-                    res is not False
+                    not res
```
elmjag:
```suggestion
                    res is not None
```
Otherwise the semantics for this if statement changes.

For example `not res` will be True even for res = "", res = 0, not only res = None. Don't know if this actually make a difference in this code.

marcus-oscarsson: `load` of the `SampleChanger` hardware object should now always return `True` or `False`. It used to be the the case, in the past that, `None` or `True` was returned for a successful load. 

Your right the test should be changed, but to `if res`. 

Thanks for spotting this one, I made a quick touch up before making the PR and made this mistake but it should be fine now.

PR #1542, opened 2025-01-31
[Adapter] - Making validation errors more visible

marcus-oscarsson:
Improved logging for adapter validation errors

### DIFF
--- a/mxcubeweb/core/adapter/beamline_adapter.py
+++ b/mxcubeweb/core/adapter/beamline_adapter.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 import logging
 
+import pydantic
+
 BEAMLINE_ADAPTER = None
 
 # Singleton like interface is needed to keep the same reference to the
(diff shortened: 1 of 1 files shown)


mxcubeweb/core/adapter/beamline_adapter.py
```diff
@@ -67,7 +69,12 @@ def dict(self):
         attributes = {}
 
         for attr_name in self.app.mxcubecore.adapter_dict:
-            _d = self.app.mxcubecore.get_adapter(attr_name).dict()
+            try:
+                _d = self.app.mxcubecore.get_adapter(attr_name).dict()
+            except pydantic.ValidationError:
+                logging.getLogger("MX3.HWR").error(f"Incorrect values in {attr_name}")
+                logging.getLogger("MX3.HWR").exception("")
```
walesch-yan, line 76: Why would we need two function calls in this case? Would `logging.getLogger("MX3.HWR").exception(f"Incorrect values in {attr_name}")` not be sufficient as `.exception()` already logs on ERROR level (with the addition of the exception info)?

marcus-oscarsson: Sometimes I like two different log entries. In this way its a bit more clear what the "user friendly" message is so that the message is not "lost" in the stack trace. But you are right that it does not make a big difference. Its simply an extra line of log.

walesch-yan: I see your point, indeed this can make the logs more readable :)

walesch-yan (approved):
Seems fine to me

PR #1535, opened 2025-01-31
Removing unused templates folder and update package.json

walesch-yan:
This PR removes the unused templates folder. The folder contained the only references to the `dustjs` related dependencies in the codebase, as such the PR removes them from `package.json` as well.

PR #1543, opened 2025-02-04
Update package.json dependencies

walesch-yan:
This PR updates some of the dependencies in `package.json`, due to possible vulnerabilities, update includes:

- `fabric` (could not update to version 6.\*, due to usage of removed functionalities, which are already deprecated in 5.\*) 
- `socket.io-client`
- `wait-on`
- removal of unused `isomorphic-fetch`

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -73,7 +73,7 @@
     "prop-types": "15.8.1",
     "vite": "5.4.12",
     "vite-plugin-eslint": "1.8.1",
-    "wait-on": "7.0.1"
+    "wait-on": "8.0.2"
   },
   "browserslist": {
     "production": [
--- a/ui/package.json
+++ b/ui/package.json
@@ -29,7 +29,6 @@
     "dygraphs": "^2.0.0",
     "fabric": "5.5.1",
     "immutability-helper": "^3.0.2",
-    "isomorphic-fetch": "^2.2.0",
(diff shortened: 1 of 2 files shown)


ui/package.json
```diff
@@ -27,9 +27,8 @@
     "bootstrap": "5.2.3",
     "classnames": "^2.3.1",
     "dygraphs": "^2.0.0",
-    "fabric": "5.3.0",
+    "fabric": "5.5.1",
     "immutability-helper": "^3.0.2",
-    "isomorphic-fetch": "^2.2.0",
```
axelboc, line 32: This is supposedly used in a couple of files:

image

but I'm not sure what these files are, so I'll let others chime in :grin: .

fabcor-maxiv: The templates were removed in <https://github.com/mxcube/mxcubeweb/pull/1535>, no?

walesch-yan: Yes they have been removed in #1535. Actually I missed to remove the dependency in the previous PR as it seems

walesch-yan: I cannot exactly tell what the initial purpose of these files was, but no content of the templates folder was ever used in the code

axelboc: Oh fantastic!

axelboc (approved):

PR #1546, opened 2025-02-11
Clean up UserMessage component

jbflo:
The Idea was to remove the deprecated " this.ref  ",
but ending  clean up a bit more the component.
- Remove old code related to overlay that are no longer use 
- Transform to functional component and css to .module.css
- Make queue-messages-title sticky and only scroll log message


ui/src/components/Notify/UserMessage.jsx
```diff
@@ -78,27 +21,17 @@ export default class UserMessage extends React.Component {
...
-        }}
-      >
-        {messages}
-      </div>
-    );
+    return msg;
   }
+  return (
+    <div id="usermessages" className={styles.messageBody}>
+      {renderMessages()}
```
axelboc, line 34: I'd recommend removing `render<Something>` functions altogether. It used to be popular in monolithic class-based components as a way to keep the `render` function under control, but it was a bit of a code smell even back then.

- First guideline: inline first, abstract later — inline all the JSX and then see what you get before deciding what to abstract/extract/compose/... You'll probably find that inlining `map` loops is actually more readable and less misdirecting than to return JSX arrays from `render<Something>` functions.
- Second guideline: with few exceptions, if a function returns JSX, it's a React component and it needs to be in its own file. This pretty much forbids `render<Something>` closures and methods altogether and highlights that the best way to remove complexity from a long JSX statement is to think more in React (components, composition, etc.)

ui/src/components/Notify/UserMessage.module.css
```diff
@@ -0,0 +1,42 @@
...
+}
+
+.message {
+  position: relative;
+  padding: 0.4em;
+  border-color: #333;
+  opacity: 1;
+}
+
+.INFO {
```
axelboc, line 17: I would use a `data-` attribute for this - e.g. `data-level="INFO"` and then:

```css
.message[data-level="INFO"] {...}
```

It should simplify the JS logic as well.

ui/src/components/Notify/UserMessage.module.css
```diff
@@ -0,0 +1,42 @@
...
+  border-top: 5px solid #d9534f;
+  background-color: rgba(217, 83, 79, 0.8);
+}
+
+.DEBUG {
+  border-top: 5px solid #d9534f;
+  background-color: rgba(217, 83, 79, 0.8);
+}
+
+.messageText {
```
axelboc, line 38: Your classes are namespaced so, technically, nothing forces you to repeat the `message` prefix. `.text` would work fine.

ui/src/components/Notify/UserMessage.module.css
```diff
@@ -0,0 +1,42 @@
+.messageBody {
```
axelboc, line 1: Maybe `.container` or `.root` or `.messages`?

ui/src/components/SampleQueue/app.css
```diff
@@ -167,6 +167,11 @@
   border-bottom: 1px solid #ddd;
   padding: 0.5rem 1rem;
 }
+.queue-messages-body {
+  height: 75%;
```
axelboc, line 171: Might be cleaner to size this element with flexbox? i.e. `display: flex` and `flex-direction: column` on the parent, and then `flex: 1` on this element so it fills up the available vertical space.

axelboc (commented):
Thanks that's awsome! I see Marcus didn't wait any time to merge, but here are some comments anyway, in case you're up for making a follow-up PR.

marcus-oscarsson:
Thanks @axelboc, yes I was maybe a bit fast this time ;). @jbflo Feel free to follow up with the suggestions from @axelboc

ui/src/components/Notify/UserMessage.module.css
jbflo: This file has been renamed.  not forces at all :innocent: . 
even odd the className could look like, don't care that much  :rofl:

PR #1547, opened 2025-02-11
Develop connection lost dialog

jbflo:
----Edited ---
Same here, Idea was to remove the deprecated " this.ref ",
but ending changing a bit more the component.

- Transform to functional component 
- give user options to reload if not automatically reconnect  "

### DIFF
--- a/ui/src/containers/ConnectionLostDialog.jsx
+++ b/ui/src/containers/ConnectionLostDialog.jsx
@@ -1,88 +1,26 @@
-/* eslint-disable react/jsx-handler-names */
-/* eslint-disable react/no-string-refs */
 import React from 'react';
-import { connect } from 'react-redux';
-import { bindActionCreators } from 'redux';
+import { useSelector } from 'react-redux';
 import { Modal, Button } from 'react-bootstrap';
-import { showConnectionLostDialog } from '../actions/general';
 
-export class ConnectionLostDialog extends React.Component {
-  constructor(props) {
-    super(props);
-    this.accept = this.accept.bind(this);
-    this.reject = this.reject.bind(this);
-    this.intervalCb = this.intervalCb.bind(this);
-    this.seconds = 60;
-    this.timerid = undefined;
-  }
+export function ConnectionLostDialog() {
+  const show = useSelector((state) => state.general.showConnectionLostDialog);
 
-  intervalCb() {
-    this.seconds--;
-
-    if (this.refs.countdown) {
-      this.refs.countdown.innerHTML = this.seconds;
-    }
-
-    if (this.seconds < 1) {
-      this.seconds = 0;
-      this.accept();
-    }
-  }
-
-  accept() {
-    this.forceUpdate();
-  }
-
-  reject() {
-    this.props.hide();
-  }
-
-  render() {
-    this.seconds = 60;
-
-    if (this.timerid) {
-      clearInterval(this.timerid);
-    }
-
-    this.timerid = setInterval(this.intervalCb, 1000);
-    setTimeout(() => {
-      clearInterval(this.timerid);
-    }, this.seconds * 1000);
-
-    return (
-      <Modal show={this.props.show} onHide={this.props.hide}>
-        <Modal.Header>
-          <Modal.Title>Connection lost</Modal.Title>
-        </Modal.Header>
-        <Modal.Body>
-          Ooops ! There seems to be a problem with the internet connection, the
-          connection to the server was lost. Trying to reconnect in{' '}
-          <span ref="countdown">{this.seconds}</span> s
-        </Modal.Body>
-        <Modal.Footer>
-          <Button onClick={this.accept}> Try again </Button>
-        </Modal.Footer>
-      </Modal>
-    );
-  }
+  return (
+    <Modal show={show}>
+      <Modal.Header>
+        <Modal.Title>Connection lost</Modal.Title>
+      </Modal.Header>
+      <Modal.Body>
+        Trying to reconnect now...
+        <br />
+        If the issue persists, please check your Internet connection or reload
+        the page.
+      </Modal.Body>
+      <Modal.Footer>
+        <Button onClick={() => window.location.reload(true)}> Reload </Button>
+      </Modal.Footer>
+    </Modal>
+  );
 }
 
-function mapStateToProps(state) {
-  return {
-    show: state.general.showConnectionLostDialog,
-  };
-}
-
-function mapDispatchToProps(dispatch) {
-  return {
-    hide: bindActionCreators(
-      showConnectionLostDialog.bind(null, false),
-      dispatch,
-    ),
-  };
-}
-
-export default connect(
-  mapStateToProps,
-  mapDispatchToProps,
-)(ConnectionLostDialog);
+export default ConnectionLostDialog;


marcus-oscarsson:
Well done !

ui/src/containers/ConnectionLostDialog.jsx
```diff
@@ -1,88 +1,78 @@
...
-      this.refs.countdown.innerHTML = this.seconds;
-    }
+  const [timerElapsed, setTimerElapsed] = useState(30);
 
-    if (this.seconds < 1) {
-      this.seconds = 0;
-      this.accept();
-    }
-  }
+  const [buttonLoading, setbuttonLoading] = useState(false);
```
axelboc: `setButtonLoading`

ui/src/containers/ConnectionLostDialog.jsx
```diff
@@ -1,88 +1,78 @@
...
-        </Modal.Footer>
-      </Modal>
-    );
+  function rejectHide() {
+    // if server still disconnect after 9 second show dialog again
+    if (serverIO.isServerDisconnect()) {
+      setTimeout(() => {
+        dispatch(showConnectionLostDialog(true));
+        setTimerElapsed(30);
+      }, 9000);
```
axelboc: Perhaps in `serverIO`, you could start a `setInterval` on line 126 (around the `dispatch` call of the `showConnectionLostDialog` action)? This would maintain the separation of concerns between `serverIO` and the connection lost dialog (which should only care about the Redux state `state.general.showConnectionLostDialog`)

> Don't forget to clean up the `setInterval` the same way as the `setTimeout`.

That being said, this feels like a hack to account for the fact that the dialog is dismissable...

ui/src/containers/ConnectionLostDialog.jsx
```diff
@@ -1,88 +1,78 @@
...
-    setTimeout(() => {
-      clearInterval(this.timerid);
-    }, this.seconds * 1000);
+  useEffect(() => {
+    if (show) {
+      const forceUpdateInterval = setInterval(intervalCb, 1000);
+      return () => clearInterval(forceUpdateInterval);
+    }
+    return null;
+  });
```
axelboc: This effect re-runs on every render because you don't pass a dependency array — i.e. you end up cleaning up and recreating `setInterval` every second. It's a bit convoluted, but it should look like this, roughly:

```suggestion
  useEffect(() => {
    if (!show) {
      return null;
    }
   
    const interval = setInterval(() => {
      setTimerElapsed(sec => sec - 1);
    }, 1000);
    
    return () => clearInterval(interval);
  }, []);
  
  useEffect(() => {
    if (timerElapsed === 1) {
      accept();
      setTimeout(() => setTimerElapsed(30), 1000);
    }
  }, [timerElapsed, accept]);
```

> You'd also need to wrap `accept` in a `useCallback`.

However, I'd prefer to remove this count down timer altogether, as mentioned.

axelboc (commented):
I think we need to rethink the UX of this a bit more to make the implementation simpler.

- First, should the dialog really be dismissable? If you dismiss it, you're in an invalid state, so you might as well reload the page.
- Since the countdown doesn't actually do anything, why show it at all? I definitely don't think the page should reload automatically after 1 min. Why not just replace the "Try again" button with a "Reload page" button?

With the above, we remove all the logic related to the timer, the fake loading state and the interval to show the dialog again if dismissed.

ui/src/containers/ConnectionLostDialog.jsx
marcus-oscarsson: Yes, I don't really know why we added this countdown timer. I also think a "Try again" button as @axelboc mentions above is nicer solution and makes things a bit simpler.

It would also be good if we could prevent the dialog from being closed as well.

jbflo: I guess the countdown was added to make the automatic page reload more intuitive for the user. This was initially implemented to handle server  and user  state synchronization   

Previously, remote users who left the screen open and returned later might see the queue still running, while the local contact reported it as finished. The auto force update with a countdown helped prevent this issue by ensuring synchronization after a timeout.

this is handle differently now.


 it seems showConnectionLostDialog is not so relevant now.

ui/src/containers/ConnectionLostDialog.jsx
jbflo: Removed .

ui/src/containers/ConnectionLostDialog.jsx
```diff
@@ -1,88 +1,38 @@
...
-    ),
-  };
+  return (
+    <Modal show={show}>
+      <Modal.Header>
+        <Modal.Title>Connection lost</Modal.Title>
+      </Modal.Header>
+      <Modal.Body>
+        Ooops ! There seems to be a problem with the internet connection, the
+        connection to the server was lost. Try realoding the page.
```
axelboc: Maybe a good time to rephrase this?

```suggestion
        Trying to reconnect now... If the issue persists, please check your Internet connection or reload the page.
```

The idea is to make it clear that the connection will be reestablished automatically if possible. It's to avoid people reloading the page right away, because if they reload before their internet connection is restored or before the server has restarted, they'll get an error page and have to try reloading again and again...

ui/src/containers/ConnectionLostDialog.jsx
```diff
@@ -1,88 +1,38 @@
...
-      this.seconds = 0;
-      this.accept();
-    }
-  }
-
-  accept() {
-    this.forceUpdate();
-  }
+export function ConnectionLostDialog() {
+  const [buttonLoading, setButtonLoading] = useState(false);
```
axelboc: I wouldn't even keep this state. If users click "Reload", we want to reload the page as quickly as possible.

ui/src/containers/ConnectionLostDialog.jsx
```diff
@@ -1,88 +1,38 @@
...
-          <Button onClick={this.accept}> Try again </Button>
-        </Modal.Footer>
-      </Modal>
-    );
+  function accept() {
+    // just a visual effect
+    setButtonLoading(true);
+    setTimeout(() => setButtonLoading(false), 500);
+    // Reload window / do a force state update maybe?
+    window.location.reload(true);
```
axelboc: Therefore, I would keep only this line and inline it in an arrow function in the JSX.

ui/src/containers/ConnectionLostDialog.jsx
```diff
@@ -1,88 +1,38 @@
...
+      </Modal.Header>
+      <Modal.Body>
+        Ooops ! There seems to be a problem with the internet connection, the
+        connection to the server was lost. Try realoding the page.
+      </Modal.Body>
+      <Modal.Footer>
+        {buttonLoading ? (
+          <Spinner animation="border" />
+        ) : (
+          <Button onClick={accept}> Reload </Button>
```
axelboc:
```suggestion
          <Button variant="outline-secondary" onClick={accept}>Reload</Button>
```

- More subtle button so people don't rush to click on it?
- Don't think the extra space around the label is needed but if you really want it, I'd prefer a Bootstrap class :wink:

axelboc (commented):
Thanks a lot for the well-needed clean-up!

ui/src/containers/ConnectionLostDialog.jsx
marcus-oscarsson: Not entirely certain that was why. I think we just tried to be informative, and it ended up to be misleading in stead. Anyways the suggested solution is nicer

axelboc: I guess showing the websocket reconnection timer would perhaps be more useful ... but definitely not worth the effort and complexity IMO :joy:

marcus-oscarsson: yes, that would be more correct. But a clear message saying that it will automatically reconnect is also fine :)

axelboc (approved):
Thanks!

PR #1551, opened 2025-02-13
Removed depricated `set-output` and using env var instead

marcus-oscarsson:
Removed depricated `set-output` and using env var instead. Also using `poetry version -s` instead of parsing `pyproject.toml`


fabcor-maxiv (approved):

fabcor-maxiv:
See also: <https://github.com/mxcube/mxcubecore/issues/1142>.

fabcor-maxiv:
Nice, looks like it did the job as expected. Git tag is here. Version has been updated in `pyproject.toml`.

PR #1552, opened 2025-02-13
Enforce modern coding style with `pyupgrade`

fabcor-maxiv:


fabcor-maxiv:
Here is an attempt at enforcing modern coding style with [_`pyupgrade`_](https://pypi.org/project/pyupgrade/).

I had a quick read through. In general this seems okay, most of the changes were expected. Some things we might want to have a closer look at:

* replacing `mock` from PyPI with `unittest.mock` from the standard library
* some shenanigans with dict comprehension [here](https://github.com/mxcube/mxcubeweb/pull/1552/files#diff-4db974c114b44f85ece8effdb39294773ec1c4dea9f4c58b1f339bdbd490d34a)

But really it is worth double-checking everything.

mxcubeweb/core/components/samplechanger.py
```diff
@@ -202,7 +201,7 @@ def mount_sample_clean_up(self, sample):
             current_queue = self.app.queue.queue_to_dict()
 
             if sample["location"] != "Manual":
-                msg = "Mounting sample: %s (%s)" % (
+                msg = "Mounting sample: {} ({})".format(
                     sample["location"],
                     sample.get("sampleName", ""),
                 )
```
dominikatrojanowska, line 207: why it didn't go further? 🤔 do we want to use fstring-s everywhere?

msg = f"Mounting sample: {sample['location']} ({ sample.get('sampleName', '')})")

mxcubeweb/core/components/user/usermanager.py
```diff
@@ -312,7 +311,7 @@ def login_info(self):
                 "useSSO": self.app.CONFIG.sso.USE_SSO,
             }
 
-            res["selectedProposal"] = "%s%s" % (
+            res["selectedProposal"] = "{}{}".format(
                 HWR.beamline.session.proposal_code,
                 HWR.beamline.session.proposal_number,
             )
```
dominikatrojanowska, line 317: as above - do we want ot get rid of .format?

mxcubeweb/core/util/convertutils.py
```diff
@@ -16,15 +16,10 @@ def convert_to_dict(ispyb_object):
...
-                            k,
-                            (convert_to_dict(x) if not isinstance(x, dict) else x),
-                        )
-                        for k, x in val.items()
-                    ]
-                )
+                val = {
+                    k: (convert_to_dict(x) if not isinstance(x, dict) else x)
+                    for k, x in val.items()
+                }
```
dominikatrojanowska, line 22: seems ok the inner part didn't change `(convert_to_dict(x) if not isinstance(x, dict) else x)`
there was a call of: `dict([(a, b), (c, d), ...])` so this shortcut seems fine

mxcubeweb/core/util/networkutils.py
```diff
@@ -161,15 +161,15 @@ def send_feedback(sender_data):
     _from = HWR.beamline.session.get_property("from_email", "")
 
     if not _from:
-        _from = "%s@%s" % (
+        _from = "{}@{}".format(
             local_user,
             HWR.beamline.session.get_property("email_extension", ""),
         )
```
dominikatrojanowska, line 167: again fstring vs .format

mxcubeweb/core/util/networkutils.py
```diff
@@ -161,15 +161,15 @@ def send_feedback(sender_data):
...
 
     # Sender information provided by user
     _sender = sender_data.get("sender", "")
     to = HWR.beamline.session.get_property("feedback_email", "") + ",%s" % _sender
-    subject = "[MX3 FEEDBACK] %s (%s) on %s" % (
+    subject = "[MX3 FEEDBACK] {} ({}) on {}".format(
         local_user,
         _sender,
         bl_name,
```
dominikatrojanowska, line 175: and again

dominikatrojanowska:
except these few comment it looks like a nice tool that made quite a job, already like it :)

marcus-oscarsson:
Looks good in general, however it did not seem to have done anything with the `__future__` imports. We only have one in `mxcubeweb` but I guess it should have been removed right ?

I'm not either a fan of the change to `str.format` seems to be more of an opinionated style change to me. It would have made more sense to move to an f-string in that case which seems to be the case for some strings so a little bit confused there. So largely agreeing with @dominikatrojanowska.

elmjag:
very nice, I like it!

fabcor-maxiv:
Yes, I thought/hoped this tool would be more aggressive, but I take any improvement in the right direction.

There are some more details on the [README](https://github.com/asottile/pyupgrade?tab=readme-ov-file). In particular:

* [f-strings](https://github.com/asottile/pyupgrade?tab=readme-ov-file#f-strings)
* [`__future__`](https://github.com/asottile/pyupgrade?tab=readme-ov-file#__future__-import-removal)

I agree that going from %-strings to formatted strings does not feel like a great step forward, rather a step to the side. I will look around if I find a tool that does the migration to f-strings. Maybe ruff...

I am always a bit reluctant to using ruff, but maybe we should just go straight to ruff.

marcus-oscarsson:
We could try `ruff` and see what the differences would be. I imagine that it would change quite big parts of the code base ? It looks like `pyupgrade` does reasonable things the only strange thing seems to be the f-string handling. 

The change to '.format' does not seem to be that numerous for `mxcubeweb` so they can maybe either be reverted or changed manually to an f-string. However, I think it might be quite a different story for `mxcubecore`

fabcor-maxiv:
I have started an attempt at migrating to ruff.

In order to ease the churn (in particular for those who maintain a fork), I want to proceed in steps. First step is to migrate the existing setup to ruff without changing anything in the Python files. Then in later steps start introducing fixes for the rule violations uncovered, probably starting with the things that ruff can fix automatically itself.

Exception for the formatter. I will take in the formatting changes straight from the start. This should be minimal since we were using black already.

fabcor-maxiv:
Attempt to use `ruff` here: <https://github.com/mxcube/mxcubeweb/pull/1555>.

I want to keep this PR open as draft, in case we do not want to move forward with the `ruff` solution.

marcus-oscarsson:
I guess it means that we can drop this one then ?

fabcor-maxiv:
> I guess it means that we can drop this one then ?

Yes. But also no, because we have not enabled the [`UP` rules](https://docs.astral.sh/ruff/rules/#pyupgrade-up) in the Ruff config yet.

But OK, let's close it, it does not make much difference...

PR #1554, opened 2025-02-14
 Resetting queue and sample list when session changes

marcus-oscarsson:
Resetting queue and sample list when session changes. The recent changes triggered some logic that cleared the queue when the user in-control logs out. This scenario happened very rarely in the past as control was passed automatically and the beamline instance always was running, and this logic was simply not executed.

Control is no longer passed automatically and we don't any longer have any requirement on having a MXCuBE web client open to run the core of the application.

### DIFF
--- a/mxcubeweb/core/components/lims.py
+++ b/mxcubeweb/core/components/lims.py
@@ -219,12 +219,9 @@ class Lims(ComponentBase):
             # Re-initialize the samplelist
             self.app.lims.init_sample_list()
 
-            # If a sample is mounted (and not already marked as such),
-            # get sample changer contents and add mounted sample to the queue
-            address = self.app.sample_changer.get_loaded_sample()
-            if not self.app.sample_changer.get_current_sample() and address:
-                self.app.sample_changer.get_sample_list()
-                self.app.server.emit("update_queue", {}, namespace="/hwr")
+            # Get sample list and send update to client
+            self.app.sample_changer.get_sample_list()
+            self.app.server.emit("update_queue", {}, namespace="/hwr")
 
             HWR.beamline.session.proposal_code = session.code
             HWR.beamline.session.proposal_number = session.number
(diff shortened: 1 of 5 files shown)


marcus-oscarsson:
I'd be happy if someone had a quick look :)

marcus-oscarsson:
:), thanks @walesch-yan

walesch-yan:
Looks good to me now:)

walesch-yan (approved):

mxcubeweb/core/components/lims.py
```diff
@@ -206,11 +206,30 @@ def select_session(self, session_id: str) -> bool:
...
-        HWR.beamline.session.proposal_code = session.code
-        HWR.beamline.session.proposal_number = session.number
-        HWR.beamline.session.session_id = HWR.beamline.lims.get_session_id()
-        HWR.beamline.session.proposal_id = session.proposal_id
-        HWR.beamline.session.set_session_start_date(session.start_date)
+        if HWR.beamline.session.session_id != HWR.beamline.lims.get_session_id():
+            # ruff: noqa: G004
+            logging.getLogger("MX3.HWR").info(
+                f"[LIMS] New session, clearing queue and sample list for {session.code}{session.number}"
+            )
```
fabcor-maxiv, line 212: I think we should probably commit to following what the linter says, even if it is annoying at first. If there is very strong and shared sentiment that we want to go against one particular linter rule then we should disable it globally, so that we have a consistent code base.

As far as I can tell, this would be the preferred way of writing this:

```suggestion
            logging.getLogger("MX3.HWR").info(
                "[LIMS] New session, clearing queue and sample list for %s%s",
                session.code,
                session.number,
            )
```

If we want to have the policy/guideline that we use f-strings for logging then we should look at the ruff documentation if there is a way to enforce it. I guess it could mean disabling `G004` globally, but that would still mean that there are two ways of doing the same thing: %-strings and f-strings. And having two ways of doing the same thing is exactly what we should avoid, or at least that is exactly what I want to avoid.

PR #1555, opened 2025-02-14
Use `ruff` as Python code formatter and linter

fabcor-maxiv:
Use `ruff` to replace:

* `autoflake`
* `black`
* `isort`
* `flake8`

Also reorganize the `pre-commit` hooks in a more logical order.


fabcor-maxiv:
It did only 2 formatting changes: adding one empty line in some place and removing one empty line in another place.

It is configured to enable all linting rules, but disabling just enough rules for specific files so that we do not have to change anything in the Python code (just yet).

It removes all the other tools we do not need any longer. All in all, the pre-commit hooks should also be faster (because Rust and because less hooks altogether).

Once this is merged, we can start with the next steps. Which would be enabling the rules that are currently disabled and do the required fixes. We might want to start by enabling the rules that can be fixed automatically by ruff itself, and then some other rules that are easy to fix manually with extremely little risk of breaking anything.

fabcor-maxiv:
As expected, this made it very clear that we have big consistency issues regarding logging, and that our handling of exceptions is abysmally bad. Then there are a whole lot of things that will be trivial to fix manually.

fabcor-maxiv:
I am also thinking, that this is only MXCuBE-Web and that the situation in MXCuBE-Core is probably much "worse". I do not know how quick the progress will be on that initiative. I can probably handle a big chunk of MXCuBE-Web on my own. And then maybe it could be a good idea to do some (online) code camp and/or sprint where everyone tries to pitch in and fix the linting issues in MXCuBE-Core.

fabcor-maxiv:
~I set it back to "draft". I need to update the contribution guidelines.~ DONE

marcus-oscarsson:
Ok, lets try.

> Once this is merged, we can start with the next steps. Which would be enabling the rules that are currently disabled and do the required fixes. We might want to start by enabling the rules that can be fixed automatically by ruff itself, and then some other rules that are easy to fix manually with extremely little risk of breaking anything.

I guess it will depend a bit on the rules :), but sounds good.

Thanks again, could you re-base please ?

fabcor-maxiv:
> Thanks again, could you re-base please ?

Done

fabcor-maxiv:
> Ok, lets try.
> 
> > Once this is merged, we can start with the next steps. Which would be enabling the rules that are currently disabled and do the required fixes. We might want to start by enabling the rules that can be fixed automatically by ruff itself, and then some other rules that are easy to fix manually with extremely little risk of breaking anything.
> 
> I guess it will depend a bit on the rules :), but sounds good.

I will try to be smart about it, to make it as smooth and uncontroversial as possible.

My personal attitude towards this is "do whatever the linter says", with few exceptions on a case-by-case basis for when I really thought it through and came to the conclusion that it is wiser to bypass the linter for that specific case.

But for MXCuBE, we can do it more conservatively, bit by bit.

marcus-oscarsson:
> But for MXCuBE, we can do it more conservatively, bit by bit.

I guess it will depend on the amount of things we need to change. Id personally be happy to just get it over with if its not too much and anything really strange.

PR #1558, opened 2025-02-17
Fix presence of package initializers

fabcor-maxiv:
Add missing package initializers `__init__.py`.
This fixes the violations of the `INP001` ruff rule.

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -20,7 +20,7 @@ select = [
 "docs/conf.py" = [
     "A001",
     "DTZ002",
-    "INP001",
+    "INP001",  # `implicit-namespace-package`: `conf.py` is not meant to be imported
     "PLR2044",
 ]
 "mxcubeweb/__init__.py" = [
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
Good thing :). thanks

fabcor-maxiv:
Yes, among other things, if I recall correctly, I think that Sphinx docs builder was either complaining about those missing `__init__.py` or something like that.

fabcor-maxiv:
In principle we should also add `__init__.py` in `/test/`:

> It is highly recommended to arrange your test modules as packages by adding `__init__.py` files to your directories containing tests. This will make the tests part of a proper Python package, allowing pytest to resolve their full name (for example `tests.core.test_core` for `test_core.py` inside the `tests.core` package).

-- <https://docs.pytest.org/en/stable/explanation/pythonpath.html>

Until yesterday I was convinced it was the other way around, that the recommendation was to not have package initializers in tests, and I was about to remove them in MXCuBE. But now I know...

I did not add `/test/__init__.py` though, because then it breaks the tests. Indeed, it would require changing the imports in the code of the tests, so I considered that it should be done in a separate pull request if we really want to do this change.

PR #1559, opened 2025-02-18
Now sends loginType to anonymous users in login_info

mockoocy:
At MAX IV we don't use proposal-based login nor SSO. This PR makes it so, that in the case of user log in appropriate text shows up ("Sign in").

Also did small cleanup to `UserManager.login_info` - now uses early return.

### DIFF
--- a/ui/src/components/LoginForm/LoginForm.jsx
+++ b/ui/src/components/LoginForm/LoginForm.jsx
@@ -11,6 +11,19 @@ import { logIn } from '../../actions/login';
 
 const SSO_PATH = '/mxcube/api/v0.1/login/ssologin';
 
+/**
+ * @param loginType {'User' | 'Proposal'} - login type; 'User' for User-type login, 'Proposal' otherwise.
+ * @param useSSO {boolean} - true if SSO is enabled
+ *
+ * @returns {string} - text used in the sign-in button
+ */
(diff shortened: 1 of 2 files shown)


ui/src/components/LoginForm/LoginForm.jsx
```diff
@@ -11,6 +11,14 @@ import { logIn } from '../../actions/login';
 
 const SSO_PATH = '/mxcube/api/v0.1/login/ssologin';
 
+/** */
```
axelboc: Empty comment here (just remove)

axelboc (approved):
LGTM

elmjag (approved):

ui/src/components/LoginForm/LoginForm.jsx
mockoocy: Something must have distracted me when writing a JSDoc there :laughing: 
Added a JSDoc instead of removing the comment.

axelboc: Haha you didn't have to, the name of the function is clear, the code it contains as well and is concise... and it's an internal function, so really not needed in this case :wink: but it's done, so no worries :grin:

marcus-oscarsson:
Thanks :)

PR #1560, opened 2025-02-19
Ignore Ruff linter rule `COM812`

fabcor-maxiv:
Ruff recommends to ignore this linter rule when using its formatter.

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -7,6 +7,7 @@ target-version = "py310"
 [lint]
 ignore = [
     "ANN",  # flake8-annotations: ignored, too many hits (1322)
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
OK, if thats what they say :)

fabcor-maxiv:
> OK, if thats what they say :)

Oh, right... I should have added a link to the reference: <https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules>

There are more rules they recommend to ignore but `COM812` was the only one throwing a warning for us.

marcus-oscarsson:
Nicer with the link, :) but i take your word for it ;)

PR #1565, opened 2025-02-21
Remove unnecessary usage of dict as kwargs

fabcor-maxiv:
This is enforced and fixed by Ruff's rule `PIE804`: <https://docs.astral.sh/ruff/rules/unnecessary-dict-kwargs/>


marcus-oscarsson:
Yes, makes sense

marcus-oscarsson:
Could you re-base and then merge please :)

PR #1566, opened 2025-02-21
Move all usage of  react-copy-to-clipboard to a single component

jbflo:
- Create a generic component from react-copy-to-clipboard
- Move usage of copy-to-clipboard to this generic component
-  small related cleanup 

[Screencast from 2025-02-25 13-53-33.webm](https://github.com/user-attachments/assets/48524fca-a03c-4794-bef0-f80674b497c8)

[Screencast from 2025-02-25 13-55-12.webm](https://github.com/user-attachments/assets/6a0bb26d-ca6f-4733-bf77-5890a03bde78)

### DIFF
--- a/ui/src/components/CopyToClipboard/CopyToClipboard.jsx
+++ b/ui/src/components/CopyToClipboard/CopyToClipboard.jsx
@@ -1,0 +1,39 @@
+import React, { useState } from 'react';
+import { Button } from 'react-bootstrap';
+import { CopyToClipboard as TriggerCopyToClipboard } from 'react-copy-to-clipboard';
+import { MdContentCopy } from 'react-icons/md';
+
+import TooltipTrigger from '../TooltipTrigger';
+import styles from './CopyToClipboard.module.css';
+
+export default function CopyToClipboard(props) {
+  const { id, tittle, text, iconColor = 'black' } = props;
+
+  const [copied, setCopied] = useState(false);
+
+  function onCopy() {
+    setCopied(true);
+    setTimeout(() => setCopied(false), 1100);
+  }
+
+  return (
+    <Button className={styles.btnContent} variant="unknown">
+      <TooltipTrigger
+        placement={copied ? 'auto' : 'right'}
+        rootClose={false}
+        id={id}
+        tooltipContent={
+          copied ? `${tittle} Copied` : `Copy ${tittle} to Clipboard`
+        }
+      >
+        <TriggerCopyToClipboard
+          onCopy={onCopy}
+          className={styles.link}
+          text={text}
+        >
+          <MdContentCopy color={iconColor} size="1em" />
+        </TriggerCopyToClipboard>
+      </TooltipTrigger>
+    </Button>
+  );
+}
--- a/ui/src/components/SampleGrid/SampleGridTableItem.jsx
+++ b/ui/src/components/SampleGrid/SampleGridTableItem.jsx
@@ -1,4 +1,3 @@
-/* eslint-disable react/jsx-handler-names */
 import React from 'react';
 import {
   ListGroup,
@@ -9,12 +8,11 @@ import {
 } from 'react-bootstrap';
 import cx from 'classnames';
 
-import { CopyToClipboard } from 'react-copy-to-clipboard';
+import CopyToClipboard from '../../components/CopyToClipboard/CopyToClipboard';
 
 import { isCollected } from '../../constants';
 
 import { BsSquare, BsCheck2Square } from 'react-icons/bs';
-import { MdContentCopy } from 'react-icons/md';
 import './SampleGridTable.css';
 import TooltipTrigger from '../TooltipTrigger';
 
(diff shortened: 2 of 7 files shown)


ui/src/components/CopyToClipboard/CopyToClipboard.jsx
```diff
@@ -0,0 +1,35 @@
...
+  const [copied, setCopied] = useState(false);
+
+  function onCopy() {
+    setCopied(true);
+    setTimeout(() => setCopied(false), 1100);
+  }
+
+  return (
+    <TooltipTrigger
+      placement={copied ? 'auto' : 'right'}
```
jbflo: This ensure the Tooltip re-render and assure repositioning and width adjust

ui/src/components/CopyToClipboard/CopyToClipboard.jsx
```diff
@@ -0,0 +1,35 @@
+import React, { useState } from 'react';
+import { Button } from 'react-bootstrap';
+import CopyToClipboard from 'react-copy-to-clipboard';
```
marcus-oscarsson: The file is called `CopyToClipboard` but the function is called `CopyToClipb` is that by purpose it would make somewhat more sense if it was called `CopyToClipboard`

jbflo: yes CopyToClipboard is also the default import  name of the react library
Maybe i can do import As

marcus-oscarsson: ah, yes did not think about that right away. perhaps nicer to either `import as` or rename the component

jbflo: I do import as, as require one file change

ui/src/components/SampleGrid/SampleGridTableItem.jsx
```diff
@@ -229,6 +198,11 @@ export class SampleGridTableItem extends React.Component {
                 {this.props.sampleData.location} {this.currentSampleText()}
               </div>
             </div>
+            <CopyToClipboard
+              text={this.sampleDisplayName()}
+              tittle="Sample Name"
+              id={this.sampleDisplayName()}
```
axelboc: Using the display name as the ID is not ideal. Just come up with a short unique ID maybe?

ui/src/components/CopyToClipboard/CopyToClipboard.jsx
```diff
@@ -0,0 +1,39 @@
+import React, { useState } from 'react';
+import { Button } from 'react-bootstrap';
+import { CopyToClipboard as TriggerCopyToClipboard } from 'react-copy-to-clipboard';
+import { MdContentCopy } from 'react-icons/md';
+
+import TooltipTrigger from '../TooltipTrigger';
+import styles from './CopyToClipboard.module.css';
+
+export default function CopyToClipboard(props) {
+  const { id, tittle, text, iconColor = 'black' } = props;
```
axelboc, line 10: I think the naming can be made clearer:

- `triggerText` instead of `text`?
- `tooltipLabel` instead of `title`?

Also, `iconColor` doesn't need to be a prop for now.

ui/src/components/CopyToClipboard/CopyToClipboard.jsx
```diff
@@ -0,0 +1,39 @@
...
+    setTimeout(() => setCopied(false), 1100);
+  }
+
+  return (
+    <TooltipTrigger
+      placement={copied ? 'auto' : 'right'}
+      rootClose={false}
+      id={id}
+      tooltipContent={
+        copied ? `${tittle} Copied` : `Copy ${tittle} to Clipboard`
```
axelboc: I wonder if it's really necessary to show this `title` in the tooltip. Would "Copy to clipboard" (lowercase "clipboard" if possible) and "Copied" suffice?

axelboc (commented):
Any chance you could please share a little screenshot or screen recording (maybe also one that shows what it looked like before for comparison, since there seem to be styling changes)?

ui/src/components/CopyToClipboard/CopyToClipboard.jsx
jbflo: yeah not necessary, but a bit more informative, just  Copied feel more advance user. 
we have this component on another application and some users prefer details.

ui/src/components/SampleGrid/SampleGridTableItem.jsx
jbflo: Right : just check and "sampleDisplayName" is being use somewhere-else as ID

jbflo:
> Any chance you could please share a little screenshot or screen recording (maybe also one that shows what it looked like before for comparison, since there seem to be styling changes)?

Yes, i am using the nice TooltipTrigger you made i guess  and i thought people see it already. basically the old code was css hardcoded triggering the tool-tip now and i remove them  . 

@axelboc  Are you using an chrome Extension  for  screen recording ?

ui/src/components/CopyToClipboard/CopyToClipboard.jsx
jbflo: tooltipLabel is good, but text is i think  clearer than triggerText, it's align with the react library. and  trigger is a bit confusing in my opinion.   

i intend to use this component somewhere  else  on an image component and the default black color may not fit, will avoid the come back modifying the style.

axelboc:
Thanks! For the screen recording, I use Ubuntu 24's built-in feature "Record a screencast interactively". I had to define a shortcut for it, I think:

image

Note that you can swap between screenshot/screencast in the UI, so you don't really need two separate shortcuts. It gives you the ability to show/hide the mouse cursor, draw selections, etc. It's pretty cool.

Before, I used to use [Peek](https://github.com/phw/peek).

marcus-oscarsson:
Nice thanks guys !

PR #1569, opened 2025-02-24
Added yellow highlight to the sessions

### LINKED ISSUE #1568: Session selection improvement
The current implement is prone to error when selecting the right proposal. This is due to the fact that multiple time slots can be allocated to the same proposal, rescheduling, moving experiments etc...

Image

The feedback received with a proposal for improvement is the following:

> 
> [...] the current MXCuBE session selection box is clearly allowing mistakes to happen as since the meeting, we have already got 3 sessions being selected by mistake, leading data to be saved in the wrong directory.
> 
>  
> 
> Can I suggest that we try the highlighting option we briefly discussed last Monday that would consist in highlighting the sessions from the current day + Day-1 and Day+1?
> 
> If it doesn’t help, we will think about another option.

antolinos:
Sessions that should be more selectable than others because they are scheduled close to the present can be highlighted.
image

### DIFF
--- a/ui/src/components/LoginForm/SelectProposal.jsx
+++ b/ui/src/components/LoginForm/SelectProposal.jsx
@@ -77,7 +77,7 @@ function SelectProposal() {
           onChange={(evt) => setFilter(evt.target.value)}
         />
 
-        <Tabs id="scheduled-tab" defaultActiveKey="active">
+        <Tabs id="session-tabs" defaultActiveKey="active">
           <Tab eventKey="active" title={`Active (${scheduledSessions.length})`}>
             <div className={styles.table}>
               <SessionTable
--- a/ui/src/components/LoginForm/SelectProposal.jsx
+++ b/ui/src/components/LoginForm/SelectProposal.jsx
@@ -77,7 +77,7 @@ function SelectProposal() {
           onChange={(evt) => setFilter(evt.target.value)}
         />
 
-        <Tabs id="scheduled-tab" defaultActiveKey="scheduled">
+        <Tabs id="scheduled-tab" defaultActiveKey="active">
           <Tab eventKey="active" title={`Active (${scheduledSessions.length})`}>
             <div className={styles.table}>
               <SessionTable
--- a/ui/src/components/LoginForm/SessionTable.jsx
+++ b/ui/src/components/LoginForm/SessionTable.jsx
@@ -3,18 +3,7 @@ import { Table } from 'react-bootstrap';
 import { LuExternalLink } from 'react-icons/lu';
 import SessionDateTime from './SessionDateTime';
 import styles from './SessionTable.module.css';
-import { isDateInRange } from './utils';
-
-const HIGHLIGHTED_TIME_RANGE = 1;
-
-function isAroundNow(session) {
-  const { actual_start_date, actual_end_date } = session;
-  return isDateInRange(
-    actual_start_date,
-    actual_end_date,
-    HIGHLIGHTED_TIME_RANGE,
-  );
-}
+import { isAroundNow } from './utils';
 
 export default function SessionTable(props) {
   const {
--- a/ui/src/components/LoginForm/SessionTable.jsx
+++ b/ui/src/components/LoginForm/SessionTable.jsx
@@ -1,38 +1,20 @@
 import React from 'react';
 import { Table } from 'react-bootstrap';
 import { LuExternalLink } from 'react-icons/lu';
-
 import SessionDateTime from './SessionDateTime';
 import styles from './SessionTable.module.css';
-
-function isDateInRange(actualStartDate, actualEndDate, rangeInDays) {
-  // Helper function to parse YYYYMMDD format into Date object
-  function parseDate(yyyymmdd) {
-    const year = yyyymmdd.slice(0, 4);
-    const month = yyyymmdd.slice(4, 6);
-    const day = yyyymmdd.slice(6, 8);
-    return new Date(`${year}-${month}-${day}`);
-  }
-
-  const startDate = parseDate(actualStartDate);
-  const endDate = parseDate(actualEndDate);
-
-  // Get today's date without time
-  const today = new Date();
-  today.setHours(0, 0, 0, 0);
-
-  // Calculate the lower and upper bounds based on the rangeInDays
-  const lowerBound = new Date(today);
-  lowerBound.setDate(today.getDate() - rangeInDays);
-
-  const upperBound = new Date(today);
-  upperBound.setDate(today.getDate() + rangeInDays);
-
-  // Check if any date in the range (today ± rangeInDays) falls within the given date range
-  return lowerBound <= endDate && upperBound >= startDate;
-}
+import { isDateInRange } from './utils';
 
 const HIGHLIGHTED_TIME_RANGE = 1;
+
+function isAroundNow(session) {
+  const { actual_start_date, actual_end_date } = session;
+  return isDateInRange(
+    actual_start_date,
+    actual_end_date,
+    HIGHLIGHTED_TIME_RANGE,
+  );
+}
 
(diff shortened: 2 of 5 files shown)


ui/src/components/LoginForm/SelectProposal.jsx
```diff
@@ -77,7 +77,7 @@ function SelectProposal() {
           onChange={(evt) => setFilter(evt.target.value)}
         />
 
-        <Tabs id="scheduled-tab" defaultActiveKey="scheduled">
```
axelboc, line 80: Might be worth changing the ID as well? To be double checked, but I think it's just for accessibility.

ui/src/components/LoginForm/SessionTable.jsx
```diff
@@ -35,7 +65,11 @@ export default function SessionTable(props) {
         {sessions.map((session) => (
           <tr
             key={session.session_id}
-            className={styles.row}
+            className={
+              isDateInRange(session.actual_start_date, session.actual_end_date, HIGHLIGHTED_TIME_RANGE)
+                ? styles.row_now
+                : styles.row
+            }
```
axelboc:
```html
data-highlight={isDateInRange(...) || undefined}
```

```css
.row[data-highlight] {
  ...
}
```

ui/src/components/LoginForm/SessionTable.jsx
```diff
@@ -5,6 +5,35 @@ import { LuExternalLink } from 'react-icons/lu';
 import SessionDateTime from './SessionDateTime';
 import styles from './SessionTable.module.css';
 
+function isDateInRange(actualStartDate, actualEndDate, rangeInDays) {
```
axelboc: Until this needs to be generic, I wouldn't bother making it so (as it might not ever need to be generic):

```js
function isAroundNow(session) {
  const startDate = parseDate(session.actual_start_date);
  const endDate = parseDate(session.actual_end_date);

  ...
}
```

Could you maybe also move this function into a `util.js` file in the same folder?

Arguably, `SelectProposal.jsx`, `SessionTable.jsx`, etc. and this `util.js` file should maybe be moved into a `SessionProposal` folder. They're not really part of the `LoginForm`.

ui/src/components/LoginForm/SessionTable.jsx
```diff
@@ -5,6 +5,35 @@ import { LuExternalLink } from 'react-icons/lu';
...
+  // Helper function to parse YYYYMMDD format into Date object
+  function parseDate(yyyymmdd) {
+    const year = yyyymmdd.slice(0, 4);
+    const month = yyyymmdd.slice(4, 6);
+    const day = yyyymmdd.slice(6, 8);
+    return new Date(`${year}-${month}-${day}`);
+  }
```
axelboc: This should be extracted to `util.js` as a separate function since it doesn't need to be a closure.

That being said, I'm surprised that you don't have access to a more normalised date/time ISO string that you could just pass to the `Date` constructor.

Moreover, I see on the screenshot that you have access to the scheduled time, not just the date. Wouldn't it be more accurate to include the time in the calculation (plus/minus 24h from now)?

ui/src/components/LoginForm/SessionTable.jsx
```diff
@@ -5,6 +5,35 @@ import { LuExternalLink } from 'react-icons/lu';
...
+  const startDate = parseDate(actualStartDate);
+  const endDate = parseDate(actualEndDate);
+
+  // Get today's date without time
+  const today = new Date();
+  today.setHours(0, 0, 0, 0);
+
+  // Calculate the lower and upper bounds based on the rangeInDays
+  const lowerBound = new Date(today);
+  lowerBound.setDate(today.getDate() - rangeInDays);
```
axelboc: `getDate()` returns the day number within the month, so it could lead to a session from the month before/after to be highlighted. Even if unlikely, maybe a simpler way would be to perform all the calculations with timestamps rather than date objects. For instance, assuming you include the time:

```js
const oneDay = 24 * 60 * 1000
const now = Date.now();

return startDate.getTime() <= now + oneDay && endDate.getTime() >= now - oneDay;
```

antolinos: I understood that JavaScript handles month overflow correctly when using setDate()

```
const today = new Date();
today.setHours(0, 0, 0, 0);
> 1740438000000

const lowerBound = new Date(today);
rangeInDays = 7

lowerBound.setDate(today.getDate() + rangeInDays);
> 1741042800000

lowerBound
> Date Tue Mar 04 2025 00:00:00 GMT+0100 (Central European Standard Time)

today
> Date Tue Feb 25 2025 00:00:00 GMT+0100 (Central European Standard Time)
```

ui/src/components/LoginForm/SessionTable.jsx
antolinos: Extracted to utils.js as there is a class called  like that ( instead of util.js)

They do not need 24h but the ~day before/after. I will check with them but I presume is fine

ui/src/components/LoginForm/SessionTable.jsx
antolinos: > Arguably, SelectProposal.jsx, SessionTable.jsx, etc. and this util.js file should maybe be moved into a SessionProposal folder. They're not really part of the LoginForm.

Completely agree. But it is probably to be addressed in a separated issue as it is beyond the current issue.
https://github.com/mxcube/mxcubeweb/issues/1568

ui/src/components/LoginForm/SessionTable.jsx
axelboc: Riiight, okay, thanks! Phew, date manipulation always hurts my brain.

ui/src/components/LoginForm/SessionTable.jsx
```diff
@@ -1,9 +1,20 @@
...
+const HIGHLIGHTED_TIME_RANGE = 1;
+
+function isAroundNow(session) {
+  const { actual_start_date, actual_end_date } = session;
+  return isDateInRange(
+    actual_start_date,
+    actual_end_date,
+    HIGHLIGHTED_TIME_RANGE,
+  );
+}
```
axelboc: My comment about not needing a generic `isDateInRange` still stands. In `utils.js`:

```js
function isAroundNow(session) {
  const startDate = parseDate(session.actual_start_date);
  const endDate = parseDate(session.actual_end_date);
  ...
  lowerBound.setDate(today.getDate() - 1); // or keep the `HIGHLIGHTED_TIME_RANGE` constant if you prefer, but use it here directly
  ...
  upperBound.setDate(today.getDate() + 1);
}
```

ui/src/components/LoginForm/utils.js
```diff
@@ -0,0 +1,26 @@
+function parseDate(yyyymmdd) {
+  const year = yyyymmdd.slice(0, 4);
+  const month = yyyymmdd.slice(4, 6);
+  const day = yyyymmdd.slice(6, 8);
+  return new Date(`${year}-${month}-${day}`);
+}
+
+export function isDateInRange(startDateStr, endDateStr, rangeInDays) {
+  // Helper function to parse YYYYMMDD format into Date object
```
axelboc: Should be moved or removed (I think removed is fine, as the function `parseDate` is clear enough IMO).

axelboc (approved):

marcus-oscarsson:
How about a light green color instead of the yellow ?, and perhaps with a legend to explain what it means. I understand it as, these are the sessions you are more likely/meant to be using ?.

Could we otherwise imagine an alternative solutions where the sessions we are "meant" to be using have normal apparence and the others like lighter color and change color once selected

marcus-oscarsson:
Closing this for now, I hope thats alright @antolinos

Issue #1567, opened 2025-02-24
Ruff linting and formating


marcus-oscarsson:
We had a discussion regarding one of the linting rule G004 in PR https://github.com/mxcube/mxcubeweb/pull/1554. Its actually a Flake-8 rule that we have not been using. The rule means that we should not use `f-strings` when logging. The main argument seems to be that using `f-strings` causes unnecessary formatting for log messages that may never be displayed. The same reasoning would apply to `%-strings` (as far as I understand). The `G` category of rules are explained here: https://docs.astral.sh/ruff/rules/#flake8-logging-format-g.

It also means that we need to update virtually all logging statements, to pass the data to be formatted as the `extra` arguments of `log`. My question is should we adhere to this rule or should we allow other ways of formatting strings ?

Something that also surprised me is that when I run a `pre-commit run --all-files` I don't get any warning for those instances where we are using for instance `f-strings`. Are there different settings between staged changes and already committed ?

fabcor-maxiv:
These are the "bad" ways (according to Ruff):

```python
# See rule `G001` <https://docs.astral.sh/ruff/rules/logging-string-format/>
logging.info("{} - Something happened".format(user))

# See rule `G002` <https://docs.astral.sh/ruff/rules/logging-percent-format/>
logging.info("%s - Something happened" % user)

# See rule `G003` <https://docs.astral.sh/ruff/rules/logging-string-concat/>
logging.info(user + " - Something happened")

# See rule `G004` <https://docs.astral.sh/ruff/rules/logging-f-string/>
logging.info(f"{user} - Something happened")
```

These are the "good" ways:

```python
logging.info("Something happened", extra=dict(user_id=user))
logging.info("%s - Something happened", user)
```

axelboc:
> Using f-strings to format a logging message requires that Python eagerly format the string, even if the logging statement is never executed (e.g., if the log level is above the level of the logging statement), whereas using the extra keyword argument defers formatting until required.

Seems like a stretch to me... One has to find a balance between performance and developer experience... `f-strings` are so much easier to comprehend and to write! They make the code a lot more readable too.

fabcor-maxiv:
If we follow Ruff, we would have two styles of string interpolation:

* `logging.info("%s - Something happened", user)` for logging and only logging
* f-strings for everything else

For me, it is a very good compromise. Linter will tell us if we use an f-string in logging by mistake. I would vote for following what Ruff recommends.


---

[...]

PR #1571, opened 2025-02-25
remove stack_traces from responses

walesch-yan:
This PR removes the stack traces from different responses in the `routes` folder. (see discussion about stack traces in responses #1393)

Instead the traces are logged to `MX3.HWR` following the ruff style discussed in #1567 

Furthermore, it fixes the logger from `HWR.MX3` to MX3.HWR` in a few places.


mxcubeweb/routes/samplechanger.py
```diff
@@ -1,3 +1,5 @@
+import logging
+
```
fabcor-maxiv, line 2: Also add something along the lines of:

```python
logger = logging.getLogger("MX3.HWR")
```

and then reuse throughout the module.

This comment is also valid for the other modules obviously. I comment here because I noticed that the logging is entirely new. Maybe there are more modules in this pull request where logging is entirely new.

walesch-yan: I'm a little hesitant to add this, since we are using logging in various other places with inline `getLogger` and I think the main reason for this is readability, as we have multiple different loggers in the applications, it helps to know, which one is used in every line. 

However, I'm also open to change this, but I think we should stay consistent and change this then for every file in the project that uses logging. In general, I guess we would need a complete revisit of the routes as well, as we  handle them in many different ways across the application.

marcus-oscarsson: @walesch-yan, indeed sounds good. It should in that case be done in one consistent way. Perhaps not even defining the logger at the top of each file but more on `App` level.

We have some ideas on how to rework the route handling mostly to get a more consistent handling of input, return data, error codes and so on. But at some point of this work logging could also become interesting to look at. We have to work on our ideas a bit more  but then we will get back to you guys as well and we can discuss what to do. Perhaps in our next `mxcubeweb` meeting

fabcor-maxiv: From my point of view it can be done now, module by module, but as you prefer.

ruff.toml
fabcor-maxiv, line 1: Thanks for working on the removal of ignores!

walesch-yan: No Problem! I was changing the log messages anyway in these files:).

walesch-yan:
Added the `escape` calls where necessary

PR #1575, opened 2025-02-27
Regenerate violations ignore rules in `ruff.toml`

fabcor-maxiv:
The original list was created by hand. Some mistakes were introduced. This version is mostly machine generated with some minor manual edits.

Also the `UP` ignores are now split.
The intention was to handle all `UP` violations at once. But this would be much more complex than initially thought. Handling the `UP` violations separately will be easier.


walesch-yan (approved):

marcus-oscarsson:
thanks guys

PR #1576, opened 2025-02-27
Fix Ruff violations from the `UP` group

fabcor-maxiv:
Fix some violations from the "pyupgrade `UP`" Ruff linter rules group:

* `UP006`
    * "Use `{to}` instead of `{from}` for type annotation"
    * <https://docs.astral.sh/ruff/rules/non-pep585-annotation/>
* `UP007`
    * "Use `X | Y` for type annotations"
    * <https://docs.astral.sh/ruff/rules/non-pep604-annotation-union/>
* `UP008`
    * "Use `super()` instead of `super(__class__, self)`"
    * <https://docs.astral.sh/ruff/rules/super-call-with-parameters/>
* `UP009`
    * "UTF-8 encoding declaration is unnecessary"
    * <https://docs.astral.sh/ruff/rules/utf8-encoding-declaration/>
* `UP015`
    * "Unnecessary mode argument"
    * <https://docs.astral.sh/ruff/rules/redundant-open-modes/>
* `UP026`
    * "`mock` is deprecated, use `unittest.mock`"
    * <https://docs.astral.sh/ruff/rules/deprecated-mock-import/>
* `UP035`
    * "Import from `{target}` instead: `{names}`"
    * <https://docs.astral.sh/ruff/rules/deprecated-import/>

These were all fixed automatically by Ruff itself.

The remaining violations in the "pyupgrade `UP`" rules group are all related to "string interpolation".

### DIFF
--- a/mxcubeweb/__init__.py
+++ b/mxcubeweb/__init__.py
@@ -9,8 +9,8 @@ import os  # noqa: E402
 import sys  # noqa: E402
 import traceback  # noqa: E402
 from pathlib import Path  # noqa: E402
+from unittest import mock  # noqa: E402
 
-import mock  # noqa: E402
 from mxcubecore import HardwareRepository as HWR  # noqa: E402
 
 from mxcubeweb.app import MXCUBEApplication as mxcube  # noqa: E402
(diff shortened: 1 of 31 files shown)


marcus-oscarsson:
looks good, thanks !

PR #1579, opened 2025-02-27
Fix Ruff "flake8-errmsg `EM`" rule violations

fabcor-maxiv:
Fix Ruff rule violations from the "flake8-errmsg `EM`" group:

* `EM101`
    * "Exception must not use a string literal, assign to variable first"
    * <https://docs.astral.sh/ruff/rules/raw-string-in-exception/>

All fixes were made automatically by Ruff itself.

### DIFF
--- a/mxcubeweb/app.py
+++ b/mxcubeweb/app.py
@@ -248,9 +248,8 @@ class MXCUBEApplication:
     server = None
 
     def __init__(self):
-        raise NotImplementedError(
-            "MXCUBEApplication is to be used as a pure static class, dont instanciate"
-        )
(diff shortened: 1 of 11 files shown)


marcus-oscarsson:
Ok, it makes the tracebacks more readable. A bit of a python issue in my opinion.

marcus-oscarsson:
Looks, good. Could you can rebase and merge please ?

fabcor-maxiv:
> Ok, it makes the tracebacks more readable. A bit of a python issue in my opinion.

Yes, I guess it's a small win. Tracebacks are indeed hard to read in Python so if it makes tracebacks more readable I am up for it.

PR #1581, opened 2025-02-27
Fix violations "flake8-return `RET`" Ruff rules

fabcor-maxiv:
Fix violations of Ruff rules in the "flake8-return `RET`" group:

* `RET502`
    * "Do not implicitly return None in function able to return non-`None` value"
    * <https://docs.astral.sh/ruff/rules/implicit-return-value/>
* `RET503`
    * "Missing explicit `return` at the end of function able to return non-`None` value"
    * <https://docs.astral.sh/ruff/rules/implicit-return/>
* `RET504`
    * "Unnecessary assignment before `return` statement"
    * <https://docs.astral.sh/ruff/rules/unnecessary-assign/>
* `RET505`
    * "Unnecessary branch after return statement"
    * <https://docs.astral.sh/ruff/rules/superfluous-else-return/>
* `RET506`
    * "Unnecessary branch after raise statement"
    * <https://docs.astral.sh/ruff/rules/superfluous-else-raise/>

All fixes done automatically by Ruff itself.

### DIFF
--- a/mxcubeweb/__init__.py
+++ b/mxcubeweb/__init__.py
@@ -143,8 +143,7 @@ def main():
     if server and cfg:
         server.run(cfg)
         return 0
-    else:
-        return 1
+    return 1
 
 
 if __name__ == "__main__":
--- a/mxcubeweb/app.py
+++ b/mxcubeweb/app.py
@@ -112,6 +112,7 @@ class MXCUBECore:
     def _get_object_from_id(_id):
         if _id in MXCUBECore.adapter_dict:
             return MXCUBECore.adapter_dict[_id]["adapter"]
+        return None
 
     @staticmethod
     def _get_adapter_id(ho):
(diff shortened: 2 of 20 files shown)


marcus-oscarsson:
Thanks !, Is there maybe a GitHub ruff plugin that can do this ?

fabcor-maxiv:
> Is there maybe a GitHub ruff plugin that can do this ?

What do you mean? Which part of the process would a plugin take care of?

marcus-oscarsson:
Creating PR's for rules that can be fixed automatically for rules that are disabled, and enabling those rules

fabcor-maxiv:
> Creating PR's for rules that can be fixed automatically for rules that are disabled, and enabling those rules

Yes, that'd be nice. I can do a quick search

marcus-oscarsson:
If not, its a cool project ;)

PR #1583, opened 2025-02-28
Fix violations of `flake8-simplify SIM` Ruff rules

fabcor-maxiv:
Fix violations of Ruff rules in the `flake8-simplify SIM` group:

* `SIM101`
    * Multiple `isinstance` calls, merge into a single call
    * <https://docs.astral.sh/ruff/rules/duplicate-isinstance-call/>
* `SIM102`
    * Use a single if statement instead of nested if statements
    * <https://docs.astral.sh/ruff/rules/collapsible-if/>
* `SIM105`
    * Use `contextlib.suppress` instead of `try`-`except`-`pass`
    * <https://docs.astral.sh/ruff/rules/suppressible-exception/>
* `SIM108`
    * Use ternary operator instead of `if`-`else`-`block`
    * <https://docs.astral.sh/ruff/rules/if-else-block-instead-of-if-exp/>
* `SIM114`
    * Combine `if` branches using logical `or` operator
    * <https://docs.astral.sh/ruff/rules/if-with-same-arms/>
* `SIM118`
    * Use `key {operator} dict` instead of `key {operator} dict.keys()`
    * <https://docs.astral.sh/ruff/rules/in-dict-keys/>
* `SIM201`
    * Use `{left} != {right}` instead of `not {left} == {right}`
    * <https://docs.astral.sh/ruff/rules/negate-equal-op/>
* `SIM210`
    * Remove unnecessary `True if ... else False`
    * <https://docs.astral.sh/ruff/rules/if-expr-with-true-false/>
* `SIM212`
    * Use `{expr_else} if {expr_else} else {expr_body}` instead of `{expr_body} if not {expr_else} else {expr_else}`
    * <https://docs.astral.sh/ruff/rules/if-expr-with-twisted-arms/>

Most fixes were done by Ruff itself automatically. For some fixes some light manual editing was done beforehand.

Violations that have no automatic fix and are left untouched:

* `SIM115`
    * Use a context manager for opening files
    * <https://docs.astral.sh/ruff/rules/open-file-with-context-handler/>

### DIFF
--- a/mxcubeweb/app.py
+++ b/mxcubeweb/app.py
@@ -396,10 +396,7 @@ class MXCUBEApplication:
             )
             uilog_file_handler.setFormatter(file_formatter)
 
-        if not log_level:
-            log_level = "INFO"
-        else:
-            log_level = log_level.upper()
+        log_level = "INFO" if not log_level else log_level.upper()
 
         custom_log_handler = MX3LoggingHandler(MXCUBEApplication.server)
         custom_log_handler.setLevel(log_level)
--- a/mxcubeweb/core/adapter/adapter_base.py
+++ b/mxcubeweb/core/adapter/adapter_base.py
@@ -1,3 +1,4 @@
+import contextlib
 import inspect
 import logging
 import traceback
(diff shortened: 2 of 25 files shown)


marcus-oscarsson (approved):

marcus-oscarsson:
Looks good, you can rebase and merge

PR #1584, opened 2025-02-28
Fix violation of Ruff rule `FLY002`

fabcor-maxiv:
Fix violation of Ruff rule `FLY002`:
<https://docs.astral.sh/ruff/rules/static-join-to-f-string/>

### DIFF
--- a/mxcubeweb/core/components/component_base.py
+++ b/mxcubeweb/core/components/component_base.py
@@ -12,12 +12,12 @@ def import_component(config, package="", module=""):
     _module = "mxcubeweb.core"
 
     if package:
-        _module = ".".join([_module, package])
+        _module = f"{_module}.{package}"
 
     if not module:
         module = config.class_name.lower()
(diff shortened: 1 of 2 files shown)


mxcubeweb/core/components/component_base.py
```diff
@@ -12,12 +12,12 @@ def import_component(config, package="", module=""):
...
 
     if package:
-        _module = ".".join([_module, package])
+        _module = f"{_module}.{package}"
 
     if not module:
         module = config.class_name.lower()
 
-    _module = ".".join([_module, module])
+    _module = f"{_module}.{module}"
```
marcus-oscarsson, line 20: ah,yes much nicer

marcus-oscarsson:
Thanks, its really nice that you take the time to do this !

marcus-oscarsson:
You can merge after re-base, thanks !

PR #1585, opened 2025-03-03
Refactor BeamlineActionsContainer

walesch-yan:
This PR refactors the `BeamlineActionsContainer` to become a functional component. Additional changes include:

- Replacing `mapStateToProps`, `mapDispatchToProps` and connect functions with `useSelector` and `UseDispatch`
- Make use of `useCallback` for component functions
- dispatching specific functions in child components to cause less re-rendering (except for `startBeamlineAction` and `stopBeamlineAction`, which are used by multiple child components) and preventing props drilling.
- using more distinguishable keys for list items (action names/ids for dropdown actions instead of list index and timestamp for messages instead of objects (objects will all be mapped to same string key which resulted in every message element having the exact same key))
- destructuring `props` at the beginning of `BeamlineActionControl` component for better overview of props elements

### DIFF
--- a/ui/src/components/BeamlineActions/BeamlineActionControl.jsx
+++ b/ui/src/components/BeamlineActions/BeamlineActionControl.jsx
@@ -6,30 +6,43 @@ import {
   twoStateActuatorIsActive,
   TWO_STATE_ACTUATOR,
 } from '../../constants';
+import { showActionOutput } from '../../actions/beamlineActions';
+import { useDispatch } from 'react-redux';
 
 export default function BeamlineActionControl(props) {
-  let variant = props.state === RUNNING ? 'danger' : 'primary';
-  let label = props.state === RUNNING ? 'Stop' : 'Run';
-  const showOutput = props.type !== TWO_STATE_ACTUATOR;
+  const {
+    actionId,
+    actionArguments,
+    handleStartAction,
+    handleStopAction,
+    state,
+    disabled,
+    type,
+    data,
+  } = props;
+  let variant = state === RUNNING ? 'danger' : 'primary';
+  let label = state === RUNNING ? 'Stop' : 'Run';
+  const showOutput = type !== TWO_STATE_ACTUATOR;
+  const dispatch = useDispatch();
 
-  if (props.type === 'INOUT') {
-    label = String(props.data).toUpperCase();
-    variant = twoStateActuatorIsActive(props.data) ? 'success' : 'danger';
+  if (type === 'INOUT') {
+    label = String(data).toUpperCase();
+    variant = twoStateActuatorIsActive(data) ? 'success' : 'danger';
   }
 
   return (
     <ButtonToolbar>
       <ButtonGroup className="d-flex flex-row" aria-label="First group">
-        {props.actionArguments.length === 0 ? (
+        {actionArguments.length === 0 ? (
           <Button
             size="sm"
             className="me-1"
             variant={variant}
-            disabled={props.disabled}
+            disabled={disabled}
             onClick={
-              props.state !== RUNNING
-                ? () => props.handleStartAction(props.actionId, showOutput)
-                : () => props.handleStopAction(props.actionId)
+              state !== RUNNING
+                ? () => handleStartAction(actionId, showOutput)
+                : () => handleStopAction(actionId)
             }
           >
             {label}
@@ -40,9 +53,9 @@ export default function BeamlineActionControl(props) {
         {showOutput ? (
           <Button
             variant="outline-secondary"
-            disabled={props.disabled}
+            disabled={disabled}
             size="sm"
-            onClick={() => props.handleShowOutput(props.actionId)}
+            onClick={() => dispatch(showActionOutput(actionId))}
           >
             <BiLinkExternal />
           </Button>
--- a/ui/src/components/BeamlineActions/BeamlineActionDialog.jsx
+++ b/ui/src/components/BeamlineActions/BeamlineActionDialog.jsx
@@ -21,7 +21,6 @@ export default function BeamlineActionDialog(props) {
     handleStartAction,
     handleStartAnnotatedAction,
     handleOnPlotDisplay,
-    handleSetActionArgument,
     plotId,
   } = props;
 
@@ -43,7 +42,6 @@ export default function BeamlineActionDialog(props) {
             actionArguments={actionArguments}
             handleStopAction={handleStopAction}
             handleStartAction={handleStartAction}
-            handleSetActionArgument={handleSetActionArgument}
           />
         )}
         {actionType === 'JSONSchema' && (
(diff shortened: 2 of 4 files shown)


ui/src/containers/BeamlineActionsContainer.jsx
```diff
@@ -1,164 +1,136 @@
...
-    this.props.hideOutput(this.props.currentAction.name);
+      const parameters = cmd.arguments.map((arg) =>
+        arg.type === 'float' ? Number.parseFloat(arg.value) : arg.value,
+      );
+
+      setPlotIdByAction((prev) => ({ ...prev, [currentAction.name]: null }));
+      dispatch(startBeamlineAction(cmdName, parameters, showOutput));
+    },
+    [actionsList, currentAction.name, dispatch],
+  );
```
walesch-yan, line 38: Also here I use `find` and `map` instead of `some` and `forEach`, as this is more understandable in my opinion. Also added an early return in case no action with the specific name was found

marcus-oscarsson:
Well done and Very nice initiative, Ill also let @axelboc have a look (but I think he might also be in vacation right ?)

marcus-oscarsson:
As I think we are both on vacation. Ill merge this now, but it does not prevent us from following up on any comments later (with an additional PR, if needed).

walesch-yan:
Thanks, yes let me know if there is anything that you think can be improved later on.

walesch-yan:
And enjoy your vacations;)

ui/src/components/BeamlineActions/BeamlineActionControl.jsx
```diff
@@ -6,30 +6,43 @@ import {
...
+  if (type === 'INOUT') {
+    label = String(data).toUpperCase();
+    variant = twoStateActuatorIsActive(data) ? 'success' : 'danger';
   }
 
   return (
     <ButtonToolbar>
       <ButtonGroup className="d-flex flex-row" aria-label="First group">
-        {props.actionArguments.length === 0 ? (
+        {actionArguments.length === 0 ? (
```
axelboc, line 36: I think this ternary and the one further down could be replaced with `&&`, since the falsy branch is just an empty string.

ui/src/components/BeamlineActions/BeamlineActionDialog.jsx
```diff
@@ -65,7 +63,7 @@ export default function BeamlineActionDialog(props) {
             {actionMessages.length > 0 && (
               <Card>
                 {actionMessages.map((message) => (
-                  <p key={message}>{message.message}</p>
+                  <p key={`message- ${message.timestamp}`}>{message.message}</p>
```
axelboc, line 66: You can even just put the timestamp: `key={message.timestamp}`.

ui/src/containers/BeamlineActionsContainer.jsx
```diff
@@ -1,164 +1,136 @@
...
+        actionType={currentAction.argument_type}
+        isActionRunning={currentActionRunning}
+        actionMessages={currentAction.messages}
+        handleStopAction={(actionId) => dispatch(stopBeamlineAction(actionId))}
+        handleStartAction={startAction}
+        handleStartAnnotatedAction={(cmdName, params, showOutput) =>
+          dispatch(startBeamlineAction(cmdName, params, showOutput))
+        }
+        handleOnPlotDisplay={newPlotDisplayed}
+        plotId={plotIdByAction[currentActionName]}
```
axelboc, line 130: Good that you didn't do it in the same PR, but if you feel up for it, it might be worth looking into reducing prop drilling here, by making use of `useSelector` and `useDispatch` inside `BeamlineActionDialog`.

It's not a bad thing to duplicate a bit of "computed state" logic, like `currentAction.state === RUNNING`, as this can be abstracted in other ways later if need be, like with a custom hook or by refactoring the structure of the state itself.

ui/src/containers/BeamlineActionsContainer.jsx
```diff
@@ -1,164 +1,136 @@
...
+        arg.type === 'float' ? Number.parseFloat(arg.value) : arg.value,
+      );
+
+      setPlotIdByAction((prev) => ({ ...prev, [currentAction.name]: null }));
+      dispatch(startBeamlineAction(cmdName, parameters, showOutput));
+    },
+    [actionsList, currentAction.name, dispatch],
+  );
+
+  const hideOutput = useCallback(() => {
```
axelboc, line 40: I wouldn't bother with the `useCallback`s and use simple `function` declarations instead. It's a form of premature optimisation here, as I don't think it prevents any re-render or avoids any heavy recomputations.

ui/src/containers/BeamlineActionsContainer.jsx
```diff
@@ -1,164 +1,136 @@
...
-    }
-
-    return (
-      <>
-        <Dropdown
-          title="Beamline Actions"
-          id="beamline-actions-dropdown"
+  const currentActionRunning = currentAction.state === RUNNING;
+  const currentActionName = currentAction.name;
+  const defaultDialogPosition = { x: -100, y: 100 };
```
axelboc, line 57: Can be extracted as a constant `DEFAULT_DIALOG_POSITION` at the top of the file.

ui/src/containers/BeamlineActionsContainer.jsx
```diff
@@ -1,164 +1,136 @@
...
-    this.startAction = this.startAction.bind(this);
-    this.hideOutput = this.hideOutput.bind(this);
-    this.newPlotDisplayed = this.newPlotDisplayed.bind(this);
-  }
+  const dispatch = useDispatch();
+  const currentAction = useSelector(
+    (state) => state.beamline.currentBeamlineAction,
+  );
 
-  startAction(cmdName, showOutput = true) {
```
axelboc, line 29: Nice refactoring here :+1: — it was needed :grin:

axelboc (commented):
Nicely done!

ui/src/components/BeamlineActions/BeamlineActionControl.jsx
walesch-yan: You are right, well spotted!

ui/src/components/BeamlineActions/BeamlineActionDialog.jsx
walesch-yan: Right!

ui/src/containers/BeamlineActionsContainer.jsx
walesch-yan: I agree that a second PR with a more in depth take on prop drilling is needed here, I can work on that and then will also add your other suggestions in that PR. Thanks for all the suggestions!

ui/src/containers/BeamlineActionsContainer.jsx
walesch-yan: Maybe I was to quick on using it here, you are right

PR #1586, opened 2025-03-04
Disallow cross-site requests by default

fabcor-maxiv:
The default behavior for *CORS* was to allow all cross-site requests, because the default value was set to `["*"]`. This is obviously a very bad default from security point of view.

The default behavior should be to disallow all cross-site requests. For an application like MXCuBE-Web, allowing same-site requests only is usually good enough.

### DIFF
--- a/demo/mxcube-web/server.yaml
+++ b/demo/mxcube-web/server.yaml
@@ -8,13 +8,7 @@ server:
 
   DEBUG: false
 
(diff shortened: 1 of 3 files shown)


mockoocy (approved):

walesch-yan (approved):

walesch-yan:
Thanks @fabcor-maxiv for updating this!
I think you can rebase and merge:)

PR #1587, opened 2025-03-10
Get sample list when no sample changer HWO loaded

fabcor-maxiv:
If the hardware object for the sample changer is not loaded properly, it is not possible for a user to authenticate properly. Actually the authentication is fine and it is possible to use MXCuBE, but the redirect from authentication page to main page is not done. The redirect is not done because it is not possible to get sample list.


marcus-oscarsson:
Oh yes, a good one, Thanks !

PR #1590, opened 2025-03-12
fix video-streamer zombie processes

### LINKED ISSUE #1577: Stopping MXCuBE-Web leaves video-streamer running
Starting MXCuBE-Web (with the demo beamline) and then stopping it (with `CTRL+C`) leaves a `video-streamer` process as well as some `ffmpeg` processes running.

walesch-yan:
This PR closes #1577 .

This PR updates `pyproject.toml` to use the version of `mxcubecore` fixing the above issue. The fix was introduced in mxcube/mxcubecore#1169 . It also moves the `video-streamer` dependency into `mxcubecore` and removes it from here. 

Additionally, I had to modify the tests, previously in our `atexit` register, we could decide to not kill any processes, as it was called in mxcubeweb. This is now different and we will see some `ProcessLookUpError` since they are not well synchronized with pytests handlers. Instead, this PR introduces a `monkeypatch` for the `atexit.register` function and to make sure to kill any child processes `psutil` is used after each test.

Bonus of this PR:
Using the latest version of mxcubecore will also fix a security vulnerability introduced by `jinja2`.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -14,8 +14,6 @@ PROJECT_PACKAGE_METADATA = importlib.metadata.metadata(PROJECT_PACKAGE_NAME)
 
 # This is necessary to for the `autoflask` directive to get access to the routes
 server, _ = mxcubeweb.build_server_and_config(test=True, argv=[])
-# This is to avoid `Server.kill_processes`, that makes the build return non-zero
-server.flask.testing = True
 
 
 # -- General configuration ------------------------------------------------
--- a/mxcubeweb/server.py
+++ b/mxcubeweb/server.py
@@ -1,7 +1,5 @@
-import atexit
 import logging
 import os
-import signal
 import traceback
 
 import flask_security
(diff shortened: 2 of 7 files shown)


walesch-yan:
So In this [line](https://github.com/mxcube/mxcubeweb/blob/0726ec0625748fe46f2eab70cd0f837c7f7ca1b5/requirements.txt#L337C1-L339C83), when running manually the pre-commits on python 3.10 it changed 
```
python_version == "3.10"
```
to 
```
python_version >= "3.10" and python_version < "3.11"
```
which was not detected by the automatic local pre-commits (3.10) but crashed the CI running on 3.11, just in case somebody encounters this issue

docs/conf.py
```diff
@@ -14,8 +14,6 @@
 
 # This is necessary to for the `autoflask` directive to get access to the routes
 server, _ = mxcubeweb.build_server_and_config(test=True, argv=[])
-# This is to avoid `Server.kill_processes`, that makes the build return non-zero
-server.flask.testing = True
```
walesch-yan, line 18: I'm removing this testing flag on a few occasions in the code, since it is now no longer needed

fabcor-maxiv:
> So In this [line](https://github.com/mxcube/mxcubeweb/blob/0726ec0625748fe46f2eab70cd0f837c7f7ca1b5/requirements.txt#L337C1-L339C83), when running manually the pre-commits on python 3.10 it changed
> 
> ```
> python_version == "3.10"
> ```
> 
> to
> 
> ```
> python_version >= "3.10" and python_version < "3.11"
> ```
> 
> which was not detected by the automatic local pre-commits (3.10) but crashed the CI running on 3.11, just in case somebody encounters this issue

That is curious...

Could it be that somewhere somehow a different version of Poetry was used? I think the version is pinned in the pre-commit hook, so it should be safe from such issues.

walesch-yan:
> That is curious...
> 
> Could it be that somewhere somehow a different version of Poetry was used? I think the version is pinned in the pre-commit hook, so it should be safe from such issues.

I thought so as well at the beginning and my first try was to use the `1.8.0` version (we have `1.9.0`), but It would give me the same faulty result, it was only when I changed the python version that this was resolved

fabcor-maxiv:
> > That is curious...
> > Could it be that somewhere somehow a different version of Poetry was used? I think the version is pinned in the pre-commit hook, so it should be safe from such issues.
> 
> I thought so as well at the beginning and my first try was to use the `1.8.0` version (we have `1.9.0`), but It would give me the same faulty result, it was only when I changed the python version that this was resolved

OK... that does not make much sense to me. Some bug somewhere maybe...

PR #1591, opened 2025-03-13
Minor fix

rhfogh:
Updated GPhL config files and fixed bug in get_entry (Raised error when getting root object with id=None)

### DIFF
--- a/mxcubeweb/core/components/queue.py
+++ b/mxcubeweb/core/components/queue.py
@@ -706,14 +706,14 @@ class Queue(ComponentBase):
         """
         Retrieves the model and the queue entry for the model node with id <id>
 
-        :param int id: Node id of node to retrieve
+        :param int _id: Node id of node to retrieve
         :returns: The tuple model, entry
         :rtype: Tuple
         """
         if _id is None:
             return (
                 HWR.beamline.queue_model.get_model_root(),
-                HWR.beamline.queue_manager
+                HWR.beamline.queue_manager,
             )
         model = HWR.beamline.queue_model.get_node(int(_id))
         entry = HWR.beamline.queue_manager.get_entry_with_model(model)
--- a/mxcubeweb/core/components/queue.py
+++ b/mxcubeweb/core/components/queue.py
@@ -707,7 +707,7 @@ class Queue(ComponentBase):
         Retrieves the model and the queue entry for the model node with id <id>
 
         :param int _id: Node id of node to retrieve
-        :returns: The tuple model, entry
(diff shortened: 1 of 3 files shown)


marcus-oscarsson:
@rhfogh, it looks like the linter strikes again ;), just some formatting issue.

mxcubeweb/core/components/queue.py
```diff
@@ -706,10 +706,15 @@ def get_entry(self, _id):
         """
         Retrieves the model and the queue entry for the model node with id <id>
 
-        :param int id: Node id of node to retrieve
+        :param int _id: Node id of node to retrieve
```
marcus-oscarsson, line 709: We should probably document that the root node is returned if `_id` is `None`

mxcubeweb/core/components/queue.py
```diff
@@ -706,10 +706,15 @@ def get_entry(self, _id):
...
         :returns: The tuple model, entry
```
marcus-oscarsson:
```suggestion
        :returns: The tuple model, entry or the root node and QueueManger if _id is None
```

beteva:
Actually this "minor" change in the queue breaks the characterisation. @rhfogh, could you, please explain why do you need it.

rhfogh:
Sorry about that. I have a meeting coming up, but just quickly:

Essentially queue.get-entry(_id) was being called with _id == None, resulting in an error when you get to the line

        model = HWR.beamline.queue_model.get_node(int(_id))

The context it was being called from made it clear that you wanted the topmost object back. I cannot quite see how the fix could break anything, unless someone is deliberately doing

try:
  self.get_entry(_id)
except TypeError:
  ...

instead of simply checking whether _id is None

If this is not enough I shall look in more detail after my meeting.

beteva:
When running characterisation, the top node seems to be the TaskNode, which is not the case of the other collection methods. TaskNode does not have state attribute, so it raises an error. I have not investigate, though, why you get as up as TaskNode, maybe because the characterisation has 3 nodes in the list, as the other have only 2.

rhfogh:
I have redone the mock test, both with and without the added code
```
        if _id is None:
            return (
                HWR.beamline.queue_model.get_model_root(),
                HWR.beamline.queue_manager,
            )
```
and they both work the same. At this point it looks like whatever causes the problem may have been fixed elsewhere - anyway it should never happen that this particular parameter was None.

So, I am happy to either leave things as they are or undo my fix, but I suspect someone else is more on top of what is going on than I am. What do people want me to do?

PR #1592, opened 2025-03-14
Making limits "Optional"

marcus-oscarsson:
Limits used to be typed `Optional`, some hardware objects return None for limits. Using `| None` handle both values and `NoneType`


beteva:
Yes, well spotted. (None, None) is the default return value from AbstravtActuator an I guess many inheriting classes.

walesch-yan (approved):

PR #1594, opened 2025-03-17
replace createStore by configureStore

walesch-yan:
This PR replaces the `createStore` function call with `configureStore`. It would be a first step into modernizing the `Redux` logic in the code. The goal of the modernized (recommended) `Redux` code would be to have a smaller and easier to maintain Codebase. Modern `Redux` still allows old code and the migration can be made step by step (This [document](https://redux.js.org/usage/migrating-to-modern-redux) explains the major implications and guides through different steps of moving to modern `Redux` patterns).

The changes in this PR include the replacement of `createStore`, `applyMiddleware`, `combineReducers`, `thunk` and `composeWithDevtools` logic by one call to `configureStore`

 In addition, per default `configureStore` adds three middleware options to the development environment. These are:
 - Immutability Check: Throws an error if state mutations happen in reducer logic
 - Serializability Check: Adds a warning to the console if state contains a non-serializable object
 - Action Creator Check: logs a warning if an action creator is dispatched without being called
More information about these checks can be found [here](https://redux-toolkit.js.org/api/getDefaultMiddleware#development)

For now I disabled the immutability check, as the thrown errors could intervene with current developments, however this option should be used in the future as well

### DIFF
--- a/ui/src/store.js
+++ b/ui/src/store.js
@@ -1,20 +1,22 @@
-import { createStore, applyMiddleware } from 'redux';
+import { configureStore } from '@reduxjs/toolkit';
 import { createLogger } from 'redux-logger';
-import thunk from 'redux-thunk';
 import rootReducer from './reducers';
-import { composeWithDevTools } from '@redux-devtools/extension';
 
 const middleware = [
-  thunk,
   ...(import.meta.env.VITE_REDUX_LOGGER_ENABLED === 'true'
     ? [createLogger()]
     : []),
 ];
 
-export const store = createStore(
-  rootReducer,
-  composeWithDevTools(applyMiddleware(...middleware)),
-);
+export const store = configureStore({
+  reducer: rootReducer,
+  middleware: (getDefaultMiddleware) => [
+    ...getDefaultMiddleware({
+      immutableCheck: false,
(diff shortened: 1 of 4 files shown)


ui/src/store.js
```diff
@@ -1,20 +1,22 @@
...
-);
+export const store = configureStore({
+  reducer: rootReducer,
+  middleware: (getDefaultMiddleware) => [
+    ...getDefaultMiddleware({
+      immutableCheck: false,
+    }),
+    ...middleware,
+  ],
+});
```
walesch-yan, line 19: According to the [documentation](https://redux-toolkit.js.org/api/getDefaultMiddleware#intended-usage) it is recommended to use chainable methods here (`.concat(...)`), since it could under some circumstances introduce issues for TS related stuff. However, since we do not using Typescript and our linter complains otherwise, I chose the array spread operator instead.

axelboc (approved):
Great idea

ui/src/store.js
marcus-oscarsson: Ok

marcus-oscarsson:
Really good initiative, well done !

PR #1595, opened 2025-03-17
[XML to YAML] - Rebased `xml_yaml_conversion` on `develop`

marcus-oscarsson:
Pending, https://github.com/mxcube/mxcubecore/pull/1177


marcus-oscarsson:
@rhfogh I rebased this on current `develop`. There where a few conflicts in the GPHL workflow related configuration files. I kept them as they where on this branch as there are syntactical and indentation changes. Could you please make sure that the files are up to date with the current configuration data ?

marcus-oscarsson:
There are a few commits to add to get the `mxcubeweb` code to run properly. I noticed something quite important that needs to be addressed as soon as possible and its these pieces of code:

https://github.com/mxcube/mxcubecore/blob/613505a24909edbffde6877881f65ead16327a40/mxcubecore/BaseHardwareObjects.py#L335

https://github.com/mxcube/mxcubecore/blob/613505a24909edbffde6877881f65ead16327a40/mxcubecore/BaseHardwareObjects.py#L1076

We have somehow been lucky to avoid this issue when we where testing on the beamline. Most likely because internal class attribute names have not been the same as configuration file property names. However you will have an issue if its the case which is clearly not acceptable. So we have to sort this out as soon as possible.

marcus-oscarsson:
I have PR for a fix to `mxcubecore` more or less ready but it needs some more testing.

demo/gphl/gphl-setup.yml
```diff
@@ -43,8 +43,8 @@ configuration:
   # HardwareRepository directories on the lookup path
 
 software_paths:
-  # Mandatory. Directory with workflow config input, e.g. instrumentation.nml
-  gphl_beamline_config: gphl/config_esrf_id30b
+    # Mandatory. Directory with workflow config input, e.g. instrumentation.nml
+    gphl_beamline_config: gphl/config_esrf_id30b
```
marcus-oscarsson: @rhfogh I had to disable the GPHL workflows for the moment because the parsing of the YAML files failed. Perhaps you know what the issue is ?

rhfogh: Actually not. Is there an error message? Is it an actual error or is it just ruff not liking the formatting?

marcus-oscarsson: Actually it was the parsing by `raumel.load` that failed for some reason. I can have another look at a later stage now when everything is else seems to back on track working. But I can't look at it right away as I have a bunch of other things to finish as well.

marcus-oscarsson: I had a look and I fixed a indentation error, not sure what the original error actually was but it seems to be loading now.

elmjag:
I did some quick tests with demo beamline, looks good to me. Ship it :)

elmjag (approved):

marcus-oscarsson:
Thanks :+1:

PR #1598, opened 2025-03-19
refactor HelpContainer

walesch-yan:
This PR refactors the `HelpContainer` into a functional component following similar approaches as for previous PRs.

### DIFF
--- a/ui/src/containers/HelpContainer.jsx
+++ b/ui/src/containers/HelpContainer.jsx
@@ -1,6 +1,15 @@
-import React from 'react';
-import { connect } from 'react-redux';
-import { Container, Card, Row, Col, Button, Form } from 'react-bootstrap';
+import React, { useEffect, useState } from 'react';
+import { useSelector } from 'react-redux';
+import {
+  Container,
+  Card,
+  Row,
+  Col,
+  Button,
+  Form,
+  Alert,
+} from 'react-bootstrap';
+import { Controller, useForm } from 'react-hook-form';
 
 import { sendFeedback } from '../api/login';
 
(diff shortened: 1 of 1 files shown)


ui/src/containers/HelpContainer.jsx
```diff
@@ -8,184 +8,161 @@ import characterisation from '../help_videos/mx3-characterisation.ogv';
...
+  const senderRef = React.useRef();
+  const contentRef = React.useRef();
 
-  handleSubmit() {
-    sendFeedback(this.sender.value, this.content.value);
-    this.sender.value = '';
-    this.content.value = '';
-  }
+  const user = useSelector((state) => state.login.user);
+  const serverVersion = useSelector((state) => state.general.serverVersion);
```
walesch-yan, line 34: The `serverVersion` in the `Redux` state would be a good example of something that in my opinion should not be in the global state, it will not change and this is the only component that uses it

ui/src/containers/HelpContainer.jsx
```diff
@@ -8,184 +8,161 @@ import characterisation from '../help_videos/mx3-characterisation.ogv';
...
 
-  localContactPanel() {
-    let panel = null;
-
-    if (this.props.login.user) {
-      const familyName = this.props.login.user.nickname || '';
-      const givenName = this.props.login.user.nickname || '';
-      const email = this.props.login.user.email || '';
-      const tel = '';
```
walesch-yan, line 32: I removed the `familyName` and `givenName` variables here and replaced them by one call to `nickname`. Also removed `tel` as it was always an empty string

ui/src/containers/HelpContainer.jsx
```diff
@@ -8,184 +8,161 @@ import characterisation from '../help_videos/mx3-characterisation.ogv';
...
-                    </Col>
-                  </Row>
-                </Card.Body>
+  const links = import.meta.env.VITE_HELP_LINKS?.map((link) => (
+    <div key={link.url}>
+      <a target="_blank" href={link.url} rel="noreferrer">
+        {link.name}
+      </a>
+    </div>
+  ));
```
walesch-yan: Unfortunately the `HELP_LINKS` were not available in the demo, hence I used the `url` for the key prop here, it would be best to have an `id` I guess.

ui/src/containers/HelpContainer.jsx
marcus-oscarsson: Makes sense

ui/src/containers/HelpContainer.jsx
marcus-oscarsson: :), good call !

marcus-oscarsson:
Nice :)

ui/src/containers/HelpContainer.jsx
```diff
@@ -8,184 +8,161 @@ import characterisation from '../help_videos/mx3-characterisation.ogv';
...
-              Name: {givenName} {familyName}
-              <br />
-              Email: {email}
-              <br />
-              Tel: {tel} <br />
-            </span>
-          </Card.Body>
-        </Card>
-      );
+  const handleSubmit = () => {
```
axelboc: I like to use the function declaration syntax for named functions: `function handleSubmit() { ... }`. I find it helps to distinguish functions from variables/constants when reading the code.

ui/src/containers/HelpContainer.jsx
```diff
@@ -8,184 +8,161 @@ import characterisation from '../help_videos/mx3-characterisation.ogv';
...
+      </a>
+    </div>
+  ));
+
+  return (
+    <Container fluid className="mt-4">
+      <Row>
+        <Col sm={12} className="d-flex">
+          <Col sm={4}>
+            {localContactPanel()}
```
axelboc: It's best to avoid "render" functions like this. It doesn't actually remove complexity from the component — on the contrary, it adds a function.

Either inline the JSX or move it into a separate component. In this case, I think inlining is fine — long JSX is okay if the logic (conditions, loops, etc.) is limited.

ui/src/containers/HelpContainer.jsx
axelboc: I think this `VITE_HELP_LINKS` feature is broken since env vars are strings but this one is used as an array in the code. I would consider removing this feature.

For the sake of discussion, storing JSX in a variable should also be avoided (for the same reason that it doesn't reduce the complexity of the component). I would have probably inlined the JSX and maybe stored and parsed the env var in a constant at the top of the file:

```js
cost HELP_LINKS = import.meta.env.VITE_HELP_LINKS; // + parse as array somehow

...
// In the JSX:
{HELP_LINKS.length > 0 && (
  <Card>
    {HELP_LINKS..map(...)}
  </Card>
)}
```

ui/src/containers/HelpContainer.jsx
```diff
@@ -8,184 +8,161 @@ import characterisation from '../help_videos/mx3-characterisation.ogv';
...
+                    <span>
+                      <b>Interleaved </b> <br />
+                      <video width="230" height="132" controls>
+                        <source src={interleaved} type="video/mp4" />
+                      </video>
+                    </span>
+                  </Col>
+                  <Col className="col-xs-4">
+                    <span>
+                      <b>Mesh </b> <br />
```
axelboc: Could you please replace the `b` and `br` tags with `h3` and remove the `span` wrappers?  Might as well improve the HTML a bit... :joy:

ui/src/containers/HelpContainer.jsx
```diff
@@ -8,184 +8,161 @@ import characterisation from '../help_videos/mx3-characterisation.ogv';
...
-export class HelpContainer extends React.Component {
-  constructor(props) {
-    super(props);
-    this.handleSubmit = this.handleSubmit.bind(this);
-    this.sender = '';
-    this.content = '';
-  }
+function HelpContainer() {
+  const senderRef = React.useRef();
+  const contentRef = React.useRef();
```
axelboc: Perhaps a good opportunity to try [React Hook Form](https://react-hook-form.com/)? It's already installed — I used it for the login form and for the manual sample creation form, if I recall.

axelboc (approved):
Nice! FYI, I think we considered removing this page at some point, which is why I hadn't touched it yet. Worth making a decision on this before addressing my comments (especially the React Hook Form one :grin: — although it could be good practice if you're interested, since it's such a simple form)

walesch-yan:
> Nice! FYI, I think we considered removing this page at some point, which is why I hadn't touched it yet. Worth making a decision on this before addressing my comments (especially the React Hook Form one 😁 — although it could be good practice if you're interested, since it's such a simple form)

Oh, maybe I missed/forgot this consideration :sweat_smile: I guess we can keep it as a draft then, I prefer to use it as a little practice if possible;)

ui/src/containers/HelpContainer.jsx
walesch-yan: Sure, I guess this was just my personal preference but for the sake of consistency I will change it (and try to not let my inner demon do that again :laughing: )

ui/src/containers/HelpContainer.jsx
walesch-yan: Well spotted! Maybe I went a little to fast over this component

ui/src/containers/HelpContainer.jsx
```diff
@@ -1,191 +1,177 @@
...
-  };
+                  {feedback.message && (
+                    <Alert
+                      className="mt-3"
+                      variant={feedback.variant}
+                      onClose={() => setFeedback({})}
+                      dismissible
+                    >
+                      {feedback.message}
+                    </Alert>
```
walesch-yan: Added an alert as feedback for the user if the form was submitted correctly or if it failed

ui/src/containers/HelpContainer.jsx
```diff
@@ -1,191 +1,177 @@
...
+                    </Form.Label>
+                    <Controller
+                      name="email"
+                      control={control}
+                      rules={{ required: 'Email is required' }}
+                      render={({ field }) => (
+                        <Form.Control
+                          type="text"
+                          id="EmailAddress"
+                          placeholder="Your contact information (email, Name or Proposal)"
```
walesch-yan, line 98: I set the type here to `text`, instead of `email`, since the label explicitly allows name or Proposals and otherwise the form could not be submitted (as the @ symbol is missing in a name). However, we could also enforce this to be an email.

walesch-yan (commented):
I added the suggestions and integrated the `react-hook-form` for this component:)

ui/src/containers/HelpContainer.jsx
```diff
@@ -1,191 +1,177 @@
...
-          </Card.Body>
-        </Card>
-      );
+  async function handleSubmit(data) {
+    try {
+      await sendFeedback(data.email, data.content);
+      setFeedback({
+        variant: 'success',
+        message: 'Feedback sent successfully',
+      });
```
axelboc: The form state has an `isSubmitSuccessful` flag that tells you if the submit was successful (i.e. no error thrown in `handleSubmit`). And if `sendFeedback` throws an error, you can catch it and call the form's `setError` function as explained here: https://react-hook-form.com/docs/useform/seterror

This means that you don't need the `feedback` state and can just declare different alerts in the JSX based on `isSubmitSuccessful` and `errors`.

ui/src/containers/HelpContainer.jsx
```diff
@@ -1,191 +1,177 @@
...
+      reset(DEFAULT_HELPFORM_VALUES);
```
axelboc: This may reset `isSubmitSuccessful` and `errors`. I think there's an option that controls this: https://react-hook-form.com/docs/useform#resetOptions

axelboc (approved):

ui/src/containers/HelpContainer.jsx
```diff
@@ -1,191 +1,176 @@
...
-            {link.name}
-          </a>
-        </div>
-      ));
+  useEffect(() => {
+    if (isSubmitSuccessful) {
+      reset();
+      setShowSuccess(true);
     }
+  }, [isSubmitSuccessful, reset]);
```
walesch-yan: According to the [docs](https://react-hook-form.com/docs/useform/reset) the recommended way of using reset is inside of `useEffect` using `isSubmitSuccessful` as trigger, I wrapped it inside the `if`, since `reset` will change the value of `isSubmitSuccessful` to `false`.

As one can see I also kept the `useState` here since there is no option to keep `isSubmitSuccessful` between resets but only `isSubmitted` (which becomes also true if the execution of submit fails). One could use `resetField` but it would go against the idea of sending multiple forms. Hence I decided to keep a local state variable.

axelboc: Makes sense, in that case, I would move `setShowSuccess(false)` (currently in `handleSubmit`) into a separate use effect as well to hide the success message when the user starts filling in the form again:

```ts
  useEffect(() => {
    if (isDirty) {
      // Hide success message if user starts filling in the form again
      setShowSuccess(false);
    }
  }, [isDirty]);
```

But it's probably overly complicating things... Forms suck.

walesch-yan: Well in terms of UX your approach is probably better, thanks for all the suggestions!

> Forms suck.

Couldn't agree more :sweat_smile:

marcus-oscarsson:
Hm, this one has been here for some time now. What would you like to do with it ?

PR #1599, opened 2025-03-19
Make energy level read-only in UI

dominikatrojanowska:
In this merge request:
- all data collection task components are made sensitive to read-only property of Energy field
- all data collection task components show the current level of energy while in read-only mode

### DIFF
--- a/ui/src/components/Tasks/fields.jsx
+++ b/ui/src/components/Tasks/fields.jsx
@@ -67,7 +67,8 @@ export function resetLastUsedParameters(formObj) {
 export function toFixed(state, hoName, parameterName = null) {
   const withTaskData =
     state.taskForm.sampleIds.constructor !== Array ||
-    state.queue.rememberParametersBetweenSamples;
+    (state.queue.rememberParametersBetweenSamples &&
+      !state.beamline.hardwareObjects[hoName].readonly);
 
   const pname = parameterName !== null ? parameterName : hoName;
 
--- a/ui/src/components/Tasks/fields.jsx
+++ b/ui/src/components/Tasks/fields.jsx
@@ -66,12 +66,10 @@ export function resetLastUsedParameters(formObj) {
 
 export function toFixed(state, hoName, parameterName = null) {
   //
-  // Check if 'remembered' value for this field will be used.
+  //  Use 'remembered' value if:
   //
-  // Don't use 'remembered' value if:
-  //  - dealing with multiple samples
-  //  - 'remember parameters between samples' option is disabled
-  //  - the field is marked as read-only
+  //  - we are dealing with a single sample
+  //  - 'remember parameters between samples' option is enabled AND the field is not marked as read-only
   //
   const withTaskData =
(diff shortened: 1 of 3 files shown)


ui/src/components/Tasks/fields.jsx
```diff
@@ -67,7 +67,8 @@ export function resetLastUsedParameters(formObj) {
 export function toFixed(state, hoName, parameterName = null) {
   const withTaskData =
     state.taskForm.sampleIds.constructor !== Array ||
-    state.queue.rememberParametersBetweenSamples;
+    (state.queue.rememberParametersBetweenSamples &&
```
marcus-oscarsson, line 76: ~Hm, what does this do ?~

marcus-oscarsson: Ah, ok I see. looks good, A comment would maybe be good.

marcus-oscarsson:
Looks, good. Maybe just add a comment in the `toFixed` logic because that quickly becomes hard to follow

ui/src/components/Tasks/fields.jsx
dominikatrojanowska: Ok, but I have no idea what is this first part is about `state.taskForm.sampleIds.constructor !== Array ||`

marcus-oscarsson: Its to check it's for one sample (an string ID) or several samples(Array), a bit weird I know. If a list of samples are passed the same template data collection is used on all samples.

axelboc: `!Array.isArray(state.taskForm.sampleIds)`

dominikatrojanowska: yes, it's wired assumption made here or at least not easy to understand it's purpose :) so in case `sampleIds` is not an array we are going to use `taskForm.taskData` as a value shown instead of `beamline.hardwareObjects` whether the parameters should be remembered between sample or not 🤔 

in that case the `constructor` is missed, so this is an aiming solution?

marcus-oscarsson: Not sure what you mean by "aiming solution" ?

Basically by checking that the constructor is `Array` we check that `sampleIds` is an Array. I don't know the internals of `isArray` but imagine that its something similar just more elegant and abstracted. 

I may have missed something but that should indeed work.

elmjag: I suggest adding this comment:

```
//
// Check if we are using 'remembered' value for this field.
//
// Don't use 'remembered' value if:
//
//  - we are dealing with multiple samples
//  - 'remember parameters between samples' option is disabled
//  - the field is marked as read-only
//
```

walesch-yan: From my understanding this comment seems to be slightly different from the implementation, as I would understand it as three ORs.
Could I suggest :
```python
// Use 'remembered' value if:
//
//  - we are dealing with a single sample
//  - 'remember parameters between samples' option is enabled AND the field is not marked as read-only
```

PR #1600, opened 2025-03-24
Refactor DeviceState

walesch-yan:
Continuing the refactoring work, this PR handles a quite small component.

### DIFF
--- a/ui/src/components/DeviceState/DeviceState.jsx
+++ b/ui/src/components/DeviceState/DeviceState.jsx
@@ -3,29 +3,24 @@ import { Badge } from 'react-bootstrap';
 
 import styles from './deviceState.module.css';
 
-export default class DeviceState extends React.Component {
-  render() {
-    let msgBgStyle = 'warning';
+function DeviceState(props) {
+  const { labelText, data } = props;
+  const msgBgStyle = data === 'READY' ? 'info' : 'warning';
 
-    if (this.props.data === 'READY') {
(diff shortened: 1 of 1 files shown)


ui/src/components/DeviceState/DeviceState.jsx
```diff
@@ -3,29 +3,20 @@ import { Badge } from 'react-bootstrap';
...
+      <Badge className={styles.msgLabelStyle} bg={msgBgStyle}>
+        {data}
+      </Badge>
+    </div>
+  );
 }
 
-DeviceState.defaultProps = {
-  labelText: '',
-  pkey: undefined,
```
walesch-yan, line 29: removed `pkey` here as it is not used in the component

axelboc (approved):
LGTM!

Only think I'd check (but maybe you have) is whether the default values are really needed in practice and what the component looks like if they are used. Because if `labelText` really falls back to `""`, then you get an empty `Badge`, so it might be safer to fall back to `undefined` (i.e. no fallback) and to have `{labelText && (<Badge>...</Badge>)}`  in the JSX.

walesch-yan:
@axelboc thanks for the review I changed it accordingly.
From my understanding they are not really needed, and the `deviceState` component is only used inside of `BeamlineSetupContainer` if it would be called with default props, it would be  a strong indication that there is a wrongly configured hardwareObject in the state, but I assume this would be catched/seen before.

PR #1603, opened 2025-03-26
Remove obsolete browserslist config

axelboc:
The `browserslist` config is a leftover from Create React App. Vite doesn't care about it and supports only [modern browsers](https://vite.dev/guide/build.html).

I don't think we've had bug reports related to old browser versions, which makes me think that all facilities use MXCuBE in up-to-date browsers. If it's not the case, then we should configure Vite accordingly, if needed by adding `@vitejs/plugin-legacy`.

There's also a [`.babelrc` file](https://github.com/mxcube/mxcubeweb/blob/develop/test/.babelrc) in the test folder. Not sure what it's doing there — is it still needed? :shrug:

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -74,18 +74,6 @@
     "vite-plugin-eslint": "1.8.1",
     "wait-on": "8.0.3"
   },
-  "browserslist": {
-    "production": [
-      ">0.2%",
-      "not dead",
-      "not op_mini all"
-    ],
-    "development": [
-      "last 1 chrome version",
-      "last 1 firefox version",
-      "last 1 safari version"
-    ]
-  },
   "pnpm": {
     "peerDependencyRules": {
       "allowedVersions": {


fabcor-maxiv:
What was going on in <https://github.com/mxcube/mxcubeweb/pull/1573> then? : D

axelboc:
Hehe excellent question. I had to think about it... So basically our linting config `eslint-config-galex` has Babel, and notably `@babel/eslint-parser`, as a dependency (to support experimental JS language features), which itself has a dependency on `@babel/core`, which itself requires `browserslist`. But the browserslist is only relevant to know which experimental language features to transpile and which to leave alone; it's not involved in the parsing step.

When we had Create React App, we had to have a `browserslist` config, because it relied on Babel for transpilation, but Vite doesn't.

Long story short, `browserslist` is now a dev dependency that no longer matters. It will be removed very soon anyway, as I'm currently working on replacing `eslint-config-galex`.

fabcor-maxiv (approved):

PR #1604, opened 2025-03-26
Switch to new ESLint config

axelboc:
As mentioned [here](https://github.com/mxcube/mxcubeweb/discussions/1544#discussioncomment-12055627), I created [my own ESLint config](https://gitlab.esrf.fr/ui/eslint-config) to replace `eslint-config-galex`. I've tested it in a few projects like H5Web, and I think it's [pretty strict](https://gitlab.esrf.fr/ui/eslint-config/-/blob/main/src/index.js?ref_type=heads) and robust, so I've gone ahead and published it to NPM as [`@esrf/eslint-config`](https://www.npmjs.com/package/@esrf/eslint-config).

> I know I still need to add documentation for it, and I'll have to eventually document its use in MXCuBE as well. I'll try to do this asap.

Thanks to the config's dynamic nature (it looks at the project's dependencies in `package.json`), once we adopt TypeScript, it will automatically start to apply TypeScript rules.

In this PR, I did not try to solve many of the issues raised by the new config. Instead I mostly turn off the rules, with the goal to re-enable them and fix them in separate PRs.

It's worth pointing out that the config does not strictly forbid React class components (which still have their [use cases](https://react.dev/reference/react/Component#componentdidcatch)) ... yet ... but it does enforce well-written class components. Of course, we should still continue the migration to function components.


ui/eslint.config.js
```diff
@@ -0,0 +1,41 @@
+import { createConfig, detectOpts } from '@esrf/eslint-config';
+import { globalIgnores } from 'eslint/config';
+
+const opts = detectOpts(import.meta.dirname);
+
+const config = [
+  globalIgnores([
+    'build/',
+    'release/',
```
axelboc: Wondering if this folder still needs to be excluded. Anyone knows how/when/if it's still generated?

ui/src/actions/queue.js
```diff
@@ -1,7 +1,7 @@
 import { showErrorPanel } from './general';
-import { mountSample } from './sampleChanger'; // eslint-disable-line import/no-cycle
```
axelboc, line 2: `no-cycle` is now disabled. It slows down linting _significantly_ because it has to perform module resolution. Vite can detect when import cycles cause issues, so it's not worth detecting this during linting.

ui/src/components/Equipment/PlateManipulator.jsx
```diff
@@ -353,7 +353,6 @@ export default function PlateManipulator(props) {
     </div>
   );
 
-  // eslint-disable-next-line sonarjs/cognitive-complexity
```
axelboc, line 356: Replaced with ESLint's built-in `complexity` rule.

ui/src/components/Plot1D.jsx
```diff
@@ -3,7 +3,6 @@ import { connect } from 'react-redux';
 import Dygraph from 'dygraphs';
 import 'dygraphs/dist/dygraph.min.css';
 
-// eslint-disable-next-line react/no-unsafe
```
axelboc, line 6: Worth noting that ESLint is now much better at detecting misused inline directives.

ui/src/components/SampleView/ContextMenu.jsx
```diff
@@ -255,10 +253,8 @@ export default class ContextMenu extends React.Component {
         {
           text: 'Go to Beam',
           action: () => {
-            {
```
axelboc, line 258: Random `{}` block here.

ui/src/components/Tasks/AddSample.jsx
```diff
@@ -9,7 +9,7 @@ import { useForm } from 'react-hook-form';
 import { hideTaskParametersForm } from '../../actions/taskForm';
 
 const REQUIRED_MSG = 'This field is required';
-const PATTERN = /^[\w+:-]*$/u;
+const PATTERN = /^[\w+\-:]*$/u;
```
axelboc, line 12: There's now a regex linting plugin. It enforces a lot of things, including that characters inside `[]` be sorted in a specific order for consistency.

marcus-oscarsson:
Wonderful, well done ! 

Ill try to find some time to look at the details tomorrow or on Friday, (I'm in the two day training thing you know ;) )

ui/eslint.config.js
```diff
@@ -0,0 +1,41 @@
...
+      'react/no-multi-comp': 'off',
+      'react/no-unused-class-component-methods': 'off',
+      'react/no-unstable-nested-components': 'off',
+      'react/prop-types': 'off',
+      'react/sort-comp': 'off',
+      'react/static-property-placement': 'off',
+      'simple-import-sort/imports': 'off',
+      'unicorn/prefer-global-this': 'off',
+      'unicorn/prefer-modern-math-apis': 'off',
+      'unicorn/switch-case-braces': 'off',
```
fabcor-maxiv, line 32: If I understood correctly these rules are disabled temporarily, until we get around to fix them, right? We do not actually want to keep them disabled, is that right? If yes, maybe there should be comment there to explain this, and to explain that developers should not add anything new to this list either.

fabcor-maxiv (approved):
Looks good to me, a non-JavaScript expert.

I assume this is still enforced in a GitHub Actions workflow, the same as before, is that right?

ui/eslint.config.js
axelboc: Some rules may remain disabled if they're too difficult to apply, of course, but yeah that's the idea. I'll add comments as I go if need be. And I'll document the whole front-end linting situation in `front-end.md` once stable.

marcus-oscarsson:
> Thanks to the config's dynamic nature (it looks at the project's dependencies in package.json), once we adopt TypeScript, it will automatically start to apply TypeScript rules.

That's really nice !

> In this PR, I did not try to solve many of the issues raised by the new config. Instead I mostly turn off the rules, with the goal to re-enable them and fix them in separate PRs.

We should perhaps do a sprint to try to fix a bunch of these. Ill try to see if we can possibly organize another MXCuBE-web code camp sometime :)

ui/eslint.config.js
marcus-oscarsson: `release` yes, it was used back in the day but not any longer

ui/src/components/Tasks/AddSample.jsx
marcus-oscarsson: Wow ;)

marcus-oscarsson (approved):

marcus-oscarsson:
Wonderful !, could you have a look at the `pacakge.json` and then we can merge :)

PR #1605, opened 2025-03-27
Fix: Invalid HSpace / VSpace value causes UI crash

mockoocy:
The bug occured, because when clearing the value of the input, `grid.cell[H|V]Space` was set to an empty string. Then attempt to call method `toFixed` of aforementioned value failed, because it's a method of type number, not string.

More specifically, the problem was the way how values were coerced from input string into a floating point value.
```javascript
    let value = Number.parseFloat(e.target.value);
    if (Number.isNaN(value)) {
      value = '';
    }
```
For an impossible coercion; that is string not representing a number (including an empty string) it returned an empty string, which later caused the crash.

### DIFF
--- a/ui/src/components/SampleView/SampleImage.jsx
+++ b/ui/src/components/SampleView/SampleImage.jsx
@@ -207,7 +207,7 @@ export default class SampleImage extends React.Component {
   setVCellSpacing(e) {
     let value = Number.parseFloat(e.target.value);
     if (Number.isNaN(value)) {
-      value = '';
+      value = 0;
     }
 
     const gridData = this.selectedGrid();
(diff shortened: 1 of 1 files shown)


ui/src/components/SampleView/SampleImage.jsx
```diff
@@ -209,7 +209,7 @@ export default class SampleImage extends React.Component {
   setVCellSpacing(e) {
     let value = Number.parseFloat(e.target.value);
     if (Number.isNaN(value)) {
-      value = '';
+      value = 0;
```
axelboc, line 210: Can the cell spacing really be set to `0`? Would it make more sense to have an early return and not call `setCellSpace` or `setCurrentCellSpace` at all maybe? Just wondering :shrug:

elmjag: I think cell spacing `0` is valid value. It means you don't have any extra space between your cells, when you do your mesh grid scans.

elmjag (approved):

ui/src/components/SampleView/SampleImage.jsx
marcus-oscarsson: Yes 0 is normally a valid value. Meaning no spacing.

PR #1608, opened 2025-03-28
Fix linting issues (2) — `func-style`, `no-shadow`

axelboc:
- [`func-style`](https://eslint.org/docs/latest/rules/func-style) enforces a consistent style for functions. More specifically, it forbids arrow function assignments in favour of function declarations. The goal, apart from consistency, is to more strongly distinguish functions from variables/constants.
- [`no-shadow`](https://eslint.org/docs/latest/rules/no-shadow) forbids shadowing variables, as doing so on purpose can make the code confusing to understand, and doing so by mistake can hide scope bugs and/or opportunities for simplification. It's usually a matter of renaming the variable in the inner scope or sometimes the one in the outer scope, but it can also be fixed in other ways depending on the situation, like with destructuring — whatever makes the code clearer.

### DIFF
--- a/ui/src/actions/remoteAccess.js
+++ b/ui/src/actions/remoteAccess.js
@@ -86,9 +86,9 @@ export function logoutUser(username) {
   };
 }
 
-export function respondToControlRequest(giveControl = true, message = '') {
+export function respondToControlRequest(giveCtrl = true, message = '') {
   return async (dispatch) => {
-    await sendRespondToControlRequest(giveControl, message);
+    await sendRespondToControlRequest(giveCtrl, message);
     dispatch(getLoginInfo());
     dispatch(getRaState());
   };
--- a/ui/src/components/SampleControls/ZoomControl.jsx
+++ b/ui/src/components/SampleControls/ZoomControl.jsx
@@ -11,9 +11,11 @@ const ZOOM_HWO_ID = 'diffractometer.zoom';
 function ZoomControl() {
   const dispatch = useDispatch();
 
-  const { state, value, commands } = useSelector(
+  const zoomObj = useSelector(
     (state) => state.beamline.hardwareObjects[ZOOM_HWO_ID],
   );
+
+  const { state, value, commands } = zoomObj;
 
   return (
     <OverlayTrigger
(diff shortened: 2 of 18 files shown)


ui/src/actions/remoteAccess.js
```diff
@@ -86,9 +86,9 @@ export function logoutUser(username) {
   };
 }
 
-export function respondToControlRequest(giveControl = true, message = '') {
+export function respondToControlRequest(giveCtrl = true, message = '') {
   return async (dispatch) => {
-    await sendRespondToControlRequest(giveControl, message);
+    await sendRespondToControlRequest(giveCtrl, message);
```
axelboc, line 91: Granted, it's not really more readable here... There's an action creator function called `giveControl` in the outer scope. I would have maybe renamed the parameter to `accept` or something, but the API expects a `giveControl` parameter, so I chose to keep the naming as close as possible.

ui/src/components/SampleControls/ZoomControl.jsx
```diff
@@ -11,10 +11,12 @@ const ZOOM_HWO_ID = 'diffractometer.zoom';
 function ZoomControl() {
   const dispatch = useDispatch();
 
-  const { state, value, commands } = useSelector(
+  const zoomObj = useSelector(
     (state) => state.beamline.hardwareObjects[ZOOM_HWO_ID],
   );
 
+  const { state, value, commands } = zoomObj;
```
axelboc, line 18: Note that the `state` variable is now declared after, so it's fine. It wouldn't have worked with the old `var` keyword.

ui/src/containers/GphlWorkflowParametersDialog.jsx
```diff
@@ -72,7 +72,7 @@ function GphlWorkflowParametersDialog(props) {
     show,
     updatedFormData,
     handleHide,
-    updateGphlWorkflowParameters,
+    updateGphlWorkflowParameters, // eslint-disable-line no-shadow
```
axelboc, line 75: Action creator imported and passed as prop. It will be refactored eventually, so no point in renaming now.

marcus-oscarsson:
Thanks !

PR #1609, opened 2025-03-28
Fix linting issues (3) — `no-unused-vars`

axelboc:
That's a big one but easy-enough. It's mostly about removing default `React` imports, which are not needed except when accessing `React.Component` (although I'd usually go with importing `{ Component } directly — but I didn't bother refactoring this since it concerns only class components).

For unused parameters that can't be removed because they come before a parameter that _is_ used, it's as easy as prefixing the parameter name with `_` or just naming it `_`.

### DIFF
--- a/ui/src/containers/SampleGridTableContainer.jsx
+++ b/ui/src/containers/SampleGridTableContainer.jsx
@@ -762,7 +762,7 @@ class SampleGridTableContainer extends React.Component {
         this.props.filterOptions.cellFilter.toLowerCase() === ''
       ) {
         const nbpuck = [];
-        cell.children.forEach((puck, puckidx) => {
+        cell.children.forEach((_, puckidx) => {
           if (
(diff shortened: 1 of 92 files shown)


ui/src/containers/SampleGridTableContainer.jsx
```diff
@@ -762,7 +762,7 @@ class SampleGridTableContainer extends React.Component {
         this.props.filterOptions.cellFilter.toLowerCase() === ''
       ) {
         const nbpuck = [];
-        cell.children.forEach((puck, puckidx) => {
+        cell.children.forEach((_puck, puckidx) => {
```
jbflo: maybe just underscore ( _ )   ? 
i kind of like the _ approche more
` cell.children.forEach((_, puckidx) => {`

jbflo (approved):

marcus-oscarsson:
Looks good :), thanks again !

PR #1610, opened 2025-03-28
Fix linting issues (4) — `simple-import-sort/imports`

axelboc:
Sort imports automatically with [`eslint-plugin-simple-import-sort`](https://github.com/lydell/eslint-plugin-simple-import-sort/).

Long overdue! :grin: 

Note that the plugin moves global CSS imports to the top. For instance, in `index.js`, the FontAwesome and ReactContextify styles imports move before the `App` component import, which means they'll end up before other components' global styles in the DOM.

This is a good thing but there's a very, very slight risk that this may lead to visual regressions. I highly doubt it, though — if anything, it's more likely that some global component styles will no longer need to overcome font awesome styles with `!important`, for instance.

Anyway, I didn't notice any regression, but I wouldn't mind another pair of eyes.

### DIFF
--- a/ui/cypress/support.js
+++ b/ui/cypress/support.js
@@ -1,5 +1,6 @@
 /* global Cypress, cy, beforeEach */
 import '@testing-library/cypress/add-commands';
+
 import installLogsCollector from 'cypress-terminal-report/src/installLogsCollector.js';
 
 installLogsCollector();
--- a/ui/eslint.config.js
+++ b/ui/eslint.config.js
@@ -17,7 +17,6 @@ const config = defineConfig([
       'react/no-multi-comp': 'off',
       'react/no-unused-class-component-methods': 'off',
       'react/prop-types': 'off',
-      'simple-import-sort/imports': 'off',
       'unicorn/prefer-global-this': 'off',
 
       /* Default is "avoid", but there are lots of complicated `switch` statements,
--- a/ui/src/actions/beamline.js
+++ b/ui/src/actions/beamline.js
@@ -1,7 +1,7 @@
 import {
-  sendSetAttribute,
   sendExecuteCommand,
   sendPrepareBeamlineForNewSample,
+  sendSetAttribute,
 } from '../api/beamline';
 import { sendLogFrontEndTraceBack } from '../api/log';
 
--- a/ui/src/actions/beamlineActions.js
+++ b/ui/src/actions/beamlineActions.js
@@ -1,8 +1,8 @@
-import { RUNNING } from '../constants';
 import {
   sendAbortBeamlineAction,
   sendRunBeamlineAction,
 } from '../api/beamline';
+import { RUNNING } from '../constants';
 
 export function addUserMessage(data) {
   return {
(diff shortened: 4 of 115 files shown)


axelboc:
I'll stop here for now, as that's already a lot to review, and if anything needs to be changed in the first PRs, I risk having trouble rebasing.

marcus-oscarsson:
Thanks for the linting work very nice @axelboc. I'm personally not a huge fan of sorting imports but I guess it possible avoids merge conflicts in the future. Anyways, that's what most projects are doing so looks good :)

fabcor-maxiv:
Side comment:

image

> Base automatically changed from ab-lint3 to develop

Where does this come from? How did this happen? Did you manage to make one pull request depend on another pull request?

marcus-oscarsson:
The first PR was made with `develop` with base and the subsequent with each other as bases and when one gets merged into `develop` and deleted the base is changed, right ?

axelboc:
Yeah, I had `develop > ab-lint1 > ab-lint2 > ...`. You just have to select the right base branch when you open the pull request. It can get a bit messy if you have to make a conflicting change in `ab-lint1` after opening `ab-lint2`, etc. so I usually stop at 2 levels when the changes are non-trivial.

After merging `ab-lint1`, the base branch of the PR for `ab-lint2` is updated automatically to `develop`, as you've noticed, but the branch is not rebased automatically. GitHub's "rebase branch" feature usually works to rebase `ab-lint2` on `develop`, but I tend to avoid it because due to the branch structure, it replays commits from `ab-lint1`, which are already in `develop` — those end up as empty commits which have to be skipped by the rebase. It usually works but again, can get messy in some scenarios (e.g. if `ab-lint1` is squash-merged).

The best is to use `git rebase --onto`:

- Pull `develop` locally
- Checkout `ab-lint2` (which needs to be rebased)
- Copy the sha of the last commit from `ab-lint1` that `ab-lint2` is based on
- Run `git rebase --onto develop <commit-sha>` => this ensures that only the commits from `ab-lint2` are replayed onto `develop`
- `ab-lint2` is now rebased => force push to update the PR.

fabcor-maxiv:
Oh wow... so it is still as complex as it ever was. I thought maybe GitHub had made improvements to the "stacked/dependent pull requests" workflow but apparently not. It is such an important workflow for me. I would use them all the time if there were supporting features directly in GitHub.

axelboc:
I don't find it complicated in practice, just need to get the hang of it. The key is to have a git graph visualizer to find the base commit sha :joy: I use VS Code's [_Git Graph_](https://marketplace.visualstudio.com/items?itemName=mhutchie.git-graph) extension for this.

fabcor-maxiv:
I mean: whatever the steps are, you still have to do all the steps manually, GitHub does not offer much help with that if any at all.

Seems like some external 3rd-party Git UI tools have some built-in functionalities for this kind of things. I read about [GitButler's "stacked branches" for example](https://blog.gitbutler.com/stacked-branches-with-gitbutler/) but I have not tried it.

I guess that for the time being I will still just stick to the extremely tedious workflow of sending one pull request, waiting until it gets merged, and then and only then sending the next pull request that depends on the previous one. At the end of the day, with all pros and cons weighed in, I feel like this is what has worked best in my experience.

PR #1612, opened 2025-04-01
Fix linting issues (5) — `unicorn/prefer-global-this`

axelboc:
Replace `window` with [`globalThis`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis), which is widely supported (FF 65+).

### DIFF
--- a/ui/src/containers/ConnectionLostDialog.jsx
+++ b/ui/src/containers/ConnectionLostDialog.jsx
@@ -16,7 +16,7 @@ export function ConnectionLostDialog() {
         the page.
       </Modal.Body>
       <Modal.Footer>
-        <Button onClick={() => window.location.reload(true)}> Reload </Button>
+        <Button onClick={() => globalThis.location.reload(true)}>Reload</Button>
       </Modal.Footer>
     </Modal>
   );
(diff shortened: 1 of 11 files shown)


ui/src/containers/ConnectionLostDialog.jsx
```diff
@@ -16,7 +16,10 @@ export function ConnectionLostDialog() {
         the page.
       </Modal.Body>
       <Modal.Footer>
-        <Button onClick={() => window.location.reload(true)}> Reload </Button>
+        <Button onClick={() => globalThis.location.reload(true)}>
+          {' '}
+          Reload{' '}
+        </Button>
```
fabcor-maxiv: That looks strange to me, but the rest seems fine.

Is it the way to keep leading and trailing empty spaces? Does the formatter force it to span over two lines?

fabcor-maxiv (approved):

ui/src/containers/ConnectionLostDialog.jsx
axelboc: Yeah there must have been a space before and after "Reload". Once you break it up on multiple lines, React no longer keeps that space so the formatter adds it back. Don't think we need those spaces though, good catch, I'll remove them.

walesch-yan (approved):
This looks good to me, Thanks!

PR #1613, opened 2025-04-02
small change on  JSX render logic

jbflo:
Fix JSX render logic
 - simplify render , manual samples were render twice.
 - Use a out scope Variable for menu ID, Instead of declaring multiple time

### DIFF
--- a/ui/src/containers/SampleGridTableContainer.jsx
+++ b/ui/src/containers/SampleGridTableContainer.jsx
@@ -46,6 +46,9 @@ const SETTINGS = {
   slidesToScroll: 6,
 };
 
+const CELL_MENU_ID = 'samples-grid-table-context-menu-cell';
+const PUCK_MENU_ID = 'samples-grid-table-context-menu-puck';
+
 class SampleGridTableContainer extends React.Component {
   constructor(props) {
     super(props);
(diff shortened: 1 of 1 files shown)


axelboc (approved):
Thanks a lot, 2-min review, love it! I'll try to be as reactive for the next ones.

walesch-yan (approved):

jbflo:
> Thanks a lot, 2-min review, love it! I'll try to be as reactive for the next ones.

you are welcome,  thanks for looking into this.
next one is a bit  complexes, but almost easy .  😇.

ui/src/containers/SampleGridTableContainer.jsx
```diff
@@ -1257,40 +1256,28 @@ class SampleGridTableContainer extends React.Component {
...
-          >
-            <div className="selection-rubber-band" id="selectionRubberBand" />
-            {this.renderManualSamples()}
-            {this.isSingleCell()
-              ? this.renderSampleTableSingleCell(2)
-              : this.renderSampleTable(6)}
-          </Row>
-        )}
+        <Row
+          className="samples-grid-table justify-content-center"
```
jbflo, line 1260: justify-content-center
Maybe that's what causing the problem [ATF] MXCuBE@ID23-1
@walesch-yan @marcus-oscarsson 
it was remove at https://github.com/mxcube/mxcubeweb/pull/1616/files which is not apply yet in the bemaline
I am not sure cause i was not able to reproduce the issue, but that's the only changed i saw.

PR #1614, opened 2025-04-02
Latest user messages should be shown on top in UI

walesch-yan:
The way the user messages were displayed in the UI was oldest on top and newest on bottom, which meant a lot of scrolling to see the latest logs.

I reversed the order of the messages in the UI and while at it I added the suggested change(#1546) to remove the render function in `UserMessage` component

### DIFF
--- a/ui/src/components/Notify/UserMessage.jsx
+++ b/ui/src/components/Notify/UserMessage.jsx
@@ -1,19 +1,16 @@
+import { useSelector } from 'react-redux';
+
 import styles from './UserMessage.module.css';
 
-export default function UserMessage(props) {
-  const { messages } = props;
+export default function UserMessage() {
+  const messages = useSelector((state) => state.logger.logRecords);
 
-  function renderMessages() {
-    const msg = [];
-
-    for (const [idx, message] of messages.entries()) {
-      const messageClass = `${styles.message} ${styles[message.severity]}`;
-
-      msg.push(
+  return (
+    <div id="usermessages" className={styles.messageBody}>
+      {[...messages].reverse().map((message) => (
         <div
-          key={`${message.id}-${idx}`}
-          ref={message.id}
-          className={messageClass}
+          key={`${message.timestamp}`}
+          className={`${styles.message} ${styles[message.severity]}`}
         >
           {message.severity === 'INFO' ? (
             <span className="fas fa-lg fa-check-circle" />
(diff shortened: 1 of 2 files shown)


ui/src/components/Notify/UserMessage.jsx
```diff
@@ -1,19 +1,16 @@
...
+import { useSelector } from 'react-redux';
 
-  function renderMessages() {
-    const msg = [];
+import styles from './UserMessage.module.css';
 
-    for (const [idx, message] of messages.entries()) {
-      const messageClass = `${styles.message} ${styles[message.severity]}`;
+export default function UserMessage() {
+  const messages = useSelector((state) => state.logger.logRecords);
```
walesch-yan, line 7: As bonus removed the `props` here and added a `useSelector`

ui/src/components/Notify/UserMessage.jsx
```diff
@@ -1,19 +1,16 @@
...
-      msg.push(
+  return (
+    <div id="usermessages" className={styles.messageBody}>
+      {[...messages].reverse().map((message) => (
         <div
-          key={`${message.id}-${idx}`}
-          ref={message.id}
-          className={messageClass}
+          key={`${message.timestamp}`}
+          className={`${styles.message} ${styles[message.severity]}`}
```
walesch-yan, line 13: The `ref` here was not used from what I saw and also there is no `id` parameter for the `logRecords` in my tests. I assume this has been removed, hence I replaced the key by the timestamp

ui/src/components/Notify/UserMessage.jsx
```diff
@@ -1,19 +1,16 @@
...
-  const { messages } = props;
+import { useSelector } from 'react-redux';
 
-  function renderMessages() {
-    const msg = [];
+import styles from './UserMessage.module.css';
 
-    for (const [idx, message] of messages.entries()) {
-      const messageClass = `${styles.message} ${styles[message.severity]}`;
+export default function UserMessage() {
```
axelboc, line 5: Might as well pluralize it (filename as well)?

axelboc (approved):
Great :)

ui/src/components/Notify/UserMessage.jsx
walesch-yan: Seems like I was too late to add this change :sweat_smile:

PR #1615, opened 2025-04-03
Call `prepare_directories` in Lims

dominikatrojanowska:
Solution for `prepare_directories` call in `Lims`.  
This solution works at least at Max IV. Passing the whole `session` object allows other facilities to pick required data.

Related PR:
* https://github.com/mxcube/mxcubecore/pull/1189 (has to be merged before)

### DIFF
--- a/mxcubeweb/core/components/lims.py
+++ b/mxcubeweb/core/components/lims.py
@@ -242,18 +242,7 @@ class Lims(ComponentBase):
             )
             self.allow_session(session)
 
-        if hasattr(HWR.beamline.session, "prepare_directories"):
-            try:
-                logging.getLogger("MX3.HWR").info(
-                    "[LIMS] Creating data directories for proposal %s%s",
-                    session.code,
-                    session.number,
-                )
-                HWR.beamline.session.prepare_directories(session)
-            except Exception:
-                logging.getLogger("MX3.HWR").info(
-                    "[LIMS] Error creating data directories, %s" % sys.exc_info()[1]
-                )
+        HWR.beamline.session.prepare_directories(session)
 
         # save selected proposal in users db
         current_user.selected_proposal = session.session_id
--- a/mxcubeweb/core/components/lims.py
+++ b/mxcubeweb/core/components/lims.py
@@ -249,7 +249,7 @@ class Lims(ComponentBase):
                     session.code,
                     session.number,
                 )
-                raise "To be implemented for those using prepare_directories"
+                HWR.beamline.session.prepare_directories(session)
             except Exception:
                 logging.getLogger("MX3.HWR").info(
                     "[LIMS] Error creating data directories, %s" % sys.exc_info()[1]
--- a/mxcubeweb/core/components/lims.py
+++ b/mxcubeweb/core/components/lims.py
@@ -245,7 +245,8 @@ class Lims(ComponentBase):
         if hasattr(HWR.beamline.session, "prepare_directories"):
             try:
                 logging.getLogger("MX3.HWR").info(
-                    "[LIMS] Creating data directories for proposal %s%s" % session.code,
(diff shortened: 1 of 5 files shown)


mxcubeweb/core/components/lims.py
```diff
@@ -245,10 +245,11 @@ def select_session(self, session_id: str) -> bool:
         if hasattr(HWR.beamline.session, "prepare_directories"):
             try:
                 logging.getLogger("MX3.HWR").info(
-                    "[LIMS] Creating data directories for proposal %s%s" % session.code,
+                    "[LIMS] Creating data directories for proposal %s%s",
+                    session.code,
                     session.number,
                 )
-                raise "To be implemented for those using prepare_directories"
+                HWR.beamline.session.prepare_directories(session)
```
marcus-oscarsson: Could this be done on the `mxcubecore` side ?

dominikatrojanowska: There is an implementation in the mxcubecore in prepare_directories` method... could you provide more details what do you mean? Do you prefer to avoid that part at all https://github.com/mxcube/mxcubeweb/blob/develop/mxcubeweb/core/components/lims.py#L245C9-L255C18 ?

marcus-oscarsson: What I thought is that it would be great if this logic could also be triggered in `mxcubecore`, so that we dont need `if hasattr(HWR.beamline.session, "prepare_directories"):`. We could add a signal/event if needed but I guess as this is when we "select_session" that it could be done in the LIMS object ?, I think that would make more sense or ?

marcus-oscarsson: Actually what you also could do is to add a method on the `Session` hardware object that simply does a "pass" and then the ones that actually needs to can implement some logic in their specific `Session` object i.e `MAXIVSession`. If it makes sense ?

beteva: I would also vote for putting the logic in mxcubecore with what @marcus-oscarsson suggests,

dominikatrojanowska: I tried to implement that logic in mxcubecore (at the end of `set_active_session_by_id` implementation), but in that case it seems that the data directory is created a bit to early...  I think it should be created after all the previous steps of `select_session` have passed (in example `allow_session` if needed)

I will follow the last comment of @marcus-oscarsson

dominikatrojanowska: https://github.com/mxcube/mxcubecore/pull/1189

mxcubeweb/core/components/lims.py
```diff
@@ -242,17 +242,7 @@ def select_session(self, session_id: str) -> bool:
...
-                logging.getLogger("MX3.HWR").info(
-                    "[LIMS] Creating data directories for proposal %s%s" % session.code,
-                    session.number,
-                )
-                raise "To be implemented for those using prepare_directories"
-            except Exception:
-                logging.getLogger("MX3.HWR").info(
-                    "[LIMS] Error creating data directories, %s" % sys.exc_info()[1]
-                )
+        HWR.beamline.session.prepare_directories(session)
```
marcus-oscarsson, line 245: :+1: Looks much nicer

test/conftest.py
```diff
@@ -56,9 +56,11 @@ def cleanup_subprocesses():
 
     for child in parent_process.children(recursive=True):
         if child.is_running():
-            child.terminate()
             try:
+                child.terminate()
                 child.wait(timeout=1)
+            except psutil.NoSuchProcess:
+                pass
```
walesch-yan, line 63: Interesting to see how the child processes continuously terminated themselves between the `is_running` and `terminate` calls. Good catch for changing this. Thanks!

walesch-yan (approved):

PR #1616, opened 2025-04-03
New Sample Flex View  (Sample List #2 )

jbflo:
This PR propose a complete remake of the SampleFlexView view
making the feature more accessible. 

- SampleFlexView was use as a Filter but it can now be use to collect easily a pre-filtered group of sample
the old code was useless because we could not view the pucks,  I'll suggest to focus on  the addition instead of the diff

- The Idea is to make this view useful enough  to be used in some beamline.
It allow you to navigate simpler from Cells and Pucks.
-  in the future and if necessary : when this view is active,  we could completely not display the table view  instead of applying a filter 
 

As the Mockup allow you to view only singleCell or only pucks samples.
you'll need to make some changes to SampleChangerMockup.py to be able to display this view. 

[Screencast from 2025-04-03 16-07-31.webm](https://github.com/user-attachments/assets/6a6c15c5-a620-4544-95a7-b05a991e454d)

- USECASE

[Screencast from 2025-04-03 16-51-36.webm](https://github.com/user-attachments/assets/381d7302-7fc2-45df-8e36-a81447eb5dbd)

### DIFF
--- a/ui/src/containers/SampleFlexView.jsx
+++ b/ui/src/containers/SampleFlexView.jsx
@@ -1,156 +1,231 @@
-import React from 'react';
+import { useEffect } from 'react';
 import { Col } from 'react-bootstrap';
-import { connect } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
 
 import { filterAction } from '../actions/sampleGrid';
 
-class NewSampleFlexView extends React.Component {
-  constructor(props) {
-    super(props);
+export default function SampleFlexView(props) {
+  const {
+    displayPuckCellContextMenu,
+    cellMenuID,
+    puckMenuID,
+    type = 'FLEX HCD',
+  } = props;
 
-    this.renderCircle = this.renderCircle.bind(this);
-    this.isCellSelected = this.isCellSelected.bind(this);
-  }
+  const filterOptions = useSelector((state) => state.sampleGrid.filterOptions);
+  const scContent = useSelector((state) => state.sampleChanger.contents);
 
-  componentDidMount() {
-    document.addEventListener('keydown', this.onKeyDown, false);
-  }
+  const numberOfCell = scContent.children.length; // Number of cells/slices
 
-  onClickCell(event, idx, disabled) {
-    if (!disabled) {
-      this.props.filter({ cellFilter: `${idx + 1}` });
-    }
+  const dispatch = useDispatch();
+
+  function handleClickOnCellPuck(event, cellID, puckID) {
+    dispatch(
+      filterAction({ cellFilter: `${cellID}`, puckFilter: `${puckID}` }),
+    );
     event.stopPropagation();
   }
 
-  isCellSelected(cell) {
-    let isCellSelected = false;
-    if (Number(this.props.filterOptions.cellFilter) === cell) {
-      isCellSelected = true;
-    }
-    return isCellSelected;
+  function isCellSelected(cellID) {
+    return Number(filterOptions.cellFilter) === cellID;
   }
 
-  renderCircle(nbCell, idx, isCellSelected) {
-    const cellOrder = [8, 7, 6, 5, 4, 3, 2, 1];
-    const percentage = (100 / nbCell) * cellOrder[idx];
-    const circumference = 2 * Math.PI * 5; // 2*pi*rayon
-    const strokeDash = (percentage * circumference) / 100;
+  function isPuckSelected(cellID, puckID) {
+    return (
+      Number(filterOptions.cellFilter) === cellID &&
+      Number(filterOptions.puckFilter) === puckID
+    );
+  }
 
-    const angle = 2 * Math.PI * (strokeDash / circumference);
+  useEffect(() => {
+    if (filterOptions.cellFilter === '') {
+      dispatch(filterAction({ cellFilter: `${1}`, puckFilter: '' }));
+    }
+  }, [filterOptions.cellFilter, dispatch]);
 
-    const strokeDashArray = `${strokeDash} ${circumference}`;
-    const rgbColorList = [
-      '#e8ebee',
-      '#cdced1',
-      '#e8ebee',
-      '#cdced1',
-      '#e8ebee',
-      '#cdced1',
-      '#e8ebee',
-      '#cdced1',
-    ];
+  async function handleDisplayPuckCellContextMenu(e, menuID, cellID, puckID) {
+    e.preventDefault();
+    await handleClickOnCellPuck(e, cellID, puckID === null ? '' : puckID);
+    displayPuckCellContextMenu(e, menuID, cellID, puckID);
+    e.stopPropagation();
+  }
 
-    const disableClasse =
-      this.props.cellSampleList(idx + 1)[0].length > 0
-        ? 'has-sample'
-        : 'empty-cell';
+  function getCellSlicePath(idxCell, radius = 8, centerX = 10, centerY = 10) {
+    // Calculate start and end angles for each slice
+    const angleStart = (idxCell * 360) / numberOfCell;
+    const angleEnd = ((idxCell + 1) * 360) / numberOfCell;
 
-    const disabled = this.props.cellSampleList(idx + 1)[0].length <= 0;
+    // Convert angles to radians
+    const startX = centerX + radius * Math.cos((angleStart * Math.PI) / 180);
+    const startY = centerY + radius * Math.sin((angleStart * Math.PI) / 180);
+    const endX = centerX + radius * Math.cos((angleEnd * Math.PI) / 180);
+    const endY = centerY + radius * Math.sin((angleEnd * Math.PI) / 180);
+
+    // Create each Cell like a pie slice path (triangle shape)
+    return `M${centerX},${centerY} L${startX},${startY} A${radius},${radius} 0 0,1 ${endX},${endY} Z`;
+  }
+
+  function getCirclesInSlice(idxCell, centerX = 10, centerY = 10) {
+    // numberOfPuck: Number of pucks/circles inside each slice/Cell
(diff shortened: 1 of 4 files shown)


ui/src/containers/SampleFlexView.jsx
```diff
@@ -1,156 +1,245 @@
...
-      '#cdced1',
-    ];
-
-    const disableClasse =
-      this.props.cellSampleList(idx + 1)[0].length > 0
-        ? 'has-sample'
-        : 'empty-cell';
-
-    const disabled = this.props.cellSampleList(idx + 1)[0].length <= 0;
+  function getCirclesInSlice(idxCell, centerX = 10, centerY = 10) {
```
jbflo, line 68: Some Value were Hard coded here.
i could apply a formula but it becomes more complicated.

ui/src/containers/SampleListViewContainer.jsx
```diff
@@ -870,7 +870,7 @@ class SampleListViewContainer extends React.Component {
                     variant="outline-secondary"
                     id="dropdown-basic"
                   >
-                    <MdGridView size="1em" /> View Mode
+                    <MdGridView size="1em" /> {this.props.viewMode.mode}
                   </Dropdown.Toggle>
                   <Dropdown.Menu>
                     {this.props.viewMode.options.map((option) => (
```
jbflo, line 876: we could also remove the one is is active from the options.  or Highlight it

ui/src/containers/SampleGridTableContainer.jsx
```diff
@@ -1257,7 +1268,7 @@ class SampleGridTableContainer extends React.Component {
           </MXContextMenu>
         ) : null}
         <Row
-          className="samples-grid-table justify-content-center"
+          className="samples-grid-table"
```
jbflo, line 1271: this cause a ugly css transition.  the transition is not smooth enough.

ui/src/containers/SampleGridTableContainer.jsx
```diff
@@ -1233,16 +1233,27 @@ class SampleGridTableContainer extends React.Component {
   }
 
   renderSampleChangerDrawing() {
-    if (this.props.type.includes('CATS')) {
+    if (this.props.type.includes('CATS') || this.isSingleCell()) {
       return <SampleIsaraView cellSampleList={this.getSampleListBydCell} />;
```
jbflo, line 1237: Force the  ISARA as  default Graphical view. as single cell is by  default in the Mockup.
. @meguiraun   are you satisfy with the  SampleIsaraView component ?

ui/src/containers/SampleFlexView.jsx
```diff
@@ -1,156 +1,245 @@
...
-    this.isCellSelected = this.isCellSelected.bind(this);
+  const filterOptions = useSelector((state) => state.sampleGrid.filterOptions);
+  const scContent = useSelector((state) => state.sampleChanger.contents);
+  const [pendingEvent, setPendingEvent] = useState(null);
+
+  const numberOfCell = scContent.children.length; // Number of cells/slices
+
+  const dispatch = useDispatch();
+
+  function handleClickonCellPuck(event, cellID, puckID) {
```
axelboc: `handleClickOnCellPuck`

ui/src/containers/SampleFlexView.jsx
```diff
@@ -1,156 +1,245 @@
...
+    return Number(filterOptions.cellFilter) === cellID;
   }
 
-  onClickCell(event, idx, disabled) {
-    if (!disabled) {
-      this.props.filter({ cellFilter: `${idx + 1}` });
+  function isPuckSelected(cellID, puckID) {
+    return (
+      Number(filterOptions.cellFilter) === cellID &&
+      Number(filterOptions.puckFilter) === puckID
```
axelboc, line 36: I see a few number/string conversions (here, in the dispatch above, etc.) This tells me that perhaps `cellFilter` and `puckFilter` should be stored as numbers in the Redux store?

ui/src/containers/SampleFlexView.jsx
```diff
@@ -1,156 +1,245 @@
...
+  const scContent = useSelector((state) => state.sampleChanger.contents);
+  const [pendingEvent, setPendingEvent] = useState(null);
+
+  const numberOfCell = scContent.children.length; // Number of cells/slices
+
+  const dispatch = useDispatch();
+
+  function handleClickonCellPuck(event, cellID, puckID) {
+    dispatch(
+      filterAction({ cellFilter: `${cellID}`, puckFilter: `${puckID}` }),
```
axelboc, line 24: `filterAction` sets `filterOptions` in the store, right? Perhaps the naming should be made explicit - e.g. `setFilterOptions`?

ui/src/containers/SampleFlexView.jsx
```diff
@@ -1,156 +1,245 @@
...
-    if (!disabled) {
-      this.props.filter({ cellFilter: `${idx + 1}` });
+  function isPuckSelected(cellID, puckID) {
+    return (
+      Number(filterOptions.cellFilter) === cellID &&
+      Number(filterOptions.puckFilter) === puckID
+    );
+  }
+
+  // Call displayPuckCellContextMenu only after Redux state updates
```
axelboc: This feels very hacky. You can call something after a Redux state update by waiting for the promise returned by a dispatch call to resolve:

```js
await dispath(...);
displayPuckCellContextMenu(...);
```

Saves you from needing that `pendingEvent` local state.

ui/src/containers/SampleFlexView.jsx
```diff
@@ -1,156 +1,245 @@
...
+
+  useEffect(() => {
+    if (filterOptions.cellFilter === '') {
+      dispatch(filterAction({ cellFilter: `${1}`, puckFilter: '' }));
     }
+  }, [filterOptions.cellFilter, dispatch]);
+
+  function handleDisplayPuckCellContextMenu(event, menuID, cellID, puckID) {
+    event.preventDefault();
+    handleClickonCellPuck(event, cellID, puckID === null ? '' : puckID);
```
axelboc: Would it be simpler to dispatch the action directly here (and `await` it as suggested in my previous comment)?

ui/src/containers/SampleFlexView.jsx
```diff
@@ -1,156 +1,245 @@
...
+        pendingEvent.menuID,
+        pendingEvent.cellID,
+        pendingEvent.puckID,
+      );
+      setPendingEvent(null);
+    }
+  }, [filterOptions, pendingEvent, displayPuckCellContextMenu]);
+
+  useEffect(() => {
+    if (filterOptions.cellFilter === '') {
```
axelboc, line 55: When/where is `cellFilter` set to `''`? If it's an event handler, you could move the `dispatch` there.

ui/src/containers/SampleFlexView.jsx
```diff
@@ -1,156 +1,245 @@
...
     event.stopPropagation();
   }
 
-  isCellSelected(cell) {
-    let isCellSelected = false;
-    if (Number(this.props.filterOptions.cellFilter) === cell) {
-      isCellSelected = true;
-    }
-    return isCellSelected;
+  function getCellSlicePath(idxCell, radius = 8, centerX = 10, centerY = 10) {
```
axelboc, line 53: With one more parameter, `numberOfCell`, this could be moved to a separate utility file. Same with `getTextPosition` I think?

ui/src/containers/SampleFlexView.jsx
jbflo: I think i try an async await and it did not work.  could not see why.

ui/src/containers/SampleFlexView.jsx
jbflo: Yeah agreed,  but maybe we can do the rename in a specific PR, because a lot more of the Action Reducers function are ending by Action  see in ../mxcubeweb/ui/src/actions/...

and this will require to touch  3 more files. maybe that's not what we wanted right now ?

axelboc: Yeah for sure, no worries :+1:

ui/src/containers/SampleFlexView.jsx
axelboc: If all else fails, `setTimeout(() => { displayPuckCellContextMenu(...) }, 0)` after `await dispath(...)` might work as a slightly cleaner workaround.

ui/src/containers/SampleFlexView.jsx
jbflo: redux Store expected a String, because of the event.taget.value from the Actual Filter option component in SampleListViewContainer .
we should have converted it to number Number(event.taget.value) .
For now if we store as number it will failed. but yeah i note that for when we will refactor the SampleListViewContainer

ui/src/containers/SampleFlexView.jsx
jbflo: i'll refactor, then I'll extract util at PR#4

ui/src/containers/SampleFlexView.jsx
jbflo: very Good suggestion, i try await again and it work without the the need of setTimeout.
When i try it the first time it behave differently maybe because of something else at that time. 

but i was happy to know the hack by the way AHAHa

ui/src/containers/SampleGridTableContainer.jsx
marcus-oscarsson: @meguiraun, no longer works at MAXIV :), its probably better to ask @elmjag

marcus-oscarsson:
Thanks @jbflo and @axelboc for the review !

PR #1619, opened 2025-04-07
Adds support for `create_point` acquisition parameter

mockoocy:
Generic tasks may define `create_point` parameter, which will trigger creating 2d point on the backend.

It's purpose is an quality of life improvement, in which user may click on a point on canvas, where no shape is defined, in order to create point and run a task on the created point.

Also it fixes order of parameters that are passed to `ContextMenu.showModal` in `createPointAndShowModal` function.

### DIFF
--- a/ui/src/components/SampleView/ContextMenu.jsx
+++ b/ui/src/components/SampleView/ContextMenu.jsx
@@ -79,6 +79,17 @@ export default class ContextMenu extends React.Component {
           key: `${task.name}`,
         });
       }
+
+      if (task.requires.includes('no_shape_2d')) {
+        genericTasks.none.push({
+          text: task.name,
+          action: () =>
+            this.createPointAndShowModal('Generic', {
+              type: tname,
+            }),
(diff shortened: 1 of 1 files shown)


mockoocy:
According to the conversation we had on zoom this morning, I changed it, so that it creates the point  if `BaseQueueEntry.REQUIRES`  contains `no_shape_2d` value.

axelboc (approved):

ui/src/components/SampleView/ContextMenu.jsx
```diff
@@ -79,6 +79,17 @@ export default class ContextMenu extends React.Component {
           key: `${task.name}`,
         });
       }
+
+      if (task.requires.includes('no_shape_2d')) {
```
marcus-oscarsson, line 83: :) ok

walesch-yan (approved):

marcus-oscarsson:
Thanks @mockoocy

PR #1620, opened 2025-04-08
Transform to Function Component (Sample List #3 )

jbflo:
this is a big file we can always do further clean up and extract some component later. 
But for now to keep it simple i just transform to functional component 
- remove bind(this)
- removed this. and this.props
- [Remove mapping options] when applying fitter

### DIFF
--- a/ui/src/containers/SampleListViewContainer.jsx
+++ b/ui/src/containers/SampleListViewContainer.jsx
@@ -1,6 +1,4 @@
-import '../components/SampleGrid/SampleGridTable.css';
-
-import React from 'react';
+import { useCallback, useEffect } from 'react';
 import {
   Button,
   Card,
@@ -12,17 +10,15 @@ import {
   InputGroup,
   Row,
 } from 'react-bootstrap';
-import ReactDOM from 'react-dom';
 import { LuSettings2 } from 'react-icons/lu';
 import { MdGridView } from 'react-icons/md';
-import { connect } from 'react-redux';
-import { bindActionCreators } from 'redux';
+import { useDispatch, useSelector } from 'react-redux';
+import { useNavigate } from 'react-router-dom';
 
 import { showConfirmClearQueueDialog } from '../actions/general';
 import {
   addSamplesToQueue,
   deleteSamplesFromQueue,
-  deleteTask,
   deleteTaskList,
   setEnabledSample,
   stopQueue,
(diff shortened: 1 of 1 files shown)


ui/src/containers/SampleListViewContainer.jsx
```diff
@@ -111,45 +99,45 @@ class SampleListViewContainer extends React.Component {
...
-      limsSamples: this.props.filterOptions.limsSamples,
-      filterText: this.props.filterOptions.text,
-      cellFilter: this.props.filterOptions.cellFilter,
-      puckFilter: this.props.filterOptions.puckFilter,
+      inQueue: filterOptions.inQueue,
+      notInQueue: filterOptions.notInQueue,
+      collected: filterOptions.collected,
+      notCollected: filterOptions.notCollected,
+      limsSamples: filterOptions.limsSamples,
+      text: filterOptions.text,
```
axelboc: Did you mean to change this property name?

ui/src/containers/SampleListViewContainer.jsx
```diff
@@ -415,9 +388,9 @@ class SampleListViewContainer extends React.Component {
       collected: { collected: e.target.checked },
       notCollected: { notCollected: e.target.checked },
       limsSamples: { limsSamples: e.target.checked },
-      filterText: { text: ReactDOM.findDOMNode(this.filterInput).value.trim() }, // eslint-disable-line react/no-find-dom-node
+      text: { text: e.target.value.trim() },
```
axelboc: Same here about the property name.

(nicely done replacing `findDOMNode` :+1: )

axelboc (approved):
Awesome, so glad to see this class component go! :tada:

ui/src/containers/SampleListViewContainer.jsx
jbflo: No , just replace  this.props.filterOptions to just filterOptions

ahh maybe we can just use  filterOptions[id] and skip the optionMap

jbflo: nice in this case the mapping is not longer need right.

axelboc: I was talking about the property name `text`, which was called `filterText` before. 

Since the E2E tests fail only in this PR, it's likely something to do with the refactoring.

ui/src/containers/SampleListViewContainer.jsx
```diff
@@ -430,27 +374,29 @@ class SampleListViewContainer extends React.Component {
...
-    });
+  function sampleGridClearFilter() {
+    dispatch(
+      filterAction({
+        inQueue: false,
+        notInQueue: false,
+        collected: false,
+        notCollected: false,
+        limsSamples: false,
+        text: '',
```
axelboc, line 385: The change from `filterText` to `text` is still present here.

ui/src/containers/SampleListViewContainer.jsx
axelboc: Could also be an action that is not dispatched or something of the sort.

axelboc: Yes, so the button is not working locally either and it's because the `onClick` handler calls the `showConfirmClearQueueDialog` action creator but does not dispatch the returned action. It's a common challenge when refactoring with `useDispatch` — it's easy to forget to replace action creators previously bound with `mapDispatchToProps` with explicit `dispatch()` calls.

ui/src/containers/SampleListViewContainer.jsx
jbflo: > The change from `filterText` to `text` is still present here.
 
Yes, filterText to text was legit, as it was define as text in the Store
`filterText: { text: ReactDOM.findDOMNode(this.filterInput).value.trim() },`
i think that why they had to do the mapping and use ReactDOM.findDOMNode.

now i think you notice  filterText is a better key name, we can always change it in the Store. 
not in the component as it was.

ui/src/containers/SampleListViewContainer.jsx
jbflo: > Yes, so the button is not working locally either and it's because the `onClick` handler calls the `showConfirmClearQueueDialog` action creator but does not dispatch the returned action. It's a common challenge when refactoring with `useDispatch` — it's easy to forget to replace action creators previously bound with `mapDispatchToProps` with explicit `dispatch()` calls.

Nice , thanks a lot, i totally missed some Dispatch !

ui/src/containers/SampleListViewContainer.jsx
axelboc: Oh I get you, thanks for clarifying :+1:

ui/src/containers/SampleListViewContainer.jsx
axelboc: I did a quick sanity check of all the action creators used in the component and they all seem to be wrapped with `dispatch` now :+1:

marcus-oscarsson:
> Awesome, so glad to see this class component go! 🎉

Same here ;) :+1:

PR #1621, opened 2025-04-08
Fix linting issues (6) — `prefer-destructuring`

axelboc:
> Apologies, I initially pushed these two commits to `develop` by mistake. I've reverted them and am now opening a proper Pull Request. I was able to push to `develop` because I'm an admin — I've now configured by local set up to prevent future mistakes with `git config branch.develop.pushRemote no_push` — cf. https://stackoverflow.com/a/57998309
> EDIT: Marcus was also able to remove the ability to push for admins.

Anyways, so I reenable and fix the `prefer-destructuring` rule, upgrade ESLint and upgrade the linting config.

ESLint 9.24.0 actually comes with an awesome new feature: there's now a way to [suppress violations](https://eslint.org/blog/2025/04/introducing-bulk-suppressions/) on existing code so that rules that are too unreasonable to fix in the entire codebase can still benefit new code. More on this soon.

The linting config upgrade relaxes a few rules, notably `no-continue`, `react/jsx-no-leaked-render` and `prefer-destructuring` (when assigning existing variables).

### DIFF
--- a/ui/src/components/SSXChip/SSXChip.jsx
+++ b/ui/src/components/SSXChip/SSXChip.jsx
@@ -328,20 +328,16 @@ export default class SSXChip extends React.Component {
   initChipCanvas() {
     const currentChipLayout =
       this.props.chipLayoutList[this.props.currentLayoutName];
-    const chipConfig = currentChipLayout.sections[0];
+    const [chipConfig] = currentChipLayout.sections;
 
     const numRows = chipConfig.number_of_rows;
     const numCols = chipConfig.number_of_collumns;
-    const blockSizeX = chipConfig.block_size[0];
-    const blockSizeY = chipConfig.block_size[0];
+    const [blockSizeX, blockSizeY] = chipConfig.block_size;
     const rowLabels = chipConfig.row_labels;
     const colLabels = chipConfig.column_lables;
 
-    const offset = chipConfig.block_spacing[0];
-    const spacing = chipConfig.block_spacing[0];
-
-    const numTargetsX = chipConfig.targets_per_block[0];
-    const numTargetsY = chipConfig.targets_per_block[1];
+    const [offset, spacing] = chipConfig.block_spacing;
+    const [numTargetsX, numTargetsY] = chipConfig.targets_per_block;
 
     const canvasWidth = numCols * (blockSizeX + spacing) + offset + blockSizeX;
     const canvasHeight = numRows * (blockSizeY + spacing) + offset + blockSizeY;
--- a/ui/src/containers/ConfirmCollectDialog.jsx
+++ b/ui/src/containers/ConfirmCollectDialog.jsx
@@ -303,13 +303,13 @@ export class ConfirmCollectDialog extends React.Component {
             </thead>
             <tbody id="table-body">
               {tasks.map((task) => {
-                let { parameters } = task;
+                const { parameters } =
+                  task.type === 'Interleaved'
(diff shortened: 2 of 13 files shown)


ui/src/components/SSXChip/SSXChip.jsx
```diff
@@ -328,20 +328,16 @@ export default class SSXChip extends React.Component {
   initChipCanvas() {
     const currentChipLayout =
       this.props.chipLayoutList[this.props.currentLayoutName];
-    const chipConfig = currentChipLayout.sections[0];
+    const [chipConfig] = currentChipLayout.sections;
 
     const numRows = chipConfig.number_of_rows;
     const numCols = chipConfig.number_of_collumns;
-    const blockSizeX = chipConfig.block_size[0];
-    const blockSizeY = chipConfig.block_size[0];
```
axelboc, line 336: I think there were bugs here and below (duplicate `[0]` access).

ui/src/containers/ConfirmCollectDialog.jsx
```diff
@@ -303,14 +303,14 @@ export class ConfirmCollectDialog extends React.Component {
...
+                const { parameters } =
+                  task.type === 'Interleaved'
+                    ? task.parameters.wedges[0]
+                    : task;
+
                 const sample = this.props.sampleGrid.sampleList[task.sampleID];
                 const sampleName = `${sample.sampleName} - ${sample.proteinAcronym}`;
 
-                if (task.type === 'Interleaved') {
-                  parameters = task.parameters.wedges[0].parameters;
```
axelboc, line 311: I've relaxed assignments to existing variables, so this actually passes now, but I thought the ternary was nicer.

marcus-oscarsson:
Thanks. So yes, now no one can push not even using `-f` to develop. If we need to Ill explicitly add a rule allowing push for when its needed and then remove it again ;)

fabcor-maxiv:
> Thanks. So yes, now no one can push not even using `-f` to develop. If we need to Ill explicitly add a rule allowing push for when its needed and then remove it again ;)

I approve of this.

And I jump on the occasion to promote the use of forks once again. :D
I believe one should never push to the main repository.
I know it is not going to happen anytime soon that we get consensus on this and enforce it.
I see it as another layer that protects myself from doing this kind of blunders. :D

marcus-oscarsson:
You do as you like, if you like to use your own forks you can, if you want to use a branch you can. I see no reason for putting a guideline for this.

> I see it as another layer that protects myself from doing this kind of blunders. :D

That would in my opinion be the wrong reason for using a fork

fabcor-maxiv:
> > I see it as another layer that protects myself from doing this kind of blunders. :D
> 
> That would in my opinion be the wrong reason for using a fork

That is THE reason of using forks, and the whole concept of how collaborations were designed to work on GitHub as I understood it... but maybe I misunderstood.

marcus-oscarsson:
I agree that GitHub was designed with collaboration in mind but one can then, as we are doing debate, which solution is the easiest to use; forks or branches.  As I mentioned earlier, we haven’t found any strong reason to recommend one over the other, so creating a guideline for it doesn't seem necessary at this point. We could add one If there are strong arguments for one solution over the other. I guess it depends on where the discussion takes us.

However I don't consider that forks are meant to be used as a protection against accidental pushes that I guess could happen in anyways, just that it would happen on the fork. Probably with less impact. 

Anyway, let's pick this discussion up another time and place.

PR #1623, opened 2025-04-09
Fix linting issues (7) — `promise/always-return`

axelboc:
I've loosened [the rule](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/always-return.md) in `@esrf/eslint-config` to ignore the [last `then` callback](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/always-return.md#ignorelastcallback), which is often used for side effects.

The option is not smart enough to identify the last `then` when it's followed by a catch, like in `login.js`. However, this prompted me to clean the promise chains in that file a little bit: instead of mutating a local `state` variable in every `then` callback, I return "state slices", which I then combine at the end with `Object.assign`.

### DIFF
--- a/ui/src/actions/login.js
+++ b/ui/src/actions/login.js
@@ -86,111 +86,78 @@ export function signOut() {
 
 export function getInitialState() {
   return async (dispatch) => {
-    const state = {};
-
-    const pchains = [
+    const initialStateSlices = await Promise.all([
       fetchUIProperties()
-        .then((json) => {
-          state.uiproperties = json;
-        })
+        .then((uiproperties) => ({ uiproperties }))
         .catch(notify),
       fetchQueueState()
-        .then((json) => {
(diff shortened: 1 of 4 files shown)


ui/src/actions/login.js
```diff
@@ -86,111 +86,78 @@ export function signOut() {
...
-        })
+        .then((general) => ({ general }))
         .catch(notify),
-    ];
-
-    await Promise.all(pchains);
+    ]);
 
-    dispatch(setInitialState(state));
+    dispatch(setInitialState(Object.assign({}, ...initialStateSlices)));
```
marcus-oscarsson, line 155: ok, yeah thats easier to follow

marcus-oscarsson:
Thanks @axelboc

PR #1624, opened 2025-04-11
adding unique identifiers to user-level logs

walesch-yan:
In #1614, I changed the key of the log messages shown to the user to be the timestamp, as the messages did not have a more unique identifier. However, this could lead to a problem when multiple log messages were added at the same time. 
See below a video on what happens if such a scenario is triggered:

problem_log_keys

In this case I added a second test log to the 3-click-centring cancelation message. As one can see the message gets duplicated and stays on top of the list. This is due to how React is updating components. Since there are multiple components with the same key, only one gets re-rendered, the others remain.
To fix this, I added a unique identifier id to each log that gets added, and use this id as key for the rendered items
 See result:
solved_log_keys

### DIFF
--- a/mxcubeweb/logging_handler.py
+++ b/mxcubeweb/logging_handler.py
@@ -1,4 +1,5 @@
 import logging
+import uuid
 
 
 class MX3LoggingHandler(logging.handlers.BufferingHandler):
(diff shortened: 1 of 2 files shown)


axelboc (approved):
Good call :100:

marcus-oscarsson:
Thanks

PR #1626, opened 2025-04-14
Refactor MXContextMenu

walesch-yan:
This PR refactors the MXContextMenu into a functional component and replaces the `.css` file with a module.

2 minor changes:
- I removed the closing animation, as it was called directly with the change to the show attribute, which meant that the element disappeared before the animation was playing anyways.
- There was a check in x direction so that the context menu would overflow. I added a check in y-direction which means less scrolling when selecting a context menu for samples displayed at the bottom of the screen


ui/src/components/GenericContextMenu/MXContextMenu.jsx
```diff
@@ -1,59 +1,58 @@
...
     }
-  }
+  }, [contextMenu]);
 
-  hideContextMenu() {
-    const ctxMenu = document.querySelector('#generic-contextMenu');
-    if (ctxMenu) {
-      ctxMenu.classList.add('generic-context-menu-close');
+  useEffect(() => {
+    function hideContextMenu() {
```
axelboc, line 37: Might be worth investigating the use of Bootstrap's `Overlay`, which has a `rootClose` prop to deal with hiding on click anywhere else. Anyway, not for this PR, it's more FYI in the broader scope of merging together the various context menu implementations present in the codebase.

ui/src/components/GenericContextMenu/MXContextMenu.jsx
```diff
@@ -1,59 +1,58 @@
...
-    const contextMenu = document.querySelector('#generic-contextMenu');
-    if (contextMenu) {
+import { useDispatch, useSelector } from 'react-redux';
+
+import { showGenericContextMenu } from '../../actions/sampleGrid';
+import styles from './MXContextMenu.module.css';
+
+export default function MXContextMenu(props) {
+  const { children } = props;
+  const contextMenu = useSelector(
```
axelboc: I would destructure this, maybe on a separate line, and then list the specific variables needed in the `useEffect` dependencies array.

axelboc (approved):
Looks good to me!

ui/src/components/GenericContextMenu/MXContextMenu.jsx
walesch-yan: Oh interesting didn't know about that, I will definitely check that out thanks!

ui/src/components/GenericContextMenu/MXContextMenu.jsx
walesch-yan: Makes sense, I will update:)

ui/src/components/GenericContextMenu/MXContextMenu.jsx
```diff
@@ -1,59 +1,58 @@
...
+
+  useEffect(() => {
+    if (show && menuRef.current) {
+      const menu = menuRef.current;
       const windowWidth = document.body.offsetWidth;
-      const menuEndPos = x + contextMenu.offsetWidth;
-
-      let posxoffset = 10;
-      if (menuEndPos > windowWidth) {
-        posxoffset = contextMenu.offsetWidth;
```
jbflo, line 30: @walesch-yan  sorry a bit late 
 i think this PR introduce a regression
the context menu lost it's click  position when the page is scroll

jbflo (commented):
a bit late but i think we need to check the Context menu position when the page scroll. that was the specificity of the MXContextMenu : be able to track the mouse click position when the page scroll .

ui/src/components/GenericContextMenu/MXContextMenu.jsx
walesch-yan: You are right. The idea was to replace the menu inside the viewable window (in y-direction) to prevent scrolling, but it is missing the value of the offset from the top of the page (I.e. when scrolling down a little, it will always pop up on top as we are below the the windowheight offset).

walesch-yan: I will make a new PR to fix this

walesch-yan: Thanks for pointing it out!:)

PR #1628, opened 2025-04-15
Transform ContextMenu into functional component

walesch-yan:
This PR transforms `ContextMenu` into a functional component and removes the `listOptions` function from the component, instead renders the elements directly in the functional component.

To keep the scope of this PR small, there will likely be follow-ups this will likely include the refactoring of `menuOptions` function and moving the `getControlAvailability` function from the parent container to this component. In fact I left it out of this PR, because otherwise the linter would like the parent container to be a functional component, which would exceed the scope of this PR.

### DIFF
--- a/ui/src/components/SampleView/ContextMenu.jsx
+++ b/ui/src/components/SampleView/ContextMenu.jsx
@@ -1,7 +1,23 @@
-import React from 'react';
 import { Dropdown } from 'react-bootstrap';
 import { createPortal } from 'react-dom';
+import { useDispatch, useSelector } from 'react-redux';
 
+import { showErrorPanel } from '../../actions/general';
+import { updateTask } from '../../actions/queue';
+import {
+  abortCentring,
+  acceptCentring,
+  add2DPoint,
+  addShape,
+  deleteShape,
+  measureDistance,
+  moveToBeam,
+  moveToPoint,
+  showContextMenu,
+  stopClickCentring,
+  toggleDrawGrid,
+} from '../../actions/sampleview';
+import { showTaskForm } from '../../actions/taskForm';
 import { getLastUsedParameters } from '../Tasks/fields';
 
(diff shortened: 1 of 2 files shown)


ui/src/components/SampleView/ContextMenu.jsx
```diff
@@ -476,62 +491,49 @@ export default class ContextMenu extends React.Component {
...
-        >
-          {optionList}
-        </Dropdown.Menu>
-      </Dropdown>,
-      document.body,
-    );
-  }
+        {optionsList.map(
+          (type) =>
+            type.text &&
```
axelboc: I would just replace this `type.text &&` condition with a `optionsList.filter(type => !!type.text).map(..)` for readability.

ui/src/containers/SampleViewContainer.jsx
```diff
@@ -179,23 +178,7 @@ class SampleViewContainer extends Component {
           <Col sm={6}>
             <DefaultErrorBoundary>
               <ContextMenu
-                {...this.props.contextMenu}
                 getControlAvailability={this.getControlAvailability}
-                sampleViewActions={this.props.sampleViewActions}
```
axelboc, line 184: One fewer use of that `sampleViewActions` option, which brings us closer to removing the namespace import `import * as sampleViewActions`. :tada:

axelboc (approved):
Nice! Great first step, looking forward to the next one.

I tried refactoring `menuOptions` once ... but gave up. :joy: Let's have a quick chat (I'll come to you).

ui/src/components/SampleView/ContextMenu.jsx
walesch-yan: Added the filter, thanks for the suggestion!:)

marcus-oscarsson:
Thanks, great initiative !

PR #1629, opened 2025-04-16
refactor QueueSettings

walesch-yan:
This PR refactors QueueSettings into a functional component and removes some dead Code, namely functions regarding the GroupfolderInput field, which are handled inside the `GroupFolderInput`.
Additionally removed the .queue-settings css style, which is only used inside this component, from the global style sheet and created a module for it

### DIFF
--- a/ui/src/containers/QueueSettings.jsx
+++ b/ui/src/containers/QueueSettings.jsx
@@ -1,150 +1,91 @@
-/* eslint-disable react/destructuring-assignment */
-/* eslint-disable react/no-unused-state */
-import React from 'react';
 import { Dropdown, Form } from 'react-bootstrap';
-import { connect } from 'react-redux';
-import { bindActionCreators } from 'redux';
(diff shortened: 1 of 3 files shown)


ui/src/containers/QueueSettings.jsx
```diff
@@ -1,148 +1,91 @@
...
+        </Dropdown.Item>
+        <Dropdown.Item>
+          <Form.Check
+            type="checkbox"
+            onChange={(e) =>
+              dispatch(
+                setCentringMethod(
+                  e.target.checked ? AUTO_LOOP_CENTRING : CLICK_CENTRING,
+                ),
+              )
```
axelboc: Some curly brackets around this maybe `{}` since it's wrapping onto multiple lines anyway?

axelboc (approved):

ui/src/containers/QueueSettings.jsx
walesch-yan: Done

PR #1630, opened 2025-04-16
Document linting config

axelboc:
As promised, adding a bit of documentation for the linting configuration and how to deal with violations. I'm also restructuring the front-end documentation file a little.

PR #1631, opened 2025-04-16
Disable remaining linting violations in legacy code

axelboc:
The remaining rules that are disabled in `eslint.config.js` are, in my opinion, not worth fixing on existing code. They can wait until the affected code is refactored (e.g. class components to function components, `useSelector`, UI revamp, etc.)

As written in the new [linting documentation](https://github.com/mxcube/mxcubeweb/pull/1630/files#diff-425da2324e83ab4adea4825089d26a6b2b0d0dc15cbd54a452499c1f4445909f), disabling a rule in `eslint.config.js` means that the rule is not applied to new code. So as a final step, I re-enable the rules and disable every violation explicitly with `eslint-disable` directives. I mostly use file-level directives to avoid cluttering the codebase.

### DIFF
--- a/ui/eslint.config.js
+++ b/ui/eslint.config.js
@@ -8,12 +8,7 @@ const config = defineConfig([
   ...createConfig(opts),
   {
     rules: {
-      'jsx-a11y/anchor-is-valid': 'off',
-      'jsx-a11y/control-has-associated-label': 'off',
-      'react/destructuring-assignment': 'off',
-      'react/no-multi-comp': 'off',
-      'react/no-unused-class-component-methods': 'off',
-      'react/prop-types': 'off',
+      'react/prop-types': 'off', // too ambitious; better to switch to TypeScript anyway
 
       /* Default is "avoid", but there are lots of complicated `switch` statements,
        * notably in Redux reducers, which benefit from clear case blocks. */
--- a/ui/package.json
+++ b/ui/package.json
@@ -14,7 +14,7 @@
     "format": "pnpm prettier --write",
     "fix": "pnpm eslint --fix",
     "prettier": "prettier . --cache --check",
-    "eslint": "eslint",
+    "eslint": "eslint --max-warnings=0",
     "e2e": "cypress run --e2e --browser firefox",
     "cypress": "cypress open --e2e --browser firefox"
   },
--- a/ui/src/components/DraggableModal.jsx
+++ b/ui/src/components/DraggableModal.jsx
@@ -11,6 +11,7 @@ function DraggableModalDialog(props) {
   );
 }
 
+// eslint-disable-next-line react/no-multi-comp
 export function DraggableModal(props) {
   const { children } = props;
 
--- a/ui/src/components/Equipment/Harvester.jsx
+++ b/ui/src/components/Equipment/Harvester.jsx
@@ -1,3 +1,4 @@
+/* eslint-disable react/destructuring-assignment */
 import React from 'react';
 import { Badge, Button, Card } from 'react-bootstrap';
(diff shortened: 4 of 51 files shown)


ui/eslint.config.js
```diff
@@ -8,12 +8,7 @@ const config = defineConfig([
   ...createConfig(opts),
   {
     rules: {
-      'jsx-a11y/anchor-is-valid': 'off',
```
axelboc, line 11: Forbids using `<a>` for interactive elements that are not actually links and should be `<button>`s instead. Not trivial to do from a styling point of view, and since the gain is minimal (accessibility), it's okay to not fix existing violations right away. Of course, if you refactor a component with such a violation, do take the time to fix it. Same applies to all the rules below.

ui/eslint.config.js
```diff
@@ -8,12 +8,7 @@ const config = defineConfig([
...
-      'jsx-a11y/control-has-associated-label': 'off',
```
axelboc, line 12: Requires elements to be labelled. While it can be as simple as adding `aria-label`, it's rarely the best solution, and the alternatives are rarely trivial. So again, this rule is important for new code but not so much for existing code.

ui/eslint.config.js
```diff
@@ -8,12 +8,7 @@ const config = defineConfig([
...
-      'react/destructuring-assignment': 'off',
```
axelboc, line 13: Lots of `this.props.<propName>` everywhere, so impractical to fix. These violations can usually be fixed when refactoring class components to function components.

ui/eslint.config.js
```diff
@@ -8,12 +8,7 @@ const config = defineConfig([
...
-      'react/no-multi-comp': 'off',
```
axelboc, line 14: Forbid having multiple components in the same file, so very minor. The whole file structure of the `ui/src` folder needs a bit of a review, so no point in creating more component files in arbitrary locations. Better to think about it on a case-by-case basis when refactoring.

ui/eslint.config.js
```diff
@@ -8,12 +8,7 @@ const config = defineConfig([
...
-      'react/no-unused-class-component-methods': 'off',
```
axelboc, line 15: I've kept all these dead methods for now in case they became dead by mistake (i.e. regressions). Again, something to look at when refactoring class components.

ui/eslint.config.js
```diff
@@ -8,12 +8,7 @@ const config = defineConfig([
...
-      'react/prop-types': 'off',
```
axelboc, line 16: The only rule I keep disabled. It's just not practical to fix, even on new code. It can be re-enabled once the entire codebase is written in TypeScript... https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/prop-types.md

ui/package.json
```diff
@@ -14,7 +14,7 @@
     "format": "pnpm prettier --write",
     "fix": "pnpm eslint --fix",
     "prettier": "prettier . --cache --check",
-    "eslint": "eslint",
+    "eslint": "eslint --max-warnings=0",
```
axelboc, line 17: No more warnings, so we can now fail if any new warning shows up!

walesch-yan:
Thanks @axelboc this PR is also a nice overview for us to see, what components still need refactoring. 
I just saw that #1628 has been merged, could you update the `ContextMenu` file changes in this PR and I will convert #1629 and #1632 into a draft until this PR has been merged?

walesch-yan (approved):

marcus-oscarsson:
Thanks, sorry for being a bit fast to merge #1628.

PR #1632, opened 2025-04-16
refactor groupfolderInput

walesch-yan:
This PR refactors the `GroupFolderInput` component using hooks, functional component and react-hook-form.

Additionally, I replaced the validationstate value, which ever only had two values: `success` and `warning`, by a specific style, when user has changed the input form but not pressed the set button, indicating the needed button press. See video below for example:
GroupFolderWarning

### DIFF
--- a/ui/src/containers/GroupFolderInput.jsx
+++ b/ui/src/containers/GroupFolderInput.jsx
@@ -1,82 +1,56 @@
-/* eslint-disable react/destructuring-assignment */
-import React from 'react';
+import { useEffect } from 'react';
 import { Button, Form } from 'react-bootstrap';
-import { connect } from 'react-redux';
-import { bindActionCreators } from 'redux';
+import { useForm } from 'react-hook-form';
+import { useDispatch, useSelector } from 'react-redux';
 
 import { setGroupFolder } from '../actions/queue';
 
-class GroupFolderInput extends React.Component {
-  constructor(props) {
-    super(props);
-    this.inputOnChangeHandler = this.inputOnChangeHandler.bind(this);
-    this.setGroupFolderInput = this.setGroupFolderInput.bind(this);
-    this.inputValue = '';
-    this.state = { validationState: 'success' };
+function GroupFolderInput() {
+  const dispatch = useDispatch();
+  const groupFolder = useSelector((state) => state.queue.groupFolder);
+
+  const {
+    register,
+    handleSubmit,
+    reset,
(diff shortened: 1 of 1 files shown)


ui/src/containers/GroupFolderInput.jsx
```diff
@@ -1,81 +1,56 @@
...
-  render() {
-    return (
-      <span>
-        <Form.Label>Group path :</Form.Label>
-        <br />
-        <Form
-          onSubmit={(e) => {
-            e.preventDefault();
+  return (
+    <Form as="div">
```
walesch-yan, line 32: I had to convert to div here, since the submit button wouldn't work otherwise, this could be related to this component being nested inside another Form element

ui/src/containers/GroupFolderInput.jsx
```diff
@@ -1,81 +1,56 @@
...
+  const groupFolder = useSelector((state) => state.queue.groupFolder);
+
+  const {
+    register,
+    handleSubmit,
+    reset,
+    formState: { isDirty },
+  } = useForm({
+    defaultValues: {
+      groupFolder: groupFolder || '',
```
axelboc, line 19: You can move the fallback inside `useSelector`

ui/src/containers/GroupFolderInput.jsx
```diff
@@ -1,81 +1,56 @@
...
-    return (
-      <span>
-        <Form.Label>Group path :</Form.Label>
-        <br />
-        <Form
-          onSubmit={(e) => {
-            e.preventDefault();
+  return (
+    <Form as="div">
+      <Form.Label>Group path :</Form.Label>
```
axelboc, line 33:
```suggestion
      <Form.Label>Group path</Form.Label>
```

Might as well :shrug:

ui/src/containers/GroupFolderInput.jsx
axelboc: Just to cross the t's: it wasn't because the `Button` was missing `type="submit"` by any chance?

axelboc (commented):
Ah damn, sorry, forgot to publish my review... Nothing much anyway, but might as well.

ui/src/containers/GroupFolderInput.jsx
walesch-yan: No, I removed the `type="submit"` only afterwards, with `onClick=...`

PR #1635, opened 2025-04-25
Added the concept of "supported types" on adapter objects.

marcus-oscarsson:
**Edited**

This is something that I felt was necessary to be done before I make the PR for the improved resource definition/handling for adapters. 

Each adapter objects can specify which hardware object it adapts via `SUPPORTED_TYPES` instead of using `adapterutils.get_adapter_cls_from_hardware_object`. Like this each adapter is more self contained and defines which hardware objects it supports.

 - Each adapter describes which types it supports
 - Removes the need for get_adapter_cls_from_hardware_object


mxcubeweb/core/adapter/adapter_base.py
```diff
@@ -40,15 +43,23 @@ def __init__(self, ho, role, app):
         self._unique = True
         self._msg = ""
 
+    @classmethod
+    def can_adapt(cls, ho):
+        return any(isinstance(ho, t) for t in cls.SUPPORTED_TYPES)
+
+    def __init_subclass__(cls, **kwargs):
```
marcus-oscarsson, line 51: Ive been thinking about how to achieve this for some time and `__init_subclass__` was the missing piece that allows for the registration of the actual existing sub classes so that we know which adapters that exist and in extension what hardware objects we support.

mxcubeweb/core/adapter/__init__.py
```diff
@@ -0,0 +1,36 @@
+import importlib
```
marcus-oscarsson, line 1: `__init_subclass__` requires that the the adapters are loaded otherwise It wont get called. So I added this sort of "boiler plate" module-load code to import all modules with the `_adapter` suffix and classes with the `Adapter` suffix

mxcubeweb/core/adapter/adapter_manager.py
```diff
@@ -0,0 +1,156 @@
...
+                ho.__class__.mro().index(t)
+                for t in supported_types
+                if t in ho.__class__.mro()
+            ),
+            default=float("inf"),
+        )
+
+    def find_best_adapter(self, ho):
+        """
+        Rank adapters by the depth of their SUPPORTED_TYPES in the
```
marcus-oscarsson, line 80: This is the next trick that allows us to match the most "specific" HardwareObject otherwise the match would be `HardwareObject` meaning basically `AdapterBase`, while we would like for instance `MotorAdapter` and not `ActuatorAdapter`

mxcubeweb/core/adapter/motor_adapter.py
```diff
@@ -7,6 +11,8 @@
 
 
 class MotorAdapter(ActuatorAdapterBase):
+    SUPPORTED_TYPES: ClassVar[list[object]] = [AbstractMotor.AbstractMotor]
```
marcus-oscarsson, line 14: We define the supported hardware objects like this. And yes RUFF strikes again with the `ClassVar` in-case one would not know that this is a class variable ?

mxcubeweb/core/adapter/adapter_manager.py
```diff
@@ -0,0 +1,156 @@
...
+        The hwr module must be imported at the very beginning of the application
+        start-up to function correctly.
+
+        This method can however be called later, so that initialization can be
+        done when one wishes.
+        """
+
+        # Is this really needed ?
+
+        # fname = os.path.dirname(__file__)
```
marcus-oscarsson: I commented this for the moment @rhfogh, I'm not sure that its needed and did it actually do what you wanted ?

mxcubeweb/core/adapter/adapter_manager.py
```diff
@@ -0,0 +1,156 @@
...
+import traceback
+
+from mxcubecore import HardwareRepository as HWR
+from mxcubecore.utils.conversion import make_table
+
+from mxcubeweb.core.adapter.adapter_base import AdapterBase
+from mxcubeweb.core.adapter.beamline_adapter import BeamlineAdapter
+
+
+class HardwareObjectAdapterManager:
```
marcus-oscarsson, line 14: I'm idle ears if you have a better name than `HardwareObjectAdapterManager` :)

mxcubeweb/app.py
```diff
@@ -32,158 +30,11 @@
 from mxcubeweb.core.components.sampleview import SampleView
 from mxcubeweb.core.components.workflow import Workflow
 from mxcubeweb.core.models.configmodels import UIComponentModel
-from mxcubeweb.core.util.adapterutils import get_adapter_cls_from_hardware_object
 from mxcubeweb.logging_handler import MX3LoggingHandler
 
 removeLoggingHandlers()
 
 
-class MXCUBECore:
```
marcus-oscarsson, line 41: This class was moved and renamed to HardwareObjectAdapterManager

mxcubeweb/core/adapter/adapter_manager.py
marcus-oscarsson: We are together with @walesch-yan  looking into a way to do this without using `mro` which we think will be a bit more robust

walesch-yan: The conclusion we came to was to have a type to adapter dictionary within the `Adapterbase` class. Every time, `__init_subclass__` is called, all the supported types from the corresponding adapter are added. 

`find_best_adapter` will find all supported types within this dictionary that are superclasses of the hardware object. If more than one is found, a check will be performed to see if there is a common deepest subclass within the list of found classes. If not, we throw an error, since multiple adapters could adapt the class.

The  scenario above could happen if an `ho` class inherits from multiple classes and no specific adapter for that object exists. Previously, this case was not treated

mxcubeweb/core/adapter/adapter_base.py
```diff
@@ -40,15 +44,27 @@ def __init__(self, ho, role, app):
...
+        for ho_class in cls.SUPPORTED_TYPES:
+            if ho_class in AdapterBase.SUPPORTED_TYPES_TO_ADAPTERS:
```
walesch-yan, line 53: The latest change also allows to early check if multiple adapters for specific types exist (Adapter/type could have been added by mistake).

marcus-oscarsson:
Ill follow up with anther PR with complete documentation when I submit the `ResourceHandler` as this changes how the resources are defined a bit

walesch-yan (approved):

PR #1640, opened 2025-05-06
UI: mark visually motor positions outside of limits

elmjag:
Changes behavior of motor widgets when position outside of motor limits is entered.

If entered position is outside of limits, change the widget style to _invalid_ mode. Block sending the _invalid_ position to the back-end. When focus is lost, restore widget to display motor's real position.

This changes the current behavior, where it's possible to enter and send a position outside of limits. The motor can refuse to move to that position, but the widget will still display the invalid value. Thus the displayed value in the UI will be different from the motor's real position, without any visual indications.

### DIFF
--- a/ui/src/components/MotorInput/BaseMotorInput.jsx
+++ b/ui/src/components/MotorInput/BaseMotorInput.jsx
@@ -19,6 +19,7 @@ function BaseMotorInput(props) {
 
   const [inputValue, setInputValue] = useState(value.toFixed(precision));
   const [isEdited, setEdited] = useState(false);
+  const [isInsideLimits, setInsideLimits] = useState(true);
 
   useEffect(() => {
     setInputValue(value.toFixed(precision));
@@ -52,6 +53,24 @@ function BaseMotorInput(props) {
     }
   }
 
+  function handleBlur() {
+    //
+    // restore widget to display motor's current position,
+    // when focus is lost
+    //
+    setInsideLimits(true);
+    setEdited(false);
+    setInputValue(value.toFixed(precision));
+  }
+
+  function handleChange(evt) {
+    const newValue = evt.target.value;
+
+    setInputValue(newValue);
+    setEdited(true);
+    setInsideLimits(min <= newValue && newValue <= max);
+  }
+
   const isReady = state === HW_STATE.READY;
   const isBusy = state === HW_STATE.BUSY;
   const isWarning = state === HW_STATE.WARNING;
@@ -71,15 +90,14 @@ function BaseMotorInput(props) {
         max={max}
         min={min}
         disabled={disabled || !isReady}
+        data-invalid={!isInsideLimits || undefined}
         data-dirty={isEdited || undefined}
         data-busy={isBusy || undefined}
         data-warning={isWarning || undefined}
         data-fault={isFault || undefined}
         onKeyDown={handleKey}
-        onChange={(evt) => {
-          setInputValue(evt.target.value);
-          setEdited(true);
-        }}
+        onBlur={handleBlur}
+        onChange={handleChange}
       />
       <input type="submit" hidden /> {/* allow submit on Enter */}
     </form>
--- a/ui/src/components/MotorInput/MotorInput.jsx
+++ b/ui/src/components/MotorInput/MotorInput.jsx
@@ -4,6 +4,7 @@ import { useDispatch, useSelector } from 'react-redux';
 
 import { setAttribute } from '../../actions/beamline';
 import { stopBeamlineAction } from '../../actions/beamlineActions';
+import { showErrorPanel } from '../../actions/general';
 import { setMotorStep } from '../../actions/sampleview';
 import { HW_STATE, QUEUE_RUNNING } from '../../constants';
 import BaseMotorInput from './BaseMotorInput';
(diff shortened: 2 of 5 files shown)


ui/src/components/MotorInput/BaseMotorInput.jsx
```diff
@@ -71,15 +90,14 @@ function BaseMotorInput(props) {
         max={max}
         min={min}
         disabled={disabled || !isReady}
-        data-dirty={isEdited || undefined}
+        data-invalid={!isInsideLimits || undefined}
+        data-dirty={(isInsideLimits && isEdited) || undefined}
```
axelboc:
```suggestion
        data-dirty={isEdited || undefined}
```

Just move the `data-invalid` style after the `data-dirty` style in the CSS so it overrides it.

axelboc (commented):
> The motor can refuse to move to that position, but the widget will still display the invalid value. Thus the displayed value in the UI will be different from the motor's real position, without any visual indications.

I feel like the changes you propose are more of a workaround for this than a fix. Sure, it's always good to avoid making an invalid request, but if this request happens, the error should be dealt with in the UI. In other words, front-end validation should be an enhancement, not a requirement.

---

To recap the problem: when the back-end receives a value outside of `[min, max]`, it should respond with an error (I assume it already does), and this error should be conveyed somehow to the user.

To do so, the `setAttribute` Redux thunk should be made asynchronous so that `dispatch(setAttribute(attribute, val))` can be awaited and wrapped in `try/catch`. Then, in `MotorInput`, you can manage an error state and maybe pass an `invalid` prop to `BaseMotorInput`.

To reset this error/invalid state, you could:

- pass an `onBlur` prop to `BaseMotorInput` to reset on blur;
- start a timer in an effect in `MotorInput` to reset after a few seconds;
- leave the field in error state until the user changes the motor value again.

Your choice of resetting on blur seems fine to me in terms of UX.

---

If this all sounds too obscure, we can have a go at it if you like.

marcus-oscarsson:
We need to be quite careful to make sure that we are consistent across the various underlying implementations (in `mxcubecore` and below) when we handle these scenarios. We also need to make sure that handle the validation in a way that is logical to the user on the UI side and make the necessary implementation on the backend to prevent incorrect values to be passed on.

I think its a great idea to show that a input is outside of the valid range, it helps the user to understand that the value is indeed out of range.

 - A `ValueError` is raised (or should be raised) when any `AbstractActator` receives a value outside of its limits. 
 
The limits of most actuators are available in the UI so we could also add an additional check. Keep in mind that, to complicate things, limits of some actuators can be a function of some other actuators position (i.e resolution). I think thats the only one at the moment.

As @axelboc points out above it should be possible to catch this error in an `async` handler and take the right actions to make sure that the value is displayed correctly and that a message is given to the user.

elmjag:
The back-end does indeed respond with an error if you try to move motor outside of the limits. 

I agree with @axelboc that we should catch error responses for motor moves and display some kind of error message in the UI. It's possible for the move to fail due to some other problem, not necessary due to hitting a limit. I'll work on adding that.

However, visually showing that the entered position is invalid is very useful feature for us. We are using MD3Up and it's easy get to the edges of motor limits. Currently it's a bit cryptic why some moves fail, you basically need to read the log.

marcus-oscarsson:
> However, visually showing that the entered position is invalid is very useful feature for us. We are using MD3Up and it's easy get to the edges of motor limits. Currently it's a bit cryptic why some moves fail, you basically need to read the log.

We completely agree, as mentioned above the limits are available on the UI side so you can easily add client side validation and some nice display of when the user tries to enter a value that is out of range.  Our concern was more that this should not replace any server side handling and be done in such a way so that it reflects the actual state of the hardware with the correct message/display. I think that is also what you have in mind ?

The `await`, `try/catch` should be there to handle the cases when/if the UI validation fails, for whatever reason.

ui/src/components/MotorInput/BaseMotorInput.jsx
elmjag: Aha! Changed as suggested.

elmjag:
I have added code that catches errors on moving motors. The errors are displayed to the user with the error panel.
https://github.com/elmjag/mxcubeweb/commit/f0d45fb1d745585c94ede65993b25c590c7fc3da

To summarize current behavior:

 - if user enters value outside of limits, the widget marks it visually
 - if back-end replies with an error on motor move, the error message will be shown to the user
 - when motor widget loses focus, the displayed position is reset to motor's current position, as reported by the back-end
 
 Is this what you have in mind?

marcus-oscarsson:
yeah, that sounds reasonable to me. Do you have a screenshot of the error display ?

ui/src/components/MotorInput/MotorInput.jsx
```diff
@@ -35,6 +36,14 @@ function MotorInput(props) {
     return null;
   }
 
+  async function handleChange(val) {
```
marcus-oscarsson, line 39: :+1:

elmjag:
> yeah, that sounds reasonable to me. Do you have a screenshot of the error display ?

It shows the standard 'error bubble' with the error message from the back-end:

Screenshot from 2025-05-12 09-07-27

axelboc (approved):
Great stuff :100:

mockoocy (approved):

PR #1641, opened 2025-05-06
Fix Characterization error

dominikatrojanowska:
When adding a `Characterisation (1 image)` task, you get an ‘Error: The task could not be added to the server’. See the attached screenshot.

image

To reproduce: right click and pick `Characterisation (1 image)` option. Click `Run Now` button.

## More detailed description of what causes the issue
Adding new Characterisation task to queue launches [`sendAddQueueItem`](https://github.com/mxcube/mxcubeweb/blob/504e2e9316a5f9082eff1106f4b8276f94fb0a99/ui/src/actions/queue.js#L401) method that tries to post data but processing them (described below) does not succeed and application shows the attached error panel. 

[Add characterization](https://github.com/mxcube/mxcubeweb/blob/504e2e9316a5f9082eff1106f4b8276f94fb0a99/mxcubeweb/core/components/queue.py#L945) creates [`refdc_model`](https://github.com/mxcube/mxcubeweb/blob/504e2e9316a5f9082eff1106f4b8276f94fb0a99/mxcubeweb/core/components/queue.py#L1482) which is a `DataCollection` model object assigned to `Characterization` model at its [`reference_image_collection`](https://github.com/mxcube/mxcubecore/blob/1e3a5c87fc4a32bdfa2f3c01b03df739fcaeb7c0/mxcubecore/model/queue_model_objects.py#L834C14-L834C40) attribute. By default those two models haven't got `_node_id`. This attribute is initialized for `Characterization` model within [`HWR.beamline.queue_model.add_child`](https://github.com/mxcube/mxcubeweb/blob/504e2e9316a5f9082eff1106f4b8276f94fb0a99/mxcubeweb/core/components/queue.py#L1506) call. But that is not happening for DataCollection, so it's `_node_id` is `None`.

The last line of [`queue_add_item`](https://github.com/mxcube/mxcubeweb/blob/504e2e9316a5f9082eff1106f4b8276f94fb0a99/mxcubeweb/core/components/queue.py#L897) causes to call sequentially: `queue_to_dict`, `queue_to_dict_rec`, `_handle_char` at some point reaching the Characterization model, `_handle_dc` with `Characterisation.reference_image_collection` with assigned the `DataModel` as the second argument,  `get_node_state` with `None` and finally `get_entry` also with `None`. `get_entry` returns [`QueueManager` object](https://github.com/mxcube/mxcubeweb/blob/504e2e9316a5f9082eff1106f4b8276f94fb0a99/mxcubeweb/core/components/queue.py#L716) and that causes the error when [program is trying to read it's `status` field](https://github.com/mxcube/mxcubeweb/blob/504e2e9316a5f9082eff1106f4b8276f94fb0a99/mxcubeweb/core/components/queue.py#L221). 

Some time ago there was no [`if-else` clause in `get_entry`](https://github.com/mxcube/mxcubeweb/blob/ea31f09/mxcubeweb/core/components/queue.py#L705) and `TypeError` raised over there (when `None` was passed instead of a number to `int` method) was caught by [`try-except Exception`](https://github.com/mxcube/mxcubeweb/blob/ea31f09/mxcubeweb/core/components/queue.py#L222) and all the following lines were not called. 

I thought also about different approach to solve this issue:
```python
def  get_node_state(self, node_id):
        """ ... """
        if not node_id:
            return (True, UNCOLLECTED)
        self.get_entry(node_id)
```
Optionally, removing try-except. But this requires clarification: what this `try-except Exception` is designed to catch (except mentioned TypeError)? Now the application may hide some important tracebacks. Maybe since there is `if-else` introduced in `get_entry` it is time to remove it?

### DIFF
--- a/mxcubeweb/core/components/queue.py
+++ b/mxcubeweb/core/components/queue.py
@@ -707,14 +707,9 @@ class Queue(ComponentBase):
         Retrieves the model and the queue entry for the model node with id <id>
 
         :param int _id: Node id of node to retrieve
-        :returns: The tuple model, entry or the root node and QueueManger if _id is None
+        :returns: The tuple of model and entry
         :rtype: Tuple
         """
-        if _id is None:
-            return (
-                HWR.beamline.queue_model.get_model_root(),
-                HWR.beamline.queue_manager,
-            )
         model = HWR.beamline.queue_model.get_node(int(_id))
         entry = HWR.beamline.queue_manager.get_entry_with_model(model)
         return model, entry
@@ -1504,7 +1499,6 @@ class Queue(ComponentBase):
 
         HWR.beamline.queue_model.add_child(sample_model, refgroup_model)
         HWR.beamline.queue_model.add_child(refgroup_model, char_model)
-        refdc_model._node_id = char_model._node_id
         refgroup_entry = qe.TaskGroupQueueEntry(Mock(), refgroup_model)
 
         refgroup_entry.set_enabled(True)
--- a/mxcubeweb/core/components/queue.py
+++ b/mxcubeweb/core/components/queue.py
@@ -1504,6 +1504,7 @@ class Queue(ComponentBase):
 
         HWR.beamline.queue_model.add_child(sample_model, refgroup_model)
         HWR.beamline.queue_model.add_child(refgroup_model, char_model)
+        refdc_model._node_id = char_model._node_id
         refgroup_entry = qe.TaskGroupQueueEntry(Mock(), refgroup_model)
 
         refgroup_entry.set_enabled(True)


marcus-oscarsson:
Thanks !

mxcubeweb/core/components/queue.py
```diff
@@ -1504,6 +1504,7 @@ def add_characterisation(self, node_id, task):
 
         HWR.beamline.queue_model.add_child(sample_model, refgroup_model)
         HWR.beamline.queue_model.add_child(refgroup_model, char_model)
+        refdc_model._node_id = char_model._node_id
```
marcus-oscarsson: Hm, what is the error on the server ?

What I'm asking myself is how it comes that it works with characterization from a centered position but not from the 2d point ?

dominikatrojanowska: I elaborated more on the [PR description](https://github.com/mxcube/mxcubeweb/pull/1641#issue-3042960981).  
I didn't check if characterization works from centered position yet... Our beamline staff reported above issue with characterization and I reproduced that also demo-upstream version.

marcus-oscarsson:
I see, I think the issue is this "fix" made by @rhfogh here: 

    https://github.com/mxcube/mxcubeweb/commit/c18c085431eb6eefca35fc978b8f88fda3e4917c

Could you please try to remove the entire `if _id is None:` statement and try if that fixes the issue ?

We have made a local fix for this issue sometime ago and I think its the same issue that you are observing. We have not yet had the time to be make into a PR. 

Marcus

elmjag:
I did a quick test, I just applied the revert commit: https://github.com/mxcube/mxcubeweb/pull/1641/commits/a5a0b7565ca4da7d0e44fe86a3e195263014baa6
It seems to solve the problem.

The question is, what do we break be removing @rhfogh changes?

marcus-oscarsson:
As far as I remember it was to solve a warning, right @rhfogh, it had no functional impact. The issue is that `QueueManager` does not have the same interface as `QueueEntry` so that would not work. I think returning None or raising en error and handling that in the calling code would be better.

dominikatrojanowska:
that's surprising but I didn't found anything broken after reverting https://github.com/mxcube/mxcubeweb/commit/a5a0b7565ca4da7d0e44fe86a3e195263014baa6. Indeed it seems the commit was made only to satisfy ruff, I guess there is some broader ruff rule (in ruff.toml ?) so the linters are not complaining any more

marcus-oscarsson:
Ok, thanks !

marcus-oscarsson:
Could you please rebase ?

dominikatrojanowska:
I missclicked and chose `Sync with merge` instead of `rebase`, that's why I produced so many commits. But rebasing's finally done :)

marcus-oscarsson:
Thanks @dominikatrojanowska

PR #1642, opened 2025-05-06
[Sample List Graphical view ] Add Circular Display Mode for Single Cell in SampleFlexView

jbflo:
The current implementation of SampleFlexView works well for visualizing multiple cells containing multiple pucks. This update introduces a new circular layout optimized for displaying a single cell, where its pucks are arranged in a circle for easier filtering and visual clarity.

- putting it in a new component/file

🔍 Targeted primarily at mockup/testing mode, but can  also  be useful for facilities using single-cell puck sample changers.

-  I also  Turn isSingleCell into a variable

image

### DIFF
--- a/ui/src/components/SampleGrid/SampleCircleView.jsx
+++ b/ui/src/components/SampleGrid/SampleCircleView.jsx
@@ -1,0 +1,117 @@
+import { useEffect } from 'react';
+import { Col } from 'react-bootstrap';
+import { useDispatch, useSelector } from 'react-redux';
+
+import { filterAction } from '../../actions/sampleGrid';
+
+export default function SampleCircleView(props) {
+  const { displayPuckCellContextMenu, puckMenuID, type = 'Mockup' } = props;
+
+  const filterOptions = useSelector((state) => state.sampleGrid.filterOptions);
+
+  const pucks = useSelector((state) => state.sampleChanger.contents.children);
+
+  const cellID = 1;
+
+  const dispatch = useDispatch();
+
+  function handleClickOnCellPuck(event, puckID) {
+    dispatch(
+      filterAction({ cellFilter: `${cellID}`, puckFilter: `${puckID}` }),
+    );
+    event.stopPropagation();
+  }
+
+  useEffect(() => {
+    if (filterOptions.cellFilter === '') {
+      dispatch(filterAction({ cellFilter: '1', puckFilter: '1' }));
+    }
+  }, [filterOptions.cellFilter, dispatch]);
+
+  function handleDisplayPuckCellContextMenu(e, menuID, puckID) {
+    e.preventDefault();
+    handleClickOnCellPuck(e, cellID, puckID === null ? '' : puckID);
+    displayPuckCellContextMenu(e, menuID, cellID, puckID);
+    e.stopPropagation();
+  }
+
+  function getSingleCellAsCircle(centerX, centerY) {
+    const totalPucks = pucks.length;
+
+    // Adjust angle range based on puck count
+    const useFullCircle = totalPucks >= 8;
+    const angleRange = useFullCircle ? 2 * Math.PI : Math.PI * 1.5;
+    const startAngle = useFullCircle ? 0 : (Math.PI - angleRange) / 2;
+
+    // Circle size and spacing
+    const baseRadius = 4;
+    const placementRadius = baseRadius + Math.max(0, (totalPucks - 16) * 0.3);
+    const puckRadius = Math.min(
+      0.7,
+      (2 * Math.PI * placementRadius) / (totalPucks * 2),
+    );
+
+    // Fix angle spacing: avoid skipping puck 1 or overlapping at end
+    const angleDivisor = useFullCircle
+      ? totalPucks
+      : Math.max(1, totalPucks - 1);
+
+    return pucks.map((_, i) => {
+      const angle = startAngle + (angleRange * i) / angleDivisor;
+      const x = centerX + placementRadius * Math.cos(angle);
+      const y = centerY + placementRadius * Math.sin(angle);
+      const puckID = i + 1;
+
+      const isPuckSelected =
+        Number(filterOptions.cellFilter) === cellID &&
+        Number(filterOptions.puckFilter) === puckID;
+
+      return (
+        <g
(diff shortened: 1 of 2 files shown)


ui/src/containers/SampleFlexView.jsx
```diff
@@ -215,11 +216,86 @@ export default function SampleFlexView(props) {
...
+      );
+    });
+  }
+
+  function getMultipleCellsAsPieSlice() {
+    return Array.from({ length: numberOfCell }, (_, idx) => getCellSlice(idx));
+  }
+
+  function getCellContents() {
+    return isSingleCell()
```
axelboc: You could turn `isSingleCell` into a variable in `SampleGridTableContainer` to avoid having to invoke the function everywhere? (Maybe for another PR, though.)

Also, this ternary could just be inlined in the JSX below, I reckon.

ui/src/containers/SampleFlexView.jsx
```diff
@@ -215,11 +216,86 @@ export default function SampleFlexView(props) {
     );
   }
 
+  function getSingleCellAsCircle(centerX = 10, centerY = 10) {
```
axelboc: Since this is brand new code, it'd be worth putting it in a new component/file to try to keep `SampleFlexView` under control. If you do, as usual, prefer `useSelector` instead of props when possible.

Also, since `centerX` and `centerY` depend on the viewbox, I would prefer if they didn't have a default value but were passed as args/props explicitly instead.

ui/src/containers/SampleGridTableContainer.jsx
```diff
@@ -479,7 +482,7 @@ export default function SampleGridTableContainer(props) {
     e.stopPropagation();
   }
 
-  function isSingleCell() {
+  function getIsSingleCell() {
```
axelboc: No need to keep this function at all, just inline the logic above when initialising the `isSingleCell` variable.

ui/src/components/SampleGrid/SampleCircleView.jsx
```diff
@@ -0,0 +1,119 @@
...
+import { Col } from 'react-bootstrap';
+import { useDispatch, useSelector } from 'react-redux';
+
+import { filterAction } from '../../actions/sampleGrid';
+
+export default function SampleCircleView(props) {
+  const { displayPuckCellContextMenu, puckMenuID, type = 'Mockup' } = props;
+
+  const filterOptions = useSelector((state) => state.sampleGrid.filterOptions);
+  const scContent = useSelector((state) => state.sampleChanger.contents);
```
axelboc:
```suggestion
  const pucks = useSelector((state) => state.sampleChanger.contents.children);
```

ui/src/components/SampleGrid/SampleCircleView.jsx
```diff
@@ -0,0 +1,119 @@
...
+  function isPuckSelected(cellID, puckID) {
+    return (
+      Number(filterOptions.cellFilter) === cellID &&
+      Number(filterOptions.puckFilter) === puckID
+    );
+  }
+
+  useEffect(() => {
+    if (filterOptions.cellFilter === '') {
+      dispatch(filterAction({ cellFilter: `${1}`, puckFilter: `${1}` }));
```
axelboc:
```suggestion
      dispatch(filterAction({ cellFilter: '1', puckFilter: '1' }));
```

ui/src/components/SampleGrid/SampleCircleView.jsx
```diff
@@ -0,0 +1,119 @@
...
+    e.preventDefault();
+    handleClickOnCellPuck(e, cellID, puckID === null ? '' : puckID);
+    displayPuckCellContextMenu(e, menuID, cellID, puckID);
+    e.stopPropagation();
+  }
+
+  function getSingleCellAsCircle(centerX = 10, centerY = 10) {
+    const pucks = scContent.children;
+    const totalPucks = pucks.length;
+    const cellID = 1;
```
axelboc: `CELL_ID` could be a constant at the top of the file

ui/src/components/SampleGrid/SampleCircleView.jsx
```diff
@@ -0,0 +1,119 @@
...
+      const puckID = i + 1;
+
+      return (
+        <g
+          key={`single_cell_puck_${puckID}`}
+          onClick={(e) => handleClickOnCellPuck(e, cellID, puckID)}
+          onContextMenu={(e) =>
+            handleDisplayPuckCellContextMenu(e, puckMenuID, cellID, puckID)
+          }
+          fill={isPuckSelected(cellID, puckID) ? '#015f9d' : '#cdced1'}
```
axelboc: Might be simpler to turn this function into a variable inside the `map` callback:

```ts
const isPuckSelected = Number(filterOptions.cellFilter) === cellID && Number(filterOptions.puckFilter) === puckID;
```

axelboc: The current implementation is fine as is, no need to change it, but FYI, another way to do this that would have kept concerns separated a bit more (markup vs styles) would have been to add a `data` attribute to the `g` element and use CSS to change the fill color:

```jsx
<g
  className={styles.puck}
  data-selected={isPuckSelected || undefined}
  //...
>
```

```css
.puck {
  fill: #015f9d;
}

.puck[data-selected] {
  fill: #cdced1;
}

.puck > text {
  fill: white;
}

.puck[data-selected] > text {
  fill: black;
}
```

ui/src/components/SampleGrid/SampleCircleView.jsx
```diff
@@ -0,0 +1,119 @@
...
+      );
+    });
+  }
+
+  return (
+    <Col sm={6}>
+      <div className="div-svg-flex">
+        <svg height="100%" width="100%" viewBox="0 1 20 20">
+          <circle fill="#6cb0f5" r="8" cx="10" cy="10" />
+          {getSingleCellAsCircle()}
```
axelboc: Just to reiterate, I would prefer for the two arguments to be passed explicitly here instead of having default values.

axelboc (approved):
Very nice, thanks!

marcus-oscarsson:
Nice :)

PR #1645, opened 2025-05-13
Add option to invert 'success'/'danger' background colors in InOutSwitch

alessandro-olivo:
### Feature
This PR introduces a new option to the InOutSwitch component that allows inverting the default background color logic based on the component’s state.

#### Current behavior:
- offValue → danger (red background)
- onValue → success (green background)

#### New behavior (when invertColors is enabled):
- offValue → success (green background)
- onValue → danger (red background)

### How to test:

Open the file `../mxcube-web/ui.yaml` and add the field `invert_bg_color` to an "InOutSwitch" element.
Example configuration (e.g. for the Fast Shutter component):
```
...
beamline_setup:
  id: beamline_setup
  components:
    ...
    - label: Fast Shutter
      attribute: fast_shutter
      invert_bg_color: true     
    ...
```

### DIFF
--- a/ui/src/components/InOutSwitch/InOutSwitch.jsx
+++ b/ui/src/components/InOutSwitch/InOutSwitch.jsx
@@ -19,6 +19,7 @@ export default function InOutSwitch(props) {
     openValue,
     offValue,
     optionsOverlay,
+    invertBgColor,
     labelText,
     isBtnLabel,
     overlayPlacement,
@@ -77,7 +78,7 @@ export default function InOutSwitch(props) {
   switch (value) {
     case openValue:
     case 'READY': {
-      msgBgStyle = 'success';
+      msgBgStyle = invertBgColor ? 'danger' : 'success';
       btn = (
         <Button
           variant="outline-secondary"
@@ -92,7 +93,7 @@ export default function InOutSwitch(props) {
     case offValue:
     case 'DISABLED':
     case 'CLOSED': {
-      msgBgStyle = 'danger';
+      msgBgStyle = invertBgColor ? 'success' : 'danger';
       btn = (
         <Button
           variant="outline-secondary"
--- a/ui/src/components/InOutSwitch/InOutSwitch.jsx
+++ b/ui/src/components/InOutSwitch/InOutSwitch.jsx
@@ -19,7 +19,7 @@ export default function InOutSwitch(props) {
     openValue,
     offValue,
     optionsOverlay,
-    invertBgColor,
+    invertBgColor = false,
     labelText,
     isBtnLabel,
     overlayPlacement,
--- a/ui/src/reducers/beamline.js
+++ b/ui/src/reducers/beamline.js
@@ -26,7 +26,6 @@ export const INITIAL_STATE = {
       state: 'undefined',
       msg: 'UNKNOWN',
       readonly: false,
-      invert_bg_color: false,
     },
     safety_shutter: {
       limits: [0, 1, 1],
@@ -35,7 +34,6 @@ export const INITIAL_STATE = {
       state: 'undefined',
       msg: 'UNKNOWN',
       readonly: false,
-      invert_bg_color: false,
     },
     beamstop: {
       limits: [0, 1, 1],
@@ -44,7 +42,6 @@ export const INITIAL_STATE = {
       state: 'undefined',
       msg: 'UNKNOWN',
       readonly: false,
-      invert_bg_color: false,
     },
     capillary: {
       limits: [0, 1, 1],
--- a/ui/src/reducers/beamline.js
+++ b/ui/src/reducers/beamline.js
@@ -26,6 +26,7 @@ export const INITIAL_STATE = {
       state: 'undefined',
       msg: 'UNKNOWN',
       readonly: false,
+      invert_bg_color: false,
     },
     safety_shutter: {
       limits: [0, 1, 1],
@@ -34,6 +35,7 @@ export const INITIAL_STATE = {
       state: 'undefined',
       msg: 'UNKNOWN',
       readonly: false,
+      invert_bg_color: false,
     },
     beamstop: {
       limits: [0, 1, 1],
@@ -42,6 +44,7 @@ export const INITIAL_STATE = {
       state: 'undefined',
       msg: 'UNKNOWN',
       readonly: false,
+      invert_bg_color: false,
     },
     capillary: {
       limits: [0, 1, 1],
--- a/mxcubeweb/core/models/configmodels.py
+++ b/mxcubeweb/core/models/configmodels.py
@@ -62,7 +62,7 @@ class UIComponentModel(BaseModel):
     value_type: str | None
     object_type: str | None
     format: str | None
-    invert_bg_color: bool | None
+    invert_color_semantics: bool | None
 
 
 class _UICameraConfigModel(BaseModel):
--- a/mxcubeweb/core/models/configmodels.py
+++ b/mxcubeweb/core/models/configmodels.py
@@ -64,6 +64,7 @@ class UIComponentModel(BaseModel):
     format: str | None
     invert_bg_color: bool | None
 
+
 class _UICameraConfigModel(BaseModel):
     label: str
     url: str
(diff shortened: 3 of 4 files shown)


axelboc:
Hi @alessandro-olivo, can you please explain why you need this feature and whether you've considered alternatives? I'm asking because it seems like another way might be to swap around the READY/CLOSED states of the hardware object and to provide more meaningful labels for those two states (but I don't know how doable that is).

I have two additional concerns:
- `invert_bg_color` does not convey the fact that your intention is to swap the semantic of the READY/CLOSED states —  `ready_is_danger` or something of the sort might be clearer in that regard.
- The new option is relevant only to the hardware objects that are displayed with the `InOutSwitch`; it doesn't make sense to add this to the `energy` object for instance. So I feel like this might not be the best place to configure this. Perhaps it  would be more appropriate to have an option in the `ui` config file that allows specifying which hardware objects need the visual display of their READY/CLOSED states to be swapped? Just a thought.

alessandro-olivo:
Hi @axelboc,
this feature comes from the need to make the background color consistent with the color logic we use in other parts of the beamline UI, in particular for the shutter.

In the beamline graphical interfaces, we use a red background when the shutter is open and green when it’s closed. That’s because red indicates that the light-path is open (i.e. “danger” from people point of view).

This is the default MXCuBE view:
image
But this is what we’d like it to look like:
image

Since it’s mainly about the meaning of "danger", I modified `InOutSwitch.jsx` and since the file `.../mxcube-web/ui.yaml` is used to configure the UI elements, I figured the `UIComponentModel` would be the right place to handle this.
There are other parameters like `precision` that don’t really make sense for certain elements like `InOutSwitch`, so I assumed that not every parameter has to be meaningful for every component, but maybe I’m wrong.

Yes, we could change the name `invert_bg_color` to something more meaningful, like `ready_is_danger`, `invert_danger_logic`, or something like that.


Thanks for your thoughts :)

marcus-oscarsson:
Very nice, I can clearly see situations where beamline staff would think that it is more appropriate to invert the "color-semantics" and that helps them to quickly identify if the beam is passing through or not. 

I would probably call the option `inverted-color-semantics` or something similar ?

marcus-oscarsson:
Additionally, I think we would like what you suggest to even be the default ;) I think our scientists would also appreciate that. @beteva, @walesch-yan and @DdSID29, @DidierNurizzo, what do you think ?

beteva:
Hi @alessandro-olivo, you can indeed change the value of the HardwareObject as @axelboc suggested. If your shutter (as many other objects) inherits from AbstractNState. You can choose what is the the value and the label displayed. This is what  you could do - in your xml file you add:
```
<vlalues>{"OPEN": "CLOSED", "CLOSED": "OPEN"}</values>
```

And than in your specific _set_value, when you receive from the GUI self,VALUES.OPEN you actually close the shutter and vice/versa.

I do agree, though, that **adding "inverse" colour semantic is a very nice feature**. As for what is by default, this I guess is institute specific - each one has its own colour code. I am not sure if we change the CLOSED  to be green in mxcube and red in the vacuum application (this one we cannot change) it will be more misleading.

I was also wondering if this is something we want rather to handle on the adapter level and have it in the HardwareObject config file.

walesch-yan:
I don't have a strong preference for any of the two configuration and the decision should I think be taken by the scientists:) But I do agree with @beteva that a green `CLOSED` sign could be misleading.

Otherwise I believe that the feature is very nice.

marcus-oscarsson:
@beteva I guess the issue with changing the `<VALUE>` tag would be that it also changes whats actually displayed. So that it would display "OPEN" with green background when its actually closed, right ?

I think what @alessandro-olivo is suggesting is quite sensible. The meaning that the color signals may sometimes be different from the standard off/closed - red and on/open - green. As, I guess, is quite clear from @alessandro-olivo example green in this case means that the there is no beam coming trough thus "safe" and green.

It could be handled on the hardware object configuration level but it feels more like a UI display option as its sort of contextual a bit like precision I guess. I don't have a strong opinion on where to add this configuration at the moment :)

I think the suggested change is fine, unless there are some compelling arguments for handling this in some other way ?

beteva:
> @beteva I guess the issue with changing the `<VALUE>` tag would be that it also changes whats actually displayed. So that it would display "OPEN" with green background when its actually closed, right ?

This will actually display whatever is the value of the dictionary, so in the above display for key OPEN it should diplay CLOSED. But this was just an example what can be done if you want to use the HO. I do agree that what @alessandro-olivo suggests is better.

> I think the suggested change is fine, unless there are some compelling arguments for handling this in some other way ?

Fine with me as well. I was just trying to see if this can be done with less changes.

alessandro-olivo:
> Very nice, I can clearly see situations where beamline staff would think that it is more appropriate to invert the "color-semantics" and that helps them to quickly identify if the beam is passing through or not.

The change was actually requested by the beamline scientist, who wanted to avoid having different color meanings

> I think what @alessandro-olivo is suggesting is quite sensible. The meaning that the color signals may sometimes be different from the standard off/closed - red and on/open - green. As, I guess, is quite clear from @alessandro-olivo example green in this case means that the there is no beam coming trough thus "safe" and green.

Yes, this is actually what I'd like to achieve

@beteva Of course, if we can achieve the same behavior in a simpler way, I’m totally fine with that, I'm not an expert, so it's possible I missed something :)

ui/src/reducers/beamline.js
```diff
@@ -26,6 +26,7 @@ export const INITIAL_STATE = {
       state: 'undefined',
       msg: 'UNKNOWN',
       readonly: false,
+      invert_bg_color: false,
```
axelboc: I think the new property ends up in the `uiproperties` state, not in `beamline.hardwareObjects`. That being said, I don't think you need initial values at all, so you can just remove the changes in this file.

ui/src/components/InOutSwitch/InOutSwitch.jsx
```diff
@@ -19,6 +19,7 @@ export default function InOutSwitch(props) {
     openValue,
     offValue,
     optionsOverlay,
+    invertBgColor,
```
axelboc: Just for explicitness:

```suggestion
    invertBgColor = false,
```

axelboc (approved):
The solution of swapping the open/closed states in the hardware object does seem counter-intuitive in the end, so **I'm fine with the proposed solution** (with the property renamed to `invert_color_semantics` as suggested by Marcus). :+1: 

> EDIT: I don't think we have sufficient information to decide whether it's safe to change the default green/red mapping.

---

Sharing my thoughts a bit more for the record: I'm always a bit wary of adding seemingly-inconsequential features like this, because it can quickly turn into logic soup in the front-end.

The two problems that this PR brings to light are that:

- it's possible to define UI properties for hardware objects that make no sense and have no effect (like `invert_color_semantics` for an object that is not an actuator)
  - The `beamline_setup` config should perhaps be reviewed to at least distinguish between beamline attributes and actuators.
- front-end components are directly coupled with hardware object states (cf. `switch/case` in `InOutSwitch`)
  - The `beamline_setup` config should perhaps provide a mapping between hardware object states and UI semantics.

Anyway, just food for thoughts, for the long term. :grin:

antolinos:
Sorry for jumping into this discussion (which I’ve found very interesting because it’s a recurring debate).

I just wanted to point out that, for the mental well-being of users who use MXCuBE at multiple institutes, I wonder if the color scheme shouldn't be consistent _across institutes_. 

For example, using green to indicate the shutter is closed at Elettra and red at the ESRF could be a bit confusing and prone to misunderstandings (from the user perspective).

marcus-oscarsson:
I completely agree @antolinos, we are trying very hard to have the same semantics and user interface across sites. Sometimes however its not only up to the MX beamlines to decide how this works, its can be a local/institute specific guideline.

marcus-oscarsson:
I also agree with @axelboc the `ui.yaml` is perhaps a bit too light at the moment we might need to add some more information to be able make a better distinction between whats what.

marcus-oscarsson:
@alessandro-olivo, could you rebase this branch ?

alessandro-olivo:
I'll apply the changes proposed by @axelboc (thanks for the explicitness) and rebase the branch

alessandro-olivo:
@marcus-oscarsson Now it should be fine! :)

PR #1647, opened 2025-05-14
Compare the entered oscillation with the current one using motor precision (In Task components)

alessandro-olivo:
This fix prevents the warning "Entered Oscillation start angle is different from current omega" from appearing when the omega motor has a precision greater than 2 and the entered value is actually the same as the current one.

If the motor is in this position:
image

The current behavior is this:
image

This happens because the comparison uses a fixed precision.

### DIFF
--- a/ui/src/components/Tasks/warning.js
+++ b/ui/src/components/Tasks/warning.js
@@ -43,18 +43,16 @@ function warn(values, props) {
   }
 
   const phiPrecision =
-    props.components.find(
-      (el) => el.attribute === 'diffractometer.phi'
-    )?.precision ?? 2;
+    props.components.find((el) => el.attribute === 'diffractometer.phi')
+      ?.precision ?? 2;
 
   if (
     Number.parseFloat(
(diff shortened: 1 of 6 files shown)


ui/src/components/Tasks/warning.js
```diff
@@ -42,10 +42,17 @@ function warn(values, props) {
...
 
+  const phiPrecision =
+    props.components.find((el) => el.attribute === 'diffractometer.phi')
+      ?.precision ?? 2;
+
   if (
     Number.parseFloat(
-      props.beamline.hardwareObjects['diffractometer.phi'].value.toFixed(2),
-    ) !== Number.parseFloat(values.osc_start)
+      props.beamline.hardwareObjects['diffractometer.phi'].value.toFixed(
```
marcus-oscarsson, line 51: Ok, nice :) Thanks !

marcus-oscarsson:
Could you please rebase this branch, you can then merge :)

alessandro-olivo:
@marcus-oscarsson  I've just rebased it :)

PR #1648, opened 2025-05-14
Add domain-based check to distinguish local and remote clients

alessandro-olivo:
This feature allows distinguishing local network clients from external ones based on their domain names. A client is considered local if the final part of its domain (i.e. the base domain or subdomain) matches an entry from a predefined list.

To configure this, just add the list of "local" domains in the `.../mxcube-web/server.yaml` file, under the `mxcube` section.

E.g.
```
...
mxcube:
  USE_EXTERNAL_STREAMER: true
  ...
  LOCAL_DOMAINS: [ 'elettra.eu' ]
  
  usermanager:
...
```

### DIFF
--- a/mxcubeweb/core/util/networkutils.py
+++ b/mxcubeweb/core/util/networkutils.py
@@ -46,15 +46,24 @@ def remote_addr():
     return str(hdr).split(",")[-1]
 
 
-def is_local_network(ip):
+def is_local_network(ip, local_domains=[]):
     localhost = socket.gethostbyname_ex(socket.gethostname())[2][0]
     localhost_range = ".".join(localhost.split(".")[0:2])
     private_address = ".".join(ip.split(".")[0:2])
+    try:
+        private_hostname = socket.gethostbyaddr(ip)[0]
+    except (socket.herror, socket.gaierror):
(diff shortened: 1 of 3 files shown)


mxcubeweb/core/util/networkutils.py
```diff
@@ -46,15 +46,23 @@ def remote_addr():
     return str(hdr).split(",")[-1]
 
 
-def is_local_network(ip):
+def is_local_network(ip, local_domains):
```
marcus-oscarsson: Ok, yes why not I guess. Could you also add a doc string and perhaps even a type hint :)

alessandro-olivo: Sure, no problem!

alessandro-olivo:
Just added docstrings and type hints and rebased :)

marcus-oscarsson:
Sorry I merger your other PR first, it would be great if you could re-bease and then merge :)

alessandro-olivo:
@marcus-oscarsson  No problem at all, just rebased!

About the merge, I think I'm not authorized to do it

image

marcus-oscarsson:
Should be fine now :), you should be able to merge

PR #1649, opened 2025-05-15
Prevent negative clicks during manual centering and display a UI message

alessandro-olivo:
The feature/fix allows handling user clicks that happen while a manual centering step is still in progress. A message is shown on the UI to inform the user.

### The issue 
After the first manual centering click, if the user clicks again before the first step is completed, a message appears on the UI and the number of available clicks is reduced by one. If the user keeps clicking, the click count can go negative, as shown in the screenshot:
image

This behavior can be tested adding a `time.sleep(3)` in the `manual_centring` method of the `DiffractometerMockup` HWO
```diff
    def manual_centring(self):
        """
        Descript. :
        """
        for click in range(3):
            self.user_clicked_event = AsyncResult()
            #self.waiting_for_click = True
            x, y = self.user_clicked_event.get()
            if click < 2:
+                time.sleep(3)
                self.motor_hwobj_dict["phi"].set_value_relative(90)
        self.last_centred_position[0] = x
        self.last_centred_position[1] = y
        centred_pos_dir = self._get_random_centring_position()
        return centred_pos_dir 
```
### The proposed change

In `mxcubeweb/core/components/sampleview.py`

```diff
    def centring_handle_click(self, x, y):
        if HWR.beamline.diffractometer.current_centring_procedure:
+            try:
                HWR.beamline.diffractometer.imageClicked(x, y, x, y)
                self.centring_click()
+            except Exception:
+                return {"clicksLeft": -1}
        else:
            if not self.centring_clicks_left():
                self.centring_reset_click_count()
```

The change allows, raising an Exception from `.imageClicked(...)`, to inform the frontend that the centering step is still in progress. If the frontend receives a negative click (i.e. "-1")  it will show the following message:
image

#### Test the feature:
Just apply the following changes in the `DiffractometerMockup`:
```diff
...
+from gevent.lock import Semaphore

from mxcubecore import HardwareRepository as HWR
from mxcubecore.HardwareObjects.GenericDiffractometer import (
    GenericDiffractometer,
    PhaseEnum,
)


class DiffractometerMockup(GenericDiffractometer):
...

    def init(self):
        """
        Descript. :
        """
        # self.image_width = 100
        # self.image_height = 100

        GenericDiffractometer.init(self)
+      self.waiting_for_click = False
+      self.click_lock = Semaphore()
        self.x_calib = 0.000444

...
    
+    def image_clicked(self, x, y, xi=None, yi=None):
+        with self.click_lock:
+            if self.waiting_for_click:
+                self.waiting_for_click = False
+                self.user_clicked_event.set((x, y))
+            else:
+                self.log.warning("User click but the last centring operation"
+                                              " was still in progress")
+                raise RuntimeError("Last centring operation is still in "
+                                                   "progress, click will be ignored")

    def manual_centring(self):
        """
        Descript. :
        """
        for click in range(3):
            self.user_clicked_event = AsyncResult()
+          self.waiting_for_click = True
            x, y = self.user_clicked_event.get()
            if click < 2:
+              time.sleep(3)
                self.motor_hwobj_dict["phi"].set_value_relative(90)

...
```

#### Thoughts

- Maybe the Exception could be a more specific User-defined Exception ? Should we use user-defined Exceptions in mxcube ?
- Is there a better solution ? Maybe with a new signal o something like that ?

### DIFF
--- a/mxcubeweb/core/components/sampleview.py
+++ b/mxcubeweb/core/components/sampleview.py
@@ -413,9 +413,11 @@ class SampleView(ComponentBase):
 
     def centring_handle_click(self, x, y):
         if HWR.beamline.diffractometer.current_centring_procedure:
-            if not HWR.beamline.diffractometer.is_moving():
+            try:
                 HWR.beamline.diffractometer.imageClicked(x, y, x, y)
                 self.centring_click()
+            except Exception:
+                return {"clicksLeft": -1}
         else:
             if not self.centring_clicks_left():
                 self.centring_reset_click_count()
--- a/mxcubeweb/core/components/sampleview.py
+++ b/mxcubeweb/core/components/sampleview.py
@@ -413,8 +413,9 @@ class SampleView(ComponentBase):
 
     def centring_handle_click(self, x, y):
         if HWR.beamline.diffractometer.current_centring_procedure:
-            HWR.beamline.diffractometer.imageClicked(x, y, x, y)
-            self.centring_click()
+            if not HWR.beamline.diffractometer.is_moving():
+                HWR.beamline.diffractometer.imageClicked(x, y, x, y)
+                self.centring_click()
         else:
             if not self.centring_clicks_left():
                 self.centring_reset_click_count()
--- a/ui/src/actions/sampleview.js
+++ b/ui/src/actions/sampleview.js
@@ -194,16 +194,16 @@ export function recordCentringClick(x, y) {
 
     const { clicksLeft } = json;
     dispatch(centringClicksLeft(clicksLeft));
-    dispatch(
-      videoMessageOverlay(
-        true,
-        `3-Click Centring: <br />${
-          clicksLeft === 0
-            ? 'Save centring or clicking on screen to restart'
-            : `Clicks left: ${clicksLeft}`
-        }`,
-      ),
-    );
+
+    let msg = '3-Click Centring: <br />';
+    if (clicksLeft === 0) {
+      msg += 'Save centring or clicking on screen to restart';
+    } else if (clicksLeft === -1) {
+      msg += 'Please wait, the centring movement is not completed yet';
+    } else {
+      msg += `Clicks left: ${clicksLeft}`;
+    }
+    dispatch(videoMessageOverlay(true, msg));
   };
 }
 


axelboc (commented):
I'm surprised we hadn't caught this bug sooner. :scream: Good catch! :fishing_pole_and_fish: It could explain why we were having issues with this in the Cypress tests...

As before, as a safeguard, I would prefer if the back-end returned a 4xx error to the front-end (which would then display that error via `showErrorPanel`). Then, to improve the user experience, I would find a way to not dispatch the click action while the 3-click centring process is in progress. As you hinted at, this would require a WebSocket message, a Redux action, and a new flag in the state.

marcus-oscarsson:
Very nice :), it would be great with error from the backend as well as @axelboc  pointed out.

marcus-oscarsson:
What you suggest is very nice and I think you did an excellent work wit this PR :)

> Maybe the Exception could be a more specific User-defined Exception ? Should we use user-defined Exceptions in mxcube ?

  - Yes, it would perhaps be preferable, or at least to use a more specific exception
  - As pointed out above this exception could then be handled on the web side and a 400
    error could be returned.

>Is there a better solution ? Maybe with a new signal o something like that ?

  You mean, instead of returning or raising an exception to send a signal and websocket with the result instead ?

  That could also work very well. I'm not sure it would make much of a difference for this kind of short interaction.
  We are not risking any timeout or similar as each click is handled separately. The logic you suggest above is IMO
  somewhat easier to follow and to handle.

axelboc:
> I'm not sure it would make much of a difference for this kind of short interaction.

Yeah, you're right, it'd probably be a bit over-engineered. So keeping the 4xx error response in mind, I think that means basically catching the error and showing that "Please wait" message instead of dispatching `showErrorPanel`, right?

marcus-oscarsson:
Yes, thats what I thought

marcus-oscarsson:
I guess this would also need the necessary changes to `MiniDiff` and `GenericDiffractometer`. Have you tested how this behaves without the changes you suggest to `GenericDiffractometer` ?

alessandro-olivo:
Actually I tried only with `DiffractometerMockup`, but I can test also for the HWOs you proposed

marcus-oscarsson:
Thanks @alessandro-olivo. The best thing would be if you could add the lock code to the two other diffractometer classes `MiniDiff` and `GenericDiffractometer`. I understand that may be complicated as you are perhaps not using them. We will be happy to test the changes to `MiniDiff`.

It would otherwise be nice if you could test without the changes to `DiffractometerMockup` so that we are sure that the changes are "backward compatible". If it turns out to be too time contusing to do the changes to `MiniDiff`, `GenericDiffractometer`.

Thanks again for the super nice work !

alessandro-olivo:
> The best thing would be if you could add the lock code to the two other diffractometer classes MiniDiff and GenericDiffractometer

Right now, the "lock logic" is only implemented in our `XRD1Diffractometer` HWO, which inherits from `GenericDiffractometer` but yes, I totally agree with you @marcus-oscarsson , if we added this feature directly to the "generic" class all subclasses should somehow benefit from it.
I was thinking about introducing a new method, maybe something like `manual_centring_step()` which you have to implement only in the `GenericDiffractometer` subclasses, or possibly modifying the existing `image_clicked()` of the `GenericDiffractometer` method keeping the current behavior if `self.click_lock` isn’t defined, and use the new "lock logic" only when it is.

In any case, I’ll try to implement this logic in both `GenericDiffractometer` and `MiniDiff` HWO as soon as I can, hopefully by the end of this week or early next week.

> It would otherwise be nice if you could test without the changes to `DiffractometerMockup` so that we are sure that the changes are "backward compatible". If it turns out to be too time contusing to do the changes to MiniDiff, GenericDiffractometer.

I've already tested these [changes](https://github.com/mxcube/mxcubeweb/pull/1649/files) with the `DiffractometerMockup` as it is (without the "lock logic") and the behavior is the same as the current one, except for the UI message shown when the frontend receives `-1` clicks

marcus-oscarsson:
Ok, sounds perfect @alessandro-olivo. So ill merge this for now then !

marcus-oscarsson:
Ah, @alessandro-olivo, could you re-base and then merge please ;)

alessandro-olivo:
Done! I think this week I should be able to give a try to the aforementioned changes

PR #1650, opened 2025-05-15
Remove X from Select Session modal

dominikatrojanowska:
Clicking on `X` button of `Select a Session` modal should cause landing on login page similarly to `Sign out` button, accordingly to documentation https://github.com/mxcube/mxcubecore/blob/develop/docs/source/dev/user-based-login-details.md

Check attached screenshot.  
Image

Unreproducible on demo - no `Select Session` modal shows up. Reproduced and tested at Max IV fork.

------------------------

Final decision was to remove misleading X button from modal showing up after logging in if no session is selected.

Provided small change in documentation regarding user-based login: https://github.com/mxcube/mxcubecore/pull/1234

### DIFF
--- a/ui/src/components/LoginForm/SelectProposal.jsx
+++ b/ui/src/components/LoginForm/SelectProposal.jsx
@@ -65,7 +65,7 @@ function SelectProposal() {
         }
       }}
     >
-      <Modal.Header closeButton>
+      <Modal.Header closeButton={showProposalsForm}>
         <Modal.Title>Select a session</Modal.Title>
       </Modal.Header>
       <Modal.Body>


axelboc (commented):
The current behaviour is as follows:

- If `loginType` is "User" and there's no proposal selected, the "select proposal" modal is **non-dismissable** to make it clear that the user cannot continue without selecting a proposal first. It's also non-dismissable so the user doesn't get logged out  inadvertently if they click on the backdrop as a reflex. The only way to proceed is to either select a proposal or to click on the "Sign out" button at the bottom of the modal (thus making this action explicit).
- If a proposal is already selected and the user opens the modal to select another proposal, then the modal is now **dismissable** — i.e. the user can click on the backdrop or the close button to close it.

I think what is confusing is that the modal has a close button even when it's non-dismissable, which it definitely shouldn't.

So I would suggest the following change on line 68 instead:

```jsx
      <Modal.Header closeButton={showProposalsForm}>
```

marcus-oscarsson:
Could you please rebase and then merge ?

Issue #1485, opened 2024-10-31
Clarify purpose of `available_methods` in `beamline_config.yml`


axelboc:
I'm trying to refactor the `ContextMenu` component and some of the code relies on the [beamline configuration](https://github.com/mxcube/mxcubeweb/blob/develop/demo/beamline_config.yml#L146-L155) below, so I'd like to understand its purpose a bit better:

```yml
available_methods:
  datacollection: true
  characterisation: true
  helical: true
  xrf_spectrum: true
  energy_scan: true
  mesh: true
  ssx_chip_collection: true
  gphlworkflow: true
  test_collection: true
```

---

This `available_methods` configuration is first made available to the front-end through the **`beamline.availableMethod`** global state, which is used to hide options from the sample view's context menu.

[The code that does this](https://github.com/mxcube/mxcubeweb/blob/2d3c486ec8862b8ba761763eb0d0d8af73a71fbf/ui/src/components/SampleView/ContextMenu.jsx#L282-L294) is unnecessarily convoluted, but basically it removes any option with a matching key in `available_methods` that is set to `false`. So if we have:

```yml
available_methods:
  datacollection: false
  ...
```

... then the context menu action with key `datacollection` doesn't get displayed. The current logic cares only about "methods" that are set to `false` and ignores the others (i.e. it's an opt-out mechanism).

---

Now the `available_methods` configuration is also used in [`Queue#get_available_tasks()`](https://github.com/mxcube/mxcubeweb/blob/2d3c486ec8862b8ba761763eb0d0d8af73a71fbf/mxcubeweb/core/components/queue.py#L2421). In this case, it's an opt-in mechanism: each task is made available only if its corresponding key is defined and set to `true` in `available_methods`.

Available tasks are made available to the front-end through the `taskForm.defaultParameters` global state:

![image](https://github.com/user-attachments/assets/4b29e830-7753-4bd6-9f07-3dbeaf448cf2)

When a "method" is set to `false` (or not set all) in `available_methods`, the corresponding task no longer appears in the `defaultParameters` object.

---

Okay, so with all this in mind, I'm curious to know how the `available_methods` is actually used in practice:

[...]

PR #1655, opened 2025-05-21
Stop relying on `available_methods` config in UI

axelboc:
Addresses #1485

### Background

The `available_methods` object in `beamline_config.yml` is used to build the `available_tasks` object in the back-end (which ends up being called `defaultParameters` in the front-end). As discussed in #1485, `available_methods` is meant as an opt-in mechanism: a method/task is available if it is defined and set to `true` — e.g. `characterisation: true`.

### Problem

Previously, as explained in #1485, the `ContextMenu` used the `available_methods` config directly to decide which options to hide from the context menu. It was an opt-out algorithm in the sense that if a method/task was not specified at all in the `available_methods` config, then the option would appear in the context menu...

Additionally, there were some unwanted (I think) side effects. For instance, one could hide a workflow option from the context menu by adding `wf-<workflow-name>: false` to `available_methods` as I had suggested in https://github.com/mxcube/mxcubeweb/issues/1485#issuecomment-2456496415.

### Proposed solution

With this PR, the `ContextMenu` now relies solely on the `defaultParameters` state (i.e. `available_tasks`) to decide which options to show — e.g. if this object has a `characterisation` property, then the "Characterisation" options show up in the context menu.

Since this component was the only one relying on the `availableMethods` state directly instead of via `defaultParameters` state, and to avoid re-introducing the issue in the future, I remove this state from the front-end entirely.

---

I'm not closing the issue for now, as there's also the problem that setting some methods/tasks to `false` in `available_methods` does not have the intended effect and may even cause errors.

### DIFF
--- a/ui/src/components/SampleView/ContextMenu.jsx
+++ b/ui/src/components/SampleView/ContextMenu.jsx
@@ -39,9 +39,7 @@ export default function ContextMenu(props) {
   const workflows = useSelector((state) => state.workflow.workflows);
   const enableNativeMesh = useSelector((state) => state.general.useNativeMesh);
   const enable2DPoints = useSelector((state) => state.general.enable2DPoints);
-  const availableMethods = useSelector(
-    (state) => state.beamline.availableMethods,
-  );
+  const availableMethods = new Set(Object.keys(defaultParameters));
   const sampleID = useSelector((state) => state.queue.currentSampleID);
   const sampleData = useSelector(
     (state) => state.sampleGrid.sampleList[sampleID],
@@ -53,6 +51,7 @@ export default function ContextMenu(props) {
 
   const dispatch = useDispatch();
 
+  // eslint-disable-next-line complexity
   function menuOptions() {
     const generalTaskNames = Object.keys(defaultParameters).filter(
       (tname) => !BESPOKE_TASK_NAMES.has(tname),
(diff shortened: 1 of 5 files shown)


ui/src/components/SampleView/ContextMenu.jsx
```diff
@@ -53,6 +51,7 @@ export default function ContextMenu(props) {
 
   const dispatch = useDispatch();
 
+  // eslint-disable-next-line complexity
```
axelboc, line 54: Of course, this component still requires further refactoring.

ui/src/components/SampleView/ContextMenu.jsx
```diff
@@ -160,28 +159,36 @@ export default function ContextMenu(props) {
...
-          text: 'Add XRF Scan',
-          action: () => showModal('xrf_spectrum'),
-          key: 'xrf_spectrum',
-        },
-        {
-          text: 'Add Energy Scan',
-          action: () => showModal('energy_scan'),
-          key: 'energy_scan',
-        },
+        availableMethods.has('datacollection')
```
axelboc, line 164: I purposely did not to keep a filtering loop (cf. below) as the goal is to make this component more declarative.

ui/src/components/SampleView/ContextMenu.jsx
```diff
@@ -314,19 +339,23 @@ export default function ContextMenu(props) {
...
-                  createPointAndShowModal('Characterisation', {
-                    num_imags: 1,
-                  }),
-                key: 6,
-              },
+              availableMethods.has('datacollection')
+                ? {
+                    text: 'Data Collection (Limited OSC)',
+                    action: () => createPointAndShowModal('DataCollection'),
+                    key: 5,
```
axelboc, line 346: The filtering loop below relied on the `key` property to hide options. Since this object's key was not called `'datacollection'`, the option would appear even if `datacollection` was set to `false` in `available_methods`.

Same with the "Characterisation (1 Image)" option below.

mockoocy (approved):

marcus-oscarsson:
It's really nice that you took the time to look at this one :). Thanks !. Its a good start but I think as you mention that there may still be some unwanted behavior when setting mesh scan to false, that needs to be sorted out.

PR #1657, opened 2025-05-27
New layout for generic task forms

mockoocy:
The layout of the form was fixed to be a 2-column grid. Now it's still the default behaviour, but it can be changed. For each field it is possible to specify `uiSchema['ui:options'].col` parameter, to a number between 1 and 12 (or auto), see [bootstrap grid system](https://getbootstrap.com/docs/4.0/layout/grid/)

This way some more sophisticated form layouts can be achieved, like using some full-width custom widgets in future.

Most importantly, it fixes some visual bugs that were present in the past, like errors breaking the UI layout.

The form fields also got changed, from having labels on the left of input fields, to having them on top of the inputs. This makes it clearer when using more than 2 columns.

**Before**
image

**After** 
image

### DIFF
--- a/ui/src/components/Tasks/CustomFieldErrorTemplate.jsx
+++ b/ui/src/components/Tasks/CustomFieldErrorTemplate.jsx
@@ -1,0 +1,26 @@
+import { BsExclamationTriangle } from 'react-icons/bs';
+
+import styles from './CustomFieldErrorTemplate.module.css';
+
+/**
+ *
+ * @param {import('@rjsf/utils').FieldErrorProps} props
+ * @returns
+ */
+export default function CustomFieldErrorTemplate(props) {
+  const { errors = [] } = props;
+  if (errors.length === 0) {
+    return null;
+  }
+
+  return (
+    <div className={styles.fieldErrorContainer}>
+      {errors.map((error) => (
+        <div key={error} className={styles.errorMessage}>
+          <BsExclamationTriangle className={styles.errorIcon} />
+          {error}
+        </div>
+      ))}
+    </div>
+  );
+}
--- a/ui/src/components/Tasks/CustomFieldTemplate.jsx
+++ b/ui/src/components/Tasks/CustomFieldTemplate.jsx
@@ -1,0 +1,75 @@
+import { BsInfoCircleFill } from 'react-icons/bs';
+
+import TooltipTrigger from '../TooltipTrigger';
+import styles from './CustomFieldTemplate.module.css';
+
+/**
+ * @typedef {Object} CustomFieldTemplateProps
+ * @property {string} id
+ * @property {Object} schema - The JSON schema for the field.
+ * @property {string} classNames - CSS classes passed to the field from RJSF.
+ * @property {Object} uiSchema - The UI schema for the field.
+ * @property {string} label - The label for the field.
+ * @property {string} rawDescription - The description for the field (from the UI Schema)
+ * @property {React.ReactNode} children - Contains the field input element.
+ * @property {React.ReactNode} errors - Contains any validation errors. (<ul> tag)
+ * @property {React.ReactNode} help - Contains help text for the field (from UI Schema).
+ * @property {boolean} required - Indicates if the field is required.
+ */
+
+/**
+ * CustomFieldTemplate is a custom field template for React JSON Schema Form (RJSF).
+ * It is used to render fields with a specific layout and styling.
+ * @param {CustomFieldTemplateProps} props
+ * @return {JSX.Element} The rendered custom field template.
+ */
+export default function CustomFieldTemplate(props) {
+  const {
+    id,
+    schema,
+    classNames,
+    uiSchema = {},
+    label,
+    rawDescription,
+    children,
+    errors,
+    help,
+    required,
+  } = props;
+
+  if (schema.type === 'object') {
+    return <div className={classNames}>{children}</div>;
+  }
+
+  const span = Number(uiSchema['ui:options']?.col) || 6;
+  const gridClass = `col-${span}`;
+
+  return (
+    <div className={`${gridClass} ${classNames}`.trim()}>
+      <div className={styles.fieldTitle}>
+        {label && (
+          <label htmlFor={id} className={styles.fieldLabel}>
+            {label}
+            {required && <span className="text-danger">*</span>}
+          </label>
+        )}
+        {rawDescription && (
+          <TooltipTrigger
+            tooltipContent={rawDescription}
+            id={`${id}_tooltip`}
+            inModal
+          >
+            <span>
+              <BsInfoCircleFill />
+            </span>
+          </TooltipTrigger>
+        )}
+      </div>
+      <div className={styles.fieldInput}>
+        {children}
+        <small className={styles.fieldHelp}>{help}</small>
+      </div>
+      <div className={styles.fieldError}>{errors}</div>
+    </div>
+  );
+}
--- a/ui/src/components/Tasks/CustomFieldTemplate.jsx
+++ b/ui/src/components/Tasks/CustomFieldTemplate.jsx
@@ -4,23 +4,9 @@ import TooltipTrigger from '../TooltipTrigger';
 import styles from './CustomFieldTemplate.module.css';
 
 /**
- * @typedef {Object} CustomFieldTemplateProps
- * @property {string} id
- * @property {Object} schema - The JSON schema for the field.
- * @property {string} classNames - CSS classes passed to the field from RJSF.
- * @property {Object} uiSchema - The UI schema for the field.
- * @property {string} label - The label for the field.
- * @property {string} rawDescription - The description for the field (from the UI Schema)
- * @property {React.ReactNode} children - Contains the field input element.
- * @property {React.ReactNode} errors - Contains any validation errors. (<ul> tag)
- * @property {React.ReactNode} help - Contains help text for the field (from UI Schema).
- * @property {boolean} required - Indicates if the field is required.
- */
-
-/**
  * CustomFieldTemplate is a custom field template for React JSON Schema Form (RJSF).
  * It is used to render fields with a specific layout and styling.
- * @param {CustomFieldTemplateProps} props
+ * @param {import('@rjsf/utils').FieldTemplateProps} props
  * @return {JSX.Element} The rendered custom field template.
  */
 export default function CustomFieldTemplate(props) {
--- a/ui/src/components/Tasks/GenericTaskForm.jsx
+++ b/ui/src/components/Tasks/GenericTaskForm.jsx
@@ -10,6 +10,7 @@ import { formValueSelector, reduxForm } from 'redux-form';
 
 import { sendUpdateDependentFields } from '../../api/queue';
 import { DraggableModal } from '../DraggableModal';
+import CustomFieldTemplate from './CustomFieldTemplate';
 import {
   FieldsHeader,
   InputField,
@@ -18,7 +19,6 @@ import {
   StaticField,
   toFixed,
 } from './fields';
-import styles from './GenericTaskForm.module.css';
 import validate from './validate';
 import warn from './warning';
 
@@ -206,49 +206,6 @@ class GenericTaskForm extends React.Component {
     return s;
   }
 
-  customFieldTemplate = ({
-    schema,
-    classNames,
-    uiSchema = {},
-    label,
-    rawDescription,
-    children,
-    errors,
-    help,
-    required,
-    id,
-  }) => {
-    if (schema.type === 'object') {
-      return <div className={classNames}>{children}</div>;
-    }
-
-    // Form fields can be displayed in a grid layout, using bootstrap `col-x` classes.
-    const span = uiSchema['ui:options']?.col || 6;
-    const gridClass = `col-${span}`;
-    return (
-      <div className={`${gridClass} ${classNames}`.trim()}>
-        <div className={styles.fieldTitle}>
-          {label ? (
-            <label htmlFor={id} className={styles.fieldLabel}>
-              {label}
-              {required ? <span className="text-danger">*</span> : null}
-            </label>
-          ) : null}
-          {rawDescription ? (
-            <small className={styles.fieldDescription}>{rawDescription}</small>
-          ) : null}
-        </div>
-
-        <div className={styles.fieldInput}>
-          {children}
-          <small className={styles.fieldHelp}>{help}</small>
-        </div>
-
-        {errors ? <div className={styles.fieldError}>{errors}</div> : null}
-      </div>
-    );
-  };
-
   columnsObjectFieldTemplate({ properties, description }) {
     return (
       <div>
@@ -352,7 +309,7 @@ class GenericTaskForm extends React.Component {
               }}
               templates={{
                 ObjectFieldTemplate: this.columnsObjectFieldTemplate,
-                FieldTemplate: this.customFieldTemplate,
+                FieldTemplate: CustomFieldTemplate,
               }}
             />
           </div>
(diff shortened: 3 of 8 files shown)


ui/src/components/Tasks/GenericTaskForm.jsx
```diff
@@ -205,48 +205,56 @@ class GenericTaskForm extends React.Component {
...
-            {required ? '*' : null}
-            {rawDescription ? ` (${rawDescription})` : null}
-          </label>
-        )}
-        {description}
-        {children}
-        {errors}
-        {help}
+      <div className={`${gridClass} ${classNames} field-container`.trim()}>
+        <div className="field-title">
```
axelboc: Would you be able to use CSS modules for all the class names?

axelboc (commented):
Awesome, with the two columns, it's indeed much clearer to have the labels above the fields.

I notice you use conditional rendering like `rawDescription && ()`, which is great, but in the screenshot, it seems that there are still gaps when fields don't have a description:

image

Would it be able to remove these gaps somehow? Otherwise, it almost looks like the label belongs with the field above it... :joy:

marcus-oscarsson:
Looks quite nice, a good start !

I think it would be really nice if we could remove the space for the components that are not rendered, for instance as @axelboc mentions the description.

It would be nice if we could do something for the input size. I know that its, at the moment, decided by the column layout but it looks a bit strange to have a input field that takes half the dialog for an exposure time that is 2 digits. Not sure if there is a way to specify the input size and then default to something reasonable.

It would also be very nice if we could have some spacing between each row so that its clear what layouts and inputs that go together and it would also be easier on the eye.

ui/src/components/Tasks/GenericTaskForm.jsx
mockoocy: Done - moved all the custom `field-*` classes to a css module.
The `field-container` class got also replaced by a *specific selector* `form-group field:not(.field-object)` which describes the same objects.
`RJFS`-related selectors are still in the plain `.css` file, as it uses the 'normal', global classes.

mockoocy:
Thanks @marcus-oscarsson, @axelboc for the feedback.
Yeah, the vertical gaps were certainly too large there. I was just bothered with situations where one field has a description and the one next to it has no description.

 In that case the input fields were misaligned. so I decided to make the length of the row containing the label and description fixed. But now I agree that it made things worse :)

I've removed some of the vertical gaps now, and it looks like this for 2-col form.
image

The input fields are probably also too wide, as Marcus pointed out.
There is no way to change the width currently, other than changing the amount of cols (the `ui:options.col` property from `uiSchema`)
If that's an issue, it could be made so that the width could be specified using `uiSchema`, if unspecified it would just take width of a whole column. What would be your idea?

axelboc (approved):
Thanks! Yeah, I don't think having the fields not lined up across the two columns is a huge deal. You could put the label and description on the same line with automatic wrapping, which would at least line things up when both the label and description are quite short like in your screenshot... No strong feelings about this, though.

I would maybe just add a tiny bit of space between labels and fields, and between rows, like 2 and 4px respectively maybe (to be tested).

marcus-oscarsson:
Already looks better, I also agree that some more space between the rows would be more pleasant. 

Would it be possible to use a tool-tip for the description and possibly the error as well, one for each with a nice little icon. Like that we would always render the rows the same ?,

It's not extremely disturbing but its somewhat annoying when the rows don't lineup.

Something that would be really neat is if we could create groups with different col layouts for each group and perhaps with an optional title. Somehow one would for instance want Cell A, B C be on one row and Cell alpha, beta, gamma on another. Space group on its own row and a nice title saying Crystal or something like that. Its perhaps pushing it a bit far for this PR, but it would be super nice.

mockoocy:
I think the idea with the tooltip is good. Will add it.

As per @marcus-oscarsson remark; I see that `rjsf` has added something like this in a new release (v6 beta): https://rjsf-team.github.io/react-jsonschema-form/docs/usage/layout-grid 
We could try using it in the future I imagine.

mockoocy:
I decided to keep the description, in a tooltip. Much better! :)
Also I did a slight 'refactor' there. I moved the component that renders the fields to be a separate functional component.


image

ui/src/components/Tasks/CustomFieldTemplate.jsx
```diff
@@ -0,0 +1,71 @@
...
+    return <div className={classNames}>{children}</div>;
+  }
+
+  const span = Number(uiSchema['ui:options']?.col) || 6;
+  const gridClass = `col-${span}`;
+
+  return (
+    <div className={`${gridClass} ${classNames}`.trim()}>
+      <div className={styles.fieldTitle}>
+        {label ? (
```
axelboc: I preferred the `&&` syntax that you had before :grin:

ui/src/components/Tasks/style.css
```diff
@@ -18,33 +18,14 @@
...
-  display: none;
+.form-group .field:not(.field-object) {
+  display: grid;
+  grid-template-rows: auto auto auto;
+}
+/* 
+  Select any element whose id starts with "root_" AND ends with "_tooltip" 
+  These should corresponds to tooltips generated for the GenericTaskForm
+  */
+[id^='root_'][id$='_tooltip'] {
```
axelboc: Not a huge fan of this. I assume it's a bit tricky to pass a custom class name to `TooltipTrigger`. What about a `data` attribute like `data-tooltip-in-modal`?

axelboc (commented):
Even better! Kudos on the TypeDoc comment :clap:

ui/src/components/Tasks/CustomFieldTemplate.jsx
mockoocy: Well, for some reason vscode automatically re-formats it using ternary operator :D

ui/src/components/Tasks/style.css
mockoocy: Oh, yeah - that's much better idea

ui/src/components/Tasks/CustomFieldTemplate.jsx
mockoocy: I just checked; it conflicts with `react/jsx-no-leaked-render` eslint rule.
so I used the `!!var && (<Component/>)` syntax :)

ui/src/components/Tasks/CustomFieldTemplate.jsx
```diff
@@ -0,0 +1,75 @@
+import { BsInfoCircleFill } from 'react-icons/bs';
+
+import TooltipTrigger from '../TooltipTrigger';
+import styles from './CustomFieldTemplate.module.css';
+
+/**
+ * @typedef {Object} CustomFieldTemplateProps
```
marcus-oscarsson: Nice :)

ui/src/components/Tasks/style.css
```diff
@@ -18,33 +18,11 @@
...
-  margin-right: 3em;
-}
-
-.json-schema-form-group-div .checkbox > label > span {
-  display: none;
-}
-
-.json-schema-form-group-div .field-description {
-  display: none;
+.tooltip[data-tooltip-in-modal='true'] {
```
marcus-oscarsson: yeah, thats easier to read :)

marcus-oscarsson:
Very nice, it would be really nice if we could look into the column layout in RJSF V6. Its probably better to do that in a separate PR though.

marcus-oscarsson:
The errors are still displayed inline I guess, do we want to keep it that way for the moment then ?

mockoocy:
> The errors are still displayed inline I guess, do we want to keep it that way for the moment then ?

I think for now it is good enough. Maybe in the future we could remove the list of errors on top of the form (unless it has some valid use case)  and perhaps think of a nicer way of displaying these.

ui/src/components/Tasks/CustomFieldTemplate.jsx
axelboc: Right, I actually ended up turning off that rule in the shared config: https://gitlab.esrf.fr/ui/eslint-config/-/merge_requests/4 — if you feel like upgrading, go for it, otherwise no worries.

axelboc:
Agreed that it's good enough for now and that the errors at the top could probably be removed later on. I think errors below the fields are fine (except perhaps for some minor styling tweaks later on), since they shouldn't ever be visible when the modal appears, so it doesn't matter too much if the layout shifts a bit once they start appearing.

ui/src/components/Tasks/CustomFieldTemplate.jsx
mockoocy: Oh, it just seems that I had outdated dependencies on my local environment. Now it works perfectly, thanks :)
Moved all these ternaries to use `&&` instead.

jbflo:
> I decided to keep the description, in a tooltip. Much better! :) Also I did a slight 'refactor' there. I moved the component that renders the fields to be a separate functional component.
> 
> image

A little margin top on the labels maybe ?  the Labels are a bit too close to the fields on top.

mockoocy:
> A little margin top on the labels maybe ? the Labels are a bit too close to the fields on top.

Good idea. 
image

marcus-oscarsson:
Starts to look pretty nice. I think its fine for this iteration. I do think we should think about the error display but thats for another discussion.  Another thing that would be really nice is to remove the generic task from from the redux-form infrastructure.

marcus-oscarsson:
There is long weekend in France starting from tomorrow. Feel free to merge if you are done ! Thanks again

mockoocy:
I've created an issue to move the `GenericTaskForm` to use `react-hook-form` instead of `redux-form`, https://github.com/mxcube/mxcubeweb/issues/1660.

Also, I included some additional changes, that you'll be able to review after the long weekend :)

First of all, I changed the JSDoc type hints, to use type definitions directly from the `rjsf` library. 
Also, I decided to include changes to how errors are displayed in this PR afterall. This is still somewhat relevant to this PR, and required me to change the css that I've introduced here.

image

marcus-oscarsson:
Looks really nice :)

marcus-oscarsson:
Thanks for creating the issue !

ui/src/components/Tasks/CustomFieldTemplate.jsx
```diff
@@ -0,0 +1,61 @@
+import { BsInfoCircleFill } from 'react-icons/bs';
+
+import TooltipTrigger from '../TooltipTrigger';
+import styles from './CustomFieldTemplate.module.css';
+
+/**
+ * CustomFieldTemplate is a custom field template for React JSON Schema Form (RJSF).
+ * It is used to render fields with a specific layout and styling.
+ * @param {import('@rjsf/utils').FieldTemplateProps} props
```
marcus-oscarsson, line 9: Oh, wow I see

ui/src/components/Tasks/CustomFieldErrorTemplate.jsx
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ *
+ * @param {import('@rjsf/utils').FieldErrorProps} props
+ * @returns
+ */
+
+export default function CustomFieldErrorTemplate(props) {
+  const { errors = [], idSchema } = props;
+  // const fieldName = schema.title || schema.name || 'Field';
```
axelboc: commented code

ui/src/components/Tasks/CustomFieldErrorTemplate.jsx
```diff
@@ -0,0 +1,34 @@
...
+  if (errors.length === 0) {
+    return null;
+  }
+
+  return (
+    <div className={styles.fieldErrorContainer}>
+      {errors.map((error, idx) => (
+        <div
+          // eslint-disable-next-line react/no-array-index-key
+          key={`${idSchema.$id}_error_${idx}`}
```
axelboc: There's no benefit to generating complicated keys here. They just need to be unique within their parent element (and ideally not based on an array index). I would simply use the error message as key: `key={error}`, as it is likely unique. (and remove the `idSchema` prop altogether)

ui/src/components/Tasks/CustomFieldErrorTemplate.jsx
```diff
@@ -0,0 +1,34 @@
...
+          className={`${styles.errorMessage} ${
+            idx < errors.length - 1 ? styles.errorMessageWithBorder : ''
```
axelboc: I would typically use a `data` attribute for this, like `data-with-border={idx < errors.length - 1 || undefined}`, and then `.errorMessage[data-with-border]`.

That being said, I think it'd look better without the border to be honest.

ui/src/components/Tasks/CustomFieldErrorTemplate.jsx
```diff
@@ -0,0 +1,34 @@
...
+          }`}
+        >
+          <BsExclamationTriangle className={styles.errorIcon} />
+          <span>{error}</span>
```
axelboc: Is the `span` needed?

ui/src/components/Tasks/CustomFieldTemplate.jsx
```diff
@@ -0,0 +1,61 @@
...
+            {required && <span className="text-danger">*</span>}
+          </label>
+        )}
+        {rawDescription && (
+          <TooltipTrigger
+            tooltipContent={rawDescription}
+            id={`${id}_tooltip`}
+            inModal
+          >
+            <span>
```
axelboc, line 48: `span` needed?

ui/src/components/Tasks/style.css
axelboc: FYI, if you want to just write `.tooltip[data-tooltip-in-modal]`, you can do this in the JSX: `data-tooltip-in-modal={inModal || undefined}`

ui/src/components/Tasks/CustomFieldErrorTemplate.jsx
```diff
@@ -0,0 +1,34 @@
+import { BsExclamationTriangle } from 'react-icons/bs';
+
+import styles from './CustomFieldErrorTemplate.module.css';
+
+/**
+ *
+ * @param {import('@rjsf/utils').FieldErrorProps} props
+ * @returns
+ */
+
```
axelboc: Likely doesn't matter, but maybe remove that extra blank line?

axelboc (commented):
Even better :+1:

axelboc:
I'm re-approving obviously. :wink: Next time, don't hesitate to click on "re-request review" if you think the new changes deserve another look over so I know to prioritise a follow-up review:

image

marcus-oscarsson:
@mockoocy I let you decide what to do with the last comments, you can merge when you feel that you are done. Unless you need to add some major changes of course.

ui/src/components/Tasks/CustomFieldErrorTemplate.jsx
mockoocy: No, it was just a leftover.

ui/src/components/Tasks/CustomFieldTemplate.jsx
mockoocy: Yes, it's needed. 
The children to `OverlayTrigger` must accept a ref, so a plain `svg` tag is not sufficient. 
https://react-bootstrap.netlify.app/docs/components/overlays/#overlaytrigger

mockoocy:
> @mockoocy I let you decide what to do with the last comments, you can merge when you feel that you are done. Unless you need to add some major changes of course.

Well, I don't have permissions to merge actually, so I will let @axelboc have one last look and merge it :)

marcus-oscarsson:
@mockoocy You should be able to merge after you accepted the invitation I just sent you ;)

mockoocy:
> @mockoocy You should be able to merge after you accepted the invitation I just sent you ;)

It works, thanks ;)

PR #1659, opened 2025-05-28
Prevent undefined property error

jbflo:
ISSUE : to reproduce add a manual sample  - > "Create new sample " Button
then click on " get Samples " Button

- new Sample list state does not contain old manual sampleID


marcus-oscarsson:
Thanks @jbflo

PR #1661, opened 2025-05-29
Adds option to draw a line between 2d points

mockoocy:
This is a feature used in MAXIV for some of the custom tasks. It adds an object type in context menu for a pair of 2D Points. Existing `HELICAL` object type did not work, as it works for a pair of 3D Point.
I could alternatively change context menu code, so that `HELICAL` object type would be common across 2D and 3D point, I am not sure if such behaviour would be desirable though.

### DIFF
--- a/ui/src/components/SampleView/ContextMenu.jsx
+++ b/ui/src/components/SampleView/ContextMenu.jsx
@@ -278,6 +278,11 @@ export default function ContextMenu(props) {
               key: 'helical',
             }
           : {},
+        {
+          text: 'Add Line',
+          action: () => createLineOnCanvas(shape.id),
+          key: 'create_line',
+        },
         ...genericTasks.line,
       ],
       LINE: [
@@ -442,6 +447,10 @@ export default function ContextMenu(props) {
         sid,
       ),
     );
+  }
+
+  function createLineOnCanvas(refs) {
+    dispatch(addShape({ t: 'L', refs }));
   }
 
   function savePoint() {
--- a/ui/src/components/SampleView/SampleImage.jsx
+++ b/ui/src/components/SampleView/SampleImage.jsx
@@ -409,7 +409,7 @@ export default class SampleImage extends React.Component {
           .filter((shape) => this.props.lines[shape.id] !== undefined)
           .map((shape) => shape.id);
 
-        if (threeDpointList.length === 2) {
+        if (threeDpointList.length === 2 || twoDPointList.length === 2) {
           ctxMenuObj = { type: 'HELICAL', id: this.props.selectedShapes };
         } else if (
           threeDpointList.length === 1 &&
--- a/ui/src/components/Tasks/warning.js
+++ b/ui/src/components/Tasks/warning.js
@@ -58,9 +58,16 @@ function warn(values, props) {
       'Entered Oscillation start angle is different from current omega';
   }
 
+  const shapeID = Array.isArray(props.pointID)
+    ? props.pointID[0]
+    : props.pointID;
+
+  const is2DPoint = shapeID !== -1 && shapeID.includes('2D');
+  // lines property is only present for helical
+  const line = props.lines?.[shapeID];
+  const is2DLine = line && line.refs.some((ref) => ref.includes('2D'));
   if (
-    props.pointID !== -1 &&
-    props.pointID.includes('2D') &&
+    (is2DPoint || is2DLine) &&
     Number.parseFloat(values.osc_range) * Number.parseFloat(values.num_images) >
       5
   ) {


marcus-oscarsson:
Good point, actually `HELICAL` would imply some rotation. What we have done so far is that its understood that a `HELICAL` without rotation is `LINE` scan. But it would perhaps be worth adding a line `LINE` type or `HELICAL` with limited rotation if that is whats needed ?

mockoocy:
> Good point, actually `HELICAL` would imply some rotation. What we have done so far is that its understood that a `HELICAL` without rotation is `LINE` scan. But it would perhaps be worth adding a line `LINE` type or `HELICAL` with limited rotation if that is whats needed ?

What's needed is an option to create a line between 2D points to interpolate along during some data collection.
Potentially an option to create a line between 2 3D points would also be nice to have (not something we'd need tho)

I don't think it's necessary to create a new object type for a line between 2D points, I was just curious whether I could hook these 2D points into the `HELICAL` type, as it was not really clear to me what the distinction was there.

beteva:
Is there a need to distinguish helical from line scan - line scan is just helical with 0 degrees rotation, isn't it? Would not it be enough to check in the acquisition parameters if omega is the same or not? Or am I missing anything?

marcus-oscarsson:
That was sort of my question. 2D points are not centered in space so it means that rotation should be limited to a certain number of degrees. That is way it has not, so far, been possible to create a helical collection between two 2D-points. The concept of 2D-points was also invented before we could do zero degree oscillations (hardware limitation). That is why there is no pure line option. Today, performing a 0-degree helical is the same as a line scan

It would be fine to me to use the helical as long as its clear that rotation is likely limited, and especially if you are doing a line.

mockoocy:
So if I understand correctly, when we have a "line" shape, then we are not sure if it is a line between 2D or 3D points - user may not know that the rotation is limited.

Well, I am not really sure what to do about it. It could be handled by the site's mxcubecore implementation via the acquisition parameters probably.

marcus-oscarsson:
@mockoocy, I think you got it. Its quite easy to handle but it depends on what you expect to do with this line. I think it would be fine If you default the oscillation range to 0 and add a warning message to the user for any oscillation bigger than for instance 5 degrees.

I imagine that the idea is to scan a Jet or chip and hence you would not need any oscillation ?

However its very likely that the crystal would rotate out of the beam if a user performs such a scan with rotation. Its quite use case specific so it would help to know what you plan to do.

mockoocy:
> that the idea is to scan a Jet or chip and hence you would not need any oscillation ?

Yes, that's right.
We do a type of collection in which data is collected along a line from crystals flowing in a viscous fluid.
There is no rotation, and user can't change it.

For the already existing `Helical` task, I think displaying this warning message (can't rotate further than e.g., 5deg) would be a good idea if it's been invoked from a line connecting 2D points. If you agree, I'll do that in this PR.

marcus-oscarsson:
Sounds very good, thanks @mockoocy

marcus-oscarsson:
So, this is still in progress then I guess, could you WIP/Draft it if its the case ?

mockoocy:
> So, this is still in progress then I guess, could you WIP/Draft it if its the case ?

It's just got done :)

marcus-oscarsson:
Ok, its just I did not find anything that displays the warning message to the user or default `osc_range` to 0 ? Its fine for me if that is done in a separate PR but I though that it was finally something you wanted to include, or did i misunderstand ?

marcus-oscarsson:
Ah, sorry. Now I see there is something for the `osc_range` but its perhaps missing a message ?

marcus-oscarsson:
Ok, never mind I see that you use the same message as for the point so thats all good. The message is perhaps a bit vague as it is but I guess we can leave it as this.  

And for the default `osc_range` what would the value be ?

mockoocy:
Yes, I just expanded the function that displays the warning messages in the frontend forms, to also consider these 2d lines.

> And for the default osc_range what would the value be ?

It is configurable in the `beamline_config` in the `default_acquisition_parameters`. 
On the `demo` config it's 0.1

marcus-oscarsson:
Ok, I see. I guess it will be slightly inconvenient as most users that perform a helical or actually line scan between two 2d-points would like `osc_range` to be 0. 

I would suggest to add a section in `default_acquisition_parameters` called `line` that set `osc_range` to 0. And add a new scan type that simply is line scan. 

The current proposal would work, so its fine. Its just that the users need to change the oscillation range. Let me know how you would like to do. We can merge this for now if you guys are happy with the current solution. I can see that we would probably like to evolve this to a specific line scan or similar if we where to use it.

mockoocy:
I think we could introduce an additional scan type in some future PR.
For now I think the way it's been introduced in this PR works good enough for us, but the UX around these line scans surely should be improved.
If the thing with the default `osc_range` value would be too bothersome, we could make the changes from this PR opt-in somehow.

marcus-oscarsson:
I think its fine as it as, especially as it adds a feature. I think we may need to change it if is frequently used.

Issue #1658, opened 2025-05-28
Get Sample List + Sync LIMS  Behavior - The functionality is bundled together,


jbflo:
*** Edited ***

🧭 Current Behavior
When a user logs in and navigates to the Sample Table, the sample list is initially empty.

✅ Goal
We want the sample list to be automatically fetched and synchronized with LIMS upon login, so the user does not have to manually click "Get Sample List" to see their samples.

axelboc:
Definitely! I feel like we talked about this "bundled" beahviour a long while ago and agreed it was confusing. 💯

marcus-oscarsson:
Yes, thats sounds like a good idea. Lets try that :)

jbflo:
after discussion it was found that The  bundled functionality is well neeed.
So we decide to focus on making the sample list to be automatically fetched and synchronized with LIMS upon login,

PR #1663, opened 2025-06-03
Get Sample List + Sync LIMS Behavior - >  Avoid Empty Sample Table by Syncing on Login

jbflo:
*** EDITED ***
This PR addresses [issue #1658](https://github.com/mxcube/mxcubeweb/issues/1658) and implements the following improvements:

The sample list is automatically fetched when a session is selected / on Login
and the Lims Filter is also automatically apply. 

For a next iteration we could define a default lims in the config .
but for now we use the fist availble "ISPYB" As it was before. 

- User still have the options to Sync with other  LIMS options or just get Sample from SC.

I also add some little side improvement. ( i add them in their own commit )
 -  Added Tooltip to view mode
 - Disable Clear Samples when Sample List is Empty
 - Add "Get Samples & Sync with LIMS" as a split button:
     - Clicking the main button performs the default sync using the first available LIMS (e.g. ISPyB).
     - The dropdown menu allows users to manually choose other sources like:
          - Sample Changer (SC)
          - Other LIMS backends (if available)

[Screencast from 2025-06-17 10-10-08.webm](https://github.com/user-attachments/assets/b6543529-86e7-4a9c-b5d3-b779ec2d5800)

**PS :  The original PR Content, tittle and description has be update after the following suggestion in the conversation Bellow.**

### DIFF
--- a/ui/src/actions/login.js
+++ b/ui/src/actions/login.js
@@ -14,6 +14,7 @@ import { fetchSampleChangerInitialState } from '../api/sampleChanger';
 import { fetchImageData, fetchShapes } from '../api/sampleview';
 import { fetchAvailableWorkflows } from '../api/workflow';
 import { applicationFetched, showErrorPanel } from './general';
+import { syncSampleListWithLims } from './sampleGrid';
 
 export function setLoginInfo(loginInfo) {
   return {
@@ -85,7 +86,7 @@ export function signOut() {
 }
 
 export function getInitialState() {
-  return async (dispatch) => {
+  return async (dispatch, getState) => {
     const initialStateSlices = await Promise.all([
       fetchUIProperties()
         .then((uiproperties) => ({ uiproperties }))
@@ -154,6 +155,18 @@ export function getInitialState() {
 
     dispatch(setInitialState(Object.assign({}, ...initialStateSlices)));
     dispatch(applicationFetched(true));
+
+    const state = getState();
+    const { sampleList } = state.sampleGrid;
+    const { loginInfo } = state.login;
+    if (Object.keys(sampleList).length === 0) {
+      // If the sample list is empty, synchronize it with LIMS
+      // Use the first LIMS available as default
+      // This is done here to ensure that the sample list is populated
+      // when the application is initialized
+      // and the user is logged in
+      dispatch(syncSampleListWithLims(loginInfo?.limsName[0]?.name));
+    }
   };
 }
 
--- a/ui/src/containers/SampleListViewContainer.jsx
+++ b/ui/src/containers/SampleListViewContainer.jsx
@@ -789,12 +789,12 @@ export default function SampleListViewContainer() {
                   className="nowrap-style"
                   variant="outline-secondary"
                   onClick={() => dispatch(showConfirmClearQueueDialog())}
-                  disabled={queue.queueStatus === QUEUE_RUNNING}
+                  disabled={
+                    queue.queueStatus === QUEUE_RUNNING ||
+                    Object.keys(sampleList).length === 0
+                  }
                 >
-                  <i
-                    className="fas fa-minus"
-                    style={{ marginRight: '0.5em' }}
-                  />
+                  <i className="fas fa-trash-alt me-1" />
                   Clear sample list
                 </Button>
               </TooltipTrigger>
--- a/ui/src/containers/SampleListViewContainer.jsx
+++ b/ui/src/containers/SampleListViewContainer.jsx
@@ -31,7 +31,7 @@ import {
   selectSamplesAction,
   setViewModeAction,
   showGenericContextMenu,
-  syncSamples,
+  syncSampleListWithLims,
 } from '../actions/sampleGrid';
 import { showTaskForm } from '../actions/taskForm';
 import TooltipTrigger from '../components/TooltipTrigger';
@@ -215,33 +215,22 @@ export default function SampleListViewContainer() {
     }
   }
 
-  async function getSamplesFromSC() {
-    const manualSamples = [];
-    Object.values(sampleList).forEach((sample) => {
-      if (sample.location === 'Manual') {
-        manualSamples.push(sample.sampleID);
-      }
-    });
-    // first need to remove manual sample from queue
-    // because they will be remove from sample List
-    await dispatch(setEnabledSample(manualSamples, false));
-
+  function getSamplesFromSC() {
     dispatch(getSamplesList());
+    sampleGridClearFilter();
   }
 
   /**
    * Synchronises samples with LIMS
-   *
-   * @property {Object} loginData
+   * @param {string} lims - LIMS name to synchronise with
+   * this also automatically filters the sample list to only show samples
+   * that have LIMS data
+   * @see syncSamples
+   * @see filterAction
+   * @see syncSampleListWithLims
    */
-  async function handleSyncSamples(lims) {
-    if (Object.keys(sampleList).length === 0) {
-      await getSamplesFromSC();
-      dispatch(syncSamples(lims));
-    } else {
-      dispatch(syncSamples(lims));
-    }
-    dispatch(filterAction({ limsSamples: true }));
+  function handleSyncSamples(lims) {
+    dispatch(syncSampleListWithLims(lims));
   }
 
   /**
--- a/ui/src/containers/SampleListViewContainer.jsx
+++ b/ui/src/containers/SampleListViewContainer.jsx
@@ -800,12 +800,17 @@ export default function SampleListViewContainer() {
               </TooltipTrigger>
               <span style={{ marginLeft: '1.5em' }} />
               <Dropdown>
-                <Dropdown.Toggle
-                  variant="outline-secondary"
-                  id="dropdown-basic"
+                <TooltipTrigger
+                  id="sync-samples-tooltip"
+                  tooltipContent="Change Samples List View Mode"
                 >
-                  <MdGridView size="1em" /> {viewMode.mode}
-                </Dropdown.Toggle>
+                  <Dropdown.Toggle
+                    variant="outline-secondary"
+                    id="dropdown-basic"
+                  >
+                    <MdGridView size="1em" /> {viewMode.mode}
+                  </Dropdown.Toggle>
+                </TooltipTrigger>
                 <Dropdown.Menu>
                   {viewMode.options.map((option) => (
                     <Dropdown.Item
--- a/ui/src/containers/SampleListViewContainer.jsx
+++ b/ui/src/containers/SampleListViewContainer.jsx
@@ -1,6 +1,7 @@
-import { useCallback, useEffect } from 'react';
+import React, { useCallback, useEffect } from 'react';
 import {
   Button,
+  ButtonGroup,
   Card,
   Col,
   Container,
@@ -547,43 +548,62 @@ export default function SampleListViewContainer() {
   }
 
   function getSynchronizationDropDownList() {
-    if (loginData.limsName.length === 1) {
-      return (
+    return (
+      <Dropdown as={ButtonGroup}>
         <TooltipTrigger
           id="sync-samples-tooltip"
-          tooltipContent={`Synchronise sample list with ${loginData.limsName[0].name}`}
+          tooltipContent={`Synchronise sample list with ${loginData.limsName[0].name},
+          and apply filter to only show with LIMS samples`}
         >
           <Button
-            className="nowrap-style"
             variant="outline-secondary"
+            className="nowrap-style"
             onClick={() => handleSyncSamples(loginData.limsName[0].name)}
           >
             <i className="fas fa-sync-alt" style={{ marginRight: '0.5em' }} />
-            Get Samples
+            Get Samples & Sync {loginData.limsName[0].name}
           </Button>
         </TooltipTrigger>
-      );
-    }
-    return (
-      <Dropdown>
-        <Dropdown.Toggle variant="outline-secondary" id="dropdown-lims">
-          <i className="fas fa-sync-alt" style={{ marginRight: '0.5em' }} />{' '}
-          Synchronize with
-        </Dropdown.Toggle>
+        <Dropdown.Toggle
+          split
+          variant="outline-secondary"
+          id="dropdown-split-samples"
+          title="Other LIMS Options"
+        />
         <Dropdown.Menu>
-          {loginData.limsName.map((lims) => (
-            <TooltipTrigger
-              key={lims.name}
-              tooltipContent={`Synchronise sample list with ${lims.name}`}
+          <TooltipTrigger tooltipContent="get samples from sample changer">
+            <Dropdown.Item
+              onClick={() => getSamplesFromSC()}
+              variant="outline-secondary"
+              className="nowrap-style"
             >
-              <Dropdown.Item
-                key={lims.name}
-                onClick={() => handleSyncSamples(lims.name)}
-              >
-                {lims.name}
-              </Dropdown.Item>
-            </TooltipTrigger>
-          ))}
+              Get Samples from SC
+            </Dropdown.Item>
+          </TooltipTrigger>
+          {loginData.limsName.length > 1 && <Dropdown.Divider />}
+          {loginData.limsName.map((lims, lindex) => {
+            if (lindex === 0) {
+              // Skip the first LIMS as it is already included in the button above
+              return null;
+            }
+
+            return (
+              <React.Fragment key={lims.name}>
+                <TooltipTrigger
+                  tooltipContent={`Synchronise sample list with ${lims.name}`}
+                >
+                  <Dropdown.Item
+                    variant="outline-secondary"
+                    className="nowrap-style"
+                    onClick={() => handleSyncSamples(lims.name)}
+                  >
+                    Get Samples & Sync {lims.name}
+                  </Dropdown.Item>
+                </TooltipTrigger>
+                {loginData.limsName.length < lindex && <Dropdown.Divider />}
+              </React.Fragment>
+            );
+          })}
         </Dropdown.Menu>
       </Dropdown>
     );
--- a/ui/cypress/e2e/app.cy.js
+++ b/ui/cypress/e2e/app.cy.js
@@ -14,7 +14,9 @@ describe('app', () => {
 
     // Samples list
     cy.findByRole('link', { name: 'Samples' }).click();
-    cy.findByRole('button', { name: 'Get Samples' }).should('be.visible');
+    cy.findByRole('button', { name: 'Add Task to Samples' }).should(
+      'be.visible',
+    );
 
     // Equipment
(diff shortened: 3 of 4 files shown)


ui/src/containers/SampleListViewContainer.jsx
```diff
@@ -77,6 +82,20 @@ export default function SampleListViewContainer() {
...
+
+  useEffect(() => {
+    if (!hasMounted.current) {
+      hasMounted.current = true;
+
+      if (!showClearSamplesButton) {
+        getSamplesFromSC();
+      }
+    }
+  });
```
axelboc: You won't need the `hasMounted` ref if you pass an empty deps array: `useEffect(()=> {...}, [])`

axelboc:
Just looking at the screen recording, it already feels like a huge usability improvement from before, congrats! 🎉 

Two points before I look at the code: 

1. By refreshing the sample list on mount, you effectively make this a side effect of navigating to the sample list page.

Is the goal to keep the sample list always up to date? If so, then it should rather be done via a polling timer or back-end signal. That being said, I might of course be wrong, but I _don't_ think we want/need to keep the sample list up to date. The issue we're trying to solve here is just to not have an empty sample list when users first log in, and this just requires prefilling the list once after logging in (in the global state initialisation code).

2. I'm wondering if we could avoid adding that checkbox that controls whether or not to apply the filter automatically after refreshing the sample list. What's the downside of never applying the filter automatically?

jbflo:
> 1. By refreshing the sample list on mount, you effectively make this a side effect of navigating to the sample list page.
> 
> Is the goal to keep the sample list always up to date? If so, then it should rather be done via a polling timer or back-end signal. That being said, I might of course be wrong, but I _don't_ think we want/need to keep the sample list up to date. The issue we're trying to solve here is just to not have an empty sample list when users first log in, and this just requires prefilling the list once after logging in (in the global state initialisation code).
>
Not for now — I believe that when scientists enter the hutch to manually change samples, they typically restart MXCuBE afterward. That’s probably why the “Clear” and “Get Samples” functionality was originally implemented.
by  trying to resolve the issue of an empty sample list when users first log in, I ended up slightly thinking of solving the backend sync behavior as well.
As you said said, a backend signal or event-based update would ideally be more effective.

> 2. I'm wondering if we could avoid adding that checkbox that controls whether or not to apply the filter automatically after refreshing the sample list. What's the downside of never applying the filter automatically?

Perhaps the scientists at MaxIV specifically wanted this behavior? I believe @meguiraun previously made changes to automatically apply the LIMS filter when syncing with ISPyB.
I decided to keep that as the default behavior, but made it optional via the checkbox, since there are known use cases where auto applying the LIMS filter is not desirable. But i am happy to remove it if it fit everyone .

ui/src/containers/SampleListViewContainer.jsx
```diff
@@ -77,6 +82,20 @@ export default function SampleListViewContainer() {
     handleSetViewMode(localStorageViewMode || viewMode.mode);
   }, [handleSetViewMode, viewMode.mode]);
 
+  const getSamplesFromSC = useCallback(() => {
+    dispatch(getSamplesList());
+  }, [dispatch]);
+
+  useEffect(() => {
```
jbflo: I could also do like this. 
```suggestion
    useEffect(() => {
    if (!showClearSamplesButton) {
      getSamplesFromSC();
    }
  // eslint-disable-next-line react-hooks/exhaustive-deps
  }, []);
```

ui/src/containers/SampleListViewContainer.jsx
```diff
@@ -214,33 +233,15 @@ export default function SampleListViewContainer() {
...
-  }
-
   /**
    * Synchronises samples with LIMS
-   *
-   * @property {Object} loginData
    */
-  async function handleSyncSamples(lims) {
-    if (Object.keys(sampleList).length === 0) {
-      await getSamplesFromSC();
```
jbflo, line 238: no longer need to remove  manual samples in queue. they get remove in backend

axelboc:
> Perhaps the scientists at MaxIV specifically wanted this behavior? I believe @meguiraun previously made changes to automatically apply the LIMS filter when syncing with ISPyB.

Yeah I would like to know why this behaviour was implemented in the first place. I'm thinking that maybe it's because the filters reset when clearing the sample list, in which case I'd suggest changing that behaviour and leaving the filters alone when retrieving/clearing samples.

axelboc:
>  As you said said, a backend signal or event-based update would ideally be more effective.

I guess with even-based updates, it would become the responsibility of the back-end to poll the state of the sample changer?

Anyway, perhaps this is worth addressing separately, what do you think? It brings up other questions, like how to convey this "liveness" of the sample list to users visually, how the LIMS synchronisation fits into it, etc. I think it would be sufficient for now (and a huge improvement already) to get the samples once on initial log in.

jbflo:
> > As you said said, a backend signal or event-based update would ideally be more effective.
> 
> I guess with even-based updates, it would become the responsibility of the back-end to poll the state of the sample changer?

yeah while performing an action like mount etc..  

> Anyway, perhaps this is worth addressing separately, what do you think? It brings up other questions, like how to convey this "liveness" of the sample list to users visually, how the LIMS synchronisation fits into it, etc. I think it would be sufficient for now (and a huge improvement already) to get the samples once on initial log in.

Agreed

fabcor-maxiv:
I guess this kind of change in UI/UX behaviour would be best discussed with the users of MXCuBE (beamlines staff, not among developers only). I will try to relay the discussion.

fabcor-maxiv:
Side note: `migueraun` is not involved with MAX IV currently, please `@`-ping someone else if needed.

Maybe we should maintain lists of GitHub user contacts.

marcus-oscarsson:
To my recollection it was scientists at both ESRF and MAIV that wanted to automatically filter out non LIMS samples. We can double check that here as well and see what the status is after having used it for some time. I think as with many things some people find it confusing others practical.

marcus-oscarsson:
Just a comment on events and update of the sample changer contents. It may be that the only one that knows that the sample contents have changed is the person filling the the sample changer. The sample changer might not be capable of detecting changes and thus can't send any event. 

The easiest and cleanest solution, without having given it a longer though about it, would be to retrieve the samples when a session is selected. (There is an event for that)

The need for somehow retrieving the sample-changer contents explicitly would probably still be needed unless there is some other elegant solution that we did not think about. Having the sample retrieval be triggered by only navigation to the `samplelist` would probably be too limiting. I think its still the case that  "synch with lims" still gets all the contents from the sample changer and then synchronizes with the lims right ?

Another question, how does this work for remote users ?, is the sample retrieval also triggered when they navigate to the samplelist ?

epanepucci:
I'll give my two cents here, hang on.

When a user first logs in and goes to the **Samples** tab, they should **only** see samples that belong to their proposal. It is very common to have samples from multiple proposals loaded in the dewar at the same time and if all samples in the dewar are fetched and made available in mxcube, it means that user from proposal AAAA could mount samples from proposal BBBB and this is a serious security/privacy issue. BTW, I just did exactly that as a test case.

In the case that a user has more pucks than the capacity of the sample changer dewar, then one could pause the beamtime, go inside the hutch, remove/load pucks and then continue. In this situation a **refresh** button might be interesting to update the list of currently available samples.

marcus-oscarsson:
Very good points @epanepucci, thanks for the input

jbflo:
> The need for somehow retrieving the sample-changer contents explicitly would probably still be needed unless there is some other elegant solution that we did not think about. Having the sample retrieval be triggered by only navigation to the `samplelist` would probably be too limiting. I think its still the case that "synch with lims" still gets all the contents from the sample changer and then synchronizes with the lims right ?

YES
> Another question, how does this work for remote users ?, is the sample retrieval also triggered when they navigate to the samplelist ?

I would say  yes, as this action does not require_control

jbflo:
> I'll give my two cents here, hang on.
> 
> When a user first logs in and goes to the **Samples** tab, they should **only** see samples that belong to their proposal. It is very common to have samples from multiple proposals loaded in the dewar at the same time and if all samples in the dewar are fetched and made available in mxcube, it means that user from proposal AAAA could mount samples from proposal BBBB and this is a serious security/privacy issue. BTW, I just did exactly that as a test case.
> 
I see — in that case, this goes beyond a simple UI filter and should be enforced at the backend level.
Right now, bundling "Get Samples" with the ISPyB sync and applying a filter in the UI is mostly cosmetic. A user can still disable the filter and access all samples, which means this PR cannot fully address the underlying security/privacy concern, but thanks i note this in a corner head. To properly restrict access, sample-level filtering based on proposal ownership should be handled server-side as part of the data access logic.

> In the case that a user has more pucks than the capacity of the sample changer dewar, then one could pause the beamtime, go inside the hutch, remove/load pucks and then continue. In this situation a **refresh** button might be interesting to update the list of currently available samples.

Indeed a **refresh**  might be interesting, for now it's 2 action clear then get.

marcus-oscarsson:
I think that we should perhaps take the larger discussion on how to re-design the sample list in a separate meeting (for the future UI/UX working group :)).

For this PR/issue what I would propose is looking at the original issue: https://github.com/mxcube/mxcubeweb/issues/1658

>
>Pressing “Get Samples” triggers:
>
>    Get Samples List + (loads the sample list into the UI)
>    A synchronization with LIMS/ISPyB
>    Then filters UI ( SHOWS ONLY LIMS/ISPyB samples )

Whats done is already a good step in the right direction. What I would recommend is the following:

-   **Get Samples List:** loads the sample list from sample changer
-   **Synchronize with LIMS/ISPyB:**: Synchronize with LIMS and filter

We remove the logic that fetches the sample list on navigation, and possibly (but not necessarily in a first step) replace it with logic that fetches the sample when a session is selected.

Would that make sense ?

fabcor-maxiv:
> > user from proposal AAAA could mount samples from proposal BBBB and this is a serious security/privacy issue.
> 
> should be enforced at the backend level.

Yes, I believe so.

fabcor-maxiv:
In the video, @epanepucci and I noticed that the "Clear samples" button transforms into a "Get samples" button [^1], and we believe that this is very bad UI/UX design. One button should not transform into another button like that. Buttons should have one and only one feature. Buttons can be enabled and disabled depending on the context, and I guess this is probably what should happen here.

[^1]: or maybe "Clear samples" is hidden while "Get samples" is made visible, but the end result is the same

jbflo:
> * **Get Samples List:** loads the sample list from sample changer
> * **Synchronize with LIMS/ISPyB:**: Synchronize with LIMS and filter
> 
> We remove the logic that fetches the sample list on navigation, and possibly (but not necessarily in a first step) replace it with logic that fetches the sample when a session is selected.
> 
> Would that make sense ?

That make good sense yeah...

marcus-oscarsson:
@fabcor-maxiv: yes I think thats more or less what we concluded as well. @jbflo is having a look. What I suggest is that he finishes the current suggestion and then we can make another video and you guys can have a look again.

jbflo:
> In the video, @epanepucci and I noticed that the "Clear samples" button transforms into a "Get samples" button [1](#user-content-fn-1-0a1e70513f65ae3262d31a198df9c5fd), and we believe that this is very bad UI/UX design. One button should not transform into another button like that. Buttons should have one and only one feature. Buttons can be enabled and disabled depending on the context, and I guess this is probably what should happen here.
> 
> ## Footnotes
> 1. or maybe "Clear samples" is hidden while "Get samples" is made visible, but the end result is the same [↩](#user-content-fnref-1-0a1e70513f65ae3262d31a198df9c5fd)

AS you notice we’re not technically transforming the same button — rather, we’re toggling visibility between two separate buttons based on the state of the sample list. But as you pointed out, the visual effect is the same, and the behavior might still be misleading.

Before seeing your comment, and as we're revisiting this PR, I was already planning to remove the toggle effect  between "Clear samples" && “Get samples”  — Since we’ve agreed that "Get samples" is better bundled with "Sync LIMS"

Now To  visit your concern. though I wouldn’t categorize this as “very bad UI/UX design.” but  One possible improvement might be to have "Clear samples" disabled when the list is already empty. This would preserve clarity and give users full control without hiding functionality.

And the question is why showing a Clear samples Button when there is nothing to clear ?

marcus-oscarsson:
@fabcor-maxiv would it work for you if we did the following:

- Get Samples List: loads the sample list from sample changer
- Synchronize with LIMS/ISPyB: Synchronize with LIMS and filter

And we leave it up to the backend, (HardwareObject implementation) to determine which samples that is retrieved when the user pushes  `Get Samples List` (depending one user currently logged in etc)

fabcor-maxiv:
> @fabcor-maxiv would it work for you id we did the following:

Best to ask @epanepucci to know what our beamlines staff wants/needs.

jbflo:
As the main goal of this PR was to avoid having an empty sample list after login and to clarify the Sync LIMS behavior, we've decided to convert it to a draft for now until we can propose a better approach.

We're currently exploring a solution where:
- The sample list is automatically fetched when a session is selected
- We may need to define a default LIMS name (limsName) in the config
— or leave the current fallback to the first available option, such as "ISPyB"

ui/src/actions/login.js
```diff
@@ -68,6 +69,12 @@ export function logIn(proposal, password) {
 
     dispatch(showErrorPanel(false));
     await dispatch(getLoginInfo());
+    const loginInfo = await fetchLoginInfo();
+    await dispatch(setLoginInfo(loginInfo));
```
axelboc: This duplicates what `await dispatch(getLoginInfo());` already does

ui/src/actions/login.js
```diff
@@ -68,6 +69,12 @@ export function logIn(proposal, password) {
...
+    if (loginInfo.limsName && loginInfo.limsName.length > 0) {
+      // synchronize the sample list with LIMS, Use First LIMS available as default
+      dispatch(syncSampleListWithLims(loginInfo.limsName[0].name));
+    }
```
axelboc: It would be good to not couple LIMS synchronisation with the login code.

When the user logs in, the `Main` component gets mounted and a `useEffect` invokes `dispatch(getInitialState());`. Perhaps this dispatch would be better placed inside this `getInitialState` action creator?

```js
export function getInitialState() {
  return async (dispatch, getState) => {
    const initialStateSlices = await Promise.all([...]);

    dispatch(setInitialState(Object.assign({}, ...initialStateSlices)));
    // HERE => call `getState` to get the LIMS name and dispatch `syncSampleListWithLims`
    dispatch(applicationFetched(true));
  };
}
```

ui/src/containers/SampleListViewContainer.jsx
```diff
@@ -547,42 +536,51 @@ export default function SampleListViewContainer() {
   }
 
   function getSynchronizationDropDownList() {
-    if (loginData.limsName.length === 1) {
-      return (
+    return (
+      <Dropdown>
```
axelboc: A dropdown is nice, but it'd be good if most users didn't have to click twice (i.e. open drop-down and click on button). Bootstrap has a [split button](https://react-bootstrap.github.io/docs/components/dropdowns#split-button-dropdowns) — I'm thinking that the main button could be "Get Samples & Sync with [first-lims]".

axelboc (commented):
Nice, definitely better to sync after login than when navigating to the sample list :100:

ui/src/actions/login.js
jbflo: I'll apply the suggestion [bellow](https://github.com/mxcube/mxcubeweb/pull/1663#discussion_r2145175844) 
Thanks :relaxed:

ui/src/actions/login.js
jbflo: > It would be good to not couple LIMS synchronisation with the login code.

That's a good idea.

> When the user logs in, the `Main` component gets mounted and a `useEffect` invokes `dispatch(getInitialState());`. Perhaps this dispatch would be better placed inside this `getInitialState` action creator?
> 
The only hic is getInitialState is execute on every page refresh, and it is maybe better to have syncSampleListWithLims auto execute once just after user login or selectSession

marcus-oscarsson: Hm, I'm wondering if its not worth investigating if the synchronization could not be done when the session is selected on the backend. There is something in place for handling the sessions selection/update, the synchronization could be done there and perhaps then simply fetched, right ?

jbflo: > Hm, I'm wondering if its not worth investigating if the synchronization could not be done when the session is selected on the backend. There is something in place for handling the sessions selection/update, the synchronization could be done there and perhaps then simply fetched, right ?
>

Yes,  where could be a good place to fetch in the UI  then?

And also where / how to fetch can also where you trigger the synch 
that is why i am wondering if it ls not simpler in the UI .

we just need to find a good place to call this function
```
export function syncSampleListWithLims(lims) {
  return async (dispatch) => {
    await dispatch(syncSamples(lims)); // sync samples with LIMS
    dispatch(filterAction({ limsSamples: true })); // auto filter samples to show only LIMS samples
  };
}
```

We can maybe put it as a callback in the Login function ?
```
export function logIn(proposal, password, callback) {

     ........

   callback()

 }
```

epanepucci:
> @fabcor-maxiv would it work for you if we did the following:
> 
> * Get Samples List: loads the sample list from sample changer
> * Synchronize with LIMS/ISPyB: Synchronize with LIMS and filter
> 
> And we leave it up to the backend, (HardwareObject implementation) to determine which samples that is retrieved when the user pushes `Get Samples List` (depending one user currently logged in etc)

I think this a good solution which prevents the wrong samples showing up.

marcus-oscarsson:
Alright, thanks for the feedback @epanepucci. We had a quick chat about all this here and we came up with the following solution that we think will make things easier to use for both users and us.

On the server we have two functions that return:

- **Get Samples List**: list  of all samples from sample changer
- **Synchronize with LIMS/ISPyB**: Synchronize sample list with LIMS (only returning samples for current user/session)

We add an option called `use_get_all_samples` or similar  which can be used to:

- Hide the button **Get samples** so that the user only can use **Synchronize with LIMS**
- Make sure that  **Get Samples List** from the back-end only returns empty list

This eliminates the need for having the **Filter LIMS** option.

marcus-oscarsson:
Did I forget something ?, @jbflo, @axelboc, @ddesanctis

marcus-oscarsson:
I will close this PR and @jbflo will open a new one

axelboc:
- In the UI, we show two buttons: **Synchronize with LIMS** and **Get samples** (unless `use_get_all_samples` is set to `false`).
- If both buttons are visible, we display them in a drop-down, with  **Synchronize with LIMS** as the main button visible without opening the drop-down.
- Both buttons clear the sample list when clicked.
- The **Synchronize with LIMS** button no longer applies the LIMS filter since this filter is removed.
- When the user navigates to the samples list and it's empty, a LIMS synchronization is performed automatically by default. Since the sample list is cleared when switching sessions, this ensures that the users always sees an up-to-date list of samples the first time they open the page.

PR #1666, opened 2025-06-05
Rhfogh develop

rhfogh:
Fixed problem in GphlWorkflowParametersDialog.jsx - reference MTZ changed from required to optional

Removed regression from components.queue.Queue.get_entry - code can now deal with node ID == None for the root object

Updated and cleaned GPhL-specific config files

UPDATED 20250606. Discussion below should now be resolved.


mxcubeweb/core/components/queue.py
```diff
@@ -707,9 +707,14 @@ def get_entry(self, _id):
         Retrieves the model and the queue entry for the model node with id <id>
 
         :param int _id: Node id of node to retrieve
-        :returns: The tuple of model and entry
+        :returns: The tuple model, entry or the root node and QueueManger if _id is None
         :rtype: Tuple
         """
+        if _id is None:
```
marcus-oscarsson, line 712: I think this fix was causing issue previously and we need to solve it in some other way.  Returning `HWR.beamline.queue_model.get_model_root()` would be fine but `HWR.beamline.queue_manager` is a `QueueManager` and not a `QueueEntry`. So the returned object `QueueEntry` and `QueueManager` does not have the same API, thus the calling code will fail later on.

I think the the problem is already that `_id` is passed `None` maybe that can be handled in someway ?

rhfogh:
The commit that was reverted contained two different fixes and I think the comment referred only to the  other one, so maybe it is not this that was causing problems

I'll give some details, and we can see if we can get somewhere.

- The error happens when you are trying to attach a Sample to the queue root
- As part of that there isa  call to `self.get_entry(parent._node_id)`, where 'parent' is the ModelRoot 
- The ModelRoot has _node_id None - that is what _node_id is initialised to in TaskNode, and _node_id is only ever set for child nodes, which the root is not.
- The code in get_entry will always break when presented with a _node_id of None. so there should be some guard.
- The call to get_entry is trying to get the QueueEntry that matches the model root. AFAICS that can only be the QueueManager. If not , what should it be? What we want is the QueueEntryContainer to which we should attach the Sample, and that is the QueueManager. We could give the ModelRoot a _node_id (e.g. 0?) and give the appropriate QueueEntry(Container) the same _id, but that will still give us the same return values as my fix. 
- The QueueManager is not a QueueEntry, but it is a QueueEntryContainer. Maybe that is enough?

It may well be that we have got the return types wrong here. Maybe we should have get_entry_container function to be correct about the return types, and distinguish the contexts where we want a container from those where we want a proper entry.  Actually, the QueueManager is the only QueueEntryContainer that is not also a BaseQueueEntry My suspicion is that the code already works (with my fix) because we only ever get the QueueManager in contexts where we only need a QueueEntryContainer, not a QueueEntry. But this change would be more far-reaching and require some more changes to the code. Is it worth embarking on? Or what else should we do?

Alternatively we could simply check up front if we were trying to get the entry to the root model and not call get_entry if that was the case. That wiould save us  adding an extra function. Again, the effect would be the same is for my fix, but it would force us to go through all the calls to get_entry and make sure that we were not calling it on the RootModel, or that we were treating the result right if we did.  That would most obviously be a job for someone who understands mxcubeweb well, since it is not immediately obvious to a casual glance whether the IDs being passed to get_entry could be that of the RootModel, and if so what should happen.

marcus-oscarsson:
I double checked and the code reverted was indeed the one that is also in this PR. You can find the reverting commit here:

   https://github.com/mxcube/mxcubeweb/pull/1641/files#diff-881eac3320962966c742cc7ac2ce302272a9749ae301a359126914c5aec70940L713

The problem was that the adding for instance a characterization failed, not sure exactly why. I think you would be able to reproduce that issue with this code, if you simply add a characterization to a centered position.

You are right that QueueEntryContainer may be sufficient in virtually all cases but it seems like adding a characterization for some reason is breaking with this change. I would suggest to investigate if that breaking logic could be fixed so that it works with the changes you propose. Otherwise I guess we have to handle this case in the calling code `queue_model_child_added` ?

rhfogh:
The problem is now (20250606)  fixed for both characterisation and adding samples to the queue top.
Please re-examine

Trouble was that unenqueued nodes had _node_id==None the same as the RootModel. get_node_status simply  called get_entry from within a try: except: so that any error was interpreted as 'not yet enqueued' (slightly dodgy coding, actually it ought to be checked that the same is not done elsewhere)).  And of course if get_entry no longer triggered an error then get_node_status would no longer work as intended. The PR now checks explicitly for _node_id zero and includes some minor clean-up as well. It runs with both characterisation and workflow in testing.

mxcubeweb/core/components/queue.py
rhfogh: Should be fixed by changes 20250606.

marcus-oscarsson:
Well done, thanks alot for having a closer look

mxcubeweb/core/components/queue.py
```diff
@@ -1764,11 +1760,19 @@ def queue_model_child_added(self, parent, child):
...
         """
-        _, parent_entry = self.get_entry(parent._node_id)
-        child_model, _ = self.get_entry(child._node_id)
+        if parent is HWR.beamline.queue_model.get_model_root():
+            parent_entry = HWR.beamline.queue_manager,
+        else:
+            node_id = parent._node_id
+            if node_id is None:
+                raise ValueError(
+                    "Trying to add child {child} to unenqueued parent {parent}"
```
marcus-oscarsson, line 1772: Just a question, did you ever see this error getting raised. `queue_model_child_added` is a signal handler so the exception is raised by the code that calls that handler. Does that behave correctly ?

rhfogh: No, I did not see this error raised particularly - and I am not quite sure I understand what the problem is, actually. But what used to happen is that an error was raised one level down, from get_entry, and that error showed in the log trace. That could still happen, this error is just a more precise error for this particular case, as opposed to teh more generic error you get if you let the error arise from get_entry.

marcus-oscarsson: Ok, I see. I'm unsure how exceptions raised in signal handlers are treated but we can leave it as this for now. Its an improvement over what we had before.

PR #1668, opened 2025-06-06
Issue_1664 Session_DB_PATH configurable

Phantominumm:
Session database path is now configurable in server.yaml and not hardcoded anymore

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "mxcubeweb"
-version = "4.383.0"
+version = "4.384.0"
 license = "LGPL-3.0-or-later"
 description = "MXCuBE Web user interface"
 authors = [


marcus-oscarsson:
The linter seems to have spotted something, you can run `ruff check --fix` and it will make the change for you. (thats why the pipelines are failing)

Phantominumm:
problem was B108 hardcode tmp dir

https://bandit.readthedocs.io/en/latest/plugins/b108_hardcoded_tmp_directory.html

marcus-oscarsson:
We just need to have a look at the conflicts as the other PR was merged and this one still contained the HOST/PORT changes.

mxcubeweb/core/models/configmodels.py
```diff
@@ -19,7 +19,7 @@ class FlaskConfigModel(BaseModel):
     ALLOWED_CORS_ORIGINS: list[str] = Field([], description="")
     SECURITY_PASSWORD_SALT: str = Field("ASALT", description="")
     SECURITY_TRACKABLE: bool = Field(True, description="")
-    USER_DB_PATH: str = Field("/tmp/mxcube-user.db", description="")
+    USER_DB_PATH: str = Field("", description="")
```
fabcor-maxiv: Why the empty strings? I am really asking, I have not used Pydantic in a while, I do not remember how it works.

Isn't it enough like this?

```suggestion
    USER_DB_PATH: str
```

mxcubeweb/server.py
```diff
@@ -178,8 +178,9 @@ def run(cfg):
                 cfg.flask.CERT_PEM, cfg.flask.CERT_KEY
             )
         elif cfg.flask.CERT == "ADHOC":
+            cert_dir = os.path.dirname(cfg.flask.USER_DB_PATH)
```
fabcor-maxiv: Should we use `pathlib` instead of `os.path`, here?

mxcubeweb/core/models/configmodels.py
Phantominumm: we need to declare that it's a Field instance for Pydantic even with nothing ("").
If we let just **str** it's just a type hint with no value and it became a required field, 
Basically the first one has a default value and the second does not.
Anyway that was the first solution but it will be changed.

mxcubeweb/server.py
Phantominumm: you are right it's better, i will change it

mxcubeweb/core/models/configmodels.py
fabcor-maxiv: > we need to declare that it's a Field instance for Pydantic even with nothing ("").
> If we let just **str** it's just a type hint with no value and it became a required field,
> Basically the first one has a default value and the second does not.

Ah right, thanks for the refresher. Indeed, I think that we want it to be optional, not required, with the default value hard-coded to `/tmp/mxcube-user.db` then, right? The empty string does not make sense to me.

mxcubeweb/core/models/configmodels.py
```diff
@@ -19,7 +20,9 @@ class FlaskConfigModel(BaseModel):
     ALLOWED_CORS_ORIGINS: list[str] = Field([], description="")
     SECURITY_PASSWORD_SALT: str = Field("ASALT", description="")
     SECURITY_TRACKABLE: bool = Field(True, description="")
-    USER_DB_PATH: str = Field("/tmp/mxcube-user.db", description="")
+    USER_DB_PATH: str = Field(
+        str(Path.home() / ".mxcube" / "tmp" / "mxcube-user.db"), description=""
```
fabcor-maxiv: Alright... okay... now we are shifting gears! :D

Then at this point I would suggest using conventional locations like the ones specified in "[XDG Base Directory](https://wiki.archlinux.org/title/XDG_Base_Directory)". So I guess we want `XDG_DATA_HOME`. Concretely, I would recommend using the [`platformdirs`](https://pypi.org/project/platformdirs/) library and in particular the output of `platformdirs.user_data_dir("mxcube")` as base directory. I guess that would be a good default.

marcus-oscarsson: Hm, I think I would prefer `~./mxcube/tmp` or `~./mxcube/` over the ie. `~./config/`. `USER_DB_PATH` is also configured in `server.yaml` so it would be somewhat strange to have it configurable in two different ways (via `server.yaml` or `$XDG_CONFIG_HOME`)

fabcor-maxiv: > I would prefer `~./mxcube/tmp` or `~./mxcube/` over the ie. `~./config/`.

It would be `~/.config/mxcube/mxcube-user.db` which is the modern convention. It can be overridden anyway, so you could set this to `~/.mxcube/mxcube-user-db` at ESRF via `USER_DB_PATH` in `server.yaml`.


> `USER_DB_PATH` is also configured in `server.yaml` so it would be somewhat strange to have it configurable in two different ways (via `server.yaml` or `$XDG_CONFIG_HOME`)

If I am not mistaken, `$XDG_CONFIG_HOME` gives you a way to override the default of `~/.config`. So if on your system you set `XDG_CONFIG_HOME=~/my-config` then `platformdirs.user_data_dir("mxcube")` would give you `~/my-config/mxcube` instead of `~/.config/mxcube`.

marcus-oscarsson: Yes, I agree that it would be good in general. But I feel that it becomes a bit overkill to add an additional library to handle XDG directories.  What we want to do is to configure this path from `server.yaml`  and at the same time provide a safe default value. 

We can perhaps default the root path to `~/.config/mxcube/` if that supports the convention better

fabcor-maxiv: > I feel that it becomes a bit overkill to add an additional library to handle XDG directories.

I agree. I guess we only support Linux, and the value can be overridden anyway, so no need for anything complex.

fabcor-maxiv: Argh, my bad, I got my wires crossed, of course "user data" belongs in `~/.local/share/mxcube` not in `.config/mxcube`.

```suggestion
        str(Path.home() / ".local" / "share" / "mxcube" / "mxcube-user.db"), description=""
```

Do we need an empty description, though?

```suggestion
        str(Path.home() / ".local" / "share" / "mxcube" / "mxcube-user.db")
```

Phantominumm: Keeping description even if it's empty is better for consistency, not necessary but just follow good practice with Pydantic

test/conftest.py
```diff
@@ -77,8 +79,11 @@ def cleanup_subprocesses():
 
 @pytest.fixture
 def client():
+    db_path = Path.home() / ".mxcube" / "tmp"
+    test_db = db_path / "mxcube-test-user.db"
+
     with contextlib.suppress(FileNotFoundError):
-        os.remove("/tmp/mxcube-test-user.db")
+        test_db.unlink()
```
fabcor-maxiv: Here for automated tests, should we consider using a temporary file (with randomly generated name) with the help from [`tempfile`](https://docs.python.org/3/library/tempfile.html), probably with a `with` context?

Something along the lines of:

```python
with tempfile.TemporaryFile() as fp:
    ...
```

mxcubeweb/__init__.py
```diff
@@ -124,9 +128,25 @@ def build_server_and_config(test=False, argv=None):
         config_path = HWR.get_hardware_repository().find_in_repository("mxcube-web")
 
         cfg = Config(config_path)
-
+        db_path = (
+            Path(cfg.flask.USER_DB_PATH)
+            if cfg.flask.USER_DB_PATH
+            else Path(FlaskConfigModel.USER_DB_PATH.default)
+        )
+        db_path.parent.mkdir(parents=True, exist_ok=True, mode=0o700)
```
fabcor-maxiv: Why do we need to set specific permission bits here?
What is wrong with the defaults?
At the very least we should document if there is a specific reason...

```suggestion
        db_path.parent.mkdir(parents=True, exist_ok=True)
```

test/conftest.py
marcus-oscarsson: That would be quite nice, maybe we can try that :+1:

mxcubeweb/__init__.py
Phantominumm: It's just that database can contain sensitive information and I wanted something really secure in case with only owner rights but if defaults permissions are fine we can change it.

test/conftest.py
```diff
@@ -77,63 +78,65 @@ def cleanup_subprocesses():
 
 @pytest.fixture
 def client():
-    with contextlib.suppress(FileNotFoundError):
-        os.remove("/tmp/mxcube-test-user.db")
+    with tempfile.TemporaryDirectory(prefix="mxcube-test-") as temp_dir:
```
fabcor-maxiv: Why `TemporaryDirectory`? Wouldn't it work with `TemporaryFile`? Python would take care of everything, finding a location, finding a unique name, deleting the file at the end (i.e. no need to explicitly unlink).

mxcubeweb/__init__.py
marcus-oscarsson: I think you are on the right track @Phantominumm. 

There is cause for concern, the normal 644 rights would be too permissive for this sort of information. I think 600 would be enough but its possible that execute is needed for some reason.  

We would not in general want either log files or any session database to be readable by any other user than the one that runs the MXCuBE process. Omitting the `mode` will probably make the warning/error disappear from bandit or Sonar Qube but it will still be an issue.

I think it would be quite good if we could keep the database permissions restrictive to i.e 700 or 600. I can understand that having the log files being readable by another user might be practical but there is no real reason for why the database should have any other permissions than 600, right ?

test/conftest.py
Phantominumm: I used tempDirectory because i needed a Path, not just a file and for the consistency during tests.
 Actually we don't need to create a path or a file here because **__init__.py** handle it. The path/file is never given to **__init__** and the function **build_server_and_config** used during the test has a **test** args and if it's true it will create the path itself.

Phantominumm: Something like this could work for the first question also(using NamedTmpFile):
with tempfile.NamedTemporaryFile(prefix="mxcube-test-user", suffix=".db") as tp:
    db_path = Path(tp.name)

mxcubeweb/__init__.py
Phantominumm: Yes i used 700 to give all the rights but there would be no reason to execute from this directory, i think 600 is more than sufficient.

fabcor-maxiv: I see the concern of course, but again it feels to me like it is not the correct way of solving this (I might be wrong, I am not a security person).

In this case, I do not think we (MAX IV) care if other Linux user accounts have access to this file or not, so I guess we are fine with setting the permission bits to `600` (or `700`).

marcus-oscarsson: The other way would be provide a separate script for setting up the log files and the databases and then its up to each one how to handle the permissions. One just need to make sure they exist before launching the application. It would make the setup/install of the application more complicated.

I personally think that handling this within the application, **if the files does not already exist** is a quite nice thing. It makes everything much easier to setup. What we want to make sure is that that functionality is as secure as we can make it. It still leaves it up to each one to configure/setup the files separately by creating the files before the application is started.

marcus-oscarsson: One could imagine a separate switch at the command line that makes the behavior/creation more explicit

test/conftest.py
```diff
@@ -132,9 +127,6 @@ def client():
 
     client.get("/mxcube/api/v0.1/login/signout/")
 
-    with contextlib.suppress(FileNotFoundError):
-        os.remove("/tmp/mxcube-test-user.db")
-
```
walesch-yan: I think we should be careful with removing this. As you said the path that we use is coded in the `init` function of the server class and creating a tempfile in here might not make sense. However, this implies that with the current configuration the file created in `/tmp/mxcube-test-user.db` will not automatically be unlinked after execution. Hence if we remove this bit of code the file will continue to exist between test runs.

walesch-yan: So I assume, it would be best to clear the file before (as you do in `__init__.py`) and after the test client logged out, i.e. here

test/conftest.py
fabcor-maxiv: I guess this comment thread is obsolete now and can be marked as "resolved", right? I got a bit lost...

test/conftest.py
fabcor-maxiv: This comment thread is obsolete as well, right?

mxcubeweb/__init__.py
fabcor-maxiv: My take on what a good way to solve this would be: create the containing directory in advance with the correct permission bits so that only allowed user accounts can access the directory and the DB file that will be created in that directory. Straightforward, no need to add complex code in MXCuBE, every facility can set the permissions they want.

marcus-oscarsson: Ok, with the proposed logic you should be free to do create the directories with the permissions you like. The permissions are only applied on created folders.

However leaving the `mode` argument out would default the mode to 777 (its default mode of `Path.mkdir` so we should probably not do that.

As we see it, the current logic provides a reasonable, more secure and functional solution that does not require that any folders are created before hand.

marcus-oscarsson:
I resolved the conflicts then we can have another look together squash the commits and merge

fabcor-maxiv (approved):
Seems fine now

Issue #1387, opened 2024-09-04
Canonical flask application or at least a --port CLI argument


woutdenolf:
The canonical way to start a flask application would be this in development

```bash
flask --app mxcubeweb.... run [--host 127.0.0.1] [--port 5000] [--debug]
```

or in production with a proper WSGI server

```bash
gunicorn mxcubeweb.... [--host 0.0.0.0] [--port 8000] [--log-level DEBUG]
```

Would it be possible to support this?

If not, the fast solution for me at the moment would be to add a --port CLI argument.

Side note: the `main` is located in `mxcubeweb/__init__.py` instead of the typical `mxcubeweb/__main__.py` (which would be run when you do `python -m mxcubeweb`). Any reason for that?

fabcor-maxiv:
I would welcome a clean up of this. I do not know why it is the way it is now, definitely looked unusual to me when I first started looking at this code. So if possible instead of adding a new CLI option, I would suggest working on making mxcubeweb behave like all other normal Flask apps.

I assume a more canonical structure would also make automated tests easier.

marcus-oscarsson:
Just for information, be aware that the server started is not the development server but a "production ready" server (https://flask-socketio.readthedocs.io/en/latest/deployment.html#embedded-server)

It would otherwise indeed be very nice if we can run the server in the standard way. I know that we have been discussing it in the past but I can't remember exactly why it was not working or decided to keep it like its currently done.

marcus-oscarsson:
I had a quick look again and do remember that we did have some issues with passing arguments. Maybe you have some idea on how that should be done as you seem to have some experience with this @woutdenolf and @fabcor-maxiv

woutdenolf:
> some issues with passing arguments

You mean you want to have additional CLI arguments in addition to when the flask or gunicorn CLI provides?

That's indeed a little tricky. I've done this for `FastAPI` but never for `Flask`. FastAPI uses `click` and you can either extend it (which is a little hacky) or use `pydantic_settings` but then the arguments need to come from environment variables. Never tried for `Flask`. It seems to use `click` as well https://flask.palletsprojects.com/en/3.0.x/cli/.

fabcor-maxiv:
I would suggest environment variables.

```none
MXCUBE_WEB_UI=/path/to/ui/build MXCUBE_HWO_REPO=/path/to/hwos gunicorn mxcubeweb ...
```

[...]

PR #1669, opened 2025-06-06
Ml fix issue 1665 binding interface

Phantominumm:
Add host and port to configuration options in server.yaml (demo/mxcube-web). Changing from 0.0.0.0 to localhost(127.0.0.1) for the binding interface problem B104. There would be no problem with the reverse proxy but in need change in the server.yaml file.

### DIFF
--- a/demo/mxcube-web/server.yaml
+++ b/demo/mxcube-web/server.yaml
@@ -9,6 +9,9 @@ server:
   DEBUG: false
 
   ALLOWED_CORS_ORIGINS: []
+  USER_DB_PATH: /tmp/mxcube_users.db
+  HOST: "0.0.0.0"
+  PORT: 8081
 
 sso:
   USE_SSO: false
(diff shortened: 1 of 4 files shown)


demo/mxcube-web/server.yaml
```diff
@@ -9,6 +9,9 @@ server:
   DEBUG: false
 
   ALLOWED_CORS_ORIGINS: []
+  USER_DB_PATH: /tmp/mxcube_users.db
```
marcus-oscarsson: I think this (the database path related things) should go into the other PR, right ?

fabcor-maxiv:
> problem B104

You mean the [Ruff rule `S104`](https://docs.astral.sh/ruff/rules/hardcoded-bind-all-interfaces/), right?

Phantominumm:
> > problem B104
> 
> You mean the [Ruff rule `S104`](https://docs.astral.sh/ruff/rules/hardcoded-bind-all-interfaces/), right?

I used this doc from bandit and the problem was B104, they are the same issues

https://bandit.readthedocs.io/en/latest/plugins/b104_hardcoded_bind_all_interfaces.html

marcus-oscarsson:
Looks good thanks !

marcus-oscarsson:
closes #1387

PR #1670, opened 2025-06-09
Fix custom tasks duplication issue

dominikatrojanowska:
Parameters of custom tasks differs in below situations:
1. If the task is added by right-click on data collection view the `experiment_name` is equal to empty string.

image

2. In case of duplication `experiment_name` is equal to `None`. Next, calling [os.path.join]( https://github.com/mxcube/mxcubeweb/blob/a673fa/mxcubeweb/core/components/queue.py#L1585) with a string as first and `None` as second argument, causes the method to crush.

image

With this change `""` is passed as a second argument to `os.path.join` in 2nd case.

### DIFF
--- a/mxcubeweb/core/components/queue.py
+++ b/mxcubeweb/core/components/queue.py
@@ -1582,7 +1582,9 @@ class Queue(ComponentBase):
             )
 
         full_path, process_path = HWR.beamline.session.get_full_path(
-            os.path.join(params.get("subdir", ""), params.get("experiment_name", "")),
+            os.path.join(
+                params.get("subdir", ""), params.get("experiment_name", "") or ""
+            ),
             task_name,
         )
         acq.path_template.directory = full_path


elmjag:
This fixes the bug where duplicating custom tasks did not work for us.

However, you write that the original task's `experiment_name` was an empty string, but the duplicated task's is `None`. Do you know where is that code that does the duplication is? It feels like that code is not doing the right thing. It feels weird that `experiment_name` is not carried over. 

I think we should take a look.

dominikatrojanowska:
> This fixes the bug where duplicating custom tasks did not work for us.
> 
> However, you write that the original task's `experiment_name` was an empty string, but the duplicated task's is `None`. Do you know where is that code that does the duplication is? It feels like that code is not doing the right thing. It feels weird that `experiment_name` is not carried over.
> 
> I think we should take a look.

Another option was to change [`experiment_name`](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/model/common.py#L20) to be a non-optional parameter with default value equal to `""`, but I'm not sure if that solution is elegant enough

marcus-oscarsson:
I agree with @elmjag, I think that the value for `experiment_name` should be copied correctly the root path would otherwise not be the same. We should probably look into why that is not happening

marcus-oscarsson:
Did you find out why `experiment_name` was not carried over ?

dominikatrojanowska:
Basically what happens:
1. When adding a custom task
> 1. If the task is added by right-click on data collection view the `experiment_name` is equal to empty string.

frontend sends a [request](https://github.com/mxcube/mxcubeweb/blob/8c7745/mxcubeweb/routes/queue.py#L168), that in case of Max IV SSX task is a list of objects:
```
[{'cell_no': 0, 'checked': True, 'code': '', 'defaultPrefix': 'sth-sample', 'defaultSubDir': 'sth/sth-sample/', 'loadable': True, 'location': 'Manual', 'proteinAcronym': 'sth', 'puck_no': 1, 'queueID': 1, 'sampleID': '1', 'sampleName': 'sample', 'state': 0, 'tasks': [{'type': 'ssx_injector', 'label': 'SSX Injector Collection', 'state': 0, 'sampleID': '1', 'sampleQueueID': 1, 'parameters': {'cellA': 0, 'cellAlpha': 0, 'cellB': 0, 'cellBeta': 0, 'cellC': 0, 'cellGamma': 0, 'cell_counting': None, 'cell_spacing': None, 'comments': '', 'create_point': False, 'detector_binning_mode': None, 'detector_distance': 0, 'detector_roi_mode': 0, 'energy': 5.2776, 'exp_time': 0.01, 'experiment_type': '', 'first_image': 1, 'helical': False, 'in_interleave': None, 'in_queue': False, 'induce_burn': False, 'inverse_beam': False, 'kappa': None, 'kappa_phi': None, 'mesh': False, 'mesh_center': None, 'mesh_range': 0, 'mesh_snapshot': None, 'num_images': 1000, 'num_images_per_trigger': 0, 'num_lines': 1, 'num_passes': 1, 'num_triggers': 0, 'osc_range': 0.1, 'osc_start': 360, 'osc_total_range': 0, 'overlap': 0, 'prefixTemplate': None, 'resolution': 6.566, 'shape': '2DP1', 'shutterless': True, 'skip_existing_images': False, 'subDirTemplate': None, 'sub_wedge_size': 10, 'take_dark_current': 1, 'take_snapshots': 4, 'take_video': False, 'transmission': 10, 'type': 'ssx_injector', 'prefix': 'sth-sample', 'name': None, 'subdir': 'sth/sth-sample/', 'cell_count': None, 'numRows': 0, 'numCols': 0, 'beam_size': 15, 'label': 'SSX Injector Collection', 'space_group': 'C2'}, 'checked': True}], 'type': 'Sample'}]
```
**no experiment_name defined**.

2. At some point [this line](https://github.com/mxcube/mxcubeweb/blob/8c7745/mxcubeweb/core/components/queue.py#L1585) is reached, and since there is no experiment_name, the `""` goes to the `os.path.join`.

3. Later on, sill processing the request, [_handle_task_node](https://github.com/mxcube/mxcubeweb/blob/8c7745/mxcubeweb/core/components/queue.py#L286C9-L286C26) method is called and it creates the response, that from [this point](https://github.com/mxcube/mxcubeweb/blob/8c7745/mxcubeweb/core/components/queue.py#L290C30-L290C45) already contains the `experiment_name`. And since it was not provided and is [optional by definition](https://github.com/mxcube/mxcubecore/blob/47eb75/mxcubecore/model/common.py#L20), it is set to `None`.
This is [returned variable assigned to queue](https://github.com/mxcube/mxcubeweb/blob/8c7745/mxcubeweb/routes/queue.py#L170C9-L170C15):
```
{'sample_order': ['1'], '1': {'sampleID': '1', 'queueID': 1, 'code': '', 'location': 'Manual', 'sampleName': 'sample', 'proteinAcronym': 'sth', 'defaultPrefix': 'sth-sample', 'defaultSubDir': 'sth/sth-sample/', 'type': 'Sample', 'checked': True, 'state': 0, 'tasks': [{'label': 'SSX Injector Collection', 'type': 'ssx_injector', 'parameters': {'num_images': 1000, 'osc_start': 360.0, 'osc_range': 0.1, 'energy': 5.2776, 'transmission': 10.0, 'resolution': 6.566, 'first_image': 1, 'kappa': None, 'kappa_phi': None, 'beam_size': 15.0, 'shutterless': True, 'selection': [], 'shape': '2DP1', 'create_point': False, 'exp_time': 0.01, 'space_group': <SpaceGroup.SG_4: 'C2'>, 'cellA': 0.0, 'cellB': 0.0, 'cellC': 0.0, 'cellAlpha': 0.0, 'cellBeta': 0.0, 'cellGamma': 0.0, 'prefix': 'sth-sample', 'subdir': 'sth/sth-sample/', 'experiment_name': None, 'skip_existing_images': False, 'take_snapshots': 4, 'type': 'ssx_injector', 'label': 'SSX Injector Collection', 'take_dark_current': 1, 'inverse_beam': False, 'num_passes': 1, 'overlap': 0.0, 'path': '/data/visitors/micromax/20230002/20250612/raw/sth/sth-sample/', 'fileName': 'sth-sample_0_%004d.h5', 'fullPath': '/data/visitors/micromax/20230002/20250612/raw/sth/sth-sample/sth-sample_0_%004d.h5'}, 'sampleID': '1', 'sampleQueueID': 1, 'taskIndex': 0, 'queueID': 3, 'checked': True, 'state': 0}]}}
```
4. So the object containing `experiment_name = None/null" is stored in global state. 

5. After duplicating the original object, the request sent to the backend **already has the `experiment_name = None`**. New [request](https://github.com/mxcube/mxcubeweb/blob/8c7745/mxcubeweb/routes/queue.py#L168C11-L168C12):
```
[{'cell_no': 0, 'checked': True, 'code': '', 'defaultPrefix': 'sth-sample', 'defaultSubDir': 'sth/sth-sample/', 'loadable': True, 'location': 'Manual', 'proteinAcronym': 'sth', 'puck_no': 1, 'queueID': 1, 'sampleID': '1', 'sampleName': 'sample', 'state': 0, 'tasks': [{'type': 'ssx_injector', 'label': 'SSX Injector Collection', 'state': 0, 'sampleID': '1', 'sampleQueueID': 1, 'parameters': {'type': 'ssx_injector', 'label': 'SSX Injector Collection', 'beam_size': 15, 'cellA': 0, 'cellAlpha': 0, 'cellB': 0, 'cellBeta': 0, 'cellC': 0, 'cellGamma': 0, 'create_point': False, 'energy': 5.2776, 'exp_time': 0.01, 'experiment_name': None, 'fileName': 'sth-sample_0_%004d.h5', 'first_image': 1, 'fullPath': '/data/visitors/micromax/20230002/20250612/raw/sth/sth-sample/sth-sample_0_%004d.h5', 'inverse_beam': False, 'kappa': None, 'kappa_phi': None, 'num_images': 1000, 'num_passes': 1, 'osc_range': 0.1, 'osc_start': 360, 'overlap': 0, 'path': '/data/visitors/micromax/20230002/20250612/raw/sth/sth-sample/', 'prefix': 'sth-sample', 'resolution': 6.566, 'selection': [], 'shape': '2DP1', 'shutterless': True, 'skip_existing_images': False, 'space_group': 'C2', 'subdir': 'sth/sth-sample/', 'take_dark_current': 1, 'take_snapshots': 4, 'transmission': 10}, 'checked': True}], 'type': 'Sample'}]
``` 

6. Thus when reaching again [this point](https://github.com/mxcube/mxcubeweb/blob/8c7745/mxcubeweb/core/components/queue.py#L1585) `None` goes to the `os.path.join`. And here the program crushes. 

So today we discussed that with @elmjag. There is no point to change [model definition](https://github.com/mxcube/mxcubecore/blob/47eb75fd3135e5129bd06b80395dcf622d85350b/mxcubecore/model/common.py#L17), remove `experiment_name` directly after appearing  in [`parameters` dictionary](https://github.com/mxcube/mxcubeweb/blob/8c7745825b83fad5a3316a762c2e9af5610187cc/mxcubeweb/core/components/queue.py#L287) in case set to `None` or rewrite [Queue class](https://github.com/mxcube/mxcubeweb/blob/8c7745825b83fad5a3316a762c2e9af5610187cc/mxcubeweb/core/components/queue.py#L42) to use some new `_handle_ssx_task` method instead of [_handle_task_node](https://github.com/mxcube/mxcubeweb/blob/8c7745825b83fad5a3316a762c2e9af5610187cc/mxcubeweb/core/components/queue.py#L286) to fit our needs.

elmjag:
To summarize Dominika's message.

The reason why `experiment_name` is sometimes empty string and sometimes `None`, is due the way back-end and front-end interact.

It's a bit inconsistent. For new queue entry, front-end does not include `experiment_name` field in it's JSON. For duplicated queue entries, it includes it, but sets it to `None`.

Arguably, this is a bug. Perhaps we should fix it, so requests are consistent.

However, I feel that's potentionally big and risky task. I think that Dominika's patch is good fix for now. It fixes a serious bug.

Let's merge this now, and perhaps refactor the queue code in the future, to be more consistent.

mxcubeweb/core/components/queue.py
```diff
@@ -1582,7 +1582,9 @@ def add_queue_entry(self, node_id, task, task_name):
             )
 
         full_path, process_path = HWR.beamline.session.get_full_path(
-            os.path.join(params.get("subdir", ""), params.get("experiment_name", "")),
+            os.path.join(
+                params.get("subdir", ""), params.get("experiment_name", "") or ""
```
elmjag, line 1586: add a comment here, that explains that `experiment_name` can either be omitted or be `None`

marcus-oscarsson:
But will it result in the a good duplicate, i.e are the root paths the same ?

elmjag:
> But will it result in the a good duplicate, i.e are the root paths the same ?

Yes. 

Both the original and duplicate will use `""` for the `experiment_name` part of the root paths. I.e. the experiment name will not be included into the paths.

marcus-oscarsson:
Ok, then its fine

marcus-oscarsson:
Could we then re-base and merge this one ?

mxcubeweb/core/components/queue.py
dominikatrojanowska: will create new PR for that

dominikatrojanowska: here it is: https://github.com/mxcube/mxcubeweb/pull/1684

PR #1671, opened 2025-06-10
Remove unused route

Phantominumm:
Remove unused function get_result from routes and the 3 sub-function used in it 'run_get_result_script', 'result_file_test' and 'apply_template'

### DIFF
--- a/mxcubeweb/routes/lims.py
+++ b/mxcubeweb/routes/lims.py
@@ -1,24 +1,14 @@
 import logging
-from os.path import (
-    isfile,
-    join,
-)
-from subprocess import check_output
 
 from flask import (
     Blueprint,
     Response,
     jsonify,
-    render_template,
     request,
 )
-from mxcubecore.model import queue_model_objects as qmo
-
-from . import signals
 
 
-# Disabling C901 function is too complex (19)
-def init_route(app, server, url_prefix):  # noqa: C901
+def init_route(app, server, url_prefix):
     bp = Blueprint("lims", __name__, url_prefix=url_prefix)
 
     @bp.route("/synch_samples", methods=["POST"])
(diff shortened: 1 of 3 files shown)


fabcor-maxiv:
This does not remove a function. This removes a route. How was it established that this route is not used?

marcus-oscarsson:
It can be further cleaned up on the UI side, but this functionality is no longer used. It was used in the past to display a mashup of results from ISPyB in MXCuBE. We used to have a `templates` folder with the templates, but they have been removed quite some time ago. What we did not do was to remove these functions.

We can also remove:

https://github.com/mxcube/mxcubeweb/blob/2f6dd074c92b968656b8263eb1093a2e238a25fb/ui/src/components/Lims/LimsResultSummary.jsx#L5

And related code, we can have a look at that together @Phantominumm or if you like to give it a go, perhaps with @axelboc

axelboc:
Clean-ups FTW <3

So up for removal in the front-end are, if I'm not mistaken: `LimsResultSummary.jsx`, `taskHasLimsData` and `fetchLimsResults`.

marcus-oscarsson:
Sounds great, @axelboc thats it. You have a good map of the project in your head :). @Phantominumm don't hesitate to give it a try when you get the time, you can always ask me or @axelboc if you have some questions :)

marcus-oscarsson:
So we will, together with @Phantominumm, try to cleanup all related parts of the result view. Ill put the PR as draft for the time being.

mxcubeweb/routes/lims.py
```diff
@@ -1,20 +1,11 @@
...
     jsonify,
-    render_template,
     request,
 )
-from mxcubecore.model import queue_model_objects as qmo
-
-from . import signals
 
 
 # Disabling C901 function is too complex (19)
```
walesch-yan: Seeing this comment for C901 I guess we can remove this as well. 
Also, since you removed a large part of this file, maybe you can take a look at the ruff exceptions for this file in `ruff.toml`. I assume most can be removed there as well:).

walesch-yan (approved):
Great seeing more cleanup being done! Thanks @Phantominumm

fabcor-maxiv:
It looks to me like it might be fine now.

I think I still have one expectation left, it is that the this gets squashed into a single commit that says that the "route" is removed (not the "function").

axelboc (approved):

PR #1672, opened 2025-06-10
B110 try_except_pass

Phantominumm:
Add logging and warning messages for try_except_pass problem

https://bandit.readthedocs.io/en/latest/plugins/b110_try_except_pass.html

### DIFF
--- a/mxcubeweb/core/adapter/wavelength_adapter.py
+++ b/mxcubeweb/core/adapter/wavelength_adapter.py
@@ -9,6 +9,7 @@ from mxcubeweb.core.util.networkutils import RateLimited
 
 logger = logging.getLogger("MX3.HWR")
 
+
 class WavelengthAdapter(ActuatorAdapterBase):
     """
     Adapter for wavelength Hardware Object, a web socket is used communicate
--- a/mxcubeweb/core/adapter/wavelength_adapter.py
+++ b/mxcubeweb/core/adapter/wavelength_adapter.py
@@ -7,7 +7,7 @@ from mxcubeweb.core.models.adaptermodels import (
 )
 from mxcubeweb.core.util.networkutils import RateLimited
 
-logger = logging.getLogger("MX3.HWR")
+hwr_logger = logging.getLogger("MX3.HWR")
 
 
 class WavelengthAdapter(ActuatorAdapterBase):
@@ -29,7 +29,7 @@ class WavelengthAdapter(ActuatorAdapterBase):
             ho.connect("stateChanged", self.state_change)
         except Exception:
             msg = f"Could not connect signal: {ho.name}"
-            logger.exception(msg)
+            hwr_logger.exception(msg)
 
     @RateLimited(6)
     def _value_change(self, pos, wl, *args, **kwargs):
@@ -52,7 +52,7 @@ class WavelengthAdapter(ActuatorAdapterBase):
             return self.get_value()
         except Exception:
             msg = "Could not set value the wavelength: {ho.name}"
-            logger.exception(msg)
+            hwr_logger.exception(msg)
 
     def get_value(self) -> FloatValueModel:
         """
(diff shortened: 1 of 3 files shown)


fabcor-maxiv:
Ah, I see, this is Bandit numbering that is used instead of the Ruff numbering, right?

Maybe it is better to use Ruff numbering (`S110`) as this is what we use currently: <https://docs.astral.sh/ruff/rules/try-except-pass/>.

mxcubeweb/core/adapter/wavelength_adapter.py
```diff
@@ -46,7 +49,8 @@ def set_value(self, value: HOActuatorValueChangeModel):
             self._ho.set_wavelength(float(value.value))
             return self.get_value()
         except Exception:
-            raise
+            msg = "Could not set value the wavelength: {ho.name}"
+            logging.getLogger("MX3.HWR").exception(msg)
```
fabcor-maxiv: I recommend setting a global variable:

```python
log = logging.getLogger("MX3.HWR")
```

And then use like this

```suggestion
            log.exception(msg)
```

This would avoid repetitions. But I also recommend to do it even when there is no repetition. This is a very common practice. Only MXCuBE, that I know of, does things like they are done here now.

Phantominumm:
> Ah, I see, this is Bandit numbering that is used instead of the Ruff numbering, right?
> 
> Maybe it is better to use Ruff numbering (`S110`) as this is what we use currently: https://docs.astral.sh/ruff/rules/try-except-pass/.

ok i will use both of them now, indeed it's better to use ruff for everyone's understanding

fabcor-maxiv:
In general, I agree that it is a very good idea to also check with Bandit, as Ruff only implements a (small?) subset of Bandit. Bandit has a larger coverage.

But we have to at least be explicit. So in commit messages we should write things like "Bandit rule B110" and/or "Ruff rule S110".

rhfogh:
@fabcor-maxiv 
`log = logging.getLogger("MX3.HWR")`
sounds fine to me. But what are you recommending for e.g. mxcubecore where we have both 
`logging.getLogger("HWR")`, `logging.getLogger("user_level_log")` and maybe others? It becomes less obvious when you have multiple log streams.

fabcor-maxiv:
> what are you recommending for e.g. mxcubecore where we have both `logging.getLogger("HWR")`, `logging.getLogger("user_level_log")` and maybe others?

```python
hwr_log = logging.getLogger("HWR")
user_log = logging.getLogger("user_level_log")
```

rhfogh:
@fabcor-maxiv 
Gets a bit wordy with multiple logs, but fine with me.

fabcor-maxiv (approved):
Looks good to me.

I am very happy this issue is being addressed. It is incredibly annoying to have an exception being raised but absolutely no trace of it at all in the logs or anywhere.

Honestly I am surprised there are not more occurrences of this in our code base, I thought there would be way more.

rhfogh:
@fabcor-maxiv A lot of errors are swallowed up but with something more complex than `pass` in the Except branch. For instance any error during configuration loading just prints out an 'error while' message and continues. Yes, it is frustrating.

marcus-oscarsson:
Thanks, very nice ! Well done !

mxcubeweb/core/adapter/wavelength_adapter.py
marcus-oscarsson: For another future PR, what do you guys think of adding to the base class or application instead, so that one can do either `self.app.mxhwr.log` or `self.mxhwr.log` ?.

fabcor-maxiv: > For another future PR, what do you guys think of adding to the base class or application instead, so that one can do either `self.app.mxhwr.log` or `self.mxhwr.log` ?.

My first impression is that it seems like it would be fine.

marcus-oscarsson: Its perhaps more convenient compared to adding `log = logging.getLogger("MX3.HWR")` in each file

fabcor-maxiv: Yes, seems like it would be more convenient.

_[But for me I usually put those logger lines at the top of each of my modules anyway (boilerplate code), so I am used to see them there.]_

PR #1673, opened 2025-06-10
Fix violations of Ruff rule S311

Phantominumm:
B311 and S311, using secrets instead of random() to generate number because it's easily predictable (crypto graphically weak)

https://docs.astral.sh/ruff/rules/suspicious-non-cryptographic-random-usage/

https://semgrep.dev/r?q=gitlab.bandit.B311

### DIFF
--- a/mxcubeweb/routes/mockups.py
+++ b/mxcubeweb/routes/mockups.py
@@ -92,7 +92,6 @@ def init_route(app, server, url_prefix):
         cm = {}
 
         if shape:
-
             for i in range(1, shape.num_rows * shape.num_cols + 1):
                 hm[i] = [
                     i,
@@ -100,8 +99,8 @@ def init_route(app, server, url_prefix):
                         int(secrets.randbelow(256)),
                         int(secrets.randbelow(256)),
                         int(secrets.randbelow(256)),
-                        0, # int(random()) was used in the original code
-                        #but always returns 0
+                        0,  # int(random()) was used in the original code
+                        # but always returns 0
                     ],
                 ]
 
@@ -112,7 +111,7 @@ def init_route(app, server, url_prefix):
                         int(secrets.randbelow(256)),
                         int(secrets.randbelow(256)),
                         int(secrets.randbelow(256)),
-                        0, # Same as above
+                        0,  # Same as above
                     ],
                 ]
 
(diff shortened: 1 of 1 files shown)


mxcubeweb/routes/mockups.py
```diff
@@ -91,27 +92,26 @@ def shape_mock_result(sid):
...
                     [
-                        int(random() * 255),
-                        int(random() * 255),
-                        int(random() * 255),
-                        int(random()),
+                        int(secrets.randbelow(256)),
+                        int(secrets.randbelow(256)),
+                        int(secrets.randbelow(256)),
+                        0,  # int(random()) was used in the original code
+                        # but always returns 0
```
fabcor-maxiv: I do not understand this...

On one hand: there is no need to comment in the code what the code used to be doing, such explanation belongs in the commit message

On the other hand: do we need a zero there?

marcus-oscarsson: Is true that the comment should be in a commit message. It was me who asked @Phantominumm to add a comment about it as the current logic was a bit odd. 

I was hoping that maybe one of you guys could answer because we don't use the RGB values for the mesh results. Could it be the alpha channel, but we though it was odd that it was always zero ?

As this is for some sort of mockup we did not want to change the logic, so we kept it as 0 for the time being.

marcus-oscarsson:
Thanks

mxcubeweb/routes/mockups.py
fabcor-maxiv: > I was hoping that maybe one of you guys could answer because we don't use the RGB values for the mesh results. Could it be the alpha channel, but we though it was odd that it was always zero ?

I'll try to get some feedback on this

fabcor-maxiv: No feedback from the rest of the team (yet).

In principle, my impression is that having zeroes here is fair, since as said already the previous code would always return zeroes.

On the other hand: is this code still needed? I have not found a way to trigger this code in a "normal" way. I can of course call the HTTP URL explicitly, but even that fails.

marcus-oscarsson: I think it was used by some test script that was triggered manually, or it was my interpretation at-least. I think we can remove the entire mockup route (actually I think it would be quite important that we did)

marcus-oscarsson: If you agree, I suggest that we close this PR and we open a new one to remove the route ?

fabcor-maxiv: I am okay with removing this route.

PR #1674, opened 2025-06-11
`FluxAdapter` no longer needed handled as a `read_only` actuator

marcus-oscarsson:
`FluxAdapter` no longer needed handled as a `read_only` actuator


fabcor-maxiv (approved):
Seems fine

PR #1675, opened 2025-06-11
Handling stop/abort actions and movements with adapters

marcus-oscarsson:
Stopping or aborting a movement/action was previously handled via `beamline.beamline_abort_action`, which delegated the stop action to the right hardware object. This was a big source of confusion as we have the adapter concepts that does the same for most other functions. 

_Right @walesch-yan, (wink wink ;), Yan wont answer because he is on vacation in Japan :ninja: but he might appreciate this PR :)_

Abort/stop is with this PR handle via the `stop` method of the corresponding adapter object.

This PR introduces a `BeamlineActionsAdapter` that will also implement the rest of the functions related to beamline actions in `beamline.py`;  `beamline_get_actions` and `beamline_rut_actions`". This PR only deals with stop the rest is handled in a separate PR.

### DIFF
--- a/mxcubeweb/core/adapter/beamline_action_adapter.py
+++ b/mxcubeweb/core/adapter/beamline_action_adapter.py
@@ -1,21 +1,37 @@
 import logging
 from enum import Enum
+from typing import ClassVar
 
-from mxcubeweb.core.adapter.adapter_base import ActuatorAdapterBase
+from mxcubecore.BaseHardwareObjects import HardwareObject
+from mxcubecore.HardwareObjects.BeamlineActions import BeamlineActions
+
+from mxcubeweb.core.adapter.adapter_base import AdapterBase
 from mxcubeweb.core.models.adaptermodels import (
     HOActuatorValueChangeModel,
     NStateModel,
-    StrValueModel,
+)
+from mxcubeweb.core.models.configmodels import AdapterResourceHandlerConfigModel
+
+resource_handler_config = AdapterResourceHandlerConfigModel(
+    commands=["stop"], attributes=["data"]
 )
 
 
-class BeamlineActionAdapter(ActuatorAdapterBase):
-    def __init__(self, ho, *args, **kwargs):
+class BeamlineActionAdapter(AdapterBase):
+    SUPPORTED_TYPES: ClassVar[list[object]] = [BeamlineActions]
+
+    def __init__(
+        self,
+        ho: HardwareObject,
+        role: str,
+        app,
+        resource_handler_config: AdapterResourceHandlerConfigModel = resource_handler_config,  # noqa: E501
+    ):
         """
         Args:
             (object): Hardware object.
         """
-        super().__init__(ho, *args, **kwargs)
+        super().__init__(ho, role, app, resource_handler_config)
         self._value_change_model = HOActuatorValueChangeModel
 
         ho.connect("valueChanged", self._value_change)
@@ -34,12 +50,6 @@ class BeamlineActionAdapter(ActuatorAdapterBase):
             and attribute.startswith("_") is False
         ]
 
-    def set_value(self, value: HOActuatorValueChangeModel):
-        self._ho.set_value(self._ho.VALUES[value.value])
-
-    def get_value(self) -> StrValueModel:
-        return StrValueModel(value=self._ho.get_value().name)
-
     def msg(self):
         try:
             msg = self._ho.get_value().name
@@ -50,6 +60,13 @@ class BeamlineActionAdapter(ActuatorAdapterBase):
             )
 
         return msg
+
+    def stop(self):
+        """
+        Stop the execution.
+        """
+        for cmd in self._ho.get_commands():
+            self._ho.abort_command(cmd.name())
 
     def data(self) -> NStateModel:
         return NStateModel(**self._dict_repr())
--- a/ui/src/components/BeamlineActions/BeamlineActionForm.jsx
+++ b/ui/src/components/BeamlineActions/BeamlineActionForm.jsx
@@ -43,7 +43,7 @@ export default function BeamlineActionForm(props) {
           <Button
             variant="danger"
             onClick={() => {
-              dispatch(stopBeamlineAction(actionId));
+              dispatch(stopBeamlineAction('beamline_actions'));
             }}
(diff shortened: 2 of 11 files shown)


mxcubeweb/core/adapter/beamline_action_adapter.py
```diff
@@ -1,21 +1,39 @@
...
+    SUPPORTED_TYPES: ClassVar[list[object]] = [BeamlineActions]
 
-class BeamlineActionAdapter(ActuatorAdapterBase):
-    def __init__(self, ho, *args, **kwargs):
+    def __init__(
+        self,
+        ho: HardwareObject,
+        role: str,
+        app,
+        default_resource_handler_config: AdapterResourceHandlerConfigModel = resource_handler_config,  # noqa: E501
```
marcus-oscarsson: So I'm not a fan of this, but I did not find any way of splitting this line

mxcubeweb/core/adapter/beamline_action_adapter.py
```diff
@@ -1,21 +1,39 @@
...
+class BeamlineActionAdapter(AdapterBase):
+    SUPPORTED_TYPES: ClassVar[list[object]] = [BeamlineActions]
 
-class BeamlineActionAdapter(ActuatorAdapterBase):
-    def __init__(self, ho, *args, **kwargs):
+    def __init__(
+        self,
+        ho: HardwareObject,
+        role: str,
+        app,
```
marcus-oscarsson, line 27: I can't import `MXCUBEApp` due to circular import, should we use Type["MXCUBEApp"] in those cases, or do you think we should deal with such cases ?

ui/src/components/BeamlineActions/BeamlineActionForm.jsx
```diff
@@ -43,7 +43,7 @@ export default function BeamlineActionForm(props) {
           <Button
             variant="danger"
             onClick={() => {
-              dispatch(stopBeamlineAction(actionId));
+              dispatch(stopBeamlineAction('beamline_actions'));
```
marcus-oscarsson, line 46: No longer passing the id of the action (`action_id`) but the hardware object responsible for handling the stop command. We only have one `beamline_action` hardware object "beamline_actions". We can also only run one of the actions at the time.

mxcubeweb/core/adapter/beamline_action_adapter.py
fabcor-maxiv: I guess it's fine to put an exception.

Otherwise I guess the most common solution is like this:

```suggestion
        default_resource_handler_config: AdapterResourceHandlerConfigModel = (
            resource_handler_config
        ),
```

mxcubeweb/core/adapter/beamline_action_adapter.py
fabcor-maxiv: What do you mean? You mean for type hint annotation?

```
if typing.TYPE_CHECK:
    import ...
```

otherwise maybe something like this:

```suggestion
        app: "mxcubeweb....MXCUBEAPP",
```

marcus-oscarsson: Yes, thats what I meant. 

I noticed that people do `if typing.TYPE_CHECK` and I guess whatever does runtime type checking has functionality to handle that in two passes ?. I have to admit that I find it really awkward , and that Python again bites its own tail. That however a discussion for another time.

How would `"mxcubeweb....MXCUBEAPP", work ?

ui/src/components/BeamlineActions/BeamlineActionForm.jsx
axelboc: `stopBeamlineAction` is used in a few other places, I think. Does `'beamline_actions'` need to be passed to it in every case? If so, then it would make more sense to hard-code this string inside the action creator itself ... or maybe even in the `sendStop` function (in which case the `type` could be hardcoded as well).

mxcubeweb/core/adapter/beamline_action_adapter.py
fabcor-maxiv: I have not needed to do this in quite a while, but a few times I had to put quotes around the type hint (so that it is a string, something like: `app: "mxcubeweb.app.MXCUBEApplication"`), to avoid things like circular imports. If I remember correctly, I am a bit fuzzy on the details because it was a while ago.

`typing.TYPE_CHECKING` is always `False` except when the code is being interpreted by  type checking tool (like `mypy`) is running. I am not sure it would help for circular imports though.

ui/src/components/BeamlineActions/BeamlineActionForm.jsx
marcus-oscarsson: It is used in a few other places and in that case the name of the hardware object is passed as a variable, for instance the name each motor or actuator. We could generalize it so that we get the name of the `BeamlineActions` hardware object. That would be the elegant solution I guess, but it also felt a bit unnecessary in this case.

`type` may become redundant in the future but I kept it for all adapters for the moment to not make a too big change from how things are currently working. We have to discuss it and think a bit about it, but I think/hope that will come from the "route cleaning/defining" discussions in the future

axelboc: Makes sense!

mxcubeweb/core/adapter/beamline_action_adapter.py
walesch-yan: I also believe the exception here is fine, but I'm also fine with @fabcor-maxiv suggestion.

I just saw that the nomenclature for `default_resource_handler_config` and `resource_handler_config` is the other way around in the base class (`adapterbase`). It's not a big issue, but it might cause confusion in the future.

walesch-yan (approved):
You are right, I do appreciate this PR a lot:D (Thanks for leaving it open until I return, so I can personally approve it;))

It definitely helps to improve this section of the code in my opinion, thanks!

axelboc (approved):

mxcubeweb/core/adapter/beamline_action_adapter.py
marcus-oscarsson: Thanks

PR #1676, opened 2025-06-11
Fix violations of Ruff rule S113

Phantominumm:
fix S113, request without timeout, the request can hang indefinitely while awaiting a response

https://docs.astral.sh/ruff/rules/request-without-timeout/

### DIFF
--- a/mxcubeweb/core/components/user/usermanager.py
+++ b/mxcubeweb/core/components/user/usermanager.py
@@ -17,7 +17,7 @@ from mxcubecore.model.lims_session import LimsSessionManager
 from mxcubeweb.core.components.component_base import ComponentBase
 from mxcubeweb.core.models.usermodels import User
 from mxcubeweb.core.util.networkutils import is_local_host
-
+HTTP_REQUESTS_TIMEOUT = 5
 
 class BaseUserManager(ComponentBase):
     """Base class for managing user-related operations
@@ -237,7 +237,7 @@ class BaseUserManager(ComponentBase):
                     "grant_type": "refresh_token",
                     "refresh_token": current_user.refresh_token,
                 },
-                timeout=5,
+                timeout=HTTP_REQUESTS_TIMEOUT,
             ).json()
         )
 
@@ -598,7 +598,7 @@ class UserManager(BaseUserManager):
                     "client_secret": self.app.CONFIG.sso.CLIENT_SECRET,
                     "refresh_token": refresh_token,
                 },
-                timeout=5,
+                timeout=HTTP_REQUESTS_TIMEOUT,
             )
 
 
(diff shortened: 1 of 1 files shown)


mxcubeweb/core/components/user/usermanager.py
```diff
@@ -237,6 +237,7 @@ def sso_token_expired(self) -> bool:
                     "grant_type": "refresh_token",
                     "refresh_token": current_user.refresh_token,
                 },
+                timeout=5,
```
fabcor-maxiv: It might be better to use a named constant

```suggestion
                timeout=HTTP_REQUESTS_TIMEOUT,
```

With something like this at the top:

```python
HTTP_REQUESTS_TIMEOUT = 5  #: Generic timeout for HTTP request in seconds
```

mxcubeweb/core/components/user/usermanager.py
```diff
@@ -17,7 +17,7 @@
 from mxcubeweb.core.components.component_base import ComponentBase
 from mxcubeweb.core.models.usermodels import User
 from mxcubeweb.core.util.networkutils import is_local_host
-
+HTTP_REQUESTS_TIMEOUT = 5
```
fabcor-maxiv:
```suggestion
HTTP_REQUESTS_TIMEOUT = 5  #: Generic timeout for HTTP requests in seconds
```

fabcor-maxiv:
Please remove the corresponding `S113` exceptions in `ruff.toml`.

mxcubeweb/core/components/user/usermanager.py
```diff
@@ -18,6 +18,8 @@
 from mxcubeweb.core.models.usermodels import User
 from mxcubeweb.core.util.networkutils import is_local_host
 
+HTTP_REQUESTS_TIMEOUT = 5  #: Generic timeout for HTTP requests in seconds
```
marcus-oscarsson: Very nice, 5 seconds should be enough but could we set it to 10 just in case. The network is sometimes really slow

mxcubeweb/core/components/user/usermanager.py
```diff
@@ -18,6 +18,8 @@
 from mxcubeweb.core.models.usermodels import User
 from mxcubeweb.core.util.networkutils import is_local_host
 
+HTTP_REQUESTS_TIMEOUT = 10  # 5 is Generic timeout for HTTP requests in seconds, 10 to be safe
```
fabcor-maxiv:
```suggestion
HTTP_REQUESTS_TIMEOUT = 10  # Generic timeout for HTTP requests in seconds
```

fabcor-maxiv (approved):
Good for me.

Commits should be squashed.

marcus-oscarsson:
:fireworks: @Phantominumm

PR #1677, opened 2025-06-11
Fix violations of Ruff rule S103

Phantominumm:
S103, bad_file_permissions, 0o666 was too permissive and may allow unintended access and arbitrary code execution.
0o600 can be not enough permissive we can discuss maybe 0o660 or 0o644 for reading only.

https://docs.astral.sh/ruff/rules/bad-file-permissions/


fabcor-maxiv:
On this and all other such pull requests (even those who have already been merged) we need to remove the corresponding exceptions in `ruff.toml`.

That is why I always try to make the point that we are clear about what we are doing in the commit messages and pull request descriptions. In this case I wish the description stated clearly "**Fix violations of Ruff rule S103**". Then it is much clearer what this is about. :)

marcus-oscarsson:
Thanks @Phantominumm  keep up the good work :). We should indeed remove the exceptions from the `ruff.toml`, for instance:

https://github.com/mxcube/mxcubeweb/blob/680078bcdd5e780a982585b800f866da17dff291/ruff.toml#L53

We can follow up with a separate  PR for the ones we forgot.

marcus-oscarsson:
Very good, ready to go as far as I can see :)

fabcor-maxiv:
Do we need to change the permissions at all? Aren't the default permissions fine already? If they are not fine, then probably the permissions should be changed on the system itself. It does not seem right to me that MXCuBE changes file permission bits.

marcus-oscarsson:
That's a good, point what are the default permissions perhaps we can remove the `chmod` ?

beteva:
> Do we need to change the permissions at all?

I guess it is good from security point of view that we are sure what the file permissions are and not rely on the system setup.

fabcor-maxiv:
> > Do we need to change the permissions at all?
> 
> I guess it is good from security point of view that we are sure what the file permissions are and not rely on the system setup.

I do not think so, but I am not a security person.

fabcor-maxiv:
> That's a good, point what are the default permissions perhaps we can remove the `chmod` ?

The default permissions are whatever you set on your system. I never remember exactly how this works, but I seem to recall that you can set permissions at some parent directory level and set some recursive and/or "sticky" bit so that the permissions are automatically applied to all the new files below that directory.

Phantominumm:
After some research, I found that the umask is generally set to 022 on linux systems, so files are usually created with permissions 644 (readable by others, writable only by the owner).
The problem can be the number of users and the sensitivity of the information stored in these files.
- If the information should not be accessible by others, we need to keep the explicit chmod to ensure the files are secured.

- If the information can be safely read by others, then we can rely on the system’s umask and avoid transforming permissions explicitly, assuming all systems have the appropriate umask configured.

fabcor-maxiv:
From my point of view, the default is fine. If facilities want different permission bits, I quite confidently assume that there are ways to do this outside of MXCuBE.

marcus-oscarsson:
> If the information should not be accessible by others, we need to keep the explicit chmod to ensure the files are secured.

Very nice summary @Phantominumm, I think the current implementation is good considering that we want to provide a safe default.

fabcor-maxiv:
No, this is a bad change. We can not override MXCuBE's behaviour, and give different permissions. There should be no `chmod` here.

marcus-oscarsson:
Hm, sorry. I misread the code a bit I thought that there was an indentation there. I thought that the `chmod` was only called on creation which I think would be fine and actually a good idea. One would not in general want the log files to be read by anyone. Ill follow up with a PR that fixes that.

PR #1678, opened 2025-06-11
Adds a way to split custom form fields into groups

mockoocy:
Based on the `ui:options.group` key in the `uiSchema` form fields will be assigned to groups.

This group-assignment is done via the `uiSchema` key instead of seemingly more obvious way - creating nested objects. In some places MXCuBE expects form to have some top-level parameters; e.g. here https://github.com/mxcube/mxcubeweb/blob/680078bcdd5e780a982585b800f866da17dff291/ui/src/components/Tasks/GenericTaskForm.jsx#L380-L384 and here: https://github.com/mxcube/mxcubeweb/blob/680078bcdd5e780a982585b800f866da17dff291/mxcubeweb/core/components/queue.py#L1020-L1027 It seems the most reasonable, to keep the form parameters flat, at least for now. Otherwise many custom forms using this feature would suffer from the problem state above.
This requires defining the form groups in some other place, like uiSchema.

This is how it looks:

image

### DIFF
--- a/ui/src/components/Tasks/CustomObjectFieldTemplate.jsx
+++ b/ui/src/components/Tasks/CustomObjectFieldTemplate.jsx
@@ -1,0 +1,43 @@
+import groupBy from 'lodash/groupBy';
+
+import { CollapsableRows, FieldsHeader } from './fields';
+/**
+ * @param {import('@rjsf/utils').ObjectFieldTemplateProps} props
+ * @returns {JSX.Element}
+ */
+export default function CustomObjectFieldTemplate({
+  properties,
+  uiSchema = {},
+}) {
+  const groups = groupBy(
+    properties,
+    (field) => uiSchema[field.name]?.['ui:options'].group,
+  );
+
+  const { undefined: ungroupedFields = [], ...groupedFieldsMap } = groups;
+  const groupNames = Object.keys(groupedFieldsMap);
+
+  return (
+    <div>
+      {ungroupedFields.length > 0 && (
+        <>
+          <FieldsHeader title="Acquisition" />
+          <div className="row">
+            {ungroupedFields.map(({ content }) => content)}
+          </div>
+        </>
+      )}
+      {groupNames.map((group) => {
+        const fields = groupedFieldsMap[group];
+        return (
+          <div key={group}>
+            <FieldsHeader title={group} />
+            <CollapsableRows>
+              <div className="row">{fields.map(({ content }) => content)}</div>
+            </CollapsableRows>
+          </div>
+        );
+      })}
+    </div>
+  );
+}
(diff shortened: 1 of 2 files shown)


marcus-oscarsson:
Hey ! Well done that starts to look really nice :). I also think its a good idea to, for now, keep the parameters structure flat

ui/src/components/Tasks/CustomObjectFieldTemplate.jsx
```diff
@@ -0,0 +1,45 @@
...
+import { CollapsableRows, FieldsHeader } from './fields';
+/**
+ * @param {import('@rjsf/utils').ObjectFieldTemplateProps} props
+ * @returns {JSX.Element}
+ */
+export default function CustomObjectFieldTemplate({
+  properties,
+  uiSchema = {},
+}) {
+  const { ungroupedFields, groupedFieldsMap } = useMemo(() => {
```
axelboc: No need for the `useMemo` here, the computations are simple.

ui/src/components/Tasks/CustomObjectFieldTemplate.jsx
```diff
@@ -0,0 +1,45 @@
...
+    const groups = Object.groupBy(
```
axelboc: `Object.groupBy` is a very recent API. Maybe just use `lodash` for now (it's already installed)?

axelboc (approved):
Looking good!

marcus-oscarsson:
Very nice and a super useful feature. 

Should we add some more space between the title i.e `processing` and the section above ? I feel that its a bit tight as it is. Also perhaps, a few pixels space between the input field label and the input box ?

I know that I mentioned it before but it would be super nice if we by default could make the number inputs shorter. Perhaps we can somehow default the bootstrap column to something slightly smaller so that we make use of the column design for this. Its perhaps worth a try ...

mockoocy:
Hm, with regards to the input widths;
The modal itself is awkwardly wide by default, so using dynamic sizing with the bootstrap cols gives us also awkward results. 

One way would be to make them a bit smaller by default (`col-4` instead of `col-6` or using `col-auto`), and the modal itself too. 
The problem is that the `Path` field on the top of the form wouldn't be fully rendered. 

image

Perhaps I could just limit the width of the modal and make it clearer that the `Path` field overflows it's container? 



Edit: alternatively the modal size could be proportional to number of columns in the form grid.

marcus-oscarsson:
I think the current suggestion is clear improvement. I;m not sure how to deal with the size of the input boxes and still keep nice "tabular/row"  display without at the same time generating lots of white space.

I'm not sure making the dialog more narrow is a good choice either, it sort of looks a bit weird to have those narrow dialogs on a otherwise fairly wide interface. 

Lets think about it, maybe some of the others has some good ideas, @axelboc, @jbflo, @ddesanctis,

marcus-oscarsson:
I'm otherwise happy to merge this one if you don't want to add something more ?

mockoocy:
I think it is good to merge this one now.
I will write down an issue regarding the input width, so that discussions regarding it will happen there :)

marcus-oscarsson:
Sure, its already much better, very nice ! Then we have to look at all the possible improvements one by one.

PR #1680, opened 2025-06-12
Fix minor 'bugs' (code cleanliness) detected by SonarQube

Phantominumm:
it's more about code cleanliness than real bug, i will list the the (resolved) problem here,
they are in different commit and doesn't change the logic:

- Unexpected missing generic font family in css files.
- Unused variable in function and useless self-assignment
- Change code so that is raises an object deriving from BaseException
- Using forEach instead of map as its return value is not being used here
- Function expects 2 args but 3 were provided

### DIFF
--- a/ui/src/components/SampleGrid/SampleCircleView.jsx
+++ b/ui/src/components/SampleGrid/SampleCircleView.jsx
@@ -30,7 +30,7 @@ export default function SampleCircleView(props) {
 
   function handleDisplayPuckCellContextMenu(e, menuID, puckID) {
     e.preventDefault();
-    handleClickOnCellPuck(e, cellID, puckID === null ? '' : puckID);
+    handleClickOnCellPuck(e, puckID === null ? '' : puckID);
     displayPuckCellContextMenu(e, menuID, cellID, puckID);
     e.stopPropagation();
   }
--- a/mxcubeweb/core/adapter/flux_adapter.py
+++ b/mxcubeweb/core/adapter/flux_adapter.py
@@ -28,9 +28,9 @@ class FluxAdapter(ActuatorAdapterBase):
             ho.connect("valueChanged", self._value_change)
 
     @RateLimited(6)
-    def _value_change(self, value, **kwargs):
-        value = "{:.2E}".format(Decimal(self._ho.get_value()))
-        self.value_change(value, **kwargs)
+    def _value_change(self,**kwargs):
+        new_value = "{:.2E}".format(Decimal(self._ho.get_value()))
+        self.value_change(new_value, **kwargs)
 
     def set_value(self, value=None):
         """Read only"""
--- a/mxcubeweb/core/adapter/flux_adapter.py
+++ b/mxcubeweb/core/adapter/flux_adapter.py
@@ -28,7 +28,7 @@ class FluxAdapter(ActuatorAdapterBase):
             ho.connect("valueChanged", self._value_change)
 
     @RateLimited(6)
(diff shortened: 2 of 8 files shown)


fabcor-maxiv:
I'd much prefer one pull request per fix, please!

marcus-oscarsson:
Ah, I will make a PR to remove the `FluxaAdapter` because its no longer needed. So you can ignore that one.

Seeing that this PR touches both python code and JS code, Id make on PR for JS and one for Python. I think that would be sufficient in this case.

Phantominumm:
they are really small changes and a PR for each seems too much, i will do one for JS and one for Python as Marcus suggested

ui/src/components/SampleGrid/SampleCircleView.jsx
```diff
@@ -30,7 +30,7 @@ export default function SampleCircleView(props) {
 
   function handleDisplayPuckCellContextMenu(e, menuID, puckID) {
     e.preventDefault();
-    handleClickOnCellPuck(e, cellID, puckID === null ? '' : puckID);
+    handleClickOnCellPuck(e, puckID === null ? '' : puckID);
```
axelboc, line 33: Not the right fix for this. Have a look at the `handleClickOnCellPuck` function :wink:

axelboc (changes requested):
The rest of the front-end fixes are fine :blush: 

(Sorry, just realised this PR was in _Draft_.)

ui/src/components/SampleGrid/SampleCircleView.jsx
Phantominumm: **cellID** is a constant define line 14,"   const cellID = 1; "
 so when we call **handleClickOnCellPuck** here the value of **cellID** is always 1.
The def of " handleClickOnCellPuck(event, puckID) " has only 2 parameters and even if **cellID** is use in the function it will always take the constant value.
I don't see why we need to keep **cellID** in the call here ?

axelboc: Indeed you're right, well spotted! Don't hesitate to comment on your own Pull Request next time to explain this sort of things, especially when it's not obvious like here. Code reviews are meant to catch human errors, so as a code reviewer, I just need to know that the changes I'm seeing are done on purpose and for valid reasons. Since I didn't know, I looked at the code quickly and thought that `handleClickOnCellPuck` was maybe missing a parameter.

Side comment, it would be good to rename `cellID` to `CELL_ID` to emphasize that it's a constant and not a local variable.

jbflo: > **cellID** is a constant define line 14," const cellID = 1; " so when we call **handleClickOnCellPuck** here the value of **cellID** is always 1. The def of " handleClickOnCellPuck(event, puckID) " has only 2 parameters and even if **cellID** is use in the function it will always take the constant value. I don't see why we need to keep **cellID** in the call here ?

 thanks, you got it right, this function was copied from another class and we forgot to remove cellID.  i guess you can even rename the function to "handleClickOnPuck" .

jbflo, line 33: EDITED :
Well spotted, these changes will be potted here : https://github.com/mxcube/mxcubeweb/pull/1681#discussion_r2142204053

marcus-oscarsson: Thanks, well done all. Its nice to see when everybody can chip in like this :smile:

PR #1681, opened 2025-06-12
Fix JS & CSS minor issues

Phantominumm:
They are the changes reported by SonarQube as bugs for css and JS files:

- add generic font family in css files (sans-serif)

- use of forEach instead of map because no use of the return value 

- remove an extra argument in the call of handleClickOnCellPuck, the function take 2 args but 3 were given and the extra one was a constant define for all the file.

### DIFF
--- a/ui/src/components/MXNavbar/MXNavbar.module.css
+++ b/ui/src/components/MXNavbar/MXNavbar.module.css
@@ -73,7 +73,7 @@
   padding-top: 1px; /* tweak vertical alignment of number */
   background-color: #085de6;
   border-radius: 50%;
-  font-family: Helvetica, sans-serif;
+  font-family: Helvetica;
   font-size: 0.75em;
   font-weight: bold;
 }
--- a/ui/src/components/MXNavbar/MXNavbar.module.css
+++ b/ui/src/components/MXNavbar/MXNavbar.module.css
@@ -73,7 +73,7 @@
   padding-top: 1px; /* tweak vertical alignment of number */
   background-color: #085de6;
   border-radius: 50%;
-  font-family: Helvetica;
+  font-family: Helvetica, sans-serif;
   font-size: 0.75em;
   font-weight: bold;
 }
--- a/ui/src/components/SampleGrid/SampleCircleView.jsx
+++ b/ui/src/components/SampleGrid/SampleCircleView.jsx
@@ -13,10 +13,10 @@ export default function SampleCircleView(props) {
 
   const CELL_ID = 1;
 
-  const DISPATCH = useDispatch();
+  const dispatch = useDispatch();
 
   function handleClickOnPuck(event, puckID) {
-    DISPATCH(
+    dispatch(
       filterAction({ cellFilter: `${CELL_ID}`, puckFilter: `${puckID}` }),
     );
     event.stopPropagation();
@@ -24,9 +24,9 @@ export default function SampleCircleView(props) {
 
   useEffect(() => {
     if (FILTEROPTIONS.cellFilter === '') {
-      DISPATCH(filterAction({ cellFilter: '1', puckFilter: '1' }));
+      dispatch(filterAction({ cellFilter: '1', puckFilter: '1' }));
     }
-  }, [FILTEROPTIONS.cellFilter, DISPATCH]);
+  }, [FILTEROPTIONS.cellFilter, dispatch]);
 
   function handleDisplayPuckCellContextMenu(e, menuID, puckID) {
     e.preventDefault();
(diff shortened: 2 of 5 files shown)


ui/src/components/SampleGrid/SampleCircleView.jsx
```diff
@@ -30,7 +30,7 @@ export default function SampleCircleView(props) {
 
   function handleDisplayPuckCellContextMenu(e, menuID, puckID) {
     e.preventDefault();
-    handleClickOnCellPuck(e, cellID, puckID === null ? '' : puckID);
+    handleClickOnCellPuck(e, puckID === null ? '' : puckID);
```
jbflo: Nice , Thanks

ui/src/components/SampleGrid/SampleCircleView.jsx
Phantominumm, line 1: Constant are now capital letter and rename handleClickOnCellPuck into handleClickOnPuck as suggested

ui/src/components/SampleGrid/SampleCircleView.jsx
```diff
@@ -7,36 +7,36 @@ import { filterAction } from '../../actions/sampleGrid';
 export default function SampleCircleView(props) {
   const { displayPuckCellContextMenu, puckMenuID, type = 'Mockup' } = props;
 
-  const filterOptions = useSelector((state) => state.sampleGrid.filterOptions);
+  const FILTEROPTIONS = useSelector((state) => state.sampleGrid.filterOptions);
```
marcus-oscarsson: `FILTEROPTIONS` and `PUCKS` are not really constants, we can IMO keep them camel-cased (sorry for making you change again).

axelboc: Indeed, "constant" is a vague term, sorry @Phantominumm. It's mostly magic numbers/strings, configuration objects, static arrays, etc. Stuff that can live outside of components. We usually put them at the top of the file they're used in so they're easy to spot, or in shared files if they are needed in multiple places.

```jsx
const CELL_ID = 1;

export default function SampleCircleView(props) {
  ...
}
```

Phantominumm: no that was me, I simply read const and change it without checking it, that was a dumb mistake.

ui/src/components/SampleGrid/SampleCircleView.jsx
```diff
@@ -7,31 +7,31 @@ import { filterAction } from '../../actions/sampleGrid';
 export default function SampleCircleView(props) {
   const { displayPuckCellContextMenu, puckMenuID, type = 'Mockup' } = props;
 
-  const filterOptions = useSelector((state) => state.sampleGrid.filterOptions);
+  const filteroptions = useSelector((state) => state.sampleGrid.filterOptions);
```
jbflo: The variable filterOptions is correctly written in camelCase
No changes are needed here.

ui/src/components/SampleGrid/SampleCircleView.jsx
```diff
@@ -7,31 +7,31 @@ import { filterAction } from '../../actions/sampleGrid';
...
 
   const pucks = useSelector((state) => state.sampleChanger.contents.children);
 
-  const cellID = 1;
+  const CELL_ID = 1;
```
jbflo: you can move this line of  code `const CELL_ID = 1;`  from line 14 to line 6
to the outscore.

jbflo (changes requested):
i think my comments are  what @axelboc and @marcus-oscarsson suggested already.

ui/src/components/MXNavbar/MXNavbar.module.css
```diff
@@ -73,7 +73,7 @@
   padding-top: 1px; /* tweak vertical alignment of number */
   background-color: #085de6;
   border-radius: 50%;
-  font-family: Helvetica;
+  font-family: Helvetica, sans-serif;
```
fabcor-maxiv: Is this one necessary if it is already set on `<body>` via `main.css`?

Phantominumm: I agree with that,
 sonarQube still shows me the error but it's a false negative i think.

axelboc: If you leave it as is, with only `Helvetica`, then it overrides the `font-family` declaration from `body`, and you lose the `sans-serif` fallback (the values don't merge), so Sonar is correct in reporting an issue.

Try removing `font-family` here altogether so as to inherit the `body`'s `font-family` declaration. Just double check the UI to make sure this `font-family` declaration was indeed not required. You'll need to login from a normal window and a private window at the same time to see the observers badge in the nav bar.

jbflo (approved):

axelboc (approved):
I'm not fussed about that `font-family` thingy — it's good to go for me as well. Thanks for the fixes :+1:

PR #1682, opened 2025-06-12
Fix python minor issues

Phantominumm:
They are the changes reported by SonarQube as bugs for python files:
1)
remove useless self-assignment and if : 
if():
img = img 
else : ...
(line 55 mxcubeweb/core/components/sampleview.py) 

2)
Change code so that is raises an object deriving from BaseException
raise msg
is now raise NoSessionError(msg) ; new class define in the file


fabcor-maxiv (approved):
Seems fine to me.

But we are in a situation where we have 1 pull request, 2 fixes, and 4 commits. So in principle we should not simply squash the whole pull request, because then we would have 2 unrelated fixes in the same commit, and then in the event where one of the 2 fixes turns out to be a bad fix, we can not simply revert the commit. But I guess you know all that already.

PR #1683, opened 2025-06-13
Remove mockups route

Phantominumm:
Remove mockups route,
the file mockups.py and the imports in server.py are deleted as discussed in  #1673


fabcor-maxiv (approved):
Seems fine to me

PR #1685, opened 2025-06-13
Remove rule already fixed in ruff.toml

Phantominumm:
Remove the rules that were fixed in previous PR in the ruff.toml file
#1683 #1676 #1672 #1669


marcus-oscarsson:
Thanks !

PR #1686, opened 2025-06-13
BeamlineActions entierly handled by BeamlineActionAdapter

marcus-oscarsson:
Continuing the cleanup of the BeamlineActions, all aspects of BeamlineActions are now handled with `BeamlineActionAdapter`. The corresponding logic have been moved/removed from the beamline "component" (which we eventually can remove)

 - The execute/run action is now handled by the `BeamlineActionAdapter` just like `stop`.
 - The retrieval of all beamline actions are done via  `BeamlineActionAdapter`

### DIFF
--- a/mxcubeweb/core/adapter/beamline_action_adapter.py
+++ b/mxcubeweb/core/adapter/beamline_action_adapter.py
@@ -4,17 +4,54 @@ from typing import ClassVar
 
 from mxcubecore.BaseHardwareObjects import HardwareObject
 from mxcubecore.HardwareObjects.BeamlineActions import BeamlineActions
+from pydantic.v1 import (
+    BaseModel,
+)
 
 from mxcubeweb.core.adapter.adapter_base import AdapterBase
+from mxcubeweb.core.components.queue import (
+    FAILED,
+    READY,
+    RUNNING,
+)
 from mxcubeweb.core.models.adaptermodels import (
+    BeamlineActionInputModel,
     HOActuatorValueChangeModel,
     NStateModel,
 )
 from mxcubeweb.core.models.configmodels import AdapterResourceHandlerConfigModel
 
 resource_handler_config = AdapterResourceHandlerConfigModel(
-    commands=["stop"], attributes=["data"]
+    commands=["stop", "run_action"], attributes=["data", "get_all_actions"]
 )
+
+
+class Argument(BaseModel):
+    name: str
+    type: str
+    items: list[str] | None = None
+
+
+class JSONSchemaArgument(BaseModel):
+    name: str
+    type: str
+    jsonschema: str = ""
+    signature: list[str]
+
+
+class Action(BaseModel):
+    name: str
+    username: str
+    state: int
+    arguments: list[JSONSchemaArgument | Argument]
+    argument_type: str
+    messages: list[str]
+    type: str
+    data: str | dict | None
+
+
+class ActionsList(BaseModel):
+    actions_list: list[Action]
 
 
 class BeamlineActionAdapter(AdapterBase):
@@ -37,29 +74,44 @@ class BeamlineActionAdapter(AdapterBase):
         ho.connect("valueChanged", self._value_change)
         ho.connect("stateChanged", self.state_change)
 
-    def _value_change(self, value):
-        v = value.name if isinstance(value, Enum) else value
-
-        self.value_change(v)
-
-    def commands(self):
-        return [
-            attribute
-            for attribute in dir(self._ho.__class__)
-            if callable(getattr(self._ho.__class__, attribute))
-            and attribute.startswith("_") is False
-        ]
-
-    def msg(self):
-        try:
-            msg = self._ho.get_value().name
-        except Exception:
-            msg = "---"
-            logging.getLogger("MX3.HWR").error(
-                "Failed to get beamline attribute message"
+        for cmd in ho.get_commands() + ho.get_annotated_commands():
+            cmd.connect(
+                "commandBeginWaitReply",
+                self._beamline_action_start,
+            )
+            cmd.connect(
+                "commandReplyArrived",
+                self._beamline_action_done,
+            )
+            cmd.connect(
+                "commandFailed",
+                self._beamline_action_failed,
             )
 
-        return msg
+    def _value_change(self, value):
+        v = value.name if isinstance(value, Enum) else value
+        self.value_change(v)
+
+    def _emit_beamline_action(self, msg):
+        self.app.server.emit("beamline_action", msg, namespace="/hwr")
+
+    def _beamline_action_start(self, name):
+        msg = {"name": name, "state": RUNNING}
+        self._emit_beamline_action(msg)
+        logging.getLogger("user_level_log").info("Command %s started.", name)
+
+    def _beamline_action_done(self, name, result):
+        msg = {"name": name, "state": READY, "data": result}
+        self._emit_beamline_action(msg)
+        logging.getLogger("user_level_log").info("Command %s done.", name)
+
+    def _beamline_action_failed(self, name):
+        msg = {"name": name, "state": FAILED}
+        self._emit_beamline_action(msg)
+        logging.getLogger("user_level_log").error("Action %s failed!", name)
+
+    def msg(self):
+        return ""
 
     def stop(self):
         """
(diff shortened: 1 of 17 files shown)


mxcubeweb/core/adapter/beamline_action_adapter.py
```diff
@@ -68,5 +69,16 @@ def stop(self):
         for cmd in self._ho.get_commands():
             self._ho.abort_command(cmd.name())
 
+    def run_action(self, value: BeamlineActionInputModel):
```
marcus-oscarsson, line 141: Input modeled as PyDantic model

mxcubeweb/core/server/resource_handler.py
```diff
@@ -51,8 +51,9 @@ def assert_pydantic_arguments(func):
             continue
 
         # Raise RuntimerError If it's not a Pydantic model
-        if not issubclass(param_type, BaseModel):
-            msg = f"Argument {param_name} of {func} are not a pydantic model"
+        if not issubclass(param_type, BaseModel | float | int | str):
```
marcus-oscarsson: Also accepting passing of types float, int and str. There is no need to model such simple types. Numbers are considered safe as they are and strings are sanitized before being passed.

mxcubeweb/core/server/resource_handler.py
```diff
@@ -355,22 +357,30 @@ def _validate_param_data(
         # If it's a Pydantic model, validate it
         if issubclass(param_type, BaseModel):
             try:
+                # We only handle a single argument when using Pydantic models
```
marcus-oscarsson, line 362: We only handle a single PyDantic model that represents the entire input. However one can pass any number of float, int or str arguments

ui/src/actions/beamlineActions.js
```diff
@@ -36,14 +35,18 @@ export function setArgumentValue(cmdName, argIndex, value) {
...
+  return (dispatch, getState) => {
     dispatch(setActionState(cmdName, RUNNING));
 
     if (showOutput) {
       dispatch(showActionOutput(cmdName));
     }
+    const state = getState();
 
-    sendRunBeamlineAction(cmdName, parameters);
+    sendExecuteCommand('beamlineaction', 'beamline_actions', 'run_action', {
```
marcus-oscarsson, line 49: Calling the generic method that runs a "command" on adapter in this case "run_action"

mxcubeweb/core/models/adaptermodels.py
```diff
@@ -75,3 +75,8 @@ class FloatValueModel(BaseModel):
 
 class StrValueModel(BaseModel):
     value: str = Field("", description="Value of actuator (position)")
+
+
+class BeamlineActionInputModel(BaseModel):
+    cmd: str
+    parameters: dict
```
marcus-oscarsson: The model is indeed not very strict at the moment

ui/src/api/hardware-object.js
```diff
@@ -1,19 +1,21 @@
...
-export function sendGetAttribute(objectId, type) {
-  return endpoint.put({}, `/${type}/${objectId}/get_value`).res();
+export function sendGetAttribute(objectId, objectType) {
+  return sendExecuteCommand(objectType, objectId, 'get_value', {});
 }
 
-export function sendStop(objectId, type) {
-  return endpoint.put({}, `/${type}/${objectId}/stop`).res();
+export function sendStop(objectId, objectType) {
+  return sendExecuteCommand(objectType, objectId, 'stop', {});
```
axelboc: Since these functions no longer have their own endpoint, they no longer make sense. I would just call `sendExecuteCommand` with the appropriate arguments from all the relevant places (just like you did for `sendRunBeamlineAction`).

mxcubeweb/core/adapter/beamline_action_adapter.py
```diff
@@ -4,19 +4,56 @@
...
 )
 from mxcubeweb.core.models.configmodels import AdapterResourceHandlerConfigModel
 
 resource_handler_config = AdapterResourceHandlerConfigModel(
-    commands=["stop"], attributes=["data"]
+    commands=["stop", "run_action"], attributes=["data", "get_all_actions"]
 )
 
 
+class Argument(BaseModel):
```
marcus-oscarsson, line 29: These models are only used by the `BeamlineActions`, I was hesitating of moving them to `models/adaptermodels` but decided to keep them in the same module for simplicity.

mxcubeweb/core/adapter/beamline_action_adapter.py
```diff
@@ -68,5 +120,113 @@ def stop(self):
...
+            raise ValueError(msg)
+
+        try:
+            self._ho.execute_command(value.cmd, value.parameters)
+        except Exception:
+            msg = f"Action cannot run: command {value.cmd} does not exist"
+            logging.getLogger("MX3.HWR").exception(msg)
+            raise
+
+    def get_all_actions(self) -> ActionsList:
```
marcus-oscarsson, line 173: This code was more or less copied as it was from `components/beamline`

mxcubeweb/core/server/resource_handler.py
```diff
@@ -437,11 +447,16 @@ def _add_exports(self, items: list[str], http_method: str) -> None:
             items: The list of commands or properties to add
             http_method: The HTTP method to use, GET, POST, PUT, DELETE
         """
+        if http_method == "GET":
+            decorators = [self._server.restrict]
```
marcus-oscarsson, line 453: Not using `require_control` for GET `requests` by default

mxcubeweb/core/server/resource_handler.py
```diff
@@ -41,7 +41,7 @@ def validate_input_str(input_string: str) -> bool:
     return bool(re.match(pattern, input_string))
 
 
-def assert_pydantic_arguments(func):
+def assert_valid_type_arguments(func):
     """Make sure that all the arguments of func are typehinted as pydantic models"""
```
walesch-yan: Maybe we should change the function description as well, as the typehints now are not necessarily pedantic models

mxcubeweb/core/server/resource_handler.py
```diff
@@ -292,12 +294,12 @@ def _view_func(object_id: str, *args, **kwargs) -> any:  # noqa: ARG001
 
         return _view_func
 
-    def _assert_pydantic_arguments(self, export):
+    def _assert_valid_type_arguments(self, export):
         """
         Ensures the method referenced in the export uses Pydantic arguments.
         """
```
walesch-yan, line 302: Same here

mxcubeweb/core/adapter/beamline_action_adapter.py
walesch-yan: I agree to keep them in this module, if we would ever need them in another place, we can always move them up

walesch-yan (approved):
From the code it looks good to me and is definitely a huge improvement towards better use of adapters:)

One minor critic point from my side is that we could maybe split this into 2 PRs, as it includes two related but different aspects of the adapters (first changes to resource_handler to allow basic types etc and then changes in BeamlineActionAdapter?) but I can also live with a single PR.

Anyways thanks a lot for the nice work!

ui/src/actions/beamlineActions.js
```diff
@@ -43,10 +46,21 @@ export function startBeamlineAction(cmdName, parameters, showOutput = true) {
...
   };
 }
 
+export function fetchGetAllActions() {
+  return sendGetAttribute(
+    'beamlineaction',
+    'beamline_actions',
+    'get_all_actions',
+  );
+}
```
axelboc, line 62: The `fetch` prefix is more for API functions in the `api` folder. I would just inline the `sendGetAttribute` call in `login.js`.

ui/src/api/hardware-object.js
```diff
@@ -2,18 +2,24 @@ import api from './api';
...
-  return endpoint.put({ args }, `/${objectType}/${objectId}/${command}`).res();
+export function sendExecuteCommand(objectType, objectId, command, value) {
+  return endpoint
+    .put({ value }, `/${objectType}/${objectId}/${command}`)
+    .safeJson();
 }
 
-export function sendSetAttribute(objectId, type, value) {
-  return endpoint.put({ value }, `/${type}/${objectId}/set_value`).res();
+export function sendGetAttribute(objectType, objectId, attributeName) {
```
axelboc, line 11: This is now a `get`, so the prefix should be `fetch`. Probably just `fetchAttribute`?

ui/src/api/hardware-object.js
```diff
@@ -2,18 +2,24 @@ import api from './api';
...
 
-export function sendStop(objectId, type) {
-  return endpoint.put({}, `/${type}/${objectId}/stop`).res();
+export function sendGetValue(objectId, objectType) {
+  return sendExecuteCommand(objectType, objectId, 'get_value', {});
+}
+
+export function sendStop(objectId, objectType) {
+  return sendExecuteCommand(objectType, objectId, 'stop', {});
 }
```
axelboc, line 25: I don't think those functions are needed. I would just call `sendExecuteCommand` directly to avoid misdirections.

Also, does the empty `value` parameter really matter? What happens if it remains undefined?

axelboc (commented):
Awesome, just a few tweaks in the front-end to match the (somewhat arbitrary) conventions I put in place for the `ui/src/api` folder.

PR #1687, opened 2025-06-13
Setting log file permissions on file creation

marcus-oscarsson:
Only setting the log file permissions when they are created

### DIFF
--- a/mxcubeweb/app.py
+++ b/mxcubeweb/app.py
@@ -33,7 +33,6 @@ from mxcubeweb.core.models.configmodels import UIComponentModel
 from mxcubeweb.logging_handler import MX3LoggingHandler
 from mxcubeweb.core.server.resource_handler import AdapterResourceHandlerFactory
 
-
 removeLoggingHandlers()
 
 
@@ -239,7 +238,6 @@ class MXCUBEApplication:
                 fpt = open(log_file, "w")
                 fpt.write(" ")
                 fpt.close()
-            os.chmod(log_file, 0o600)
             Path(log_file).touch()
 
             log_file_handler = TimedRotatingFileHandler(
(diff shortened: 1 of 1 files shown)


fabcor-maxiv (dismissed):
I do not see why MXCuBE should set specific permission bits at all. What is the reasoning?

It seems to me like these lines should be removed.

marcus-oscarsson:
The reasoning is that that the default umask is often 022 resulting in permissions of 644 when we by default would like 600. Then its up to each one to change the permissions of the log files or make sure they exist with the desired permisions before MXCuBE is launched.

fabcor-maxiv:
At the very least the rationale should be added as comment to the code.

I am still far from convinced that the `chmod` operations are necessary. Maybe I need to read up on this and discuss this with colleagues. Do you mind if we hold on this for a bit?

fabcor-maxiv:
At MAXIV, I am pretty sure that we want log files to be readable by other users. And it seems to me like setting such restrictive permissions would not allow that, and it seems to me like these permission bits set by MXCuBE can not be overridden easily, if at all.

marcus-oscarsson:
We don't mind. Basically, you don't want these files to be readable by other users on the system (especially not "others"). In case some one gets access to these files and are able to spot some weakness exception message, dump data etc.

**Edit**: With this change the permissions are only set when the files are created, I assume you already have the log files created in the right place with the rights you need ?

Actually, perhaps we can merge this in the meantime and you can get back to us

fabcor-maxiv (commented):
I forgot that we use `supervisord` and it handles the logging (from `stdout` and `stderr`). So I guess this change will have no impact for us after all. It might come a surprise to other facilities, though. In any case, I believe a comment in the code is necessary.

marcus-oscarsson:
So for the time being I removed the calls to `chmod` and then we have to think about how to better handle this in the meantime. I think it would still be better to set quite restrictive permissions on file creation. However, setting the same permissions every time the server starts is probably not the best idea. So finally not handling the permissions at all  would be the middle way

fabcor-maxiv (approved):

PR #1689, opened 2025-06-16
Remove GET from /clear endpoint

Phantominumm:
Removing GET method from /clear endpoint since it performs a destructive operation (clearing the queue). Only PUT should be used for state-modifying operations..
the methods use were PUT and GET and it's unsafe to allow both of them in the same time.

### DIFF
--- a/mxcubeweb/routes/queue.py
+++ b/mxcubeweb/routes/queue.py
@@ -89,7 +89,7 @@ def init_route(app, server, url_prefix):  # noqa: C901
         server.emit("queue", msg, namespace="/hwr")
         return Response(status=200)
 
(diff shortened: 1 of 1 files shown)


walesch-yan (approved):
LGTM!
Also, `Get` method is not used in the code base anyway

marcus-oscarsson:
Well done !

fabcor-maxiv:
@Phantominumm, the rationale in the pull request description is very good, but it is not in the git commit message. The best place for such explanation is in the git commit message. When I work on the project I look at the git commit messages first to understand why a change was made. In some cases I do not even have access to internet and GitHub.

PR #1690, opened 2025-06-16
Remove subclass Exception redudant

Phantominumm:
it’s a bit redundant because catching ImportError alone will also catch ModuleNotFoundError (since it inherits from it).

### DIFF
--- a/mxcubeweb/__version__.py
+++ b/mxcubeweb/__version__.py
@@ -21,7 +21,7 @@ except PackageNotFoundError:
     try:
         # Python >= 3.11 standard library
         from tomllib import load as load_toml  # pyright: ignore[reportMissingImports]
-    except (ImportError, ModuleNotFoundError):
+    except ImportError:
         # Python <= 3.10 predecessor library
         from tomli import load as load_toml  # pyright: ignore[reportMissingImports]
 


walesch-yan (approved):
Good catch!

Phantominumm:
I saw more of it, I will resolve all of them before reviewing

ui/src/components/Equipment/SampleChanger.jsx
```diff
@@ -207,8 +207,6 @@ export default function SampleChanger(props) {
         <i className="fas fa-stop" /> Abort
       </Button>
     );
-  } else {
-    abortButton = '';
   }
```
walesch-yan: Maybe we can replace this whole if statement with an inline jsx condition, i.e.: remove the if else completely and use 
``` 
        {props.state === 'MOVING' && (
          <span style={{ marginLeft: '1em' }}>
            {' '}
            <Button
              variant="danger"
              className={styles.abortButton}
              onClick={abort}
            >
              <i className="fas fa-stop" /> Abort
            </Button>
          </span>
        )}
```
in the return statement

fabcor-maxiv:
Are these things related? I do not see the relation between the change in the Python code and the changes in the JavaScript code.

walesch-yan:
I think the changed parts share that it is some redundant code. Indeed I share @fabcor-maxiv opinion, that we could split this into 2 or 3 PRs

Phantominumm:
yes, you are right, i made a mistake with branches this morning. I will keep only the python change here and do another PR for JS.

fabcor-maxiv (approved):

walesch-yan (approved):
Thanks @Phantominumm

PR #1691, opened 2025-06-17
Ml fix generic exception

Phantominumm:
Replace generic exception class with a more specific one.

### DIFF
--- a/mxcubeweb/core/components/beamline.py
+++ b/mxcubeweb/core/components/beamline.py
@@ -7,10 +7,6 @@ from mxcubeweb.core.components.component_base import ComponentBase
 from mxcubeweb.core.components.queue import READY
 
 
-class BeamlineActionError(Exception):
-    """Raised when a beamline action fails to execute."""
-
-
 class Beamline(ComponentBase):
     def __init__(self, app, config):
         super().__init__(app, config)
@@ -245,7 +241,7 @@ class Beamline(ComponentBase):
             HWR.beamline.beamline_actions.execute_command(name, params)
         except Exception as ex:
             msg = "Action cannot run: command '%s' does not exist" % name
-            raise BeamlineActionError(msg) from ex
+            raise AttributeError(msg) from ex
 
     def get_beam_info(self):
         """
--- a/mxcubeweb/core/components/beamline.py
+++ b/mxcubeweb/core/components/beamline.py
@@ -6,6 +6,8 @@ from mxcubeweb.core.adapter.beamline_adapter import BeamlineAdapter
 from mxcubeweb.core.components.component_base import ComponentBase
 from mxcubeweb.core.components.queue import READY
 
+class BeamlineActionError(Exception):
+    """Raised when a beamline action fails to execute."""
 
 class Beamline(ComponentBase):
     def __init__(self, app, config):
@@ -241,7 +243,7 @@ class Beamline(ComponentBase):
             HWR.beamline.beamline_actions.execute_command(name, params)
         except Exception as ex:
             msg = "Action cannot run: command '%s' does not exist" % name
-            raise Exception(msg) from ex
+            raise BeamlineActionError(msg) from ex
 
     def get_beam_info(self):
         """
--- a/mxcubeweb/core/components/beamline.py
+++ b/mxcubeweb/core/components/beamline.py
@@ -6,8 +6,10 @@ from mxcubeweb.core.adapter.beamline_adapter import BeamlineAdapter
 from mxcubeweb.core.components.component_base import ComponentBase
 from mxcubeweb.core.components.queue import READY
 
+
 class BeamlineActionError(Exception):
     """Raised when a beamline action fails to execute."""
+
 
 class Beamline(ComponentBase):
     def __init__(self, app, config):
--- a/mxcubeweb/core/components/queue.py
+++ b/mxcubeweb/core/components/queue.py
@@ -39,18 +39,6 @@ READY = 0
 ORIGIN_MX3 = "MX3"
 
 
-class QueueError(Exception):
-    """Base exception for queue-related errors."""
-
-
-class SampleNotFoundError(QueueError):
-    """Raised when a requested sample cannot be found in the queue."""
-
-
-class MethodNotFoundError(QueueError):
-    """Raised when a requested method cannot be found for a sample."""
-
-
 class Queue(ComponentBase):
     def __init__(self, app, config):
         super().__init__(app, config)
@@ -2205,7 +2193,7 @@ class Queue(ComponentBase):
         else:
             msg = "[QUEUE] Sample with id %s not in queue, can't update" % sid
             logging.getLogger("MX3.HWR").error(msg)
-            raise QueueError(msg)
+            raise ValueError(msg)
 
     def toggle_node(self, node_id):
         node = HWR.beamline.queue_model.get_node(node_id)
@@ -2410,7 +2398,7 @@ class Queue(ComponentBase):
         if not sample:
             msg = "[QUEUE] sample info could not be retrieved"
             logging.getLogger("MX3.HWR").error(msg)
-            raise SampleNotFoundError(msg)
+            raise KeyError(msg)
         # Find task with queue id method_id
         for task in sample.tasks:
             if task["queueID"] == int(method_id):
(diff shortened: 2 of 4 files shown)


rhfogh:
Is this not a bit pointless? The message says what the user needs to know. What do we gain by having a slew of new classes just to say the same thing? More complicated code and no more information? Does anyone plan to catch all these new exceptions separately and act differently on them? 

Or is this just to keep ruff happy?

axelboc:
While I'm definitely in favour of throwing more specific errors in general, I'm just wondering if it's really worth creating one exception class per exact error thrown. If there's no instance check, then isn't the error `msg` sufficient to identify the exact exception thrown? What does the subclass bring?

Using exception subclasses probably makes the code a bit more self-documenting and the logs more readable, but I think that can be achieved just as well with fewer subclasses. In JS, we have a few built-in error classes like `TypeError`, `RangeError`, etc. — they're more specific than just `Error`, yet they remain quite high-level.

fabcor-maxiv:
From my point of view...

This is a good initiative to look at exceptions. The way we deal with exceptions in the MXCuBE code base is quite bad. So any improvement in this area is welcome. This particular set of changes is maybe not where I would have started, though. Indeed it might be unnecessary to have very specific exception classes if they are never caught.

I think it might be a case where improvements have to be done case by case with thorough investigation, instead of in bulk. Things to consider ("to consider", not "to take for granted that this is the right thing to do"):

* Create a base MXCuBE-Web exception class `class MxcubeWebError(Exception): pass`
  * And then either raise `MxcubeWebError` or a custom exception that inherits from it
* When creating custom exception classes, maybe we should consider the level of specificity we want to go with, maybe it might make sense to group multiple sources of exceptions under one same custom exception class (especially if we have no intention of catching any of those anyway)
* Make sure that all exceptions are properly logged (if that makes sense), either at the point where they are raise or at the point where they are caught
* Make sure that logging of exceptions is done properly, with stack trace if that makes sense
  * For example in a `catch` block, consider using `logging.exception()` because it automatically takes care of printing the stack trace
* Make sure that things like `except: pass` are justified, in other words do not "swallow" exceptions without any trace, maybe a log entry would be nice even at debug level
* And then of course: investigate where we should/could actually catch exceptions and do something useful

On the other hand, we could also go along with this change. We define specific exception classes, and take it as incentive to write the code that catches those in the future. I would be fine with that.

rhfogh:
@fabcor-maxiv 
Thinking through the way we want to handle exceptions first, and then making the changes - as proposed by @fabcor-maxiv - sounds like a good idea. My proposal would be a handful (max) of special errors with clearly defined scope. For detailed information there are the messages.

On the other hand I would strongly oppose the idea to 'define specific exception classes, and take it as incentive to write the code that catches those'. Do we want 1) a separate exception class for every error, 2) the work of coding and catching all these exceptions individually? I would say absolutely NO! It would lead to code bloat, add a lot of extra lines and details that would need to be maintained - and I cannot see a single use case for that. Why would we want it?? What would it gain us? After all, Python itself defines TypeError, not 'StringInsteadOfFloatError' or 'StringNotAcceptedByMotorError'

This looks like part of a general trend. Ruff is supposed to help us, by pointing out places where the code might need improving, and prodding us to rewrite it properly. It is not us who are supposed to help Ruff, by just doing the simplest fix that will avoid linting failures.

marcus-oscarsson:
Very nice start @Phantominumm. So I think it would be good to try to as far as possible use the predefined exceptions:

https://docs.python.org/3/library/exceptions.html 

`MethodNotFoundError` and `QueueError` could be replaced by `ValueError`. They are both related to that the entity asked for somehow does not exist.

`AlreadyLoggedInError` and `InhouseAccessError` could be replaced by `RuntimeError` 

The main reason for being more specific with the exception classes is to also be able to handle them separately and make sure that we don' handle something incorrectly because the exception class  (or catch) is to broad.

**Edit**: These exceptions are not explicitly being caught, so there is no real need for creating one exception class for each exception unless one really need to handle that exception separately. It might be worth creating `MxcubeWebError` or `MxcubeError`, as @fabcor-maxiv mentions, and then creating more specific exception classes from those. 

I would however only do that if the built in exceptions for some reason are not enough or that we specifically need to handle that exception in a certain way (and discriminate it from the rest).

axelboc:
I think this comes from SonarQube (@Phantominumm maybe clarify this next time, and provide a link to the rule :pray:): https://rules.sonarsource.com/python/RSPEC-112/?search=exception

> From a consumer perspective, it is generally a best practice to only catch exceptions you intend to handle. Other exceptions should ideally not be caught and let to propagate up the stack trace so that they can be dealt with appropriately.

Phantominumm:
Yes, sorry for the confusion, it indeed come from SonarQube. As for the specific exception i understand that it's useless in this case, i will change them to the predefined from python for now waiting for the result of creating more specific one or not.

fabcor-maxiv:
Where is this Sonarqube thing running? Can we see the reports?

rhfogh:
Actually, I think that unless it is pretty clear that an error falls within the standard Python error types, it might be better to raise RuntimeError. That is supposed to be for " when an error is detected that doesn’t fall in any of the other categories." Maybe that could be the default option for replacing 'raise Exception'. 

Looking at these cases again, the beamline_run_action case actually has a problem that is rather more serious than just raising a bare Exception (which SonarQube then pointed us to, which is good). The line of code:
`            HWR.beamline.beamline_actions.execute_command(name, params)`
could raise an error for a number of reasons. It could be that the command does not exist, but it could equally well be that the command does exist but that you get an error when you execute it.  As the code is written the original error is swallowed up and replaced by a possibly misleading alternative. Arguably it would be better if you just sent out a log message and then re-raised the original error - at least people couild then see what actually went wrong. If we want better error handling than that (and we might well), it would be better to go into BeamlineActions.execute_command (and its sister commands) and improve the handling of command-not-found there.

Phantominumm:
SonarQube is currently running locally on my machine.
Regarding the report, it's a bit tricky. I can export the project. However, importing the project requires an upgraded version of SonarQube:

**_Projects cannot be imported. This feature is only available starting from Enterprise Edition._**

Another solution would be to manually report it myself

fabcor-maxiv:
> SonarQube is currently running locally on my machine. Regarding the report, it's a bit tricky. I can export the project. However, importing the project requires an upgraded version of SonarQube:
> 
> **_Projects cannot be imported. This feature is only available starting from Enterprise Edition._**
> 
> Another solution would be to manually report it myself

From my side, it's fine, I can do without the full Sonarqube report. An explicit mention in the pull request description that the change is triggered by a suggestion from Sonarqube with the link to the specific Sonarqube rule(s) would be much appreciated, though.

marcus-oscarsson:
The confusion is probably my fault. I should have mentioned earlier that @Phantominumm is a summer intern that is working with us to make security related fixes to the code base. Part of that project is to run SonarQube and Bandit to identify security related issues, categorize and fix these. 

The idea was to mention this during our next monthly developers meeting but its true that its a bit too long until then. Well now introductions have been made ;) and you know where these fixes come from.

For those that don't already know SonarQube is an open-source tool that can be used for static code analysis that is often used to scan code for vulnerabilities. It can be integrated in the CI and it should give the same report once configured the same. So you can also setup an instance of this server and get the same report.

@rhfogh We have thought about taking a closer/deeper look at the exception handling as well and do the sort of investigation that you mention above. We are already using `raise from` in many places so as far as I can see, in this particular case, the exception would not get "swallowed". You would get the full chain of exceptions. Anyhow I agree that we can improve on this.

rhfogh:
@marcus-oscarsson @Phantominumm "_We are already using raise from in many places so as far as I can see, in this particular case, the exception would not get "swallowed"._" Ooops! I sure missed that one. 
Well, it was obviously unreasonable of me to suggest that a newcomer with a time-limited task should undertake some kind of deep refactoring. My apologies.

marcus-oscarsson:
No problem at all, it's a great suggestion and we have also thought about it. Actually @Phantominumm is working really fast and we had the good luck to find someone that was able to dive into the code right away. So one never knows what we will tackle next, but the list is long and there are also other things to look at :)

fabcor-maxiv:
Should we get Sonarqube running in our GitHub Actions workflows? As a **non-blocking** job! Maybe that would be helpful if we could all see what advice Sonarqube has to offer. From a very quick look it seems possible; it would require some credentials though.

* <https://github.com/SonarSource/sonarqube-scan-action>

PR #1692, opened 2025-06-17
Ml fix redundant assignment of variables

Phantominumm:
Did a bit of cleanup:

  - Removed variable assignments where the value was overwritten with the same thing.

  -  In the MachInfo file, also removed an unnecessary if statement that was redundant because of the for loop.

### DIFF
--- a/ui/src/components/Equipment/SampleChanger.jsx
+++ b/ui/src/components/Equipment/SampleChanger.jsx
@@ -214,8 +214,8 @@ export default function SampleChanger(props) {
         >
           <i className="fas fa-qrcode" /> Scan all containers
         </Button>
-        {props.state === 'MOVING' && (
-          <span style={{ marginLeft: '1em' }}>
+        <span style={{ marginLeft: '1em' }}>
+          {props.state === 'MOVING' && (
             <Button
               variant="danger"
               className={styles.abortButton}
@@ -223,8 +223,8 @@ export default function SampleChanger(props) {
             >
               <i className="fas fa-stop" /> Abort
             </Button>
-          </span>
-        )}
+          )}
+        </span>
         {current}
         <div style={{ marginBottom: '1em' }} />
         <div style={{ maxHeight: '60vh', overflowY: 'auto' }}>{nodes}</div>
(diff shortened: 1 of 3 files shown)


ui/src/components/Equipment/SampleChanger.jsx
```diff
@@ -225,7 +214,17 @@ export default function SampleChanger(props) {
         >
           <i className="fas fa-qrcode" /> Scan all containers
         </Button>
-        <span style={{ marginLeft: '1em' }}>{abortButton}</span>
```
axelboc, line 228: Now that you render the `span` conditionally, when there's no abort button, there's no longer an extra `1em` space. To be checked if this space was important or not.

ui/src/components/MachInfo/MachInfo.jsx
```diff
@@ -16,30 +16,26 @@ function MachInfo(props) {
 
   if (info.attention === true) {
     variant = 'danger';
-  } else {
-    variant = 'info';
   }
 
   for (propName in info) {
```
axelboc, line 21: This loop doesn't make any sense, right? It's effectively overriding the same variables on every iteration, so only the last iteration (i.e. the last `propName in info`) really matters. :thinking:  Anyway, not the point of this PR. Your change takes us a step forward in refactoring this loop later on, which is great. :grin:

axelboc (approved):
Thanks!

ui/src/components/Equipment/SampleChanger.jsx
Phantominumm: Ah, yes. Didn’t think about the spacing the <span> was adding before.
To preserve that layout,i can move the conditional inside the <span> so the spacing remains consistent, regardless of whether the Abort button is rendered.

 # <i className="fas fa-qrcode" /> 
        </Button>
        <span style={{ marginLeft: '1em' }}>
          {props.state === 'MOVING' && (
            <Button
              variant="danger"
              className={styles.abortButton}
              onClick={abort}
            >
              <i className="fas fa-stop" /> Abort
            </Button>

ui/src/components/MachInfo/MachInfo.jsx
walesch-yan: Actually, in each iteration, the new `popContent` includes its previous value inside the span, so it should in the end be a list of all generated `msg`. However, I agree that it could definitely be refactored and simplified. And you are right should be done in a later PR.

axelboc: Oh wow I looked at it 3 times and didn't even see it :joy:

axelboc (approved):

PR #1694, opened 2025-06-17
Improved path handling for session datbase and adhoc certificates (Orginally Issue_1664 Session_DB_PATH)

marcus-oscarsson:
I struggled a bit with the rebase of the original PR, https://github.com/mxcube/mxcubeweb/pull/1668. Here is the squashed  and rebased version :). Sorry for the mess !

### DIFF
--- a/mxcubeweb/__init__.py
+++ b/mxcubeweb/__init__.py
@@ -2,6 +2,8 @@ from gevent import monkey
 
 monkey.patch_all(thread=False)
 
+from mxcubeweb.core.models.configmodels import FlaskConfigModel  # noqa: I001 E402
+
 # Disabling E402 (module level import not at top of file)
 # for the lines below as we are monkey patching
 import argparse  # noqa: E402
@@ -124,9 +126,19 @@ def build_server_and_config(test=False, argv=None):
         config_path = HWR.get_hardware_repository().find_in_repository("mxcube-web")
 
         cfg = Config(config_path)
-
+        db_path = (
+            Path(cfg.flask.USER_DB_PATH)
+            if cfg.flask.USER_DB_PATH
+            else Path(FlaskConfigModel.USER_DB_PATH.default)
+        )
+        db_path.parent.mkdir(parents=True, exist_ok=True, mode=0o600)
         if test:
-            cfg.flask.USER_DB_PATH = "/tmp/mxcube-test-user.db"
+            test_db = db_path.parent / "mxcube-test-user.db"
+            cfg.flask.USER_DB_PATH = str(test_db)
+
+            # Clean up existing test database if it exists
+            if test_db.exists():
+                test_db.unlink()
 
         server.init(cmdline_options, cfg)
         mxcube.init(
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -4,6 +4,10 @@ from gevent import monkey
 
 monkey.patch_all(thread=False)
 
+
+from pathlib import Path  # noqa: I001
+
+
 import atexit
 import copy
 import json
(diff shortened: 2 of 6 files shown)


marcus-oscarsson:
@Phantominumm does it look ok  ? :)

Phantominumm:
seems good, i don't see any problem here

mxcubeweb/__init__.py
```diff
@@ -124,9 +128,25 @@ def build_server_and_config(test=False, argv=None):
...
+                cfg.flask.USER_DB_PATH = str(test_db)
+                cfg.test_db_path = str(test_db)
+
+                # Clean up existing test database if it exists
+                if test_db.exists():
+                    test_db.unlink()
+            except AttributeError as e:
+                msg = f"Failed to set test database path: {e}"
+                logging.getLogger(msg)
+                raise
```
fabcor-maxiv: This does not log anything

```suggestion
            except AttributeError as e:
                msg = "Failed to set test database path"
                logging.getLogger(SOME_LOGGER_NAME).exception(msg)
                raise
```

fabcor-maxiv (changes requested):

mxcubeweb/__init__.py
fabcor-maxiv: Also, I have no idea what `except AttributeError` is supposed to catch.

Phantominumm: That was an older change that I kept and forgot to modify. We can change it to something like FileNotFoundError or/and OSError/PermissionError (subclass) in this case.

marcus-oscarsson: Thanks guys. I made some simplifications after a quick chat here :)

mxcubeweb/__init__.py
```diff
@@ -1,5 +1,7 @@
 from gevent import monkey
 
+from mxcubeweb.core.models.configmodels import FlaskConfigModel
+
```
fabcor-maxiv: As fas as I know `from gevent import monkey; monkey.patch_all(...)` should be done as early as possible before anything else including imports. So this import should be moved after the Gevent monkey patching.

* <https://www.gevent.org/api/gevent.monkey.html>

test/conftest.py
```diff
@@ -1,5 +1,7 @@
 """Helper functions for pytest"""
 
+from pathlib import Path
+
```
fabcor-maxiv: Move after the Gevent monkey-patching.

* <https://www.gevent.org/api/gevent.monkey.html>

fabcor-maxiv (changes requested):

mxcubeweb/__init__.py
marcus-oscarsson: Oh, yes I guess that stems from our dear `isort` :)

mxcubeweb/__init__.py
```diff
@@ -2,6 +2,8 @@
 
 monkey.patch_all(thread=False)
 
+from mxcubeweb.core.models.configmodels import FlaskConfigModel  # noqa: I001 E402
+
```
fabcor-maxiv, line 6: Better now, after the monkey patching, but still it should be with the rest of the `from mxcubeweb` imports.

test/conftest.py
```diff
@@ -4,6 +4,10 @@
 
 monkey.patch_all(thread=False)
 
+
+from pathlib import Path  # noqa: I001
```
fabcor-maxiv, line 8: Same here. Better now, after the Gevent monkey patching, but would be even better with the rest of standard library imports. I think in this case it should be right after `import sys` (no empty line).

fabcor-maxiv (approved):

PR #1695, opened 2025-06-18
Add permissions to workflow

Phantominumm:
Adding security permissions to our workflow file. This makes sure each job has exactly what it needs

https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions

I think all the necessary permissions are given but if you think I missed one, tell me.

### DIFF
--- a/.github/workflows/ui.yaml
+++ b/.github/workflows/ui.yaml
@@ -10,6 +10,11 @@ name: UI
   pull_request: null
   push: null
 
+permissions:
+  contents: read    # Required for checkout
+  actions: read     # for cache
+  packages: read    # for package access
+
 jobs:
   format:
     name: JS-Format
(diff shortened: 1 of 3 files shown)


fabcor-maxiv:
I had a quick investigation into this as well. I quickly tried [*zizmor*](http://docs.zizmor.sh/) for this, it gave good results, I can recommend it.

fabcor-maxiv (approved):
It is a good idea to look into this.

I think it might be fine as it is, but if you (or anyone) feels like bringing this further, I recommend checking what [zizmor](https://docs.zizmor.sh/) has to say about this stuff, and for example one low-hanging fruit would probably be to move the permissions to the job level instead of the workflow level.

.github/workflows/ui.yaml
```diff
@@ -84,6 +89,9 @@ jobs:
     name: E2E-Test
     runs-on: ubuntu-latest
 
+    permissions:
+      id-token: write       # for artifact upload
```
marcus-oscarsson, line 93: Very nice, thanks. Also good with the comments so we know why its needed !

PR #1696, opened 2025-06-18
fix Dependabot failed to update your dependencies poetry version

Phantominumm:
There are two vulnerability alerts in Dependabot with the following error:

**Dependabot encountered an unknown error:
Dependabot failed to update your dependencies because an unexpected error occurred. See the logs for more details.**

https://github.com/mxcube/mxcubeweb/security/dependabot/151

https://github.com/mxcube/mxcubeweb/security/dependabot/152

The issue was caused by a mismatch between the Poetry version installed in the environment (2.1.1) and the version specified in the workflow (2.1.3).
To resolve this, I added the Poetry version explicitly in pages.yaml to ensure consistency across environments and allow Dependabot to proceed with the updates.

### DIFF
--- a/.github/workflows/pages.yaml
+++ b/.github/workflows/pages.yaml
@@ -47,6 +47,14 @@ jobs:
         with:
           version: "2.1.3"
 
+      - name: "Cache Poetry dependencies"
+        uses: actions/cache@v3
+        with:
+          path: ~/.cache/pypoetry
+          key: ${{ runner.os }}-poetry-${{ hashFiles('**/poetry.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-poetry-
+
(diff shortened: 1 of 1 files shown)


fabcor-maxiv:
I am not following what the problem is. We do install [Poetry version `2.1.3`](https://github.com/mxcube/mxcubeweb/blob/50235a156cd372cb4dc3517f2356a1c1e473c163/conda-environment.yml#L33).

Phantominumm:
Dependabot appears to be using its own internal Poetry version when trying to resolve dependencies. With more research found this is actually a bug from Dependabot itself and not with the configuration.
I will close this PR as it's useless to change anything

fabcor-maxiv:
OK, thanks :)

PR #1697, opened 2025-06-18
SelectProposal: Add a way to show/hide tabs

mockoocy:
At MAX IV we don't distinguish between different types of sessions (scheduled, unscheduled, other beamline session).
Our sessions are created "on the fly" and are not tied to a particular beam time, thus it's not necessary to split them.

This commit makes it possible to opt-out of displaying some of the tabs via `ui.yaml` config file.

### DIFF
--- a/docs/source/config/ui.rst
+++ b/docs/source/config/ui.rst
@@ -9,6 +9,7 @@ The ``ui.yaml`` contains following sections:
 * sample_view_video_controls_
 * beamline_setup_
 * camera_setup_
+* session_picker_
 
 Each section groups settings for related UI features.
 See below an `example  ui.yaml <ui_yaml_example_>`_ file.
@@ -273,3 +274,25 @@ Below is an example of a UI configuration file ``ui.yaml``, as used by the mocku
 
 .. literalinclude:: ../../../demo/mxcube-web/ui.yaml
    :language: yaml
+
+.. _session_picker:
+
+Configures the "Select session" dialog.
+It controls which of the tabs are shown in the dialog.
+Possible tabs are:
+* "Active"
+* "Scheduled"
+* "Other beamlines"
+* "All"
+
+By default all tabs, except the last one, are shown.
+The section has the following syntax:
+.. code-block:: yaml
+
+    session_picker:
+      id: session_picker
+      tabs:
+        active: <show>
+        scheduled: <show>
+        other_beamlines: <show>
+        all_sessions: <show>
--- a/mxcubeweb/core/models/configmodels.py
+++ b/mxcubeweb/core/models/configmodels.py
@@ -108,11 +108,24 @@ class UISampleViewVideoControlsModel(UIPropertiesModel):
     ]
 
 
+class UiSessionPickerTabs(BaseModel):
+    active: bool = True
+    scheduled: bool = True
+    other_beamlines: bool = True
+    all_sessions: bool = False
+
+
+class UISessionPickerModel(BaseModel):
+    id: Literal["session_picker"] = "session_picker"
+    tabs: UiSessionPickerTabs = UiSessionPickerTabs()
+
+
 class UIPropertiesListModel(BaseModel):
     sample_view: UIPropertiesModel
     beamline_setup: UIPropertiesModel
     camera_setup: UICameraConfigModel | None
     sample_view_video_controls: UISampleViewVideoControlsModel | None
+    session_picker: UISessionPickerModel = UISessionPickerModel()
 
 
 class UserManagerUserConfigModel(BaseModel):
--- a/ui/src/components/LoginForm/SelectProposal.jsx
+++ b/ui/src/components/LoginForm/SelectProposal.jsx
@@ -11,6 +11,12 @@ import ActionButton from './ActionButton';
 import SessionTable from './SessionTable';
 import styles from './SessionTable.module.css';
 
+/**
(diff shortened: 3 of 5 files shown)


walesch-yan (approved):
Looks like a good addition to me, thanks @mockoocy

mxcubeweb/core/models/configmodels.py
```diff
@@ -104,11 +104,31 @@ class UISampleViewVideoControlsModel(UIPropertiesModel):
     ]
 
 
+class UISessionPickerTabModel(BaseModel):
+    id: Literal[
+        "active_tab", "scheduled_tab", "other_beamlines_tab", "all_sessions_tab"
```
fabcor-maxiv: Maybe this for consistency, also this would match what is in the documentation:

```suggestion
        "active", "scheduled", "other_beamlines", "all"
```

mxcubeweb/core/models/configmodels.py
```diff
@@ -104,11 +104,31 @@ class UISampleViewVideoControlsModel(UIPropertiesModel):
...
+    ]
+    show: bool = True
+
+
+class UISessionPickerModel(BaseModel):
+    id: Literal["session_picker"] = "session_picker"
+    components: list[UISessionPickerTabModel] = Field(
```
axelboc: Could this property be called `tabs` maybe?

ui/src/components/LoginForm/SelectProposal.jsx
```diff
@@ -27,6 +38,17 @@ function SelectProposal() {
...
+  const tabs =
+    useSelector((state) => state.uiproperties.session_picker.components) || [];
+
+  /**
+   * @type {Record<TabId, boolean>}
+   */
+  const showTab = Object.fromEntries(tabs.map((tab) => [tab.id, tab.show]));
+
+  const defaultActiveTab =
+    TAB_PRIORITY.find((tabId) => showTab[tabId]) ?? 'scheduled_tab';
```
axelboc:
```suggestion
    TAB_PRIORITY.find((tabId) => showTab[tabId]) ?? TAB_PRIORITY[0];
```

docs/source/config/ui.rst
```diff
@@ -273,3 +274,29 @@ Below is an example of a UI configuration file ``ui.yaml``, as used by the mocku
...
+      id: session_picker
+      components:
+        - id: active
+          show: <show>
+        - id: scheduled
+          show: <show>
+        - id: other_beamlines
+          show: <show>
+        - id: all
+          show: <show>
```
axelboc: I feel like this could be simplified to:

```yml
components: # (or tabs:)
  active: <bool>
  scheduled: <bool>
  ...
```

marcus-oscarsson:
Really nice work, Id just like to double check with @antolinos that this is not also overlapping with some of this changes ?

mxcubeweb/core/models/configmodels.py
mockoocy: Done, but I've kept name "all_sessions" instead of just "all" - the latter sounds like it were to just open all the possible tabs

docs/source/config/ui.rst
mockoocy: It indeed looks prettier.
I had the `components` section there, because it was required [there](https://github.com/mxcube/mxcubeweb/blob/2eaceacd981fe5b5e9a9a0e387884ad738f3b0e9/mxcubeweb/app.py#L308)

But I've changed this line since.

antolinos:
Dear @mockoocy 

At the ESRF, after several months (almost a year) using the tab system that organizes sessions based on when and where they were scheduled, the feedback we’ve received is perfectly aligned with your PR — we wanted to move away from using tabs and simply display a list of sessions instead.

Basically, we want to have a table with the available sessions and, on top of the table, an easy way to filter the sessions based on different factors such as dates and beamline.

If you're creating sessions on the fly, I don't think this would be useful for you, but I'm wondering if you think it would be compatible with your developments?

mockoocy:
> If you're creating sessions on the fly, I don't think this would be useful for you, but I'm wondering if you think it would be compatible with your developments?

I believe it would be.
This PR would allow you to just render the tab containing all sessions. The config would look something like that:
```yaml
# ...
session_picker:
  id: session_picker
  tabs:
    active: false
    scheduled: false
    other_beamlines: false
    all_sessions: true
```
The existing filters are still applied, adding additional ones should still work :)

image

marcus-oscarsson:
Thanks guys, well done. Looks good !

docs/source/config/ui.rst
```diff
@@ -273,3 +274,25 @@ Below is an example of a UI configuration file ``ui.yaml``, as used by the mocku
 
 .. literalinclude:: ../../../demo/mxcube-web/ui.yaml
    :language: yaml
+
+.. _session_picker:
+
+Configures the "Select session" dialog.
```
marcus-oscarsson, line 280: Nice, thanks for the documentation

ui/src/components/LoginForm/SelectProposal.jsx
```diff
@@ -11,6 +11,12 @@ import ActionButton from './ActionButton';
 import SessionTable from './SessionTable';
 import styles from './SessionTable.module.css';
 
+/**
+ * @typedef {"active"|"scheduled"|"other_beamlines"|"all_sessions"} TabId
```
marcus-oscarsson, line 15: Nice

PR #1699, opened 2025-06-18
Make phase control optional

alessandro-olivo:
These changes make the "Phase control" optional, that is, you can remove the "Phase control" section from the frontend. This was requested by the beamline scientists (they don't use the button so they prefer not to display it at all).
<p align="center">
  
  &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
  
</p>

## HowTo
In order to disable the "Phase control" you have to edit the `beamline_config.yml` like this:
```diff
configuration:
  # Non-object attributes:
  advanced_methods:
    ...
  ...
  use_native_mesh: true
  enable_2d_points: true
+ enable_phase_control: false
 . ..
  ```

The `demo/beamline_config.yml` has been changed as well to provide a concrete example

> ⚠️ Make sure to merge the corresponding branch (see this PR https://github.com/mxcube/mxcubecore/pull/1291) in `mxcubecore` first.

### DIFF
--- a/demo/beamline_config.yml
+++ b/demo/beamline_config.yml
@@ -82,6 +82,7 @@ configuration:
   mesh_result_format: PNG
   use_native_mesh: true
   enable_2d_points: true
+  enable_phase_control: true
 
   default_acquisition_parameters:
     default:
--- a/mxcubeweb/core/models/generic.py
+++ b/mxcubeweb/core/models/generic.py
@@ -25,12 +25,18 @@ class AppSettingsModel(BaseModel):
             "be dis-activated to not clash with i.e workflow mesh,"
         ),
     )
-
     enable_2d_points: bool = Field(
         True,
         description=(
             " Enable features to work with points in the plane, called2D-points,"
             " (none centred positions)"
+        ),
+    )
+    enable_phase_control: bool = Field(
+        True,
+        description=(
+            "Allow to hide or show the PhaseControl object in the frontend,"
+            " True by default"
         ),
     )
 
--- a/mxcubeweb/routes/main.py
+++ b/mxcubeweb/routes/main.py
@@ -50,6 +50,7 @@ def init_route(app, server, url_prefix):
                 "mesh_result_format": _blc.mesh_result_format,
                 "use_native_mesh": _blc.use_native_mesh,
                 "enable_2d_points": _blc.enable_2d_points,
+                "enable_phase_control": _blc.enable_phase_control,
             }
         )
 
--- a/ui/src/containers/SampleViewContainer.jsx
+++ b/ui/src/containers/SampleViewContainer.jsx
@@ -117,7 +117,10 @@ class SampleViewContainer extends Component {
             <DefaultErrorBoundary>
               {this.props.enablePhaseControl && (
                 <div className={motorInputStyles.container}>
-                  <label className={motorInputStyles.label} htmlFor="PhaseInput">
+                  <label
+                    className={motorInputStyles.label}
+                    htmlFor="PhaseInput"
+                  >
                     Phase Control
                   </label>
                   <PhaseInput />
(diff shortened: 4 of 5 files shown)

### CODE REVIEW

**Comment by marcus-oscarsson** [comment]
Nice one :)

**Comment by marcus-oscarsson** [comment]
How are you using `enable_phase_control` is it only for display only or is it also for the collection logic somehow ?. We already have all of the display logic in the `ui.yaml`. So that would probably a more logical place for this option in that case ?

**Comment by axelboc** [comment]
Nice indeed. If the phase control field is to remain visible by default, may I suggest inverting the configuration option to `disable_phase_control` (or maybe `hide_phase_control`). Otherwise, if `enable_phase_control` is not defined, one might think that it is `false`.

**Comment by alessandro-olivo** [comment]
> How are you using `enable_phase_control` is it only for display only or is it also for the collection logic somehow ?. We already have all of the display logic in the `ui.yaml`. So that would probably a more logical place for this option in that case ?

Actually, it is for display only, so I can see the point of moving it to `ui.yaml`. I'll give it a try :)


> Nice indeed. If the phase control field is to remain visible by default, may I suggest inverting the configuration option to `disable_phase_control` (or maybe `hide_phase_control`). Otherwise, if `enable_phase_control` is not defined, one might think that it is `false`.

Another good point!


I'll be back with these changes soon. Right now, I'm trying to rebase two other branches :)

**Comment by alessandro-olivo** [comment]
I started moving this into `ui.yml`, but since "Phase Control" isn't already there, a few questions/thoughts came up...

1) Should "Phase Control" be added to the `sample_view.component` list or to the `beamline_setup.components` list ?

>  If we go with `sample_view.component` , we’d also need to update the UI logic that uses index-based rendering to show the motors: https://github.com/mxcube/mxcubeweb/blob/2bcea2b0ef7a0ecb48c7d21c3729fb0ab523894e/ui/src/components/SampleView/MotorControls.jsx#L32  (maybe we can use roles instead?)

2) Should we introduce a `show` attribute or  stick with the current behavior, that is, what is defined in the `ui.yml` is shown in the UI.

3) If we add this new "Phase Control" item, all the partners will need to update their `ui.yml`

**Comment by marcus-oscarsson** [comment]
Very good questions :). I guess it would be logical to stick with the idea that whats listed in the `ui.yaml` file is what gets rendered. 

We could perhaps move all the motors to a section called `sample_view_motors` just like we have `sample_view_video_controls`. We can then still have  `sample_view` section that only has i.e:

```
  components:
    - id: phase_control
      show: true
    - id: beam_size
      show: true
```

Like this we don't need to change the motor render logic too much. and we stay with the same philosophy. What do you think ? I'm very much open to other ideas as well :), @walesch-yan, @beteva, @axelboc

beteva:
> 1. Should "Phase Control" be added to the `sample_view.component` list or to the `beamline_setup.components` list ?

I would say sample_view is what would be the easiest. Definitely not in the beamline_setup, as it will be miss leading. The best is, though, as @marcus-oscarsson suggests, is to go for a new setup item for the motors

> 2\. Should we introduce a `show` attribute or  stick with the current behaviour, that is, what is defined in the `ui.yml` is shown in the UI.
I would vote for whatever is configured, is shown. An extra logical level (show/noshow) would only make things more complicated, with no real gain. I find it awkward to configure something, but decide not to show it on the GUI. What would it be used for?

alessandro-olivo:
I would vote "for whatever is configured, is shown" too and I like what @marcus-oscarsson proposed :)

axelboc:
> We could perhaps move all the motors to a section called sample_view_motors just like we have sample_view_video_controls.

Seems legit.

I was going to go off on a tangent about `ui.yaml` but I think it's best to keep the discussion going in a dedicated issue.

marcus-oscarsson:
:) @axelboc. We can do a small review and cleanup of that at some point as well.

alessandro-olivo:
So, mind if I give a try with the @marcus-oscarsson's proposal?

I like your proposal too @axelboc (i.e. https://github.com/mxcube/mxcubeweb/issues/1725) but for this PR I'd use the current `ui.yml` structure, hope that’s okay.. :)

alessandro-olivo:
> So, mind if I give a try with the @marcus-oscarsson's proposal?

I gave it a try :) https://github.com/mxcube/mxcubeweb/pull/1740

PR #1703, opened 2025-06-19
Sample List  [ Part 1 ]

jbflo:
While working on the previous PR related to the Sample List, I noticed a few small improvements that are worth addressing here first, before we proceed with implementing the main solution:

✅ Removed getSamplesFromSC call from handleSyncSamples
The backend already handles fetching samples from the Sample Changer, so the extra call is redundant.

✅ Temporarily removed getSamplesFromSC
Since it's currently unused, it was removed to satisfy the linter. We’ll reintroduce it later when we implement the use_get_all_samples option.

✅ Disable "Clear Samples" button when the sample list is empty
Prevents unnecessary actions and improves UI clarity.

✅ Added tooltip to the Sample List view mode selector

### DIFF
--- a/ui/src/containers/SampleListViewContainer.jsx
+++ b/ui/src/containers/SampleListViewContainer.jsx
@@ -26,7 +26,6 @@ import {
 import { showConfirmCollectDialog } from '../actions/queueGUI';
 import {
   filterAction,
-  getSamplesList,
   selectSamplesAction,
   setViewModeAction,
   showGenericContextMenu,
@@ -214,32 +213,12 @@ export default function SampleListViewContainer() {
     }
   }
 
-  async function getSamplesFromSC() {
-    const manualSamples = [];
-    Object.values(sampleList).forEach((sample) => {
-      if (sample.location === 'Manual') {
-        manualSamples.push(sample.sampleID);
-      }
-    });
-    // first need to remove manual sample from queue
-    // because they will be remove from sample List
-    await dispatch(setEnabledSample(manualSamples, false));
-
-    dispatch(getSamplesList());
-  }
-
   /**
    * Synchronises samples with LIMS
-   *
-   * @property {Object} loginData
+   * @param {string} lims - LIMS name to synchronise with
    */
-  async function handleSyncSamples(lims) {
-    if (Object.keys(sampleList).length === 0) {
-      await getSamplesFromSC();
-      dispatch(syncSamples(lims));
-    } else {
-      dispatch(syncSamples(lims));
-    }
+  function handleSyncSamples(lims) {
+    dispatch(syncSamples(lims));
     dispatch(filterAction({ limsSamples: true }));
   }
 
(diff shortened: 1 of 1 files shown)


ui/src/containers/SampleListViewContainer.jsx
```diff
@@ -214,32 +213,12 @@ export default function SampleListViewContainer() {
...
-
   /**
    * Synchronises samples with LIMS
-   *
-   * @property {Object} loginData
+   * @param {string} lims - LIMS name to synchronise with
    */
-  async function handleSyncSamples(lims) {
-    if (Object.keys(sampleList).length === 0) {
-      await getSamplesFromSC();
```
jbflo, line 238: D'ont need to call getSamplesFromSC
it's redundant see  in https://github.com/mxcube/mxcubeweb/blob/2eaceacd981fe5b5e9a9a0e387884ad738f3b0e9/mxcubeweb/core/components/lims.py#L278

ui/src/containers/SampleListViewContainer.jsx
```diff
@@ -214,32 +213,12 @@ export default function SampleListViewContainer() {
     }
   }
 
-  async function getSamplesFromSC() {
```
rhfogh, line 217: If this will be needed later, would it not be better to leave it in and put in an 'ignore' directive for the linter?

ui/src/containers/SampleListViewContainer.jsx
```diff
@@ -789,23 +768,28 @@ export default function SampleListViewContainer() {
...
                 </Button>
               </TooltipTrigger>
               <span style={{ marginLeft: '1.5em' }} />
               <Dropdown>
-                <Dropdown.Toggle
-                  variant="outline-secondary"
-                  id="dropdown-basic"
+                <TooltipTrigger
+                  id="sync-samples-tooltip"
+                  tooltipContent="Change Samples List View Mode  from Graphical to Table and vice versa"
```
axelboc:
```suggestion
                  tooltipContent="Change Samples List View Mode from Graphical to Table and vice versa"
```

ui/src/containers/SampleListViewContainer.jsx
```diff
@@ -214,32 +213,12 @@ export default function SampleListViewContainer() {
     }
   }
 
-  async function getSamplesFromSC() {
-    const manualSamples = [];
-    Object.values(sampleList).forEach((sample) => {
-      if (sample.location === 'Manual') {
```
jbflo, line 220: note : Manual sample are automatically delete , also redundant
and as getSamplesFromSC now being use in the PR, remove it for now.

axelboc (approved):
Brilliant :tada:

ui/src/containers/SampleListViewContainer.jsx
axelboc: Neh, that's what git is for :+1:

marcus-oscarsson:
Well done @jbflo

PR #1704, opened 2025-06-19
display alert when user_level_log is critical

epanepucci:
- this adds a mechanism for mxcubecore objects to enforce
  an error message being displayed to the user
- showErrorPanel(msg) is called if 
     => logging.getLogger("user_level_log").**critical**(msg) is issued
- currently no user_level_log in the code is issuing messages on the critical level so this should be safe 👀

### DIFF
--- a/ui/src/serverIO.js
+++ b/ui/src/serverIO.js
@@ -14,7 +14,7 @@ import {
   plotEnd,
   setActionState,
 } from './actions/beamlineActions';
-import { showConnectionLostDialog } from './actions/general';
+import { showConnectionLostDialog, showErrorPanel } from './actions/general';
 import {
   setHarvesterState,
   updateHarvesterContents,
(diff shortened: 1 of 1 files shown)


axelboc:
No problem with the code obviously, but just wondering if this is the right way to do this as opposed to adding a dedicated WebSocket message. Would all errors that need to be displayed to users be "critical" and therefore warrant writing a critical log? Might be good to have a concrete example of such an error.

marcus-oscarsson:
Interesting idea, I think that can work for the time being. Its a very convenient way to send this sort of messages to the user. We can also implement a custom level if we want to be more specific and avoid possible conflicts.

Anyhow, I guess message/log display also partly falls into the UX/UI working group. We have been thinking of moving the log "window" to the bottom of the browser window instead of having it below the queue. Or in a separate drawer style pane to the right, with some nice notification icons/system. Perhaps that would be worth to look into for the future, I don't think its something that would demand a very big development effort, but we need to integrate it into the overall layout re-work

I think this is a good solution meanwhile :)

axelboc:
A custom log level sounds wiser to me. Coupling the `critical` log level with showing an error to the user really seems problematic to me:
- If a developer doesn't know about this side effect, they could expose critical information to the user.
- Conversely, a developer could use this log level to show a minor error to the user, which would then make the logs less accurate.

PR #1707, opened 2025-06-20
Update setuptools to fix path traversal vulnerability

Phantominumm:
Updating setuptools to a newer version because the version used had a vulnerability that didn't check if the URL given is safe enough and someone can manipulate the path.

https://github.com/pypa/setuptools/security/advisories/GHSA-5rjg-fvgr-3xxf


walesch-yan (approved):

marcus-oscarsson:
Thanks

PR #1710, opened 2025-06-20
Custom manual centring method name

alessandro-olivo:
These changes allow the name of the manual centering procedure to be customized.
One of the beamlines using MXCuBE has a custom `2-click` procedure, so it was necessary to make the name displayed in the UI configurable (see screenshots).

<p float="left">
  
  
  
</p>

## HowTo
In order to change the name showed in the UI you have to edit the `mxcube-web/ui.yaml` like this:
```diff
sample_view_video_controls:
  id: sample_view_video_controls
  components:
    - id: snapshot
      ...
    ...
    - id: 3_click_centring
      show: true
+     label: 2-click
    ...
```
If the `label` key is not defined, the UI will show the `3-click` as default

### DIFF
--- a/mxcubeweb/core/components/sampleview.py
+++ b/mxcubeweb/core/components/sampleview.py
@@ -384,11 +384,11 @@ class SampleView(ComponentBase):
                     "Aborting current centring ..."
                 )
                 HWR.beamline.diffractometer.cancel_centring_method(reject=True)
-
-            logging.getLogger("user_level_log").info(
+            msg = (
                 "Centring using %s centring"
                 % HWR.beamline.diffractometer.MANUAL3CLICK_MODE.lower()
             )
+            logging.getLogger("user_level_log").info(msg)
 
             HWR.beamline.diffractometer.start_centring_method(
                 HWR.beamline.diffractometer.MANUAL3CLICK_MODE
--- a/mxcubeweb/core/components/sampleview.py
+++ b/mxcubeweb/core/components/sampleview.py
@@ -386,7 +386,8 @@ class SampleView(ComponentBase):
                 HWR.beamline.diffractometer.cancel_centring_method(reject=True)
 
             logging.getLogger("user_level_log").info(
-                "Centring using %s centring" % HWR.beamline.diffractometer.MANUAL3CLICK_MODE.lower()
+                "Centring using %s centring"
+                % HWR.beamline.diffractometer.MANUAL3CLICK_MODE.lower()
             )
 
             HWR.beamline.diffractometer.start_centring_method(
--- a/mxcubeweb/core/components/sampleview.py
+++ b/mxcubeweb/core/components/sampleview.py
@@ -385,7 +385,9 @@ class SampleView(ComponentBase):
                 )
                 HWR.beamline.diffractometer.cancel_centring_method(reject=True)
 
-            logging.getLogger("user_level_log").info("Centring using 3-click centring")
+            logging.getLogger("user_level_log").info(
+                "Centring using %s centring" % HWR.beamline.diffractometer.MANUAL3CLICK_MODE.lower()
+            )
 
             HWR.beamline.diffractometer.start_centring_method(
                 HWR.beamline.diffractometer.MANUAL3CLICK_MODE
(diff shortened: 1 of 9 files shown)


mxcubeweb/core/components/sampleview.py
```diff
@@ -384,8 +384,11 @@ def start_manual_centring(self):
                     "Aborting current centring ..."
                 )
                 HWR.beamline.diffractometer.cancel_centring_method(reject=True)
-
-            logging.getLogger("user_level_log").info("Centring using 3-click centring")
+            msg = (
+                "Centring using %s centring"
+                % HWR.beamline.diffractometer.MANUAL3CLICK_MODE.lower()
+            )
+            logging.getLogger("user_level_log").info(msg)
```
fabcor-maxiv, line 391: This is kind of the worst of all the logging solutions available to us :D

Also I do not see the point of the `lower()` call here.

My personal preference is something like this:

```suggestion
            msg = "Centring using %s centring"
            logging.getLogger("user_level_log").info(msg, HWR.beamline.diffractometer.MANUAL3CLICK_MODE)
```

But seems like the general consensus is to use f-strings:

```suggestion
            msg = f"Centring using {HWR.beamline.diffractometer.MANUAL3CLICK_MODE} centring"
            logging.getLogger("user_level_log").info(msg)
```

mxcubeweb/routes/signals.py
```diff
@@ -227,10 +227,13 @@ def sc_maintenance_update(*args):
 def centring_started(method, *args):
     msg = {"method": method}
 
-    if method in ["Computer automatic"]:
+    if method in [HWR.beamline.diffractometer.C3D_MODE]:
```
fabcor-maxiv, line 230: That line change seems unrelated, isn't it?

fabcor-maxiv (commented):
I wonder if maybe we wouldn't be better off changing all the labels to `N-click centering` and be done with this...

marcus-oscarsson:
> I wonder if maybe we wouldn't be better off changing all the labels to `N-click centering` and be done with this...

I think its nice to display how many clicks that is expected it tells the user what method thats being used. Perhaps just `Click centering` if we want to omit the number of clicks. However I think its nice to be a bit more informative when we can.

marcus-oscarsson:
Looks, good. You can perhaps do something with the log message if you like. It would be great if you could rebase and merge when you are done. If you have no other additions to make to this PR, that is.

fabcor-maxiv:
> > I wonder if maybe we wouldn't be better off changing all the labels to `N-click centering` and be done with this...
> 
> I think its nice to display how many clicks that is expected it tells the user what method thats being used. Perhaps just `Click centering` if we want to omit the number of clicks. However I think its nice to be a bit more informative when we can.

Sure, it's nice, but one always has to ask themselves "is it worth the added complexity?". In this case the UI shows the number of remaining clicks anyway. And yes, MXCuBE seems to know the number of clicks so wouldn't it be also possible to show this value in the label (`f"{N}-click centering"`) instead of a string label from the configuration files? I guess it is mostly food for thought, I do not have any opposition to merging this.

marcus-oscarsson:
Ok, you can rebase and merge this one and the one on mxcubecore (https://github.com/mxcube/mxcubecore/pull/1293#pullrequestreview-2961282251). 

Thanks alot !

marcus-oscarsson:
Just keep in mind that it would be great if you then could update `pyproject.toml` with the latest version of mxcubecore. Perhaps by merging the `mxcubecore` PR first and then adding a updated `pyproject.toml` to this PR.

mxcubeweb/core/components/sampleview.py
alessandro-olivo: I'm in the f-string team, but when I committed the changes, it seemed like ruff didn’t like them, turns out I was wrong (I just tried it again now).

I find f-strings more readable, but yeah, this:
```python
msg = "Centring using %s centring"
logging.getLogger("user_level_log").info(msg, HWR.beamline.diffractometer.MANUAL3CLICK_MODE)
```
can be a cleaner solution when you’re dealing with long variable names.

I guess I’ll just decide case by case, depending on the length :)

...about the `lower()` I just wanted to maintain the case

mxcubeweb/routes/signals.py
alessandro-olivo: Yes, you are right! I changed just to have the same "accordance" with the other if statement, but I can throw it away if it not make sense to do this change in this PR

alessandro-olivo:
> ... MXCuBE seems to know the number of clicks so wouldn't it be also possible to show this value in the label (`f"{N}-click centering"`) instead of a string label from the configuration files? I guess it is mostly food for thought, I do not have any opposition to merging this.


It seems a very good idea! I'll give a try

alessandro-olivo:
I've just opened another PR (https://github.com/mxcube/mxcubeweb/pull/1730) to implement the changes proposed by @fabcor-maxiv.

PR #1698, opened 2025-06-18
clean-up css styles of ConfirmCollectDialog

walesch-yan:
This is a first PR to try to refactor `ConfirmCollectDialog`. It focuses on the CSS part of the component.

It includes the following changes:
 - Adding a scrollbar to the table of samples to collect (from comments it seems that this functionality was there previously but it does not seem to exist on the current version)
 -  transform CSS file into module
 - replace corresponding `className` and (unnecessary ) `id` props
- Remove (dead) code regarding table resize

See UI differences below:

 Previously:
FixedTable2

Now: 
Scrollbar

PR #1711, opened 2025-06-20
split TaskTable from confirmCollectDialog

walesch-yan:
This PR continues the refactoring work of `ConfirmCollectDialog` (see #1698 ). It creates specific functional components for the table rendering methods and moves the logic into specific files. It creates components for the table itself and the designated popovers per task.

While at it, I fixed the following issues in the logic related to the table and popover rendering:
 - Popover logic existed, but popovers were not visible as they were rendered behind the modal
 - Popover would move when scrolling, while the visibility only changes once the user hovers over another element. Now, a component's internal state is used to fix this behavior.

See popover showing behavior below:
fixedPopover

### DIFF
--- a/ui/src/components/ConfirmCollectDialog/TaskTable.jsx
+++ b/ui/src/components/ConfirmCollectDialog/TaskTable.jsx
@@ -1,0 +1,87 @@
+import { useState } from 'react';
+import { OverlayTrigger, Popover, Table } from 'react-bootstrap';
+import { useSelector } from 'react-redux';
+
+import TaskOverlayTable from './TaskOverlayTable';
+import styles from './TaskTable.module.css';
+
+export default function TaskTable(props) {
+  const { tasks } = props;
+  const sampleList = useSelector((state) => state.sampleGrid.sampleList);
+  const rootPath = useSelector((state) => state.login.rootPath);
+  // We handle the show state of each Overlay internally to hide them when scrolling
+  const [popoverShown, setPopoverShown] = useState({});
+
+  function handleScroll() {
+    setPopoverShown({});
+  }
+
+  if (tasks.length === 0) {
+    return (
+      <div className={styles.emptyTable}>
+        No tasks added to any of the samples, you have the possibility to add
+        tasks while the queue is running. <br />
+        The queue is executed sample by sample and will wait until
+        <b> Mount Next Sample </b> is pressed before mounting the next sample{' '}
+        <br />
+      </div>
+    );
+  }
+
+  return (
+    <div className={styles.scroll} onScroll={handleScroll}>
+      <Table responsive bordered hover className={styles.tableStriped}>
+        <thead>
+          <tr>
+            <th>Type</th>
+            <th>Sample</th>
+            <th>Path</th>
+            <th># Images</th>
+          </tr>
+        </thead>
+        <tbody>
+          {tasks.map((task) => {
+            const parameters =
+              task.type === 'Interleaved'
+                ? task.parameters.wedges[0].parameters
+                : task.parameters;
+
+            const sample = sampleList[task.sampleID];
+            const sampleName = `${sample.sampleName} - ${sample.proteinAcronym}`;
+            const relativePath = parameters.fullPath.split(rootPath).pop();
+
+            return (
+              <OverlayTrigger
+                key={task.queueID}
+                placement="bottom"
+                show={!!popoverShown[task.queueID]}
+                onToggle={(show) =>
+                  setPopoverShown((prev) => ({
+                    ...prev,
+                    [task.queueID]: show,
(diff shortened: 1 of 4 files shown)


ui/src/components/ConfirmCollectDialog/TaskOverlayTrigger.jsx
```diff
@@ -0,0 +1,77 @@
...
+      break;
+    }
+  }
+  return [headers, elements];
+}
+
+export default function TaskOverlayTrigger(props) {
+  const { children, setShow, show, task } = props;
+
+  const [headers, elements] = getTaskDetails(task.type, task.parameters);
```
axelboc: Sorry, I think I liked the previous implementation in `taskPopover` better :sweat_smile: — it was more declarative. Roughly turning the `taskPopover` method into a component would probably suffice in my opinion. Here a full suggestion so you see what I mean:

```jsx
// In `TaskTable`: keep `OverlayTrigger` inlined, just extract the
// rendering of the task table in the overlay

return (
  <OverlayTrigger
    key={task.queueID}
    placement="bottom"
    show={!!popoverShown[task.queueID]}
    setShow={(show) =>
      setPopoverShown((prev) => ({
        ...prev,
        [task.queueID]: show,
      }))
    }
    overlay={
      <Popover className={styles.taskPopover}>
        <TaskOverlayTable task={task} />
      </Popover>
    }
  >
    {children}
  </OverlayTrigger>
);
```

```jsx
// In `TaskOverlayTable: keep a similar logic to `taskPopover but without
// the `Popover` wrapper (already in `TaskTable`) and with early returns

if (task.type === 'energy_scan') {
  return (
    <Table striped bordered hover>
      // ...
    </Table>
  );
}

if (task.type === 'xrf_spectrum') {
  return (
    <Table striped bordered hover>
      // ...
    </Table>
  );
} 

return (
  <Table striped bordered hover>
      // ...
  </Table>
);
```

ui/src/containers/ConfirmCollectDialog.jsx
```diff
@@ -133,159 +125,6 @@ export class ConfirmCollectDialog extends React.Component {
...
-   * Returns the markup for a table containing summary/details for each task
-   * in the queue
-   *
-   * @property {Object} sampleGrid
-   * @property {Object} queue
-   * @return {ReactDomNode} Table Markup
-   */
-  taskTable() {
-    const tasks = this.tasksToCollect();
-    const summary = this.collectionSummary();
```
axelboc, line 217: Haha so `collectionSummary` itself also calls `tasksToCollect` just to get the number of tasks... Nice clean-up :joy:

ui/src/components/ConfirmCollectDialog/TaskTable.jsx
```diff
@@ -0,0 +1,81 @@
...
+import styles from './TaskTable.module.css';
+
+export default function TaskTable(props) {
+  const { tasks } = props;
+  const sampleList = useSelector((state) => state.sampleGrid.sampleList);
+  const rootPath = useSelector((state) => state.login.rootPath);
+  const [popoverShown, setPopoverShown] = useState({});
+
+  function handleScroll() {
+    setPopoverShown({});
```
axelboc, line 16: Oh and good call on hiding the popover on scroll — it's always annoying to have a popover that moves away from the cursor and no longer reflects what's underneath it... :clap:

ui/src/containers/ConfirmCollectDialog.jsx
walesch-yan: Yes, this was indeed a very useless call here :sweat_smile:

ui/src/components/ConfirmCollectDialog/TaskTable.jsx
walesch-yan: I agree it was pretty annoying to me as well :smile:

ui/src/components/ConfirmCollectDialog/TaskOverlayTrigger.jsx
walesch-yan: I see that is indeed a quite clever way to do it. In fact, I moved the `OverlayTrigger` wrapper to `TaskOverlayTable` because otherwise, you would need to handle quite a few props exchanged between `OverlayTrigger` and `Popover`, but silly me didn't think about moving the `Popover` wrapper into the `TaskTable` part, which is definitely the better option. Thanks @axelboc :smiley:

walesch-yan: For the content of the `TaskOverlayTable` I am fine with keeping the multiple `if`s structure if it is more declarative :)

axelboc (approved):
:100:

PR #1716, opened 2025-06-24
Display error Messages on sample mount/unmount

walesch-yan:
This PR addresses two problems with the error message handling in the UI related to the `SampleChanger`:

- The Error Panel displayed never showed the error message. Fixed it to show the generic message that is returned upon error
       before: 
       Screenshot from 2025-06-24 15-46-48
       after: 
       Screenshot from 2025-06-24 15-45-08

- On unmount error status code 200 was returned, changed it to be 409

### DIFF
--- a/mxcubeweb/routes/samplechanger.py
+++ b/mxcubeweb/routes/samplechanger.py
@@ -64,35 +64,20 @@ def init_route(app, server, url_prefix):  # noqa: C901
     @server.restrict
     def unmount_current():
         try:
-            res = app.sample_changer.unmount_current()
+            return jsonify(app.sample_changer.unmount_current()), 200
         except Exception:
             logging.getLogger("MX3.HWR").exception("Cannot unload sample")
-            res = (
-                "Cannot unload sample",
-                409,
-                {
-                    "Content-Type": "application/json",
-                },
-            )
-        return jsonify(res)
+            return jsonify(message="Cannot unload sample"), 409
 
     @bp.route("/mount", methods=["POST"])
     @server.require_control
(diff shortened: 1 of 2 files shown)


mxcubeweb/routes/samplechanger.py
```diff
@@ -74,7 +74,7 @@ def unmount_current():
                     "Content-Type": "application/json",
                 },
             )
-        return jsonify(res)
+        return res
```
walesch-yan: When an Exception occured res was only in the format of a response with status code 409, using jsonify moved the status code into a json body and the returned response got status code 200 by default

axelboc (approved):

mxcubeweb/routes/samplechanger.py
fabcor-maxiv: Reminder that we have this issue opened: <https://github.com/mxcube/mxcubeweb/issues/1393>. If anyone feels in a mood to tackle this. :)

walesch-yan: It's a good call, I changed it, so that the content of the error response is actually a json and not a plaintext with json header

walesch-yan:
After a few more tests, I converted this into a draft. There seems to be an error thrown every time we unmount the sample on develop (at least on the mockup). The issue seems to be that the sample is removed and the address cannot be accessed afterwards (see the traceback log below). The changes here would prevent the sample to be removed in the front-end (except if you reload the page), as the error would now be correctly catched and handled. 
Screenshot from 2025-06-25 11-15-31

PR #1717, opened 2025-06-24
Remove LIMS Filtering Functionality and return only LIMS samples 

jbflo:
This PR removes the filter related to LIMS samples.

What Changes: 

✅ (Commit 1)
The LIMS filter option is removed from the UI.

USE Case
The user clicks “Get Samples”, or Chooses to Synchronize with LIMS/ISPyB
- The assumption is that only relevant samples (for the current user/session) will be returned directly from the backend
- As a result, an explicit filter for LIMS data is no longer necessary in the frontend.

✅ (Commit 2)
Only samples relevant to the user/session are returned .
Any required filtering is handled server-side.

[Screencast from 2025-06-26 16-12-44.webm](https://github.com/user-attachments/assets/bd85d87e-05de-4357-a649-8b9e60aea39f)

### DIFF
--- a/mxcubeweb/core/components/lims.py
+++ b/mxcubeweb/core/components/lims.py
@@ -323,3 +323,32 @@ class Lims(ComponentBase):
                 self.sample_list_sync_sample(sample_info)
 
         return self.sample_list_get()
+
+    def get_lims_samples(self, lims_name: str) -> dict:
+        """
+        Gets samples from LIMS and filters to include only LIMS-linked entries.
+
+        This method synchronizes the sample list with the specified LIMS,
+        filters for entries that have a `limsID`, and sets the new sample list
+        accordingly.
+
+        Args: Name of the LIMS system to synchronize with.
+
+        Returns: The updated sample list with only LIMS samples.
+        """
+        self.synch_with_lims(lims_name)
+        new_sample_list = {"sampleList": {}, "sampleOrder": []}
+
+        try:
+            for loc, data in self.app.SAMPLE_LIST.get("sampleList", {}).items():
+                if data.get("limsID"):
+                    new_sample_list["sampleList"][loc] = data
+                    new_sample_list["sampleOrder"].append(loc)
+        except Exception:
+            logging.getLogger("MX3.HWR").exception(
+                "Error while filtering LIMS samples for '%s':", lims_name
+            )
+            return {"sampleList": {}, "sampleOrder": []}
+
+        self.sample_list_set(new_sample_list)
+        return self.app.SAMPLE_LIST
--- a/ui/src/api/lims.js
+++ b/ui/src/api/lims.js
@@ -3,7 +3,9 @@ import api from './api';
 const endpoint = api.url('/lims');
 
 export function fetchLimsSamples(lims) {
-  return endpoint.post({ lims }, '/synch_samples').safeJson();
+  return endpoint
+    .get(`/lims_samples?lims=${encodeURIComponent(lims)}`)
+    .safeJson();
 }
 
 export function fetchLimsResults(qid) {
--- a/mxcubeweb/routes/lims.py
+++ b/mxcubeweb/routes/lims.py
@@ -11,16 +11,16 @@ from flask import (
 def init_route(app, server, url_prefix):
     bp = Blueprint("lims", __name__, url_prefix=url_prefix)
 
-    @bp.route("/synch_samples", methods=["POST"])
+    @bp.route("/lims_samples", methods=["GET"])
     @server.restrict
-    def proposal_samples():
+    def get_proposal_samples():
         try:
-            lims_name = request.get_json().get("lims", None)
-            res = jsonify(app.lims.synch_with_lims(lims_name))
+            lims_name = request.args.get("lims")
+            res = jsonify(app.lims.get_lims_samples(lims_name))
         except Exception:
-            logging.getLogger("MX3.HWR").exception("Could not synchronize with Lims")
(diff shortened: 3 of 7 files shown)


dominikatrojanowska (approved):

ui/src/api/lims.js
```diff
@@ -3,6 +3,10 @@ import api from './api';
 const endpoint = api.url('/lims');
 
 export function fetchLimsSamples(lims) {
+  return endpoint.post({ lims }, '/get_lims_samples').safeJson();
+}
+
+export function fetchSyncLimsSamples(lims) {
```
axelboc: Can this one be removed?

ui/src/api/lims.js
```diff
@@ -3,6 +3,10 @@ import api from './api';
 const endpoint = api.url('/lims');
 
 export function fetchLimsSamples(lims) {
+  return endpoint.post({ lims }, '/get_lims_samples').safeJson();
```
axelboc: It's a `POST`, so the function should technically be called `sendGetLimsSamples` (looks like it was already wrong before, my bad :sweat_smile:).

That being said, I would prefer making it a `GET` request (with endpoint `/lims_samples`), in which case the name could remain `fetchLimsSamples`.

axelboc (commented):
Thanks for the review request, this one had slipped through, sorry!

ui/src/api/lims.js
jbflo: Got IT, Thanks !

mxcubeweb/core/components/lims.py
```diff
@@ -323,3 +323,30 @@ def synch_with_lims(self, lims_name):
                 self.sample_list_sync_sample(sample_info)
 
         return self.sample_list_get()
+
+    def get_lims_samples(self, lims_name: str) -> dict:
```
marcus-oscarsson, line 327: I think we can make the comment or doc string clearer but mentioning that this only returns LIMS samples and NOT any other samples.

marcus-oscarsson:
Looks good as far as I can see @jbflo. I think you can undraft it :)

marcus-oscarsson (approved):

jbflo:
> Looks good as far as I can see @jbflo. I think you can undraft it :)

thanks @marcus-oscarsson , i'll un-draft  Monday next week.

mxcubeweb/core/components/lims.py
```diff
@@ -323,3 +323,30 @@ def synch_with_lims(self, lims_name):
...
+        """
+        Get samples from LIMS.
+        This method synchronizes the sample list with LIMS and returns
+        the updated sample list.
+        :param lims_name: Name of the LIMS to synchronize with.
+        :return: Updated sample list.
+        :rtype: dict
```
fabcor-maxiv: Our contribution guidelines say to use Google-style docstrings (not Numpy-style).

mxcubeweb/core/components/lims.py
```diff
@@ -323,3 +323,30 @@ def synch_with_lims(self, lims_name):
...
+
+        try:
+            for loc, data in self.app.SAMPLE_LIST["sampleList"].items():
+                # If the sample has a limsID, it means it is a valid LIMS sample
+                # and should be added to the new only lims sample list
+                if data.get("limsID"):
+                    new_sample_list["sampleList"][loc] = data
+                    new_sample_list["sampleOrder"].append(loc)
+        except Exception as e:
+            logging.getLogger("MX3.HWR").info("Error while getting LIMS samples: %s", e)
```
fabcor-maxiv: Why not show the whole stack trace here?
```suggestion
            logging.getLogger("MX3.HWR").exception("Error while getting LIMS samples")
```
If the `try` block was one single statement, or if we were catching specific exceptions, maybe it would be acceptable like this. But here we have a whole bunch of code in the `try` block (including a loop!), and we catch all exceptions(!). So we have very little control over what could possibly be raised and caught. Having a stack trace seems the bare minimum to me.

mxcubeweb/core/components/lims.py
jbflo: thanks!

marcus-oscarsson: Great !

mxcubeweb/core/components/lims.py
marcus-oscarsson: Well spotted :)

mxcubeweb/core/components/lims.py
```diff
@@ -323,3 +323,34 @@ def synch_with_lims(self, lims_name):
...
+    def get_lims_samples(self, lims_name: str) -> dict:
+        """
+        Gets samples from LIMS and filters to include only LIMS-linked entries.
+
+        This method synchronizes the sample list with the specified LIMS,
+        filters for entries that have a `limsID`, and sets the new sample list
+        accordingly.
+
+        Args:
+            lims_name (str): Name of the LIMS system to synchronize with.
```
fabcor-maxiv:
```suggestion
            lims_name: Name of the LIMS system to synchronize with.
```

Skip the type in the docstring if there is already a type hint annotation. Sphinx documentation generator and all the other tools will know to do the right thing.

mxcubeweb/core/components/lims.py
```diff
@@ -323,3 +323,34 @@ def synch_with_lims(self, lims_name):
...
+
+        Returns:
+            dict: The updated sample list with only LIMS samples.
```
fabcor-maxiv:
```suggestion
            The updated sample list with only LIMS samples.
```

PR #1718, opened 2025-06-25
Fix typo in `core/components/sampleview.py`

alessandro-olivo:
Just a typo... :)

Let me know if the PR is the right tool for typos, I think a double check is usually better, but I don't want to waste your time :)

### DIFF
--- a/mxcubeweb/core/components/sampleview.py
+++ b/mxcubeweb/core/components/sampleview.py
@@ -394,9 +394,9 @@ class SampleView(ComponentBase):
             self.centring_reset_click_count()
         else:
             logging.getLogger("user_level_log").warning(
-                "Diffracomter is busy, cannot start centering"
+                "Diffractometer is busy, cannot start centering"
             )
-            msg = "Diffracomter is busy, cannot start centering"
+            msg = "Diffractometer is busy, cannot start centering"
             raise RuntimeError(msg)
 
         return {"clicksLeft": self.centring_clicks_left()}


fabcor-maxiv:
From my point of view, yes even typo fixes have to go through a pull request and review.

When I did my big "typo fixes" pull request on MXCuBE-Core (<https://github.com/mxcube/mxcubecore/pull/1279>), I paid extra care to avoid anything that was actual code, exception messages, and log messages.

My rationale was that maybe somehow someone somewhere relies on the typos in the exception messages and log messages. Maybe to do some `grep`-ing for example. Now I do not know if it is actually the case or not, maybe it is safe to fix those typos as well. I would like to fix those typos. I just did not want to run the risk of breaking this kind of things in this kind of big pull request.

alessandro-olivo:
I can see your point, definitely for code and exceptions, maybe less so for logs; but as you said, maybe somehow, somewhere, someone...
I don't have a clear answer, so maybe we should wait for other thoughts.

fabcor-maxiv:
On the other hand, every code change we make, every pull request we merge potentially breaks someone. This is no different.

So yes, such typo fixes should go through the normal pull request and review process. If no one objects, this gets merged. Same as everything else :)

fabcor-maxiv (approved):

rhfogh:
How about combining these fixes into large batches, or bundling them with other changes you make? I make it a policy to read all the notification mails I get, to make sure I am keeping track. If I am to get 3-4 notification messages every time someone wants to change 'teh' to 'the', I might have to delete notifications unseen.

alessandro-olivo:
Sure! I'll create something like `ao-fix-typos` branch and commit to it as I find typos. I'll try to keep the commits granular so it's easy to roll back anything that might break something.

marcus-oscarsson:
Thanks @alessandro-olivo

PR #1719, opened 2025-06-25
Bring back  get samples from SC and make it Optional

jbflo:
This PR reintroduces the “Get Samples from SC” option, allowing users to retrieve the full list of samples from the sample changer, independent of LIMS/ISPyB.

🖱️ UI Behavior
The UI now provides two buttons:

1 : Get Samples  -> Default option to Synchronizes with LIMS/ISPyB and Returns only LIMS-tracked samples
see [PR](https://github.com/mxcube/mxcubeweb/pull/1717)

2 : Get Samples from SC (" Menu Inside split Button")
   it Fetches all samples currently in the sample changer
   Useful when samples are not registered in LIMS
   ⚙️ Configurable Option
   This feature is optional : You can disable the second button by setting:  "use_get_all_samples"  show : false
   
"use_get_all_samples"  show : false
image

"use_get_all_samples"  show : true
image

### DIFF
--- a/mxcubeweb/core/models/configmodels.py
+++ b/mxcubeweb/core/models/configmodels.py
@@ -169,6 +169,12 @@ class MXCUBEAppConfigModel(BaseModel):
             "True to use video stream produced by external software, false otherwise"
         ),
     )
+    USE_GET_SAMPLES_FROM_SC: bool = Field(
+        True,
+        description=(
+            "True to acticvate or be able to get samples from the sample changer, false otherwise"
+        ),
+    )
     mode: ModeEnum = Field(
         ModeEnum.OSC, description="MXCuBE mode OSC, SSX-CHIP or SSX-INJECTOR"
     )
--- a/ui/src/containers/SampleListViewContainer.jsx
+++ b/ui/src/containers/SampleListViewContainer.jsx
@@ -1,6 +1,7 @@
 import { useCallback, useEffect } from 'react';
 import {
   Button,
+  ButtonGroup,
   Card,
   Col,
   Container,
@@ -27,6 +28,7 @@ import { showConfirmCollectDialog } from '../actions/queueGUI';
 import {
   filterAction,
   getLimsSamples,
+  getSamplesList,
   selectSamplesAction,
   setViewModeAction,
   showGenericContextMenu,
@@ -55,6 +57,10 @@ export default function SampleListViewContainer() {
   const sampleChanger = useSelector((state) => state.sampleChanger);
   const sampleChangerType = useSelector((state) =>
     state.sampleChanger.contents ? state.sampleChanger.contents.name : 'Mockup',
+  );
+
+  const showGetSamplesFromSC = useSelector(
+    (state) => state.general.useGetSamplesFromSC,
   );
 
   const handleSetViewMode = useCallback(
@@ -211,6 +217,13 @@ export default function SampleListViewContainer() {
         );
       }
     }
+  }
+
+  /**
+   * Fetches samples from the SampleChanger
+   */
+  function getSamplesFromSC() {
+    dispatch(getSamplesList());
   }
 
   /**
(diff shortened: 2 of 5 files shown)


mxcubeweb/core/models/configmodels.py
```diff
@@ -80,7 +80,7 @@ class _UICameraConfigModel(BaseModel):
     height: int | None
 
 
-class _UISampleViewVideoControlsModel(BaseModel):
```
jbflo: I change this name to be more generic, other class can use it .

ui/src/components/SampleGrid/utils.js
```diff
@@ -0,0 +1,9 @@
+import { useSelector } from 'react-redux';
+
+export function useComponent(name) {
+  return useSelector((state) =>
```
jbflo: Copied from the samples control util class. this can be useful for other component related to sample List.

demo/mxcube-web/ui.yaml
```diff
@@ -98,6 +98,12 @@ sample_view_video_controls:
     - id: video_size
       show: true
 
+sample_list:
+  id: sample_list
+  components:
+    - id: use_get_samples_from_sc
+      show: true
```
axelboc: Did we talk about making this a more global configuration option, since it will need to be used in the back-end (for the get samples endpoitn to return an empty list)?

ui/src/containers/SampleListViewContainer.jsx
```diff
@@ -526,43 +538,67 @@ export default function SampleListViewContainer() {
...
+                className="nowrap-style"
               >
-                {lims.name}
+                Get Samples from SC
               </Dropdown.Item>
             </TooltipTrigger>
-          ))}
+          )}
+          {loginData.limsName.length > 1 && <Dropdown.Divider />}
+          {loginData.limsName.map((lims, lindex) => {
```
axelboc:
```suggestion
          {loginData.limsName.slice(1).map((lims, lindex) => {
```

Otherwise you can split the LIMS beforehand:

```js
const firstLims = ...
const extraLims = ...
```

ui/src/containers/SampleListViewContainer.jsx
```diff
@@ -526,43 +538,67 @@ export default function SampleListViewContainer() {
...
+                >
+                  <Dropdown.Item
+                    variant="outline-secondary"
+                    className="nowrap-style"
+                    onClick={() => handleSyncSamples(lims.name)}
+                  >
+                    Get Samples & Sync {lims.name}
+                  </Dropdown.Item>
+                </TooltipTrigger>
+                {loginData.limsName.length < lindex && <Dropdown.Divider />}
```
axelboc: I don't think you need dividers, to be honest — just my opinion.

axelboc (approved):
Looks good :blush: — mostly just the question about where to put the config option.

demo/mxcube-web/ui.yaml
jbflo: >.(for the get samples endpoitn to return an empty list)?

😄 I may have forgotten part of what we discussed — or maybe I just found a simpler way to approach it!

If [PR #1717](https://github.com/mxcube/mxcubeweb/pull/1717) gets approved, I think this solution should be fine as a UI component config.

Let me know what you think!

marcus-oscarsson: We did indeed say that it could be a server side setting more as a feature than a UI setting. Like this the flag could be used also by hardware objects.

marcus-oscarsson: Ok, lets go with this solution for the moment, and then we might evolve it to be a "feature" setting instead of a UI option.

marcus-oscarsson:
@jbflo Could you have a look at the conflicts ?

demo/mxcube-web/ui.yaml
jbflo: oupss, i missed your comment
i made it global, accessible in general state. 

Now i think i remember the (for the get samples endpoitn to return an empty list)

PR #1720, opened 2025-06-25
[DRAFT ] Add Rate Limiter

Phantominumm:
Add RateLimitedBlueprint, a subclass from Blueprint that automatically enforces rate limits on all its endpoints. It tracks requests per user (or IP otherwise) and endpoint, to limit the numbers of requests.

The rate limiter is thread-safe, configurable (default: 60 requests/minute), and self-cleaning. 
For testing, a test_mode disables the limit so the tests work.

it's still pretty basic, to see what may be necessary to add or change.

**Simply replace Blueprint with RateLimitedBlueprint.**

bp = RateLimitedBlueprint('api', __ name__, 
rate_limit=60 _[nb of request]_, 
         rate_period=60 _[time in seconds]_)

@bp.route('/endpoint')
def endpoint():


mxcubeweb/core/server/ratelimited_blueprint.py
```diff
@@ -0,0 +1,85 @@
...
+        if self._testing:
+            return None
+
+        identifier = _get_user_identifier()
+        now = time.time()
+        window_start = now - self._rate_period
+        endpoint = request.endpoint or "unknown"
+        key = (identifier, endpoint)
+
+        with _request_times_lock:
```
walesch-yan: I'm a little bit hesitant to use a single lock for all traffic that passes through, would that not make us more vulnerable to attacks like DoS, because the locks are acquired every time a request is made, even if the rate limit has been reached (as the check happens afterwards). 
Maybe, I'm wrong since these checks don't take very long but a single bot making continuous requests could still result in major delays and blocking all endpoints.
I assume we could have a lock per identifier, let me know your thoughts about this @Phantominumm

mxcubeweb/core/server/ratelimited_blueprint.py
```diff
@@ -0,0 +1,85 @@
...
+                return jsonify(
+                    {"error": "Rate limit exceeded", "retry_after": retry_after}
+                ), 429
+            times.append(now)
+            _request_times[key] = times
+
+            if not times:
+                del _request_times[key]
+            else:
+                _request_times[key] = times
```
walesch-yan: I don't really see how this part of the code is reached, isn't times always initialized to be at least an empty list and we always add the current time? In addition the code from the `else` block seems to be duplicated from the line before the `if`?

walesch-yan (commented):
In general this looks like a very good draft for the rate limiter. My only concern is the use of a single lock for all traffic in this implementation

mxcubeweb/core/server/ratelimited_blueprint.py
Phantominumm: the else was a paste from an old version that I missed, for the logic of clean old entries, I'm working on it to have a better structure than that.

PR #1722, opened 2025-06-25
Remove directly unused props from SampleQueueContainer

walesch-yan:
This PR cleans up the prompts passed from `SampleQueueContainer`  to the child components, by removing all the props that are not used in the child components (Or passed further down). 

This reduces the number of unnecessary redraws and makes it clearer which information is passed down. It is a first step in cleaning up the queue. Next steps include some further prevention of prop drilling in the child components and general refactoring


axelboc (approved):
Wow that's a lot of unused props... :joy:

marcus-oscarsson:
Nice !

PR #1724, opened 2025-06-26
Refactor QueueControl

walesch-yan:
This PR refactors `QueueControl` component. As such, the following changes have been done:

- `QueueControl` has been rewritten as functional component
- The `QUEUE_STARTED` constant which was only used in the lookupTable here and never actually set by any signal in the code (Seems to be what `QUEUE_RUNNING` was called previously) has been removed
- Removed not anymore used props from parent component
- Instead of using two lookuptables, one per rendered button, they have been merged to one
- Removed unused states from lookuptables
- removed button render function
- simplified logic for rendering `Next Sample` button

### DIFF
--- a/ui/src/components/SampleQueue/QueueControl.jsx
+++ b/ui/src/components/SampleQueue/QueueControl.jsx
@@ -1,205 +1,37 @@
-/* eslint-disable react/destructuring-assignment */
 import './app.css';
 
-import React from 'react';
-import { Button, Nav, Navbar } from 'react-bootstrap';
+import { Nav, Navbar } from 'react-bootstrap';
+import { useSelector } from 'react-redux';
 
-import {
-  QUEUE_PAUSED,
-  QUEUE_RUNNING,
-  QUEUE_STARTED,
-  QUEUE_STOPPED,
-} from '../../constants';
+import { QUEUE_RUNNING } from '../../constants';
 import QueueSettings from '../../containers/QueueSettings';
 import loader from '../../img/busy-indicator.gif';
+import QueueControlOptions from './QueueControlOptions';
 
-export default class QueueControl extends React.Component {
-  constructor(props) {
-    super(props);
+export default function QueueControl() {
+  const queueStatus = useSelector((state) => state.queue.queueStatus);
 
-    this.nextSample = this.nextSample.bind(this);
+  const showBusyIndicator = queueStatus === QUEUE_RUNNING ? 'inline' : 'none';
 
-    this.state = {
-      options: {
-        [QUEUE_STARTED]: [
-          {
-            text: 'Stop',
-            class: 'btn-danger',
-            action: props.stopQueue,
-            key: 1,
-          },
-        ],
-        [QUEUE_RUNNING]: [
-          {
-            text: 'Stop',
-            class: 'btn-danger',
-            action: props.stopQueue,
-            key: 1,
-          },
-        ],
-        [QUEUE_STOPPED]: [
-          {
-            text: 'Run Queue',
-            class: 'btn-success',
-            action: props.startQueue,
-            key: 1,
-          },
-        ],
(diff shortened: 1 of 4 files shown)


ui/src/components/SampleQueue/QueueControl.jsx
```diff
@@ -1,205 +1,147 @@
...
-          },
-        ],
-        NoSampleMounted: [
-          {
-            text: 'New Sample',
-            class: 'btn-primary',
-            action: this.showForm,
-            key: 1,
-          },
-        ],
```
walesch-yan, line 93: The NoSampleMounted option did not exist, hence completely removed

ui/src/components/SampleQueue/QueueControl.jsx
```diff
@@ -1,205 +1,147 @@
...
-        LastSample: [
-          {
-            text: 'Unmount',
-            class: 'btn-primary',
-            action: this.nextSample,
-            key: 1,
```
walesch-yan, line 99: The fact that the key here was 1 had as a consequence that we would render two buttons with the same key (where the keys were numerical values, which is already not quite how it should be). The new implementation makes the `keys` unnecessary, since no map function is used anymore

ui/src/components/SampleQueue/QueueControl.jsx
```diff
@@ -1,205 +1,147 @@
...
+      <Nav
+        className="me-auto my-2 my-lg-0"
+        style={{ maxHeight: '100px' }}
+        navbarScroll
+      >
+        <Nav.Item>
+          {optionsToRender.map((option, index) => (
+            <span
+              style={{ marginRight: '0.5em' }}
+              key={` ${index === 0 ? 'First' : 'Second'} QueueControlButton`}
```
walesch-yan: I'm open to suggestions for the unique key assignments here :smile:

axelboc (commented):
Sorry, the diff is a bit hard to review and I'm off on holidays :sunglasses:, so I'll leave it to someone else. :grin:

From a quick glance, my recommendation would be to refactor the code in a preparatory PR, before converting `QueueControl` to a functional component.

By refactoring the code, I mean extracting sub-components (which can be written as function components from the get go) and replacing those options records with pure JSX:

```jsx
          <span style={{ marginRight: '0.6em' }}>
            <QueueOptions status={status} />
          </span>
          <span>}}>
            <SampleQueueOptions status={status} />
          </span>
```

```jsx
function QueueOptions(props) {
  const { status } = props;

  if (status === 'QUEUE_STOPPED') {
    return (
      <>
        <Option ... />
        <Option ... />
      </>
    )
  }
   ...
}
```

See y'all in August! :wave:

walesch-yan:
@axelboc Sorry, you are right, I should have probably split this into two PRs from the beginning, to make it easier to review :)
Enjoy your vacation! :beach_umbrella:

walesch-yan:
So I split the component in two `QueueControl` and `QueueControlOptions`. Unfortunately, since the remaining logic in `QueueControl` only consists of rendering the `NavBar`, the linter would complain if I do the refactoring into a functional component in a second PR. Hence, I cannot split into 2 PRs currently, but I hope splitting the logic makes it easier to review

walesch-yan:
One thing that remains to do for this component is to refactor the CSS file, which is used in all Queue related components, but this will be done in another PR

ui/src/constants.js
```diff
@@ -3,7 +3,6 @@
 // are not used within the same file. So disable eslint for this
 // section
 
-export const QUEUE_STARTED = 'QueueStarted';
```
walesch-yan, line 6: This constant was removed as well, as it was only used in the old implementation of the component, but never set in the applications live-time (seems like a processor of `QUEUE_RUNNING`)

marcus-oscarsson:
Well done, thanks guys !

PR #1726, opened 2025-06-27
Flask Rate limiter

Phantominumm:
- Adds rate limiting using Flask-Limiter

- Allows disabling the rate limiter in **configmodels**, notably for tests.

- Updates the CI workflow to include the environment variable **MXCUBEWEB_TEST_MODE** during E2E tests (Cypress).

To add a limit on a route, simply add the decorator (example in login.py):
 **@server.limiter.limit("5 per minute")**
with the format "x per y", where x = number of allowed requests, y = time unit (second, minute, hour, day).


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "mxcubeweb"
-version = "4.408.0"
+version = "4.407.0"
```
Phantominumm: I pulled a recent version at the time, but since I added Flask-Limiter with Poetry, I didn't check the version and ended up with an old file

mxcubeweb/server.py
```diff
@@ -72,6 +75,8 @@ def init(cmdline_options, cfg):
         )
         Server.flask_socketio.init_app(Server.flask)
 
+        Server.limiter = init_limiter(Server.flask)
```
marcus-oscarsson: We can simply avoid calling this when `cfg.flask.RATE_LIMITER_ENABLED = False` Like this you don't need the `DummyLimiter`

mxcubeweb/routes/login.py
```diff
@@ -18,6 +18,7 @@ def init_route(app, server, url_prefix):
     bp = Blueprint("login", __name__, url_prefix=url_prefix)
 
     @bp.route("/", methods=["POST"])
+    @server.limiter.limit("5 per minute")
```
marcus-oscarsson: This would probably be to restrictive, do we at all need it ? Does `default_limits` above in the call to `Limiter` apply for all routes in general or do we need to specify on which routes its added ?

mxcubeweb/core/server/limiter.py
```diff
@@ -0,0 +1,36 @@
...
+def init_limiter(app):
+    """Initialize the rate limiter"""
+    if not app.config.get("RATE_LIMITER_ENABLED", True):
+        return DummyLimiter()
+
+    limiter = Limiter(
+        app=app,
+        key_func=get_remote_address,
+        default_limits=app.config.get(
+            "RATELIMIT_DEFAULT", "2000 per day;500 per hour"
```
marcus-oscarsson: We should probably increase the default limits with an order of magnitude. Like 6000 per hour. ie 100 per minute and roughly 150000 per day.

mxcubeweb/core/server/limiter.py
```diff
@@ -0,0 +1,36 @@
...
+
+        return decorator
+
+
+def init_limiter(app):
+    """Initialize the rate limiter"""
+    if not app.config.get("RATE_LIMITER_ENABLED", True):
+        return DummyLimiter()
+
+    limiter = Limiter(
```
marcus-oscarsson, line 21: A question, would it be possible to configure it so that we get a warning message printed when a certain limit is reached. Like at 80 or 90 % of the limit.  Like "Warning 80 requests on / recorded during the last minute". Is there such a feature in this library

mxcubeweb/__init__.py
```diff
@@ -140,6 +140,9 @@ def build_server_and_config(test=False, argv=None):
             if test_db.exists():
                 test_db.unlink()
 
+        if test or os.environ.get("MXCUBEWEB_TEST_MODE") == "1":
```
marcus-oscarsson: The rate limiter probably comes with a solid test suite. But it would perhaps still be interesting to run the test with the rate limiter and add a test that check if the rate limiter works as we have configured it :).

marcus-oscarsson:
Wonderful, very nice. I think this will be quite useful :). I made a few comments, we can have a closer look on Monday. Have a nice WE @Phantominumm :)

mxcubeweb/server.py
marcus-oscarsson:
```suggestion
        if app.config.get("RATE_LIMITER_ENABLED", False):
            Server.limiter = init_limiter(Server.flask)
```

marcus-oscarsson: Perhaps something like that ?

marcus-oscarsson: It will probably not work if we want to use the decorator, but I thought that we could avoid using the decorator and add something general on `Bluprint` or `Server` level. I think the decorator approach is to fine grained for the moment

pyproject.toml
marcus-oscarsson: Well done

mxcubeweb/routes/login.py
Phantominumm: That was just to show how to use the decorator waiting for the discussion, default_limits is applied to all routes.

mxcubeweb/core/server/limiter.py
Phantominumm: Ok, that was the default config given by Flask, I will change it

mxcubeweb/__init__.py
Phantominumm: Yes of course I will add tests, the problem is that it was too restrictive for the tests, so for the moment I disable it just for the tests

mxcubeweb/core/server/limiter.py
Phantominumm: Flask doesn't support it, if we want to have a warning we have to create a Subclass or a Before the request handler or a Middleware

mxcubeweb/core/server/limiter.py
```diff
@@ -0,0 +1,23 @@
...
+    return jsonify({"error": "Too many requests", "message": str(e.description)}), 429
+
+
+def init_limiter(app):
+    """Initialize the rate limiter"""
+    limiter = Limiter(
+        app=app,
+        key_func=get_remote_address,
+        default_limits=app.config.get(
+            "RATELIMIT_DEFAULT", "150000 per day;6000 per hour"
```
fabcor-maxiv, line 25: Would something like that make sense?

```suggestion
            "RATELIMIT_DEFAULT", configmodels.FlaskConfigModel.RATELIMIT_DEFAULT.split(";")
```

In order to avoid the repetition.

Phantominumm: The problem is that **RATELIMIT_DEFAULT** and **configmodels.FlaskConfigModel.RATELIMIT_DEFAULT** both come from configmodels.
If there is an issue with configmodels and the value cannot be read, the rate limiter will not initialize because there is no value available. The hardcoded (and repeated) value is just a safety fallback in case this happens.

mxcubeweb/core/server/limiter.py
```diff
@@ -0,0 +1,23 @@
+from flask import jsonify
+from flask_limiter import Limiter
+from flask_limiter.util import get_remote_address
+
+
+def rate_limit_error_handler(e):
+    """Handler for rate limit errors"""
+    return jsonify({"error": "Too many requests", "message": str(e.description)}), 429
```
marcus-oscarsson: My apologies for taking some time to get back to this PR, looks really good. We should perhaps look when this happens, or do we get a log message from `flask_limiter` ?

test/test_rate_limiter.py
```diff
@@ -0,0 +1,70 @@
...
+    mock_error = MagicMock()
+    mock_error.description = "Rate limit exceeded: 100 per minute"
+
+    with limiter_app.app_context():
+        response, status_code = rate_limit_error_handler(mock_error)
+        data = json.loads(response.data)
+
+        assert status_code == 429
+        assert data["error"] == "Too many requests"
+        assert data["message"] == "Rate limit exceeded: 100 per minute"
```
marcus-oscarsson: :+1: Great

mxcubeweb/core/server/limiter.py
Phantominumm: Flask does not log a message, I just add a logging when the limit is reached

fabcor-maxiv: I am not a security person, but my gut feeling is that publishing this kind of information (in this case, the exact rate limit) is counter-productive to security, this feels more like: "if you want to hack us, this is exactly how you should do it". Responding with the HTTP code 429 is more than enough from my point of view.

test/test_rate_limiter.py
fabcor-maxiv: I am not a security person, but my gut feeling is that publishing this kind of information (in this case, the exact rate limit) is counter-productive to security, this feels more like: "if you want to hack us, this is exactly how you should do it". Responding with the HTTP code 429 is more than enough from my point of view.

Phantominumm: That's a valid point, but in practice, anyone can easily discover the rate limit by probing until they hit a 429. So hiding it doesn’t add much real protection. On the other hand, including it in the message makes things easier for users and developers.

fabcor-maxiv: I disagree. I do not think it matters enough to block this change from being merged, but for the sake of the argument: if our rate limits block legitimate usage by legitimate users and developers, then probably our rate limits are not carefully chosen and need tuning.

Phantominumm: I agree that legitimate users shouldn’t get blocked. That said, sometimes developers might hit the rate limit when testing new features for different reasons. In those cases, it’s useful to know what’s going on.

Right now, the rate limit is set really high, so it’s not something people would normally reach. If those numbers get changed, I think it makes sense to keep them visible, so it’s easier to figure out what’s causing the problem. And hiding them doesn’t really add any extra security.

It might be good to hear others’ thoughts and preferences on this.

walesch-yan: I think both of you have valid points. In my opinion, we should always strive for a solution that exposes as little information as possible, so it would be best not to send the rate limit, even if in this case, it is only a small detail. For the users, I think a general message like "Too many requests" or "request limit exceeded, try again later" is more than enough. For developers, we can log it on the server side. 

> anyone can easily discover the rate limit by probing until they hit a 429

This is absolutely correct for this version of the rate limiter, resp. for our case, which is why I said it is only a small detail here (especially since the code of our rate limiter is publicly available). However, for the general case, one could imagine a dynamically set rate limit (depending on peak hours or IP-specific (reaching the rate limit will be penalised with lower limits in the future, for example)). In this case, the information becomes kinda crucial.

walesch-yan: That being said, since this message is only in the test file and not the standard message, which you can modify (very nicely implemented btw :+1: ) I don't think that we need to change anything in this PR

fabcor-maxiv (approved):
Seems fine to me.

marcus-oscarsson:
Well done @Phantominumm, :+1:

PR #1728, opened 2025-06-27
transform ConfirmCollectDialog into a functional component

walesch-yan:
This PR transforms ConfirmCollectDialog into a functional component

### DIFF
--- a/ui/src/containers/ConfirmCollectDialog.jsx
+++ b/ui/src/containers/ConfirmCollectDialog.jsx
@@ -1,14 +1,9 @@
-/* eslint-disable react/destructuring-assignment */
-import React from 'react';
 import { Button, Form, Modal } from 'react-bootstrap';
-import { connect } from 'react-redux';
-import { bindActionCreators } from 'redux';
+import { useDispatch, useSelector } from 'react-redux';
 
 import {
-  runSample,
   setAutoMountSample,
   setCentringMethod,
-  setNumSnapshots,
   startQueue,
 } from '../actions/queue';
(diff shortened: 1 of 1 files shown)


ui/src/containers/ConfirmCollectDialog.jsx
```diff
@@ -20,193 +15,128 @@ import {
...
+    dispatch(showConfirmCollectDialog(false));
   }
 
-  onCancelClick() {
-    this.props.hide();
-  }
-
-  setNumSnapshots(n) {
-    this.props.setNumSnapshots(n);
-  }
```
walesch-yan, line 49: This was function was never used, hence removed it

ui/src/containers/ConfirmCollectDialog.jsx
```diff
@@ -20,193 +15,128 @@ import {
...
-      (this.props.queue.currentSampleID || this.props.queue.queue[0])
-    ) {
-      numSamples = 1;
-    }
+  function collectText(numTasks) {
+    const numSamples =
+      !autoMountNext && (currentSampleID || queue[0]) ? 1 : queue.length;
 
-    return { numSamples, numTasks };
-  }
```
walesch-yan, line 111: So the `collectionSummary` only returned the number of samples and the number of texts. Because, we used it, there were some unnecessary function calls to `tasksTocollect` I changed the code so that tasksToCollect is only called once in the component and the result is passed, when we need to get the length. NumSamples is a one line calculation

marcus-oscarsson:
Thanks @walesch-yan  :fireworks:

PR #1729, opened 2025-06-27
Removed unnecessary `phaseChanged` signal handler from diffractometer

marcus-oscarsson:
Removed unnecessary `phaseChanged` signal handler from diffractometer already handled by the Diffractometer adapter: https://github.com/mxcube/mxcubeweb/blob/768c784cb2a560ebf2e9466ba8a9ee6eb2b38687/mxcubeweb/core/adapter/diffractometer_adapter.py#L30


walesch-yan (approved):
LGTM!

PR #1731, opened 2025-06-27
Moved beam signal handlers from beamline "component" to "beam" adapter

marcus-oscarsson:
Moved beam signal handlers from beamline "component" to "beam" adapter

### DIFF
--- a/mxcubeweb/core/adapter/beam_adapter.py
+++ b/mxcubeweb/core/adapter/beam_adapter.py
@@ -18,6 +18,22 @@ class BeamAdapter(ActuatorAdapterBase):
     def __init__(self, ho, role, app):
         super().__init__(ho, role, app, default_resource_handler_config)
 
+        for sig in [
+            "beamPosChanged",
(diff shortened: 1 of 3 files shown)


mxcubeweb/core/adapter/beam_adapter.py
```diff
@@ -18,6 +18,19 @@ class BeamAdapter(ActuatorAdapterBase):
     def __init__(self, ho, role, app):
         super().__init__(ho, role, app, default_resource_handler_config)
 
+        for sig in [
+            "beamPosChanged",
```
marcus-oscarsson, line 22: We might have to revise these signals abit. I think there are some redundancy and that what we want to use is probably `valueChanged`, perhaps `beamPosChanged` and possible `stateChanged` (what that now would mean for beam ?). `beamInfoChanged` could probably be removed.

But these signals are all present and emitted from `mxcubecore` as it is right now. So I kept them as they are

PR #1737, opened 2025-07-01
Fixed casing of `imageClicked`, should be `image_clicked`

marcus-oscarsson:
An old "alias" of `image_clicked` was used in `sampleview.py`. I guess the camel cased method name have lingered in some site specific or mock code. It should be removed or change to `image_clicked`


fabcor-maxiv (approved):
Seems fine to me

PR #1738, opened 2025-07-01
Add CSP 

Phantominumm:
Introduces Content Security Policy (CSP). 
The changes include adding a middleware to enforce CSP headers, updating the configuration model to include CSP settings, and creating a route to handle CSP violation reports.

###Implementation:

* **Configuration Updates**:
  - Added new fields to the `FlaskConfigModel` for CSP settings, including `CSP_ENABLED`, `CSP_POLICY`, `CSP_REPORT_ONLY`, and `CSP_REPORT_URI`. These fields allow enabling/disabling CSP, defining the policy, enabling report-only mode, and specifying a URI for violation reports. 
  - (`mxcubeweb/core/models/configmodels.py`)

* **Middleware for CSP Enforcement**:
  - Introduced a `CSPMiddleware` class to add CSP headers to HTTP responses based on the specified configuration. The middleware dynamically constructs the CSP header string and supports both enforcement and report-only modes. 
  - (`mxcubeweb/core/server/csp.py`)

* **CSP Violation Reporting**:
  - Implemented a new route in `csp_report.py` to handle CSP violation reports. The route logs violations using a dedicated logger and returns a response indicating the report was received.
  -  (`mxcubeweb/routes/csp_report.py`)

### DIFF
--- a/demo/mxcube-web/server.yaml
+++ b/demo/mxcube-web/server.yaml
@@ -13,6 +13,38 @@ server:
   PORT: 8081
   USER_DB_PATH: /tmp/mxcube-user.db
 
+  CSP_ENABLED: true
+  CSP_POLICY:
+    default-src:
+      - "'self'"
+    script-src:
+      - "'self'"
+      - "'unsafe-inline'"
+      - "'unsafe-eval'"
+    style-src:
+      - "'self'"
+      - "'unsafe-inline'"
+    img-src:
+      - "'self'"
+      - "data:"
+      - "blob:"
+    font-src:
+      - "'self'"
+    connect-src:
+      - "'self'"
+      - "wss:"
+      - "ws:"
+    frame-src:
+      - "'self'"
+    object-src:
+      - "'none'"
+    base-uri:
+      - "'self'"
+    form-action:
+      - "'self'"
+  CSP_REPORT_ONLY: false
+  CSP_REPORT_URI: ""
+
 sso:
   USE_SSO: false
   ISSUER: https://websso.[site].[com]/realms/[site]/
--- a/demo/mxcube-web/server.yaml
+++ b/demo/mxcube-web/server.yaml
@@ -16,32 +16,32 @@ server:
   CSP_ENABLED: true
   CSP_POLICY:
     default-src:
-      - "'self'"
+      - "self"
     script-src:
-      - "'self'"
-      - "'unsafe-inline'"
-      - "'unsafe-eval'"
+      - "self"
+      - "unsafe-inline"
+      - "unsafe-eval"
     style-src:
-      - "'self'"
-      - "'unsafe-inline'"
+      - "self"
+      - "unsafe-inline"
     img-src:
-      - "'self'"
+      - "self"
       - "data:"
       - "blob:"
     font-src:
-      - "'self'"
+      - "self"
     connect-src:
-      - "'self'"
+      - "self"
       - "wss:"
       - "ws:"
     frame-src:
-      - "'self'"
+      - "self"
     object-src:
-      - "'none'"
+      - "none"
     base-uri:
-      - "'self'"
+      - "self"
     form-action:
-      - "'self'"
+      - "self"
   CSP_REPORT_ONLY: false
   CSP_REPORT_URI: "/mxcube/api/v0.1/csp/report"
 
--- a/demo/mxcube-web/server.yaml
+++ b/demo/mxcube-web/server.yaml
@@ -43,7 +43,7 @@ server:
     form-action:
       - "'self'"
   CSP_REPORT_ONLY: false
-  CSP_REPORT_URI: ""
+  CSP_REPORT_URI: "/mxcube/api/v0.1/csp/report"
 
 sso:
   USE_SSO: false
--- a/mxcubeweb/app.py
+++ b/mxcubeweb/app.py
@@ -249,6 +249,7 @@ class MXCUBEApplication:
             "user_logger": logging.getLogger("user_level_log"),
             "queue_logger": logging.getLogger("queue_exec"),
             "mx3_ui_logger": logging.getLogger("MX3.UI"),
+            "csp_logger": logging.getLogger("csp"),
         }
 
         stdout_log_handler = StreamHandler(sys.stdout)
--- a/mxcubeweb/core/models/configmodels.py
+++ b/mxcubeweb/core/models/configmodels.py
@@ -38,6 +38,30 @@ class FlaskConfigModel(BaseModel):
         description="One of the strings ['SIGNED', 'ADHOC', NONE]",
     )
 
+    CSP_ENABLED: bool = True
+    CSP_POLICY: dict[str, list[str]] = Field(
+        {
+            "default-src": ["'self'"],
+            "script-src": ["'self'", "'unsafe-inline'", "'unsafe-eval'"],
+            "style-src": ["'self'", "'unsafe-inline'"],
+            "img-src": ["'self'", "data:", "blob:"],
+            "font-src": ["'self'"],
+            "connect-src": ["'self'", "wss:", "ws:"],
+            "frame-src": ["'self'"],
+            "object-src": ["'none'"],
+            "base-uri": ["'self'"],
+            "form-action": ["'self'"],
+        },
+        description="Content Security Policy directives",
+    )
+    CSP_REPORT_ONLY: bool = Field(
+        False,
+        description="Set to True to enable report-only mode (won't block content)",
+    )
+    CSP_REPORT_URI: str = Field(
+        "", description="URI for CSP violation reports (empty to disable reporting)"
+    )
+
 
 class SSOConfigModel(BaseModel):
     USE_SSO: bool = Field(False, description="Set to True to use SSO")
--- a/mxcubeweb/core/models/configmodels.py
+++ b/mxcubeweb/core/models/configmodels.py
@@ -39,21 +39,7 @@ class FlaskConfigModel(BaseModel):
     )
 
     CSP_ENABLED: bool = True
-    CSP_POLICY: dict[str, list[str]] = Field(
-        {
-            "default-src": ["'self'"],
-            "script-src": ["'self'", "'unsafe-inline'", "'unsafe-eval'"],
-            "style-src": ["'self'", "'unsafe-inline'"],
-            "img-src": ["'self'", "data:", "blob:"],
-            "font-src": ["'self'"],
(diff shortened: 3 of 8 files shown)


mxcubeweb/core/models/configmodels.py
```diff
@@ -38,6 +38,30 @@ class FlaskConfigModel(BaseModel):
         description="One of the strings ['SIGNED', 'ADHOC', NONE]",
     )
 
+    CSP_ENABLED: bool = True
+    CSP_POLICY: dict[str, list[str]] = Field(
```
marcus-oscarsson: We could also store the policy in a file, would that be easier ?

mxcubeweb/routes/csp_report.py
```diff
@@ -0,0 +1,29 @@
...
+            logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+        )
+        csp_logger.addHandler(handler)
+        csp_logger.setLevel(logging.WARNING)
+
+    @bp.route("/report", methods=["POST"])
+    def csp_report():
+        """Endpoint to collect CSP violation reports"""
+        report = request.get_json()
+        csp_logger.warning("CSP Violation: %s", report.get("csp-report", {}))
```
Phantominumm, line 19: This part is to automatically receive and log content security policy violations reported by browsers. This helps detect script injection attempts or other XSS vulnerabilities.

fabcor-maxiv:
Can we have some more context and info? What prompted this pull request? If it is another recommendation from Sonarqube, can we have a link to the Sonarqube documentation? Can we have links to what CSP recommendations we are aiming to follow here if any besides Sonarqube.

Phantominumm:
It didn’t come from SonarQube; it is intended to enhance security.
The CSP rules follow OWASP's recommendations from their [Content Security Policy Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Content_Security_Policy_Cheat_Sheet.html)

The policy is strict, allowing only 'self' and the additional sources that were necessary. The configuration is stored in a YAML file, making it easy to update if needed.

marcus-oscarsson:
@fabcor-maxiv: The CSP will be a bit site dependent. We are currently applying these on the reverse proxy but its very useful to have the possibility to do this on the flask server.

fabcor-maxiv:
Would it make sense to add some documentation then? Maybe on [the "Deployment" page](https://mxcubeweb.readthedocs.io/en/stable/dev/deployment.html)?

mxcubeweb/routes/csp_report.py
```diff
@@ -0,0 +1,29 @@
...
+    bp = Blueprint("csp", __name__, url_prefix=url_prefix)
+
+    csp_logger = logging.getLogger("csp")
+    if not csp_logger.handlers:
+        handler = logging.StreamHandler()
+        handler.setFormatter(
+            logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+        )
+        csp_logger.addHandler(handler)
+        csp_logger.setLevel(logging.WARNING)
```
walesch-yan: Maybe for now, you could add csp as a logger in the [`init_logging`](https://github.com/mxcube/mxcubeweb/blob/develop/mxcubeweb/app.py#L204) method, then you would keep the same formatter/handler for it as the other loggers have at the moment. 
I guess it might make sense to visit this method later and add specific logger configurations (specific log_file locations, levels, etc), but for now, I think it is okay and removes the necessity to instantiate it here. Also, it makes it easier to have all the logic in one place

mxcubeweb/routes/csp_report.py
walesch-yan: That's a nice addition, could you set the report URI in the demo then to match the location of this when running the mockup?

mxcubeweb/app.py
```diff
@@ -194,31 +194,11 @@ def init_signal_handlers():
...
-
-        try:
-            MXCUBEApplication.harvester.init_signals()
-        except Exception:
-            sys.excepthook(*sys.exc_info())
+        MXCUBEApplication.queue.init_signals(HWR.beamline.queue_model)
+        MXCUBEApplication.sample_view.init_signals()
+        MXCUBEApplication.sample_changer.init_signals()
+        MXCUBEApplication.beamline.init_signals()
+        MXCUBEApplication.harvester.init_signals()
```
walesch-yan: In the PR, there seem to be a few diffs that are not related to the PR itself and are already merged onto develop. Could you maybe rebase on the latest, that makes it easier to review/merge :)

Phantominumm: Yes, I made some mistakes, but normally it's settled

demo/mxcube-web/server.yaml
```diff
@@ -13,6 +13,38 @@ server:
...
+      - "wss:"
+      - "ws:"
+    frame-src:
+      - "'self'"
+    object-src:
+      - "'none'"
+    base-uri:
+      - "'self'"
+    form-action:
+      - "'self'"
```
fabcor-maxiv: Why do all these have both double quotes and single quotes? Is that on purpose? Do we need quotes at all?

Phantominumm: Yes it follows CSP syntax requirements, CSP requires certain values to be enclosed in single quotes and to have single quotes include, we have to wrap in double quotes in the YAML.

fabcor-maxiv: That seems off to me and prone to errors. I would recommend adding the single quotes when building the policy string in `_build_policy_string`.

fabcor-maxiv: Ah OK, now I see that some of them are not in single quotes. I had not seen that... Now I understand that things are more complex than I thought... and maybe your previous version was somewhat better after all. I do not know. I guess I should understand more about CSP to be able to give useful feedback.

Would it make sense to write the whole policy string in the YAML file? Is it necessary to have the policy string be built at run-time? For example, could we have a list like this in the YAML?

```yaml
CSP_POLICY:
  - "connect-src 'self' wss: ws:"
  - "img-src 'self' data: blob:"
```

(I guess double quotes are necessary because of the trailing colon `:`)

marcus-oscarsson: Ill drop by later @Phantominumm, too many things to do today ;). How about just storing the the csp in a .txt file and running some validator on the file ?

marcus-oscarsson: Maybe this project will work https://github.com/gwen001/csp-analyzer

Phantominumm: > Ah OK, now I see that some of them are not in single quotes. I had not seen that... Now I understand that things are more complex than I thought... and maybe your previous version was somewhat better after all. I do not know. I guess I should understand more about CSP to be able to give useful feedback.
> 
> Would it make sense to write the whole policy string in the YAML file? Is it necessary to have the policy string be built at run-time? For example, could we have a list like this in the YAML?
> 
> ```yaml
> CSP_POLICY:
>   - "connect-src 'self' wss: ws:"
>   - "img-src 'self' data: blob:"
> ```
> 
> (I guess double quotes are necessary because of the trailing colon `:`)

It's not strictly necessary but it's easier to modify or extend via code, more readable this way and less error-prone with parsing the string.
It's possible to have list like that but it has to be handle when loading the config (not really complicated but necessary)

Phantominumm: > Ill drop by later @Phantominumm, too many things to do today ;). How about just storing the the csp in a .txt file and running some validator on the file ?

Storing the CSP as a .txt file is simple and works fine if the policy doesn’t change much (as it would be in our case) and you’re fine checking it with csp-analyzer (became dependent of it so no easily validation directly in the app)  . But if things get more complex or need to change often, using YAML makes it easier to manage and validate.

Phantominumm:
@marcus-oscarsson about the URI: 
It's exactly what we talked.

If a browser detects a violation of your CSP, it will send a report to the endpoint on your server. (the current endpont is /report by default)
you can set CSP_REPORT_URI to an absolute URL pointing to another server.

**Note:**
The external server must be accessible from clients’ browsers.
Make sure the external server is configured to accept and process CSP reports.
Some browsers may restrict sending reports to a different origin due to privacy or security reasons, but most modern browsers support absolute URLs for report-uri or report-to

demo/mxcube-web/server.yaml
fabcor-maxiv: I am really not familiar with this topic, so hard to guess if the code change is good or not.

> It's not strictly necessary but it's easier to modify or extend via code, more readable this way and less error-prone with parsing the string.

Do we need to modify or extend via code? In which scenario would we want to do that?

Does MXCuBE need to parse the CSP policy string before sending it via headers? Why?


> It's possible to have list like that but it has to be handle when loading the config (not really complicated but necessary)

Why is that necessary? What is necessary exactly? What kind of handling does MXCuBE need to do? Can you point at code and/or documentation, maybe show with an example what you mean?

marcus-oscarsson:
Well done :clap: :partying_face:

demo/mxcube-web/server.yaml
Phantominumm: @fabcor-maxiv Sorry, I'm late for the answers but here they are: 

**_Do we need to modify or extend via code? In which scenario would we want to do that?_**

It would be rare to do that but in these cases: 
- To enable or disable certain sources or directives based on environment
- To add extra sources for debugging or testing
- To dynamically adjust the policy based on user roles, features, or plugins

Currently, the policy is static, but having it as a a dict makes it easy to add/remove sources if you ever need to.

**_Does MXCuBE need to parse the CSP policy string before sending it via headers? Why?_**

Yes, MXCuBE needs to parse it into a string for the HTTP header because browsers expect the CSP as a single string in the HTTP header like : 
- Content-Security-Policy: default-src 'self'; script-src 'self' 'unsafe-inline' ...

the dict is for more readability and practicality for us as it's pretty easy to understand how to add/remove policy.

_**Why is that necessary? What is necessary exactly? What kind of handling does MXCuBE need to do? Can you point at code and/or documentation, maybe show with an example what you mean?**_

-I notice I wasn't clear and wrote a quick reply as we talked about this IRL with marcus, It's necessary to be handle in the same way, the dict is handled, we need to build the string in the same way I show in 2nd paragraph, if we want to have list of string like that, we have to change the function that build the string policy (not complicated or long). 
Sorry this part was really unclear.

fabcor-maxiv: Thanks for the answers, @Phantominumm :)

I truly do not know enough about CSP to understand what is going on in this pull request, so I will trust you that this complexity is useful if not necessary.

PR #1741, opened 2025-07-02
Add in control user to userlist

walesch-yan:
The Userlist which was showed on the remote access tab was never complete as the user who was in control was not displayed, only observers.

This PR changes the way the User list is displayed, adding the in control user, and highlighting him and the current logged in User with different Tags (In Control and (You))

For this to work, the information of the user in control was added to the state under `remoteAccess`

Additionally, I made the Columns in the remote access page and Userlist more reactive to take up the space according to the number of columns (previously there was always some space left for elements that were only displayed to the user in control)

### DIFF
--- a/ui/src/components/RemoteAccess/UserList.jsx
+++ b/ui/src/components/RemoteAccess/UserList.jsx
@@ -8,34 +8,60 @@ function UserList() {
 
   const user = useSelector((state) => state.login.user);
   const observers = useSelector((state) => state.remoteAccess.observers);
+  const operator = useSelector((state) => state.remoteAccess.operator);
+
+  const columnSize = user.inControl ? 4 : 6;
 
   return (
     <Card className="mb-3">
       <Card.Header>Users</Card.Header>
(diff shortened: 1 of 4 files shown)


ui/src/components/RemoteAccess/UserList.jsx
```diff
@@ -8,34 +8,60 @@ function UserList() {
...
+              <span>&nbsp;</span>
+            </Col>
+          )}
+        </Row>
+        <Row key={operator.username} className="mt-3">
+          <Col sm={columnSize}>
+            <span style={{ lineHeight: '24px' }}>
+              <span style={{ fontWeight: 'bold' }}>In Control </span>
+              {user.inControl && (
+                <span style={{ fontWeight: 'bold', fontStyle: 'italic' }}>
```
walesch-yan, line 37: Maybe somebody has a better idea for the styling of these Tags, this is what they would look like:
Screenshot from 2025-07-02 17-38-32

PR #1752, opened 2025-07-07
transform Todotree into functional component

walesch-yan:
As the title suggests, this PR transforms the `todotree` into a functional component

### DIFF
--- a/ui/src/components/SampleQueue/TodoTree.jsx
+++ b/ui/src/components/SampleQueue/TodoTree.jsx
@@ -1,75 +1,65 @@
-/* eslint-disable react/destructuring-assignment */
-
 import './app.css';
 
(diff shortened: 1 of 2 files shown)


ui/src/components/SampleQueue/TodoTree.jsx
```diff
@@ -1,75 +1,65 @@
-/* eslint-disable react/destructuring-assignment */
-
 import './app.css';
```
walesch-yan, line 1: I did not yet change this import as multiple components rely on this `app.css` file. So this will be done in a separate PR

marcus-oscarsson: Hehe, yeees ;)

PR #1753, opened 2025-07-08
Cleanup TaskItems and collect similar container code

walesch-yan:
This PR is a first (of probably many) PRs to try to clean up the different `TaskItems` components by doing two things essentially:

 - Introduce the `TaskItemContainer`: Every component had the exact same header rendering logic, so I created a single component to extract and centralise this code and make it easier to refacture/maintain. To make it easy to review, I did not make any changes to the code itself yet. (Will definitely in a later PR)
 - Remove dead code in `TaskItems`: 
   This includes:
    - Now unused imports
    - `deleteTask`, `headerOnClick` and headerOnContext` methods which were moved to `TaskItemContainer`
    - `deleteButton`, `toggleChecked` methods and internal `state`, which were present in every component but never used
    - logic related to showing 1D plots in the `XRFTaskItem`: it existed, but the Modal was never shown (always remained in `falsy` state) for now I removed it as part of a cleanup, but this could be added and/or fixed in a later PR

So while there are a lot changes in this PR, it is mainly one repetitive change in every file

Next steps are (not necessary in this order):
 - cleanup CSS 
 - remove unused `props`
 - remove logic from global state that could be handled in component states (collapse Item for example)
 - remove unused actions related to this section
 - fix 1D plotting (if necessary)
 - refactor components

### DIFF
--- a/ui/src/components/SampleQueue/CharacterisationTaskItem.jsx
+++ b/ui/src/components/SampleQueue/CharacterisationTaskItem.jsx
@@ -1,21 +1,16 @@
 /* eslint-disable react/destructuring-assignment */
 /* eslint-disable jsx-a11y/anchor-is-valid */
 /* eslint-disable react/no-unused-prop-types */
-/* eslint-disable react/no-unused-state */
 
 /* eslint-disable react/no-array-index-key */
 
 import PropTypes from 'prop-types';
 import { Component } from 'react';
-import { Button, Collapse, ProgressBar, Table } from 'react-bootstrap';
+import { Button, Table } from 'react-bootstrap';
 
-import {
-  TASK_COLLECT_FAILED,
-  TASK_COLLECTED,
-  TASK_RUNNING,
-  TASK_UNCOLLECTED,
-} from '../../constants';
+import { TASK_COLLECTED } from '../../constants';
 import TooltipTrigger from '../TooltipTrigger';
+import TaskItemContainer from './TaskItemContainer';
 
 export default class TaskItem extends Component {
   static propTypes = {
@@ -26,17 +21,9 @@ export default class TaskItem extends Component {
   constructor(props) {
     super(props);
     this.showForm = this.showForm.bind(this);
-    this.deleteTask = this.deleteTask.bind(this);
-    this.toggleChecked = this.toggleChecked.bind(this);
-    this.taskHeaderOnClick = this.taskHeaderOnClick.bind(this);
-    this.taskHeaderOnContextMenu = this.taskHeaderOnContextMenu.bind(this);
     this.getResult = this.getResult.bind(this);
     this.showDiffPlan = this.showDiffPlan.bind(this);
     this.pointIDString = this.pointIDString.bind(this);
-    this.state = {
-      overInput: false,
-      selected: false,
-    };
   }
 
   getResult(_state, data) {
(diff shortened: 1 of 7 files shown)


ui/src/components/SampleQueue/TaskItemContainer.jsx
```diff
@@ -0,0 +1,140 @@
...
+      case TASK_COLLECTED: {
+        pbarBsStyle = 'success';
+        if (
+          specialTaskCSS === 'Characterisation' &&
+          diffractionPlanLength === undefined
+        ) {
+          taskCSS += ' warning';
+          break;
+        }
+        taskCSS += ' success';
```
walesch-yan, line 77: So there is a special style for characterisations that is applied if the Task has been collected but no diffraction plan was chosen

ui/src/components/SampleQueue/CharacterisationTaskItem.jsx
```diff
@@ -233,184 +188,113 @@ export default class TaskItem extends Component {
...
-      paddingRight: '10px',
-      color: '#d9534f',
-      cursor: 'pointer',
-    };
-
-    let taskCSS = this.props.selected
-      ? 'task-head task-head-selected'
-      : 'task-head';
-
-    taskCSS += ' uncollected';
```
walesch-yan, line 294: So the css applied to characterisation items was slightly different from the others, due to the warning style that I explained in another comment. The `uncollected` class does also only exist in this component. However, this one only sets the background to the same color as `task-head` does, hence I removed it (it was unnecessary)

ui/src/components/SampleQueue/XRFTaskItem.jsx
```diff
@@ -73,97 +48,6 @@ export default class XRFTaskItem extends Component {
...
-    }
-
-    return (
-      <div className="float-end">
-        <a href="#" onClick={this.showXRFPlot}>
-          Plot available
-        </a>
-      </div>
-    );
-  }
```
walesch-yan, line 98: So this component was supposed to render a link for the plot, but it is never rendered, hence I removed all logic related to 1d plots in `XRFTaskItem`

marcus-oscarsson: Ok, it would be super nice if we could try to get those plots work again :).

ui/src/components/SampleQueue/TaskItemContainer.jsx
marcus-oscarsson: Ok

ui/src/components/SampleQueue/TaskItemContainer.jsx
```diff
@@ -0,0 +1,140 @@
+import { Component } from 'react';
+import { Collapse, ProgressBar } from 'react-bootstrap';
+
+import {
```
marcus-oscarsson, line 4: Nice, :)

marcus-oscarsson:
I guess some of the CSS could be moved to a separate css file but I think this PR is good as it is. It could be done in a future PR.

walesch-yan:
> I guess some of the CSS could be moved to a separate css file but I think this PR is good as it is. It could be done in a future PR.

Oh Yes definitely, that's why it was on my next steps task list;) Just wanted to keep this PR as simple as possible:)

axelboc (commented):
image

Wow, that was _a lot_ of duplicated code! :open_mouth: 

Nice clean-up here and in other PRs generally like #1790. Always satisfying. :grin:

PR #1754, opened 2025-07-08
Add job in workflow for Snyk update pnpm-lock

Phantominumm:
Updates the CI workflow to include a new job for syncing `pnpm-lock.yaml` with `package.json` because when Snyk updated dependencies, only `package.json` is changed :

### Continuous Integration (CI) Enhancements:
* Added a new `sync-lockfile` job in `.github/workflows/ui.yaml` to automatically update `pnpm-lock.yaml` when `package.json` changes in pull requests. This job includes steps for checking changes, updating the lockfile, and committing the updates.
* Updated the `format`, `lint`, and `e2e` jobs in `.github/workflows/ui.yaml` to depend on the `sync-lockfile` job. [[1]](diffhunk://#diff-d5e44b190fc7455fb881305ab41a824a0df4c19b7def67f908add8e1fa0d0b8dR22-L29) [[2]](diffhunk://#diff-d5e44b190fc7455fb881305ab41a824a0df4c19b7def67f908add8e1fa0d0b8dR56) [[3]](diffhunk://#diff-d5e44b190fc7455fb881305ab41a824a0df4c19b7def67f908add8e1fa0d0b8dR92)


fabcor-maxiv:
> * Added a new `sync-lockfile` job in `.github/workflows/ui.yaml` to automatically update `pnpm-lock.yaml` when `package.json` changes in pull requests.

Should we have that? I feel like that is not something that we want to have. For sure we want a test that checks if the lock file is in sync with `package.json`. If that test fails, then we want to trigger the update of the lock file manually and re-do the pull request (amend the commit).

_[Personally I want to avoid the CI suite to do commits on its own (for example, I am a bit wary of the "bump version" job).]_

Now I see that this pull request has been closed already, but I thought I would leave the comment anyway in case there is a plan to go forward with this idea.

PR #1755, opened 2025-07-08
remove toggleCheckbox action/route

walesch-yan:
This PR removes code related to the `toggleCheckBox` logic from the queue, it includes:
- actions `toggleCheckBox` and `toggleChecked` (funnily enough the related reducer did not exist)
- backend function `toggle_node`
- route `/queue/<node_id>/toggle`
- appearances in component props

Additional Info: the functions which called this action/route have been removed In #1753 (they were already unreachable before)


marcus-oscarsson:
Excellent !

PR #1762, opened 2025-07-09
Don't use deprecated Beamline.get_id() method

elmjag:
Fixes following warning on MXCuBE-Web start-up:

>   UserWarning: Beamline.get_id is Deprecated. Use hwobj.id instead


marcus-oscarsson (approved):

marcus-oscarsson:
Thanks @elmjag

PR #1763, opened 2025-07-09
Added resource handler (`ComponentResourceHandler`) for Components

marcus-oscarsson:
I've had this code for the `ComponentResourceHandler` more or less ready for some time and I decided that its time to at-least share it for input and ideas. 

The priority is still to migrate the components that are directly related to hardware objects to adapters, at-least where its feasible. That basically concerns all components except: `log`, `csp_report`, `login`, `ra`, `main`. The bigger components like `queue` and `lims` may have to wait and remain components for some time. 

The concept is exactly the same as for the `AdapterResourceHandler`, and it was possible to re-use the same configuration/definition.

The two classes  `ComponentResourceHandler` and `AdapterResourceHandler` shares the same base class which means that all the validation, serialization, exception handling of input and output are the same. It also means that the `routes` module as can be removed. No more route decorators :partying_face: !

For the sake of example I configured one component in [__log.py__](mxcubeweb/core/components/log.py)
Summary:

-  The plan is to migrate components->adapters
-  Use the ComponentResourceHandler on remaining components
-  The ComponentResourceHandler and AdapterResourceHandler uses the same mechanism for data validation and error handling
-  The routes module can finaly be removed.


mxcubeweb/core/server/resource_handler.py
```diff
@@ -114,14 +120,14 @@ def register_with_server(cls, flask_server):
             rh.register_blueprint(flask_server)
 
 
-class AdapterResourceHandler:
+class ResourceHandler:
```
marcus-oscarsson, line 123: The main logic of `AdapterResourceHandler` is now in base class called `ResourceHandler`. There are two sub classes `AdapterResourceHandler` and `ComponentResourceHandler`

mxcubeweb/core/server/resource_handler.py
```diff
@@ -511,3 +495,142 @@ def register_blueprint(self, parent_bp) -> None:
...
+            name, url_prefix, handler_dict, app, exports, commands, attributes
+        )
+
+    def _get_handler_object_route(self, export) -> str:
+        """
+        Returns the base route for the resource handler.
+        """
+        return f"{export['url']}"
+
+    def _create_view_func(self, export: dict[str, str]) -> Callable:
```
marcus-oscarsson, line 527: The creation of the view function is slightly different between the two resource handlers. The most important difference is that there is only one instance of a `Component`, while an Adapter handles several instance of `HardwareObjects`. The hardware object id string is hence not needed for the `Component`.

mxcubeweb/core/server/resource_handler.py
```diff
@@ -511,3 +495,142 @@ def register_blueprint(self, parent_bp) -> None:
...
+        app: object,
+        exports: list[dict[str, str]],
+        commands: list[str],
+        attributes: list[str],
+    ) -> None:
+        super().__init__(
+            name, url_prefix, handler_dict, app, exports, commands, attributes
+        )
+
+    def _get_handler_object_route(self, export) -> str:
```
marcus-oscarsson, line 585: The Adapters manage several instance of HardwareObjects, which means that need to be able to pass an id

mxcubeweb/core/components/log.py
```diff
@@ -0,0 +1,58 @@
...
+from mxcubeweb import logging_handler
+from mxcubeweb.core.components.component_base import ComponentBase
+from mxcubeweb.core.models.configmodels import (
+    ResourceHandlerConfigModel,
+)
+
+hwr_logger = logging.getLogger("MX3.HWR")
+
+
+class Log(ComponentBase):
```
marcus-oscarsson, line 13: log.py in routes no longer needed, everything is contained in the `Component` route definition and handler methods. The input and return data are serialized and handled the same way as for Adapters, making it much easier to handle the "data flow"

mxcubeweb/core/components/log.py
```diff
@@ -0,0 +1,58 @@
...
+                    {
+                        "attr": "log",
+                        "method": "GET",
+                        "url": "/",
+                        "decorators": [app.server.restrict],
+                    },
+                    {
+                        "attr": "log_frontend_traceback",
+                        "method": "POST",
+                        "url": "/",
```
walesch-yan: This URL should be `/log_frontend_traceback`

walesch-yan (commented):
I think this is a great change, that works well with what is already in place. 

There is only a misconfigured URL in the `log.py` file. (The description should also say `log.py` and not `workflow.py` for those who are wondering:))

walesch-yan (approved):

PR #1767, opened 2025-07-10
Fixed GPL license files

rhfogh:
You need these two files, with (strongly recommended) these names


fabcor-maxiv (approved):
I am not a lawyer.

rhfogh:
@beteva, @fabcor-maxiv 
I am not a lawyer either, but the GPL site has made it pretty clear. If you look at
https://www.gnu.org/licenses/gpl-howto.html they tell you that  you need a license notice in each file (saying you use LGPL), and copyright statements with years on them. They also say that since the LGPL is a set of additional permissions added to the GPL, you need both of the license files present if you use LGPL.

The main thing is that if **any** of MXCuBE is GPL, **all** of it must be GPL - that is how the GPL is written. And that would mean that non-GPL code, like that from the Global Phasing workflow, cannot link to it.

Qt is a separate problem. The license conditions mean that any group coding to PyQt must have a commercial license, and that the program can only use the PyQt versions for which there was  a commercial license. More precisely, if  you have a PyQt commercial license in 2022 your program can use the Qt versions that were current in 2022 (and that you therefore have as license for). It cannot later download and use a non-commercial 2024 version

beteva:
> The main thing is that if **any** of MXCuBE is GPL, **all** of it must be GPL - that is how the GPL is written. And that would mean that non-GPL code, like that from the Global Phasing workflow, cannot link to it.

@rhfogh Isn't that a problem for you? I start recalling that this was the reason we've chosen LGPL.

rhfogh:
@beteva 
If any of the code was GPL-licensed it would indeed be a **huge** problem for us. But the way to set the license as LGPL is to have a statement in each file (and in the project header, but it has to be in each file) saying that the license is LGPL. The mere presence of a file called COPYING containing the GPL is 1) not what determines how the code is licensed, 2) required by the FSF, since the LGPL license, in effect, says "_In addition to what it says in the GPL license, you have the following additional rights ..._'. Specifically they say that for LGPL licensing you must have both COPYING and COPYING.LESSER present.

PR #1774, opened 2025-07-11
Add workflows for Bandit and SonarQube

Phantominumm:
This pull request adds a new GitHub Actions workflow to integrate security scanning with Bandit and code quality analysis with SonarCloud. The workflow is scheduled to run weekly and includes steps for scanning, caching, and uploading results.

### New GitHub Actions workflow for security and code quality:

* `.github/workflows/bandit_sonar.yaml`: 
  - Introduced a new workflow named "Bandit and SonarQube" to perform security scanning using Bandit and code quality analysis using SonarCloud.
  - Configured the workflow to run on a weekly schedule (every Wednesday at 7:30 AM) and also support manual triggers (`workflow_dispatch`).
  - Added a Bandit job to scan Python code with medium severity and high confidence, excluding the `./test/` directory, and upload the results as a SARIF file to GitHub.
  - Added a SonarCloud job to perform code quality analysis with caching for SonarCloud packages and environment variables for authentication.

There is some questions to configure:
- Do we scan for every push/PR (can be too much, maybe only scan when push on develop ? )
- Other way can be to schedule analysis (once per week for example, ...)
- The level of threat that is reported (`level: MEDIUM` , `confidence: HIGH`) (confidence parameter is the confidence that Bandit know it's a vulnerability, I think if it's too low, there would be too many false positive)

**Bandit** always generates a report that can be downloaded from GitHub Actions. The report is readable but quite long and not specifically designed for our needs. However, when a vulnerability is detected, Bandit adds a comment to the pull request explaining the issue, and it's also accessible under **Security > Code Scanning** in GitHub.

**SonarQube / SonarCloud** results are only accessible through the SonarCloud project page. Unfortunately, the free plan is limited to 5 users. Be aware that SonarQube often reports many issues that are not critical and can be safely ignored (they are mostly about having "clean" or "perfect" code).

### DIFF
--- a/.github/workflows/bandit_sonar.yaml
+++ b/.github/workflows/bandit_sonar.yaml
@@ -5,8 +5,6 @@ name: Bandit and SonarQube
 "on":
   pull_request: null
   push: null
-  schedule:
-    - cron: '30 7 * * 3'
   workflow_dispatch: null
 
 permissions:
@@ -19,14 +17,26 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+
+      - name: Generate Bandit skips from Ruff
+        id: bandit_skips
+        run: |
+          pip install ruff jq
+          SKIPS=$(ruff rule --all --output-format json \
+            | jq -r '.[] | select(.code | test("^S[0-9]{3}$")) | .code' \
+            | sed 's/S/B/' \
+            | paste -sd, -)
+          echo "skips=$SKIPS" >> $GITHUB_OUTPUT
+
       - name: Bandit Scan
-        uses: shundor/python-bandit-scan@ab1d87dfccc5a0ffab88be3aaac6ffe35c10d6cd   # yamllint disable rule:line-length
+        uses: shundor/python-bandit-scan@ab1d87dfccc5a0ffab88be3aaac6ffe35c10d6cd
         with:
           exit_zero: true
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           excluded_paths: ./test/
           level: MEDIUM
           confidence: HIGH
+          skips: ${{ steps.bandit_skips.outputs.skips }}
 
       - name: Upload SARIF file to GitHub
         uses: github/codeql-action/upload-sarif@v3
--- a/.github/workflows/bandit_sonar.yaml
+++ b/.github/workflows/bandit_sonar.yaml
@@ -1,5 +1,4 @@
 ---
-# yamllint disable rule:line-length
 
 name: Bandit and SonarQube
 
@@ -21,7 +20,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Bandit Scan
-        uses: shundor/python-bandit-scan@ab1d87dfccc5a0ffab88be3aaac6ffe35c10d6cd
+        uses: shundor/python-bandit-scan@ab1d87dfccc5a0ffab88be3aaac6ffe35c10d6cd   # yamllint disable rule:line-length
         with:
           exit_zero: true
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/bandit_sonar.yaml
+++ b/.github/workflows/bandit_sonar.yaml
@@ -1,0 +1,63 @@
+---
+# yamllint disable rule:line-length
+
+name: Bandit and SonarQube
+
+"on":
+  pull_request: null
+  push: null
+  schedule:
+    - cron: '30 7 * * 3'
+  workflow_dispatch: null
+
+permissions:
+  contents: read
+  security-events: write
+  actions: read
+
+jobs:
+  bandit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Bandit Scan
+        uses: shundor/python-bandit-scan@ab1d87dfccc5a0ffab88be3aaac6ffe35c10d6cd
+        with:
+          exit_zero: true
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          excluded_paths: ./test/
+          level: MEDIUM
+          confidence: HIGH
+
+      - name: Upload SARIF file to GitHub
+        uses: github/codeql-action/upload-sarif@v3
+        with:
+          sarif_file: results.sarif
+          category: bandit
+
+  sonarcloud:
+    name: SonarCloud Scan
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Java
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: '17'
+
+      - name: Cache SonarCloud packages
+        uses: actions/cache@v3
+        with:
+          path: ~/.sonar/cache
+          key: ${{ runner.os }}-sonar
+          restore-keys: ${{ runner.os }}-sonar
+
+      - name: SonarCloud Scan
+        uses: SonarSource/sonarcloud-github-action@v2
+        with:
+          args: >
+            -Dsonar.organization=mxcubeweb
+            -Dsonar.projectKey=mxcube_mxcubeweb
+        env:
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/bandit_sonar.yaml
```diff
@@ -0,0 +1,63 @@
+---
+# yamllint disable rule:line-length
```
fabcor-maxiv: Why?

```suggestion
```

.github/workflows/bandit_sonar.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+name: Bandit and SonarQube
+
+"on":
+  pull_request: null
+  push: null
+  schedule:
+    - cron: '30 7 * * 3'
```
fabcor-maxiv: I recommend to add a comment explaining in human language what the schedule is.

.github/workflows/bandit_sonar.yaml
walesch-yan: I assume the line that includes the commit hash for the bandit-scan is too long?

If that is the case, I would add 
```yaml
# yamllint disable-line rule:line-length
```
to the corresponding line instead of the whole file

.github/workflows/bandit_sonar.yaml
```diff
@@ -0,0 +1,63 @@
...
+  security-events: write
+  actions: read
+
+jobs:
+  bandit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Bandit Scan
+        uses: shundor/python-bandit-scan@ab1d87dfccc5a0ffab88be3aaac6ffe35c10d6cd
```
walesch-yan, line 32: I see that you pin here to a specific commit instead of to a version. I'm wondering if there is a specific reason of doing it differently for this action compared to the sonarcloud for example, which is pinned to a version.

walesch-yan:
I think this is a very nice addition, and the workflow can provide some exceptional advantages to the repo in terms of cybersecurity. Very nice work!

> Do we scan for every push/PR (can be too much, maybe only scan when push on develop ? )
Other way can be to schedule analysis (once per week for example, ...)

I believe that the probably best way would be to scan every PR that tries to push to develop. I think we should aim to have a repository that has as few vulnerabilities as possible, and if we can detect something on a possible change, it is better to prevent the change from happening than to have to revert it later.  

> The level of threat that is reported (level: MEDIUM , confidence: HIGH) (confidence parameter is the confidence that Bandit know it's a vulnerability, I think if it's too low, there would be too many false positive)

I think that those are good to start with, maybe we can try to lower the level, once we get used to the workflow, at least if that would not block other PRs directly due to existing vulnerabilities in the code. Regarding the confidence, I totally agree with a high value here. One idea could also be to have one with high confidence, which is required, and a second workflow with lower confidence that is not required to pass, so it would be the developer's job to check if that is a false positive or not. But for now, the high confidence one is more than enough

walesch-yan:
I still have two questions concerning already existing workflows

- Currently, the SonarCloud code analysis is already running on the repo directly, so if we merge the workflow, should we remove it again?
- `Ruff` is also checking for bandit rules as of now, I did not check, but I assume that the workflow scan will do a more in-depth job, do you maybe happen to have a short explanation on what the differences are @Phantominumm

.github/workflows/bandit_sonar.yaml
Phantominumm: Yes that was for the line, but this line is a problem I didn't saw while I was working in my fork with Github auto configuration, if necessary i will add just to the corresponding line as suggested

.github/workflows/bandit_sonar.yaml
Phantominumm: Sure, I was waiting to know the configuration we will use, I don't know if we will keep the schedule scan so I was thinking of doing comments later.

.github/workflows/bandit_sonar.yaml
Phantominumm: That was suggested by Github when importing bandit workflow and I didn't pay attention to that but I will change it to a version instead

Phantominumm: So i tried to use the v1 but it doesn't work, he can't find it so I think we can stay with this commit until I can find something else.

walesch-yan: There is nothing wrong with using the commit hash, I was just wondering, since all the other workflows use the version tag. I guess it is a tradeoff between getting the version patches and having an always stable running workflow but we might have to update more frequently. 

> So i tried to use the v1 but it doesn't work, he can't find it so I think we can stay with this commit until I can find something else.

According to the project's [README](https://github.com/shundor/python-bandit-scan/blob/main/README.md), the workflow name is slightly different from the repo name, so it should work if you use `shundor/bandit-action@v1`  ;)

Phantominumm:
> I still have two questions concerning already existing workflows
> 
>     * Currently, the SonarCloud code analysis is already running on the repo directly, so if we merge the workflow, should we remove it again?
> 
>     * `Ruff` is also checking for bandit rules as of now, I did not check, but I assume that the workflow scan will do a more in-depth job, do you maybe happen to have a short explanation on what the differences are @Phantominumm

1:  SonarCloud automatic analyze is disable, using our workflow is better than the automatic analyze as we have more flexibility and it integrates better. We can't remove it because the CI is bind with SonarCloud Project so the reports are available there. The current state is fine for me.

2: Bandit tool covers more security rules and certain special patterns than ruff with Bandit rules and has the SARIF output (the report) integrates with Github code scanning that is not always the case for Ruff.
There are surely some rules that are checked twice but Bandit come with more security 
(I will search if it's possible to remove rules checking with Bandit so the rules already check by Ruff will be skipped)

.github/workflows/bandit_sonar.yaml
Phantominumm: but the number are in the order: "minutes hour day_of_the_month month day_of_the_week", currently this config is 7:30 am each Wednesday.

.github/workflows/bandit_sonar.yaml
Phantominumm: I tried this but it didn't work, I will try it again, my branch had a problem maybe this cause a problem.

 Error: Unable to resolve action `shundor/bandit-action@v1`, unable to find version `v1`

Here the message. I think that's the reason Github used a commit instead of the v1

walesch-yan: I see, then let's stick to the commit

Phantominumm:
Remove rules already check by Ruff: 
https://docs.astral.sh/ruff/rules/#flake8-bandit-s
I let the rules that were not stable ( still in preview from Ruff)

Phantominumm:
@walesch-yan , Rules used by Ruff are selected with : ```select = ["ALL",]``` in `ruff.toml` and I don't find another way to have them. so we have to keep it hardcoded like that

.github/workflows/bandit_sonar.yaml
```diff
@@ -0,0 +1,62 @@
...
+      - uses: actions/checkout@v4
+      - name: Bandit Scan
+        uses: shundor/python-bandit-scan@ab1d87dfccc5a0ffab88be3aaac6ffe35c10d6cd   # yamllint disable rule:line-length
+        with:
+          exit_zero: true
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          excluded_paths: ./test/
+          level: MEDIUM
+          confidence: HIGH
+          skips: B101,B102,B103,B104,B105,B106,B107,B108,B110,B112,B113,B201,B202,B301,B302,B303,B304,B305,B306,B307,B308,B310,B311,B312,B313,B314,B315,B316,B317,B318,B319,B321,B323,B324,B501,B502,B503,B504,B505,B506,B507,B508,B509,B601,B602,B603,B604,B605,B606,B607,B608,B609,B610,B611,B612,B701,B702,B704
```
walesch-yan: I let you decide on how to deal with this as you did more research on it, than I did. However, I think that we could use the bash command `ruff rule --all --output-format json` in a separate script, then filter and double-check this with the ignored and selected rules from `ruff.toml` to get the list of rules to skip to prevent as from hadrcoding the rules to skip(which might change over time).

I can also agree on merging this workflow first and adding something like this in a follow-up PR.:)

marcus-oscarsson:
Very nice !

walesch-yan (approved):
Thanks! It looks good to me!

fabcor-maxiv:
So when does this workflow run now?

I do not think it is necessary to skip the rules that are already checked by Ruff. Or is there some kind of disruptive behaviour that we are want to avoid?

PR #1775, opened 2025-07-11
split SampleQueue css file into modules

walesch-yan:
This PR removes the general CSS file called `app.css` that was placed in the `SampleQueue` component folder and included all styles for components related to the queue. 

The classes in said CSS file have been either removed, because they were not used anymore (More than 10 unused classes!), resp. overwritten in the same file or added to one of the newly generated `module.css` files.

Since a lot of the components in the queue have similar styles I grouped the styles of the `CurrentTree` and the `TodoTree` into one file (called `Tree.module.css`), as well as the styles of the different `TaskItems` and the `TodoItem`. `QueueControl` and `SampleQueueContainer` still have their own modules

Below I add comments if there was a more specific change regarding a style class.

There will be a follow-up PR to remove undefined classnames in components and pushing hardcoded styles to the modules if needed.


ui/src/components/SampleGrid/SampleGridTable.css
```diff
@@ -1,3 +1,23 @@
...
+  justify-content: center;
+  min-height: 24em;
+  width: 100%;
+  height: 100%;
+  position: absolute;
+  top: 0;
+  left: 0;
+  background-color: #000;
+  opacity: 0.5;
+}
```
walesch-yan, line 19: So these two were the only ones that were not only used within the Queue components. But both were also used in `SampleListViewContainer`, hence I moved them into the file, where all the other classes of this component (and others) are defined until further clean-up.

While `center-in-box` has also been copied to the `SampleQueueContainer` module, `img-centerd was only moved here, since it was not used within Queue components.

ui/src/components/SampleQueue/Item.module.css
```diff
@@ -10,15 +10,63 @@
...
+  background-color: rgba(150, 198, 29, 0.52) !important;
+}
+.taskWarning {
+  background: linear-gradient(to bottom, #ffffff 0%, rgba(255, 255, 255, 0) 5%);
+  background-color: rgba(255, 212, 0, 0.39) !important;
+}
+.taskError {
+  background: linear-gradient(to bottom, #ffffff 0%, rgba(255, 255, 255, 0) 5%);
+  background-color: rgba(217, 83, 79, 0.56) !important;
+}
```
walesch-yan, line 72: These were previously named running, warning, success, etc. I renamed them to give them a name that matches better with their actual functionality

ui/src/components/SampleQueue/QueueControl.module.css
```diff
@@ -0,0 +1,9 @@
+.mTree {
+  position: relative;
+  user-select: none;
+  background-color: white;
+  padding: 0.5em;
+  border: 1px solid #aaa;
+  border-radius: 3px;
+  box-shadow: 0 1px 1px rgba(0, 0, 0, 0.05);
+}
```
walesch-yan, line 9: This one was overwritting its own style inside the app.css file, this is the result of both definitions

ui/src/components/SampleQueue/app.css
```diff
@@ -1,260 +0,0 @@
-*,
-*:before,
-*:after {
-  box-sizing: border-box;
-}
```
walesch-yan, line 5: This was added to the style of the div element at the top of the `SampleQueueContainer`

ui/src/components/SampleQueue/app.css
```diff
@@ -1,260 +0,0 @@
...
-.m-tree.queue-body > ul > li.active {
-  background-color: rgba(0, 0, 0, 0) !important;
-}
-.queue-controlls {
-  background-color: white;
-  border-radius: 0;
-}
-.fa {
-  margin-left: 5px;
-}
```
walesch-yan, line 66: This was a weird one, because it targets font-awesome items. I removed this, because a style that is targeting that many objects in the whole should not be defined in a CSS file that is hidden in a specific folder. I checked for a lot of icons before I removed it and they still seemed to be well placed (most were overwritting this anyway)

ui/src/containers/SampleQueueContainer.module.css
```diff
@@ -0,0 +1,72 @@
...
+  display: flex;
+  flex-direction: column;
+  width: 100%;
+}
+
+.container *,
+.container *::before,
+.container *::after {
+  box-sizing: inherit;
+}
```
walesch-yan, line 12: So this container is a merge of the styles that were directly added to the top-level div element and the rules that should be applied to all components in the queue

ui/src/containers/SampleQueueContainer.module.css
```diff
@@ -0,0 +1,72 @@
...
+  background-color: white;
+  padding-bottom: 0.5em;
+  padding-top: 0;
+  margin-top: 10px;
+  overflow-x: hidden;
+  flex-grow: 1;
+  border: 1px solid #aaa;
+  border-radius: 3px;
+  box-shadow: 0 1px 1px rgba(0, 0, 0, 0.05);
+}
```
walesch-yan, line 26: the queueBody here is a merge of the styles that previously defined m-tree(which already had two definitions in app.css) and queue-body, because the only component that used any of those two used both

ui/src/components/SampleQueue/app.css
axelboc: Huge win, I was too scared of doing this myself, so I'm glad you did :joy:

ui/src/components/SampleQueue/Item.module.css
axelboc: For future refactoring, maybe switching to a data attribute — e.g. `data-variant="success"`

ui/src/components/SampleQueue/app.css
axelboc: It's quite a common practice to have this block somewhere as a global style so it's applied to all elements, so I was a bit worried that you'd moved it.

It turns out that Bootstrap already applies this style, so we don't need to specify it again ourselves. 

image

In other words, you don't even need to have it in `SampleQueueContainer.module.css`. Notice how it's actually applied twice:

image

ui/src/components/SampleQueue/app.css
```diff
@@ -1,260 +0,0 @@
...
-  background-color: rgba(150, 198, 29, 0.52) !important;
-}
-.warning {
-  background: linear-gradient(to bottom, #ffffff 0%, rgba(255, 255, 255, 0) 5%);
-  background-color: rgba(255, 212, 0, 0.39) !important;
-}
-.error {
-  background: linear-gradient(to bottom, #ffffff 0%, rgba(255, 255, 255, 0) 5%);
-  background-color: rgba(217, 83, 79, 0.56) !important;
-}
```
axelboc, line 197: Slightly worried about removing these because it's the sort of global class names that I could see Bootstrap using inside it's components. Fortunately, it looks like Bootstrap typically prefixes its class names a little — e.g. when you pass `variant="success"` to a `Button`, it sets a class called `btn-success` and not `success`. I'm hopeful this is the case with other elements like `Alert`, `Badge`, etc. so I think we're good.

ui/src/containers/SampleQueueContainer.module.css
axelboc: As mentioned, I think this is safe to remove.

axelboc (commented):
Bold but necessary move, love it :star_struck:

PR #1781, opened 2025-07-15
Disable readonly HardwareObject-derived fields

mockoocy:
In the Task forms (DataCollection, Mesh, etc.) if energy is set to read-only, then corresponding form controls are disabled. It makes it so that, the same behaviour is applied for transmission and energy.

For context: at MicroMAX we set transmission to read-only, so we don't want this field to be editable.

### DIFF
--- a/ui/src/components/Tasks/Characterisation.jsx
+++ b/ui/src/components/Tasks/Characterisation.jsx
@@ -127,6 +127,9 @@ class Characterisation extends React.Component {
                 list={[1, 2, 4]}
               />
               <InputField
+                disabled={
+                  this.props.beamline.hardwareObjects.transmission.readonly
+                }
                 propName="transmission"
                 type="number"
                 label="Transmission"
--- a/ui/src/components/Tasks/DataCollection.jsx
+++ b/ui/src/components/Tasks/DataCollection.jsx
@@ -246,6 +246,9 @@ class DataCollection extends React.Component {
                 label="Exposure time (s)"
               />
               <InputField
+                disabled={
+                  this.props.beamline.hardwareObjects.transmission.readonly
+                }
                 propName="transmission"
                 type="number"
                 label="Transmission"
--- a/ui/src/components/Tasks/Helical.jsx
+++ b/ui/src/components/Tasks/Helical.jsx
@@ -142,6 +142,9 @@ class Helical extends React.Component {
                 label="Exposure time (s)"
               />
               <InputField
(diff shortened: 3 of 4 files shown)


elmjag (approved):

elmjag:
LGTM!

Maybe @axelboc want to take a look, before we merge it.

ui/src/components/Tasks/Characterisation.jsx
```diff
@@ -139,6 +142,9 @@ class Characterisation extends React.Component {
                 label="Exposure time (s)"
               />
               <InputField
+                disabled={
+                  this.props.beamline.hardwareObjects.resolution.readonly
+                }
```
beteva: resolution depends on the detector distance, not just the energy. I will be very surprised if it is ever read only

ui/src/components/Tasks/DataCollection.jsx
```diff
@@ -259,6 +262,9 @@ class DataCollection extends React.Component {
                 label="Energy"
               />
               <InputField
+                disabled={
+                  this.props.beamline.hardwareObjects.resolution.readonly
+                }
```
beteva: same as above.

ui/src/components/Tasks/Helical.jsx
```diff
@@ -155,6 +158,9 @@ class Helical extends React.Component {
                 label="Energy"
               />
               <InputField
+                disabled={
+                  this.props.beamline.hardwareObjects.resolution.readonly
+                }
```
beteva: same as above

ui/src/components/Tasks/Mesh.jsx
```diff
@@ -165,6 +168,9 @@ class Mesh extends React.Component {
                 label="Energy"
               />
               <InputField
+                disabled={
+                  this.props.beamline.hardwareObjects.resolution.readonly
+                }
```
beteva: same as above

ui/src/components/Tasks/Characterisation.jsx
mockoocy: Fair, I've added the option to disable resolution inputs for consistency with other fields that have their values derived from hardware objects (energy & transmission). I agree that, most likely, resolution will never be set as read only - removing this to avoid unnecessary confusion.

marcus-oscarsson:
Thanks @mockoocy

PR #1782, opened 2025-07-15
Hide the processing section of "Tasks" dialogs based on backend configuration

alessandro-olivo:
This feature allows hiding the "processing" section in the Task dialogs (i.e. DataCollection, Helical, and Mesh) based on the new acquisition parameter `disable_processing` that can be configured from the backend.

## HowTo
Update your `beamline_config.yaml` file like this:
```diff
   ...
  default_acquisition_parameters:
    default:
      ...
      take_snapshots: true
      sub_wedge_size: 10
+     disable_processing: false
      ...

      helical:
        # Defaults for helical scan. Missing values are taken from default
        number_of_images: 100
+       disable_processing: true
```
> The first `disable_processing: false` is not actually needed since "false" is the default value when not explicitly set in this file.

## Requirements
⚠️ This PR depends on https://github.com/mxcube/mxcubecore/pull/1331 (merge this first)


ui/src/components/Tasks/DataCollection.jsx
```diff
@@ -290,46 +290,50 @@ class DataCollection extends React.Component {
...
-                <InputField
-                  col1="1"
-                  col2="5"
-                  propName="cellGamma"
-                  label="&gamma;"
+          {!this.props.initialValues.disable_processing && (
+            <FieldsHeader title="Processing" />
+          )}
+          {!this.props.initialValues.disable_processing && (
+            <CollapsableRows>
```
walesch-yan: Maybe it is my personal opinion but I think it is a little cleaner and easier to change in the future if we wrap the header and rows in a div, which would result in not repeating the conditional logic, so something like
```js
{!this.props.initialValue.disable_processing && (
   <div>
        <FieldsHeader title="Processing" />
        <CollapsableRows>
        ....
        </CollapseRows>
   </div>
}
```

alessandro-olivo: I totally agree with you, it's a cleaner solution. Actually, I would have preferred to use only one if statement, but as far as I know, JSX doesn't allow two `tags` in the if statement. 
I'm not an expert of JSX, maybe there is another solution.

Anyway, the solution you suggested, @walesch-yan, makes sense to me.

alessandro-olivo: If that's okay with "everyone", I'd go with this solution

> 
> ```js
> {!this.props.initialValue.disable_processing && (
>    <div>
>         <FieldsHeader title="Processing" />
>         <CollapsableRows>
>         ....
>         </CollapseRows>
>    </div>
> }
> ```

walesch-yan:
Since this depends on mxcube/mxcubecore#1331, which has been merged by now, could you update the version of mxcubecore in `pyproject.toml` to the one that includes your backend changes? (should be 1.335.0) 

Afterwards, I guess you can rebase and merge:)

marcus-oscarsson:
Perfect, yes please go-ahead and re-base and merge

alessandro-olivo:
I've updated the tests, but there's still an issue with the CI checks. I think it might be related to the configuration, since the warning says: `Set the SONAR_TOKEN env variable.`

walesch-yan:
To me it seems that the problem is not directly the configuration, but the SonarQube token is needed for running the SonarQube CI. The token itself is set in the repo's secrets for the action to pick it up. However, GitHub is not passing secrets to a fork; hence, you cannot run the workflow on your PR, but it runs on all PRs done from a branch inside the repo. However, it will correctly run when we push the changes to develop.

 This is indeed a problem, we have not thought about when introducing it recently #1774 . As of now, I recon we can ignore it, the check is not required to pass anyways. Then we probably need to find a solution like only running it on a push.

walesch-yan:
Sorry, I was merging the current open PRs but should have started with this one. Could you rebase again @alessandro-olivo ?

alessandro-olivo:
@marcus-oscarsson Done! :)

walesch-yan:
Thanks @alessandro-olivo

Issue #1761, opened 2025-07-09
use `demo.yaml` configuration in CI E2E test


elmjag:
As far as I understand, when we run E2E tests in CI we are using the [`demo`](https://github.com/mxcube/mxcubeweb/blob/develop/.github/workflows/ui.yaml#L146) folder, that is the XML configuration files.

We should also run tests using `demo.yaml` folder, so that we check YAML configs as well.

As long as we support both XML and YAML configuration files, we should test _both_ of them. Obviously, once we drop support for XML configs, we can switch to test only YAML configs.

rhfogh:
I absolutely agree

PR #1783, opened 2025-07-15
Test both xml and yaml demo 

Phantominumm:
#1761 
As suggested in this issue, I change the workflow so the tests covers demo.yaml as well.

It just runs a second time with `demo.yaml` instead of `demo` when starting the server.

### DIFF
--- a/.github/workflows/ui.yaml
+++ b/.github/workflows/ui.yaml
@@ -141,7 +141,7 @@ jobs:
       - name: Build UI
         run: "${MAMBA_EXE} run --name mxcubeweb pnpm --prefix ui build"
 
-      - name: Start MXCuBE-Web server
+      - name: Start MXCuBE-Web server (XML config)
         run: |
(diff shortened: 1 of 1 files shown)


rhfogh (approved):
Looks good (not that I can evaluate any details, of course)

mockoocy (approved):

.github/workflows/ui.yaml
```diff
@@ -145,9 +145,16 @@ jobs:
         run: |
           ${MAMBA_EXE} run --name mxcubeweb mxcubeweb-server -r ./demo/ --static-folder $(pwd)/ui/build/ -L debug &
           ${MAMBA_EXE} run --name mxcubeweb pnpm --prefix ui exec wait-on http://127.0.0.1:8081
+          ${MAMBA_EXE} run --name mxcubeweb pnpm --prefix ui e2e
+        env:
+          CYPRESS_BASE_URL: http://127.0.0.1:8081
```
walesch-yan, line 150: Maybe we should change the name of this process from `Start MXCuBE-Web server` to `E2E (XML config)` as well then? otherwise looks good to me

PR #1786, opened 2025-07-16
cleanup styles of queue components Part 2

walesch-yan:
This PR is a follow-up to #1775. It includes the following changes concerning the CSS styles of the queue components:

- gathered duplicated styles applied in many files in the corresponding modules
- removed duplicated style parameters
- moved large style definitions to modules
- improved conditional styling in `TaskItemContainer`

### DIFF
--- a/ui/src/components/SampleQueue/CharacterisationTaskItem.jsx
+++ b/ui/src/components/SampleQueue/CharacterisationTaskItem.jsx
@@ -32,18 +32,7 @@ export default class TaskItem extends Component {
       return <span />;
     }
 
-    return (
-      <div
-        style={{
-          borderLeft: '1px solid #DDD',
-          borderRight: '1px solid #DDD',
-          borderBottom: '1px solid #DDD',
-          padding: '0.5em',
-        }}
-      >
-        {this.getDiffPlan(data)}
-      </div>
-    );
+    return <div className={styles.resultBody}>{this.getDiffPlan(data)}</div>;
   }
 
   getDiffPlan(data) {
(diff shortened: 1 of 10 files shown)


ui/src/components/SampleQueue/CharacterisationTaskItem.jsx
```diff
@@ -217,58 +206,42 @@ export default class TaskItem extends Component {
...
                     paddingTop: padding,
                   }}
                 >
-                  <div
-                    style={{
-                      borderTop: '1px solid #DDD',
-                      padding: '0.5em',
-                      display: 'flex',
-                      justifyContent: 'space-around',
-                      alignItems: 'center',
```
walesch-yan, line 243: Here and also in `TaskItem` I merged the styles and created 1 single div for both. That is also the reason why I increased the conditional padding in both cases by `0.5em`

ui/src/components/SampleQueue/TaskItem.jsx
```diff
@@ -197,51 +190,34 @@ export default class TaskItem extends Component {
...
-                      onClick={() => {
-                        navigator.clipboard.writeText(
-                          `${wedge.parameters.path}`,
-                        );
-                      }}
-                    >
-                      <i
-                        style={{ marginLeft: 0 }}
-                        className="fa fa-copy"
-                        aria-hidden="true"
```
walesch-yan, line 234: The removal of this `marginLeft` style is a consequence of removing the applied style on all font-awesome icons in #1775. Hence, setting `marginLeft` to 0 is not necessary anymore, as it is the default.

ui/src/components/SampleQueue/TaskItemContainer.jsx
```diff
@@ -37,54 +43,19 @@ export default class TaskItemContainer extends Component {
...
 
-    const delTaskCSS = {
-      display: 'flex',
-      marginLeft: 'auto',
-      alignItems: 'center',
-      paddingLeft: '10px',
-      paddingRight: '10px',
-      color: '#d9534f',
-      cursor: 'pointer',
-    };
```
walesch-yan, line 58: moved this to module

ui/src/components/SampleQueue/TaskItemContainer.jsx
```diff
@@ -37,54 +43,19 @@ export default class TaskItemContainer extends Component {
...
-      case TASK_COLLECT_FAILED: {
-        taskCSS += ` ${styles.taskError}`;
-        pbarBsStyle = 'danger';
-        break;
-      }
-      // No default
-    }
+    taskCSS += ` ${
+      warningTaskCSS ? styles.taskWarning : stateBasedStyles[state] || ''
+    }`;
```
walesch-yan, line 58: A few comments on this:
- instead of passing `specialTaskCSS` and `diffractionPLanLength`, now only one prop (`warningTaskCSS`) is passed if the warning style should be applied. Only `CharacterisationTaskItem` uses this prop.
- replaced the switch statement with a look-up table for styles.
- removed `pbarStyle` because the `ProgressBar` is only rendered if `state` equals `TASK_RUNNING` hence the conditional setting was unnecessary

ui/src/components/SampleQueue/TaskItemContainer.jsx
```diff
@@ -99,36 +70,26 @@ export default class TaskItemContainer extends Component {
...
+                  striped
+                  className={styles.progressBar}
+                  min={0}
+                  max={1}
+                  animated={progress < 1}
+                  label={`${(progress * 100).toPrecision(3)} %`}
+                  now={progress}
+                />
+              )}
+
```
walesch-yan, line 88: replaced `span` and `b` tags here with proper styling

PR #1787, opened 2025-07-16
update img-responsive to img-fluid

walesch-yan:
I saw this while working on #1786, but decided to make a separate PR since the scope for this is the whole repository. The utility class `img-responsive` is from [bootstrap v3](https://getbootstrap.com/docs/3.4/css/#images). In [bootstrapv5](https://getbootstrap.com/docs/5.3/content/images/) (which we use), `img-fluid` should be used.


jbflo:
Nice, i'll merge this easy one 😆..

PR #1788, opened 2025-07-16
Remove the incorrect 'defusedxml' dependency from `pyproject.toml`

alessandro-olivo:


marcus-oscarsson:
Thanks @alessandro-olivo, I was just about to do this :). Ill by you and @jbflo a beer on next MXCuBE meeting :beer:  :). My apologies again for not paying attention.

marcus-oscarsson:
@Phantominumm, I'm not sure why but it seems like we are now missing a SONAR_TOKEN env variable. It ran smoothly on your PR yesterday, so I'm not sure whats going on. Do perhaps need to update some token somewhere ?

Issue #1773, opened 2025-07-11
`limsResultData` disappeared from task state


dominikatrojanowska:
As in the title, `limsResultData` disappeared from task signal:
* https://github.com/mxcube/mxcubeweb/blob/7231e6076b64c0ffccbe8dca05ae639e42101349/mxcubeweb/routes/signals.py#L221
but fronted still looks for it:
* https://github.com/mxcube/mxcubeweb/blob/7231e6076b64c0ffccbe8dca05ae639e42101349/ui/src/serverIO.js#L181
* https://github.com/mxcube/mxcubeweb/blob/7231e6076b64c0ffccbe8dca05ae639e42101349/ui/src/serverIO.js#L212

As a result this link leads to nowhere:

<img width="1030" height="595" alt="Image" src="https://github.com/user-attachments/assets/51091f9b-3c15-4814-84ec-6f21b100193d" />

Was it made by mistake or decision?

dominikatrojanowska:
Seems like this commit https://github.com/mxcube/mxcubeweb/commit/6af4e5de242adc3affc62df2c43e79521671ac32 removed `limsResultData` from the event

marcus-oscarsson:
I don't think the change was made directly by me but touched by me when re-basing. The change was indeed intended, the link/view had not been functioning properly for some time. The front end bits can consequently be removed the link is no longer needed.

dominikatrojanowska:
ok, I will remove that feature from the front-end

marcus-oscarsson:
Thanks @dominikatrojanowska

PR #1790, opened 2025-07-17
Removing `View results` and `View results in IPSYB`

dominikatrojanowska:
Accordingly to the discussion https://github.com/mxcube/mxcubeweb/issues/1773 `limsResultData` was removed from the backend signals' messages in https://github.com/mxcube/mxcubeweb/commit/6af4e5de242adc3affc62df2c43e79521671ac32

This PR removes it and related features from the frontend. `limsResultData` is not stored in global state, and there is no option of `View results` and `View results in ISPYB` any more

Screenshot from 2025-07-17 15-00-23
Screenshot from 2025-07-17 15-00-35
.

### DIFF
--- a/ui/src/components/SampleGrid/TaskItem.jsx
+++ b/ui/src/components/SampleGrid/TaskItem.jsx
@@ -5,14 +5,12 @@ import React from 'react';
 import { Badge, OverlayTrigger, Popover } from 'react-bootstrap';
 
 import {
-  isUnCollected,
   TASK_COLLECT_FAILED,
   TASK_COLLECT_WARNING,
   TASK_COLLECTED,
   TASK_RUNNING,
   TASK_UNCOLLECTED,
 } from '../../constants';
-import loader from '../../img/busy-indicator.gif';
 import { LimsResultSummary } from '../Lims/LimsResultSummary';
 
 export class TaskItem extends React.Component {
@@ -30,7 +28,6 @@ export class TaskItem extends React.Component {
     this.summary = this.summary.bind(this);
     this.title = this.title.bind(this);
     this.stateClass = this.stateClass.bind(this);
-    this.result = this.result.bind(this);
   }
 
   tagName() {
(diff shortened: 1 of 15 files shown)


marcus-oscarsson:
Great @dominikatrojanowska, thank. Starts to look quite complete

marcus-oscarsson:
I think this is rather complete, you are missing this though: 
https://github.com/mxcube/mxcubeweb/blob/9428bf7d23dc604067d6c7e7be024a466567f2e0/ui/src/components/SampleQueue/TaskItem.jsx#L29

For all Queue items except for Characterization, where we still use the result (but its not the LIMS result)

dominikatrojanowska:
> I think this is rather complete, you are missing this though:
> 
> https://github.com/mxcube/mxcubeweb/blob/9428bf7d23dc604067d6c7e7be024a466567f2e0/ui/src/components/SampleQueue/TaskItem.jsx#L29
> 
> For all Queue items except for Characterization, where we still use the result (but its not the LIMS result)

it is related to showing up the `LimsResultDialog` and `LimsResultSummary`. Should I remove them as well?
image

walesch-yan:
So, I'm not sure, but since we removed the link, isn't the `LimsResultDialog` respectively `LimsResultSummary` not showing only the parameters used for the task? I don't know how useful that is. Maybe somebody has a strong opinion on keeping this?

dominikatrojanowska:
> So, I'm not sure, but since we removed the link, isn't the `LimsResultDialog` respectively `LimsResultSummary` not showing only the parameters used for the task? I don't know how useful that is. Maybe somebody has a strong opinion on keeping this?

I've just checked twice, indeed it doesn't make sense to keep it

ui/src/components/SampleGrid/TaskItem.jsx
```diff
@@ -271,17 +159,10 @@ export class TaskItem extends React.Component {
               className="p-2"
               id="taskSummaryPopover"
               title={<b>{this.title()}</b>}
-            >
-              <LimsResultSummary taskData={this.props.taskData} scale="0.5" />
-            </Popover>
+            />
```
walesch-yan: Oh I see, I was not aware of this. This here might be useful as it is a popover for the tasks in the sampleGrid tab:
Screenshot from 2025-07-21 16-20-11
So, I think we can keep that, as it can serve as a reminder, for the parameters set in the scheduled tasks. 
Sorry, I was overlooking this before. I also don't really know if anybody uses this, so I would advocate to keep it but maybe change the name of `LimsResultSummary` in that case, as it's more of a summary of parameters? But I do not have a strong opinion on this, so if you think it makes sense to remove it, we can do that

dominikatrojanowska: ok, I wasn't aware of it too, I can revert this part back then.

walesch-yan: No, it's in this repo and available in the mockup.

To see it, you have to:
- Add a task to the queue (i.e. either from SampleView or by right-clicking on the Item in `SampleGrid` (the item framed in the bordered line above and click add DataCollection))
- A small Badge (In the picture the one saying 'DC') will be added to the Item, which you can hover over to see the overview.

Normally, for every task you add, there will be a Badge. In the SampleView you have a better overview of them in general, but for the `SampleGrid` tab, this Popover might be useful:)

walesch-yan (approved):
Thanks @dominikatrojanowska, this looks pretty complete to me now

dominikatrojanowska:
https://github.com/mxcube/mxcubeweb/pull/1790#discussion_r2219388122
Renamed to `DataCollectionResult...` - I think this is last commit until this PR will get more comments :)

walesch-yan:
Thanks @dominikatrojanowska, I will merge this then:)

PR #1796, opened 2025-07-18
cleanup imports and pyproject.toml

walesch-yan:
This PR cleans up the `pyproject.toml` file by 

- removing unused dependencies from the file
- adding dependencies, that only existed as transitive dependencies in the codebase (imported in files, but only installed as subdependencies)

Additionally, I removed an import from `importlib_metadata` which was only used as a backport to python 3.7, which we dropped support for some time ago

FYI I used [deptry](https://deptry.com/) to identify the dependencies

### DIFF
--- a/mxcubeweb/__version__.py
+++ b/mxcubeweb/__version__.py
@@ -1,13 +1,7 @@
-try:
-    # Python3.8+ standard library
-    from importlib.metadata import (  # pyright: ignore[reportMissingImports]
-        PackageNotFoundError,
-        version,
-    )
-except ImportError:
-    # Python3.7 module backport
-    from importlib_metadata import version  # pyright: ignore[reportMissingImports]
-    from importlib_metadata import PackageNotFoundError
+from importlib.metadata import (
+    PackageNotFoundError,
+    version,
+)
 
 __version__: str
 try:
--- a/poetry.lock
+++ b/poetry.lock
@@ -48,65 +48,6 @@ test = ["anyio[trio]", "coverage[toml] (>=7)", "exceptiongroup (>=1.2.0)", "hypo
 trio = ["trio (>=0.26.1)"]
 
 [[package]]
-name = "argon2-cffi"
-version = "23.1.0"
-description = "Argon2 for Python"
-optional = false
-python-versions = ">=3.7"
-groups = ["main"]
-files = [
-    {file = "argon2_cffi-23.1.0-py3-none-any.whl", hash = "sha256:c670642b78ba29641818ab2e68bd4e6a78ba53b7eff7b4c3815ae16abf91c7ea"},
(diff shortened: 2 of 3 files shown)


marcus-oscarsson:
Very nice @walesch-yan, feel free to rebase and merge

walesch-yan:
So a short update on a few dependencies that were initially installed directly, namely `redis`, `flask-sqlalchemy` and `bcrypt`. Those were not directly used by us, but internally from `Flask-Security-Too` resp. `Flask-limiter`(depending on the config) for a better declaration in the `pyproject.toml` file, I added them as extras.

For the other removed packages a short explanation below:
- `gevent-websocket`: not used, and also no other dependency might use it as extra
- `mock`: was replaced in the repo by a module from `unittest`, which is in the standard lib
- `Pydispatcher` and `jsonschema`: those are direct dependencies of `mxcubecore` and should hence not be included here
- `argon2cffi`: this is an alternative to using bcrypt for password hashing with `Flask-security-Too` and also becomes the default installed with the common extra in more recent version. So in my opinion we could also bump `Flask-Security-Too` to use argon2 as default.

fabcor-maxiv:
Thanks for working on this! : )

I wish this list was alphabetically sorted. But maybe this should be a separate pull request.

Seems like `Flask-Security-Too` is now the same as `Flask-Security` (without the `-Too` suffix), but I do not know since which version. Maybe something to investigate and fix (if possible) in the context of this pull request.

walesch-yan:
@fabcor-maxiv my understanding is that they are the same from version 5.5.1 onwards, as it is the first version of `flask-security` that has been published to pypi since version 3.0.0. (you can checkout the [version history](https://pypi.org/project/Flask-Security/#history))

I have to resolve conflicts anyways due to merging other PRs (I wanted to wait for ongoing work to finish before merging this), so I can also order them alphabetically. I think PRs like this, that might introduce conflicts on whatever part you work on should be kept to a minimum. But thanks for the remark, I will update.

walesch-yan:
So conflicts are resolved and dependencies alphabetically sorted. On second thought I also removed the redis extra from flask-limiter, as the default is memory and `redis` would have to be configured on the machine the project runs on anyways. I'm not 100% sure if this is the way to go and I would like to hear another opinion on this.

For flask-security-too, since our version does not correspond to a version that also exists as flask-security, I did not change yet, as I guess the version update should a separate PR.

fabcor-maxiv (approved):
Seems fine to me.

> On second thought I also removed the redis extra from flask-limiter, as the default is memory and redis would have to be configured on the machine the project runs on anyways. I'm not 100% sure if this is the way to go and I would like to hear another opinion on this.

I guess/hope it will be easy enough to notice if turns out that this dependency is missing for someone and easy enough to add this dependency back in.

walesch-yan:
It normally should be the exception thrown should tell you that a certain module is not installed, but expected then one can always do `pip/poetry install` for a quick fix and/or add it back in this file

PR #1805, opened 2025-07-28
Fix  eslint destructuring-assignment

jbflo:
removed repeated ".props", access props from variable.

### DIFF
--- a/ui/src/components/ImageViewer/GalleryModal.jsx
+++ b/ui/src/components/ImageViewer/GalleryModal.jsx
@@ -1,3 +1,4 @@
+import PropTypes from 'prop-types';
 import { useEffect, useRef } from 'react';
 import { Modal } from 'react-bootstrap';
 import { TransformComponent, TransformWrapper } from 'react-zoom-pan-pinch';
@@ -35,13 +36,10 @@ export default function GalleryModal(props) {
     context.lineWidth = 2;
 
     context.beginPath();
-
     context.moveTo(imgTargetX - 20, imgTargetY - 20);
     context.lineTo(imgTargetX + 20, imgTargetY + 20);
-
     context.moveTo(imgTargetX + 20, imgTargetY - 20);
     context.lineTo(imgTargetX - 20, imgTargetY + 20);
-
     context.stroke();
   }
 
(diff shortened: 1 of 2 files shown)


walesch-yan (approved):
Looks good to me! 

Regarding the failing check: Sonarcloud would like us to use proptypes or typescript with type validation for the props. Since we are not doing this in the rest of the project, I think it's fine here. Probably something we can introduce later down the road

jbflo:
> Looks good to me!
> 
> Regarding the failing check: Sonarcloud would like us to use proptypes or typescript with type validation for the props. Since we are not doing this in the rest of the project, I think it's fine here. Probably something we can introduce later done the road

Yeah,  I added the proptype to satisfy Sonacube.
but it was a bit weird seeing this rule.

there is a lot of /* eslint-disable react/destructuring-assignment */
That mean people will have to comply by typing the props when fixing destructuring-assignment

ui/src/components/ImageViewer/GalleryModal.jsx
```diff
@@ -55,14 +60,22 @@ export default function GalleryModal(props) {
...
         </TransformWrapper>
       </Modal.Body>
     </Modal>
   );
 }
+
+GalleryModal.propTypes = {
+  handleClose: PropTypes.func.isRequired,
+  isOpen: PropTypes.bool.isRequired,
+  drawTarget: PropTypes.bool,
```
jbflo, line 77: I am wondering if we should use this fix. 
but as we are not moving to TypeScript soon

mockoocy: I think it does not hurt to have this anyway.
Alternatively it could possibly be replaced with a JSDoc type hint, something like

```js
/**
* Summary
*
* @typedef {object} GalleryModalProps
* @property {() => void} handleClose description
* @property {boolean} isOpen description
* ...
* 
* @param {GalleryModalProps} props
*/
```
I am not sure if it's understood by Sonarqube though ( but at least is understood by popular code editors during static code analysis (via typescript language server AFAIK) ).

jbflo: i'll leave the propTypes for now. i think it might be easier to switch to typeScript when we upgrade to react 19.

mockoocy: Since React docs mention [some way](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-deprecated-react-apis) to convert prop-types to typescript interfaces, I'm okay with it.
For most files where we just define a component, having the type definitions just for props could be enough.
But in other places we could think of using `JSDoc` since [it can be checked](https://www.typescriptlang.org/docs/handbook/intro-to-js-ts.html) by TypeScript and supports the same type operations (like intersection types, conditional types, etc.)

mockoocy (approved):

PR #1807, opened 2025-07-29
[refactor]  DataCollectionResult Transform to function component 

jbflo:
- [refactor] Transform to function component 
 - fix destructuring-assignment
 - Transform to module.css
 - Access props directly from state with useSelector and useDispatch]

### DIFF
--- a/ui/src/components/DataCollectionResult/DataCollectionResultDialog.jsx
+++ b/ui/src/components/DataCollectionResult/DataCollectionResultDialog.jsx
@@ -1,25 +1,43 @@
+import { useCallback } from 'react';
 import { Button, Modal } from 'react-bootstrap';
+import { useDispatch, useSelector } from 'react-redux';
 
+import { showDialog } from '../../actions/general';
 import styles from './dataCollectionResult.module.css';
 import { DataCollectionResultSummary } from './DataCollectionResultSummary';
 
-export function DataCollectionResultDialog(props) {
-  const { show, onHide, taskData } = props;
+export function DataCollectionResultDialog() {
+  const dispatch = useDispatch();
+
+  const taskData = useSelector((state) => state.general?.dialogData || {});
+  const show = useSelector(
+    (state) => state.general?.dialogType === 'LIMS_RESULT_DIALOG',
+  );
+
+  const handleHideDialog = useCallback(() => {
+    dispatch(showDialog(false));
+  }, [dispatch]);
+
+  if (!show) {
+    return null;
+  }
 
   return (
     <Modal
       dialogClassName={styles.dc_result_dialog}
       show={show}
-      onHide={onHide}
+      onHide={handleHideDialog}
     >
       <Modal.Header closeButton>
-        <Modal.Title>Result summary : {taskData?.label}</Modal.Title>
+        <Modal.Title>
+          Result summary : {taskData?.label || 'Untitled Collection'}
+        </Modal.Title>
       </Modal.Header>
       <Modal.Body>
-        {taskData && <DataCollectionResultSummary taskData={taskData} />}
+        <DataCollectionResultSummary />
       </Modal.Body>
       <Modal.Footer>
-        <Button onClick={onHide}>Close</Button>
+        <Button onClick={handleHideDialog}>Close</Button>
       </Modal.Footer>
     </Modal>
   );
--- a/ui/src/components/DataCollectionResult/DataCollectionResultDialog.jsx
+++ b/ui/src/components/DataCollectionResult/DataCollectionResultDialog.jsx
@@ -1,9 +1,9 @@
 /* eslint-disable react/destructuring-assignment */
-import './DataCollectionResultDialog.css';
 
 import React from 'react';
 import { Button, Modal } from 'react-bootstrap';
 
+import styles from './dataCollectionResult.module.css';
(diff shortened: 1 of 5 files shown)


ui/src/components/DataCollectionResult/DataCollectionResultSummary.jsx
```diff
@@ -42,3 +44,19 @@ export function DataCollectionResultSummary(props) {
...
+    parameters: PropTypes.shape({
+      fullPath: PropTypes.string,
+      osc_range: PropTypes.number,
+      first_image: PropTypes.number,
+      osc_start: PropTypes.number,
+      num_images: PropTypes.number,
+      exp_time: PropTypes.number,
+      transmission: PropTypes.number,
+      energy: PropTypes.number,
+      resolution: PropTypes.string,
```
beteva: Is the resolution really a string?

jbflo: > Is the resolution really a string?

thanks @beteva

ui/src/components/DataCollectionResult/DataCollectionResultDialog.jsx
```diff
@@ -1,40 +1,36 @@
...
-  constructor(props) {
-    super(props);
-    this.onHide = this.onHide.bind(this);
-  }
-
-  onHide() {
-    this.props.onHide();
-  }
+export function DataCollectionResultDialog(props) {
+  const { show, onHide, taskData } = props;
```
walesch-yan: Maybe, we can make use of `useSelector` and `useDispatch` for the props. It would remove the necessity to pass them from the parent, as all they are already in the state. Additionally, we could then restrict the general selector in [`Main.jsx`](https://github.com/mxcube/mxcubeweb/blob/6bf391b2937eaef665745ab8acab750807a3f98c/ui/src/components/Main.jsx#L35) to `general.applicationFetched`

jbflo: very good idea. i wanted to avoid touching file from other component for now. to keep the PR as light as possible. in another PR maybe ?

walesch-yan: I see, I guess it can also be a follow-up PR if you prefer that:)

jbflo: > I see, I guess it can also be a follow-up PR if you prefer that:)

hmmm Wait , i'll give it a try, maybe that will prevent us from validate with PropType

jbflo: > Maybe, we can make use of `useSelector` and `useDispatch` for the props. It would remove the necessity to pass them from the parent, as all they are already in the state. Additionally, we could then restrict the general selector in [`Main.jsx`](https://github.com/mxcube/mxcubeweb/blob/6bf391b2937eaef665745ab8acab750807a3f98c/ui/src/components/Main.jsx#L35) to `general.applicationFetched`

@walesch-yan  i follow your suggestion and use the state instead of passing the props. and i as i thought it pass SonarCube without the need of Proptype.

walesch-yan (approved):

ui/src/components/DataCollectionResult/DataCollectionResultDialog.jsx
walesch-yan: Perfect, thanks @jbflo

mockoocy (approved):

ui/src/components/DataCollectionResult/DataCollectionResultDialog.jsx
```diff
@@ -1,40 +1,44 @@
...
-          ) : null}
-        </Modal.Body>
-        <Modal.Footer>
-          <Button onClick={this.onHide}>Close</Button>
-        </Modal.Footer>
-      </Modal>
-    );
+  const handleHideDialog = useCallback(() => {
+    dispatch(showDialog(false));
+  }, [dispatch]);
```
axelboc, line 19: I wouldn't have bothered with a `useCallback` (just an inline arrow function) but no biggy.

ui/src/components/DataCollectionResult/dataCollectionResult.module.css
axelboc, line 1: Filename should be uppercase by convention

ui/src/components/DataCollectionResult/DataCollectionResultSummary.jsx
```diff
@@ -1,42 +1,47 @@
...
-          <span className="col-sm-3">Resolution</span>
-          <span className="col-sm-3">{`${task.parameters.resolution} Å`}</span>
-        </div>
-      </div>
-    );
-  }
+      <span className="col-sm-3">Exposure time:</span>
+      <span className="col-sm-3">{`${exp_time}s`}</span>
+      <span className="col-sm-3">Transmission</span>
+      <span className="col-sm-3">{`${transmission} %`}</span>
```
axelboc, line 39: FYI, I think this does the same thing:

```jsx
<span className="col-sm-3">{transmission} %</span>
```

axelboc (commented):
_Two weeks later_

Awesome! :grin:

PR #1808, opened 2025-07-30
Removes unused props from `SampleViewContainer`

mockoocy:
The `cellSpacing` property is no longer used.
Cell spacing is now managed by the `drawGridPlugin` in the `gridData` property.

### DIFF
--- a/ui/src/containers/SampleViewContainer.jsx
+++ b/ui/src/containers/SampleViewContainer.jsx
@@ -58,7 +58,6 @@ class SampleViewContainer extends Component {
     if (!('sample_view_motors' in uiproperties)) {
       return null;
     }
-
     const { sourceScale, imageRatio } = this.props.sampleViewState;
     const { currentSampleID } = this.props;
     const [points, lines, grids, twoDPoints] = [{}, {}, {}, {}];
@@ -207,7 +206,6 @@ class SampleViewContainer extends Component {
                 grids={grids}
                 selectedGrids={selectedGrids}
                 cellCounting={this.props.cellCounting}
-                cellSpacing={this.props.cellSpacing}
                 busy={this.props.queueState === QUEUE_RUNNING}
                 setAttribute={this.props.setAttribute}
                 displayImage={this.props.displayImage}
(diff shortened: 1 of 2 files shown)


ui/src/containers/SampleViewContainer.jsx
```diff
@@ -242,8 +240,6 @@ function mapStateToProps(state) {
     hardwareObjects: state.beamline.hardwareObjects,
     defaultParameters: state.taskForm.defaultParameters,
     shapes: state.shapes.shapes,
-    cellCounting: state.taskForm.defaultParameters.mesh.cell_counting,
```
walesch-yan, line 245: I do get the removal of `cellSpacing`, but `cellCounting` is still passed and used right? Maybe you removed this line by accident?

mockoocy: This was not intended for this PR  - but it does not seem to break anything :D
This comes  from an 'investigation' of these two lines
```js
cellCounting: state.taskForm.defaultParameters.mesh.cell_counting,
cellSpacing: state.taskForm.defaultParameters.mesh.cell_spacing,
```
These two seem to always evaluate to `undefined`.
Since they're read from `beamline_config.yml` in the [`default_acquisiton_parameters.mesh`](https://github.com/mxcube/mxcubeweb/blob/6bf391b2937eaef665745ab8acab750807a3f98c/demo.yaml/beamline_config.yml#L121-L122)
These seem to be read from config [there](https://github.com/mxcube/mxcubeweb/blob/6bf391b2937eaef665745ab8acab750807a3f98c/mxcubeweb/core/components/queue.py#L2180-L2214)
And saved in the redux store [there](https://github.com/mxcube/mxcubeweb/blob/6bf391b2937eaef665745ab8acab750807a3f98c/ui/src/reducers/taskForm.js#L34-L41)
So they should be available under 
```js
cellCounting: state.taskForm.defaultParameters.mesh.acq_parameters.cell_counting,
cellSpacing: state.taskForm.defaultParameters.mesh.acq_parameters.cell_spacing,
```
I've made sure that `cellSpacing` is safe to remove, but I am yet to look deeper into `cellCounting`  :)

_I've removed the `cellCounting` change from this PR_

walesch-yan: Very interesting, it is indeed looking in the wrong place in the state. Thanks for taking a look into that @mockoocy :)

walesch-yan (approved):

PR #1810, opened 2025-07-30
refactor TaskItemContainer into functional component

walesch-yan:
This PR transforms the `TaskItemContainer` into a functional component while reducing prop drilling through multiple levels of components.

I also moved down functions that only affected `TaskItemContainer` from `CurrentTree` down to this component

### DIFF
--- a/ui/src/components/SampleQueue/TaskItemContainer.jsx
+++ b/ui/src/components/SampleQueue/TaskItemContainer.jsx
@@ -1,6 +1,9 @@
-import { Component } from 'react';
 import { Collapse, ProgressBar } from 'react-bootstrap';
+import { contextMenu } from 'react-contexify';
+import { useDispatch, useSelector } from 'react-redux';
 
+import { deleteTask } from '../../actions/queue';
+import { collapseItem, selectItem } from '../../actions/queueGUI';
 import {
   TASK_COLLECT_FAILED,
   TASK_COLLECTED,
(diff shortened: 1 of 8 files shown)


ui/src/components/SampleQueue/TaskItemContainer.jsx
```diff
@@ -15,88 +18,116 @@ const stateBasedStyles = {
...
-    taskHeaderOnClickHandler(e, index);
+  function handleContextMenu(e) {
+    e.preventDefault();
+    contextMenu.show({
+      id: 'currentSampleQueueContextMenu',
+      event: e,
+      props: {
+        taskIndex: index,
+      },
+    });
```
walesch-yan, line 64: So before there were two contextmenu handlers. One for showing the `contextmenu` which uses react-contexify and is defined in `Currenttree` component. And another for setting the internal state of `Currenttree` to use the index of the task that tries to show the context menu. Therefore we had to pass the function that sets the state down the component tree. 

This solution sends the index as a prop to the contextmenu directly making use of `react-contexify`. It prevents some more prop drilling and we don't have to keep track of the internal stat of the `Currenttree` component anymore

ui/src/components/SampleQueue/TaskItemContainer.jsx
```diff
@@ -15,88 +18,116 @@ const stateBasedStyles = {
...
+  // if a diffraction task is collected but no diffraction plan is present
+  if (
+    data.type === 'Characterisation' &&
+    taskState === TASK_COLLECTED &&
+    data.diffractionPlan.length === undefined
+  ) {
+    taskCSS += ` ${styles.taskWarning}`;
+  } else {
+    taskCSS += ` ${stateBasedStyles[taskState] || ''}`;
+  }
```
walesch-yan, line 101: Here I made the check directly in this component instead of passing the prop, as there is only one case, were the special css is used

ui/src/components/SampleQueue/TaskItemContainer.jsx
```diff
@@ -15,88 +18,135 @@ const stateBasedStyles = {
...
+          onKeyDown={taskHeaderOnKeyDown}
+          aria-expanded={!displayData.collapsed}
+          aria-controls={`collapse-${data.queueID}`}
+          className={taskCSS}
+          style={{ display: 'flex' }}
         >
-          <div
-            onClick={this.taskHeaderOnClick}
-            onContextMenu={this.taskHeaderOnContextMenu}
-          >
```
walesch-yan, line 71: This change was added after `Sonarqube` popped up a warning about accessibility of the header, which is a valid proposal for a change, in my opinion. This change makes the element accessible for screen-readers and keyboard-only users. I did not change to a real button, as this would have completely messed up the styles, that we use on the item

ui/src/components/SampleQueue/TaskItemContainer.jsx
marcus-oscarsson: Great !

marcus-oscarsson:
Thanks, very nice !

PR #1811, opened 2025-07-31
remove withNavigate function

walesch-yan:
This removes the `withNavigate` function from the project. It was initially used to wrap class components that could not use `useNavigate`; however, all those that used it have been transformed to functional components in the meantime. Hence, making this function redundant.


marcus-oscarsson:
Good initiative, well done !

PR #1813, opened 2025-08-01
[Fix] Prevents a crash when clicking the table inside a task item in the queue

alessandro-olivo:
Thanks to @walesch-yan for pointing out the issue suggested by the title. Here's a short video explaining it: [Screencast from 2025-07-28 17-29-10.webm](https://github.com/user-attachments/assets/6112581d-7153-4a8e-a26f-ac940d5aabd4)

This bug was introduced by me :sweat_smile:  with this PR #1723

I decided to use the `detector_mode_list` parameter directly from `acq_parameters`, instead of the one from `taskData`, since it's not really task data, at least it is not something that can be changed like the `detector_roi_mode`. 
This list is more like the `limits` of the range of the motors.

### DIFF
--- a/ui/src/components/Tasks/Characterisation.jsx
+++ b/ui/src/components/Tasks/Characterisation.jsx
@@ -173,11 +173,11 @@ class Characterisation extends React.Component {
                 <InputField propName="kappa_phi" type="number" label="Phi" />
               </FieldsRow>
               <FieldsRow>
-                {this.props.initialValues.detector_mode_list.length > 0 && (
+                {this.props.detector_mode_list.length > 0 && (
                   <SelectField
                     propName="detector_roi_mode"
                     label="Detector mode"
-                    list={this.props.initialValues.detector_mode_list}
+                    list={this.props.detector_mode_list}
                   />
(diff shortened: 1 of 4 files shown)


alessandro-olivo:
For PR like this, should I also open a related issue? Just asking to understand the best practice. :)

fabcor-maxiv:
> For PR like this, should I also open a related issue? Just asking to understand the best practice. :)

From my point of view, this is not necessary to open an issue ticket, since you already have the pull request with the fix.

walesch-yan (approved):
Looks good to me! Indeed taking the list from the `defaultparameters` makes more ense, as you suggest.

Regarding the question about opening an issue: I agree with @fabcor-maxiv, as you already have a solution for it. I think creating an issue would be rather something you should do when you find a problem, that you would like to have the opinion of others on it or you think it might take some time to implement a solution and want others to be aware of it resp. prevent a scenario where multiple people work on the same issue
Long story short: I believe the way you handled this is perfectly fine by shortly restating the problem in the description and opening a seperate issue is not necessary :)

PR #1814, opened 2025-08-04
Fix sample order not being set in some cases

mockoocy:
There are two different ways the `/queue_state` response is handled in MXCuBE.
First one is on initial load of the application. The endpoint is being called, then both `sampleOrder` and `sampleList` are saved into redux store here:
https://github.com/mxcube/mxcubeweb/blob/da92c63661ab502512a5fa1707e9b95872cceaf8/ui/src/reducers/sampleGrid.js#L315-L319 The other way occurs when a signal is emitted from back-end (either 'queue' or 'update_queue')
These call `getQueue` action:
https://github.com/mxcube/mxcubeweb/blob/da92c63661ab502512a5fa1707e9b95872cceaf8/ui/src/actions/queue.js#L495-L504 which then invokes `SET_QUEUE` in `sampleGrid` reducer - it only saves the `sampleList` in the redux store, the information about `sampleOrder` is ignored.
It seems wrong and leads to a confusing bug when using session type login.
On session selection, `getQueue` after `update_queue` is emitted https://github.com/mxcube/mxcubeweb/blob/390d60a5493a137bec01eae3aefdb0d6b5bdcfd9/mxcubeweb/core/components/lims.py#L223 This leads to this issue:
1. User logs in
2. `sampleList` and `sampleOrder` are initialized with empty values
3. User selects a session
4. Only `sampleList` is set, `sampleOrder` is ignored.
5. Samples are displayed, but due to order being empty, nothing happens

### DIFF
--- a/ui/src/reducers/sampleGrid.js
+++ b/ui/src/reducers/sampleGrid.js
@@ -42,6 +42,7 @@ function sampleGridReducer(state = INITIAL_STATE, action = {}) {
       return {
         ...state,
         sampleList: { ...action.sampleList },
+        order: action.sampleOrder,
       };
     }
     // Set the list of samples (sampleList), clearing any existing list


jbflo (approved):

walesch-yan (approved):
Thanks @mockoocy that's quite a good catch. If you want you can rebase and merge:)

mockoocy:
> Thanks @mockoocy that's quite a good catch. If you want you can rebase and merge:)

I've done it, but a CI job failed on the `develop` branch; see: https://github.com/mxcube/mxcubeweb/issues/1815

marcus-oscarsson:
Thanks, well done. I don't know why but I thought there was already a fix made for this ... perhaps to much salt water during vacation :)

jbflo:
@mockoocy 
This Pr was very good, as it make sure the Sample Order from the backend match the Ui State

I think some logic on the Ui was not expected that sampleOrder to change
SO It seems this fix cause at least 2 issues, 

1 - You can no longer use the mouse to select multiple items that are not in Sample Order in the Sample Grid.
2 - You can’t mount items that are not in Sample Order.

have you notice those issues  ?

marcus-oscarsson:
There is indeed an issue with the change introduced in this PR. We have reverted this commit in our production branch, otherwise we do indeed have the issues the @jbflo outlines above.

I have not yet had the time to get back to you regarding this, but well now @jbflo beat me to it ;)

mockoocy:
> 1 - You can no longer use the mouse to select multiple items that are not in Sample Order in the Sample Grid. 2 - You can’t mount items that are not in Sample Order.
>  2 - You can’t mount items that are not in Sample Order.
> have you notice those issues ?

We've lately found some similar issue, where doing a data collection caused similar issues. It's probably related.
I'll try to take a look what may be causing it.

PR #1816, opened 2025-08-06
fix poetry version in create tag workflow

### LINKED ISSUE #1815: Create/Update Tag CI job fails
After merging this PR: https://github.com/mxcube/mxcubeweb/pull/1814 the pipeline on `develop` branch fails on the Update/Create tag job. The job output is available [here](https://github.com/mxcube/mxcubeweb/actions/runs/16772334430/job/47490249445)

An interesting bit from this error is that part

```
This project requires Poetry 2.1.3, but you are using Poetry 2.1.4
Error: Process completed with exit code 1.
```
Possibly we are pulling the latest version of poetry in our CI job somewhere? It would make sense, the 2.1.4 version has been released [yesterday](https://pypi.org/project/poetry/#history)

walesch-yan:
This PR closes #1815.

As mentioned in the issue, the workflow for creating the version tag was failing due to a poetry mismatch. This PR fixes this by fixing the version to be the same as in the conda environment. For further information take a look at #1815 or mxcube/mxcubecore#1344

### DIFF
--- a/.github/workflows/bump_version_and_tag.yaml
+++ b/.github/workflows/bump_version_and_tag.yaml
@@ -8,46 +8,50 @@ name: Create/Update Tag
     branches:
       - develop
 
-permissions:
-  contents: write
-  packages: write
-  id-token: write
-
 jobs:
   create-version-tag:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      packages: write
     steps:
       - uses: actions/checkout@v4
         with:
           token: ${{ secrets.CI_TOKEN }}
-      - name: Set up Python 3.10
-        uses: actions/setup-python@v5
+
+      - name: Set up Conda environment with Micromamba
+        uses: mamba-org/setup-micromamba@v2
         with:
-          python-version: "3.10"
+          cache-environment: true
+          create-args: >-
+            python=3.10
+          environment-file: conda-environment.yml
+          post-cleanup: all
+
       - name: Update version
         id: update-version
         run: |
-          pip install --upgrade pip
-          pip install poetry
-          poetry version minor
+          ${MAMBA_EXE} run --name mxcubeweb poetry version minor
           git config --global user.email "oscarsso@esrf.fr"
           git config --global user.name "Marcus Oskarsson"
           git add -A
           git commit -m "[skip ci] Bumped minor version"
           git push -f
-          poetry build
+          ${MAMBA_EXE} run --name mxcubeweb poetry build
+
       - name: Publish package to PyPI
-        id: publish-pacakge
+        id: publish-package
         run: |
-          poetry config pypi-token.pypi ${{ secrets.PYPI }}
-          poetry publish
+          ${MAMBA_EXE} run --name mxcubeweb poetry config pypi-token.pypi ${{ secrets.PYPI }}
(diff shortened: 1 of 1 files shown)


.github/workflows/bump_version_and_tag.yaml
```diff
@@ -8,46 +8,50 @@ name: Create/Update Tag
...
-  contents: write
-  packages: write
-  id-token: write
-
 jobs:
   create-version-tag:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      packages: write
```
walesch-yan, line 16: Moved the permissions down to the job level and removed the id-token permission, it does not seem to be needed for this workflow

.github/workflows/bump_version_and_tag.yaml
```diff
@@ -8,46 +8,50 @@ name: Create/Update Tag
...
           git config --global user.email "oscarsso@esrf.fr"
           git config --global user.name "Marcus Oskarsson"
           git add -A
           git commit -m "[skip ci] Bumped minor version"
           git push -f
           poetry build
+
       - name: Publish package to PyPI
-        id: publish-pacakge
+        id: publish-package
```
walesch-yan, line 43: Also took the opportunity to fix the typo

.github/workflows/bump_version_and_tag.yaml
```diff
@@ -8,46 +8,50 @@ name: Create/Update Tag
...
-          pip install --upgrade pip
-          pip install poetry
-          poetry version minor
+          ${MAMBA_EXE} run --name mxcubeweb poetry version minor
           git config --global user.email "oscarsso@esrf.fr"
           git config --global user.name "Marcus Oskarsson"
           git add -A
           git commit -m "[skip ci] Bumped minor version"
           git push -f
           poetry build
```
fabcor-maxiv: Missed this one?

```suggestion
          ${MAMBA_EXE} run --name mxcubeweb poetry build
```

.github/workflows/bump_version_and_tag.yaml
```diff
@@ -8,46 +8,50 @@ name: Create/Update Tag
     branches:
       - develop
 
-permissions:
-  contents: write
-  packages: write
-  id-token: write
```
fabcor-maxiv, line 14: Should the `id-token` permission be moved somewhere? Or was it not needed to begin with?

.github/workflows/bump_version_and_tag.yaml
walesch-yan: Yes, thanks!

.github/workflows/bump_version_and_tag.yaml
walesch-yan: As of my understanding it is not needed to begin with, it should be needed for communicating with third party services. So maybe the idea was for communication with Pypi, but we are using the github secret in the workflow and no id-token for that. So maybe I'm wrong but if we compare with the CI in `mxcubecore`, we do not seem to need the permission there either for the exact same workflow.

fabcor-maxiv: OK

fabcor-maxiv (approved):
Seems fine to me

mockoocy (approved):

PR #1740, opened 2025-07-02
Make the "Phase Control" and "Beam Size" sections of the UI configurable from the "ui.yaml" file

alessandro-olivo:
This PR is a follow-up to this one: https://github.com/mxcube/mxcubeweb/pull/1699

## The feature
It allows configuring the "Phase Control" and "Beam Size" sections from the `ui.yaml` file. It is possible to hide one or both the sections if needed or change the label. 

### Examples:
1. Change "Beam Size" label 
   In `ui.yaml`
   > ```yaml
   > sample_view:
   >   id: sample_view
   >   components:
   >     - attribute: phase_control
   >       label: Phase Control
   >     - attribute: beam_size
   >       label: Beam Size (um)
   >  ```
   

2. Hide "Phase Control" 
   In `ui.yaml`
   > ```yaml
   > sample_view:
   >  id: sample_view
   >  components:
   >    - attribute: beam_size
   >      label: Beam Size
   > ```
   

## Notes:
* A new `UISampleViewMotorsModel` has been introduced to avoid changing too much the motor rendering logic. 
* The logic "whatever is configured in the `ui.yaml` gets shown" has been preserved
* I used this
   ```yaml
   sample_view:
     id: sample_view
     components:
       - attribute: phase_control
         label: Phase Control
    ```
    instead of this
    ```yaml
    sample_view:
      id: sample_view
      components:
        - id phase_control
          show: true
    ```
   to make fewer changes and avoiding extra logical `show/noshow` level (as suggested by @beteva) furthermore, I already know that the "label" customization is something that will be used on our site :) and this attribute was already part of the `UIComponentModel`

Issue #1725, opened 2025-06-26
Structure of `ui.yaml`


axelboc:
A discussion started in #1699 concerning the structure of `ui.yaml`. I think I've mentioned before that I'm not a huge fan of the following structure:

```yml
<namespace>:
  id: <namespace>
  components:
    - id: <some-id>
      <some-properties>
    - id: <some-id>
      <some-properties>
```

Two reasons:

- In the front-end, we end up having to loop through the components to find the one with the ID we want.
- If/when we introduce typing, it will be a nightmare, because it will look something like: `components: { id: string; [key: string]: any }[]`, which is not very helpful.

I would prefer to have a clear, static structure that is tailored to the needs of the front-end:

```yml
sampleViewVideoControls: # camelCase
  snapshot:
    show: true
  drawGrid:
    show: true
    showVspace: false
    showHspace: true
  ...  
```

```ts
const showSnapshot = useSelector(state => state.uiproperties.sampleViewVideoControls.snapshot.show);

// instead of
const showSnapshot = useSelector(state => state.uiproperties.sampleViewVideoControls.components.find(c => c.id === 'snapshot').show);
```

---

I would also be in favor of having a near-zero config experience out of the box, which means having a default config that gets merged/overridden by the content of `ui.yaml` (in JS below, but could also be done in the back-end of course):

```ts
const DEFAULT_CONFIG = {
  sampleViewVideoControls: {
    snapshot: {
      show: true
    },
    // ...
  },
  // ...
};

const uiproperties = merge(DEFAULT_CONFIG, uiConfig);
```

**EDIT:** To clarify, this idea of a default config is to avoid putting that logic throughout the codebase. For instance, we really don't want the sample view component to have to know that if `snapshot: { show: true }` is not defined in the config, it means that the snapshot control should be shown...

---

Finally, we may want to think about versioning the configuration. Without versioning, we tend to fear making breaking changes to the config structure, and this is when technical debt creeps up. With versioning, we can support a few versions easily to give people time to upgrade — upgrading the config can then be done separately from upgrading MXCuBE-Web:

```js
function upgradeUiConfig(uiConfig) {
  if (uiConfig.version === '2.1') {
    return {
      ...omit(uiConfig, 'oldProperty'),
      newProperty: uiConfig.oldProperty
    }
  }
}
```

marcus-oscarsson:
work started in: #2068

PR #1819, opened 2025-08-08
Make SampleListViewContainer viewMode options configurable from ui.yaml

matyson:
At MaxIV we want to be able to hide the `Graphical View` from the SampleListView available `viewMode` options. So, my proposal was to make the whole view configurable from the `ui.yaml` file, inspired by this [PR](https://github.com/mxcube/mxcubeweb/pull/1740).

So, i introduced this new section `sample_list_view` in the yaml file and moved the `viewMode` away from `sampleGrid`  reducer, making necessary adjustments:

```yaml
sample_list_view:
  id: sample_list_view
  options:
    - Table View
    - Graphical View
```

Then, by removing the `Graphical View` from this options list we display only the `Table View` for the users, disabling the dropdown menu selector.

image

image

I aimed at a solution with few changes, but it didn't seemed have an easy one, so if you have other suggestions i am willing to try them 😄


mxcubeweb/core/models/configmodels.py
```diff
@@ -118,6 +118,14 @@ class UISampleViewMotorsModel(UIPropertiesModel):
     components: list[UIComponentModel]
 
 
+class UISampleListViewModel(BaseModel):
+    id: Literal["sample_list_view"] = "sample_list_view"
+    options: list[str] = Field(
+        ["Table View", "Graphical View"],
+        description="List of available views for the sample list",
+    )
+
```
walesch-yan: Since there are only two possible values for the sample list view, maybe it would be better to use an `enum` and make options, then a list of said `enum`?

Currently, the front-end Component only checks for "Graphical View"; everything else will be rendered as "Table View" (I'm talking about the components in the grid not the name in the options field). Having the enum would ensure that the correct terminology is used (before it was hardcoded, so that was less of a problem :sweat_smile: )

ui/src/containers/SampleListViewContainer.jsx
```diff
@@ -72,15 +74,22 @@ export default function SampleListViewContainer() {
...
-  }, [handleSetViewMode, viewMode.mode]);
+    if (
+      localStorageViewMode &&
+      viewModeOptions.includes(localStorageViewMode)
+    ) {
+      handleSetViewMode(localStorageViewMode);
+    } else {
+      handleSetViewMode(viewModeOptions[0]);
+    }
+  }, [handleSetViewMode, viewMode, viewModeOptions]);
```
walesch-yan: I think you can remove `viewMode` from the list of triggers, otherwise useEffect will be called everytime you switch the mode instead of only once when the component is loaded

walesch-yan (commented):
Well changing the `ui.yaml` or `server.yaml` automatically comes with changing quite a few files. However, I think you chose the best approach for your type of change.

Also good thinking for directly removing the now redundant action and state field :)

ui/src/containers/SampleListViewContainer.jsx
matyson: you're right

mxcubeweb/core/models/configmodels.py
matyson: ok, makes sense and it's more reliable, but you're proposing to validate it on the frontend as well? i would have to replicate the schema on the frontend too, right?

docs/source/config/ui.rst
```diff
@@ -57,6 +57,37 @@ All auxiliary controls are optional. If you want to hide one from the UI, just r
...
+* Graphical View
+
+Here is an example of ``sample_list_view`` section:
+.. code-block:: yaml
+
+    sample_list_view:
+      id: sample_list_view
+      options:
+        - Table View
+        - Graphical View
```
fabcor-maxiv: This probably needs empty lines before and after the code blocks. The rendered result is here: <https://mxcubeweb--1819.org.readthedocs.build/en/1819/config/ui.html#sample-list-view> and it does not look as expected, see screenshot below.

```suggestion
This section configures the sample list view, more specifically the view modes available.
The section has the following syntax:

.. code-block:: yaml

    sample_list_view:
      id: sample_list_view
      options:
        - <view_mode_1>
        - <view_mode_2>
        - ...

The ``options`` key contains a list of view modes available in the sample list view.
Each view mode is a string, which is displayed in the UI in a dropdown menu.

The default view mode is the first one in the list. If the list has only one item, the dropdown will be disabled.
The default view modes are:
* Table View
* Graphical View

Here is an example of ``sample_list_view`` section:

.. code-block:: yaml

    sample_list_view:
      id: sample_list_view
      options:
        - Table View
        - Graphical View
```

image

docs/source/config/ui.rst
```diff
@@ -57,6 +57,40 @@ All auxiliary controls are optional. If you want to hide one from the UI, just r
...
+        - <view_mode_2>
+        - ...
+
+The ``options`` key contains a list of view modes available in the sample list view.
+Each view mode is a string, which is displayed in the UI in a dropdown menu.
+
+The default view mode is the first one in the list. If the list has only one item, the dropdown will be disabled.
+The default view modes are:
+* Table View
+* Graphical View
```
fabcor-maxiv: Looks like lists require empty lines as well:

```suggestion
The default view mode is the first one in the list. If the list has only one item, the dropdown will be disabled.
The default view modes are:

* Table View
* Graphical View
```

image

ui/src/containers/SampleListViewContainer.jsx
```diff
@@ -49,7 +52,10 @@ export default function SampleListViewContainer() {
   const loading = useSelector((state) => state.queueGUI.loading);
   const selected = useSelector((state) => state.sampleGrid.selected);
   const sampleList = useSelector((state) => state.sampleGrid.sampleList);
-  const viewMode = useSelector((state) => state.sampleGrid.viewMode);
+  const [viewMode, setViewMode] = useState(null);
```
axelboc: The default view mode here is `null`. Looking at `SampleGridTableContainer`, I think that pretty much means "Table View". This means that for one render (i.e. until the effect runs), users will get the table view. This "flash" may not look great, especially if the table view is disabled in the config. It would be better to initialise the state correctly on first render.

My suggestion would be to create a utility function outside of this component (in a `utils.js` file for instance) with something like:

```js
function getInitialViewMode(options) {
  const mode = localStorage.getItem('view-mode');
    if (mode && options.includes(mode)) {
      return mode;
    }

    return options[0];
}
``` 

Then:

```js
const [viewMode, setViewMode] = useState(getInitialViewMode(viewModeOptions));
```

... and remove the `useEffect` altogether.

axelboc (commented):
Hi @matyson! (Thanks for the bump @fabcor-maxiv :wink: )

- No problem on my end with the general idea of enabling/disabling the two sample list views from the config.
- The front-end code looks great too — just one suggestion to avoid a subtle flash on initial load.
- I have a few minor concerns with the config syntax:
  1. First I would move the new entry above `sample_view` to keep the "Sample view" entries together (since the "Sample list" is a separate screen from the "Sample view").
  2. I don't see the term "view modes", which could be confusing.
  3. Being able to list the modes freely in an array is problematic to me because the app expects only two, pre-determined ones.

What do you think of a config that looks more like this:

```yml
sample_list_view_modes:
  id: sample_list_view_modes
  table_view: true # `false` if omitted (so it's opt-in)
  graphical_view: true
```

... or if we want to remain closer to other config entries if it's easier:

```yml
sample_list_view_modes:
  id: sample_list_view_modes
  components:
    - id: 'table_view'
      show: true # again, `false` if omitted (so it's opt-in)
    - id: 'graphical_view'
      show: true
```

axelboc:
Forgot something: maybe it's already the case, but it would be good to have a check in the back-end, when the config is parsed, to ensure that at least one of the two views is enabled in `ui.yaml`. This should help with discoverability of the new config entry too.

ui/src/containers/SampleListViewContainer.jsx
matyson: Hi, i made some adjustments based on your suggestions, but there's still this workaround mapping the uiproperties from the backend to the frontend. I wanted to make a precomputed property in the `BaseModel` so that we have the correct view modes from the reducer itself, but i didn't succeed in doing so, do you have a nicer way of accomplishing this? Right now im stuck with this mapping of yaml component's ids into an object `SAMPLE_LIST_VIEW_MODES` in the frontend, which i feel is not ideal

axelboc: I'm liking your current solution; I don't think it's a workaround. What's better from before is that it doesn't rely on "display text" to map between the back-end config and the front-end logic. It now relies on internal IDs (`table_view` and `graphical_view`). I would actually even keep the keys of the `SAMPLE_LIST_VIEW_MODES` object in lowercase to match the back-end IDs exactly and save calling `toUpperCase`.

matyson: Right. I just felt that the lowercase ones were not so "javascript names" haha, but just personal preference. I will change them then

demo.yaml/mxcube-web/ui.yaml
```diff
@@ -1,5 +1,13 @@
 ---
 
+sample_list_view:
```
axelboc: Maybe just rename to `sample_list_view_modes` so it's really clear and consistent with the front-end?

I get that the page itself is called "Sample list", but the fact that the other page is called "Sample view" makes the term `sample_list_view` slightly confusing to me.

axelboc (approved):
Config and front-end looks good on my end! :100:

fabcor-maxiv:
@matyson, is there some feedback that has not been addressed yet? If every feedback has been addressed, and since this change has at least one approval already, then if you can rebase this pull request (on top of `develop`) I can merge it.

marcus-oscarsson:
Hm, I'm sorry guys but I unforunetly think you where a bit to fast with this one. 

  - Whats the reason for hiding the graphical view ?.
  - What happens of there is no `sample_list_view_modes` section what would the default behavior be ?

fabcor-maxiv:
> Whats the reason for hiding the graphical view ?.

It's a choice. Beamline staff at MAX IV does not want users to see this graphical view (it is irrelevant to the users where pucks and samples are in the dewar).

> What happens of there is no sample_list_view_modes section what would the default behavior be ?

@matyson?

marcus-oscarsson:
I see. 

> It's a choice. Beamline staff at MAX IV does not want users to see this graphical view (it is irrelevant to the users where ?> > pucks and samples are in the dewar).

The aim of the project is to a quite large extent to provide the same or very similair user experience across the different institutes. There is clearly need for configuration especially when it comes to functional differences of the beamlines. It would have been good to have a discussion on why its irrelevant for your users or staff to not have the graphical view. So that we can understand what we missed or if it really falls into the category of something that needs to be configurable.

fabcor-maxiv:
> > It's a choice. Beamline staff at MAX IV does not want users to see this graphical view (it is irrelevant to the users where pucks and samples are in the dewar).
> 
> The aim of the project is to a quite large extent to provide the same or very similair user experience across the different institutes. There is clearly need for configuration especially when it comes to functional differences of the beamlines. It would have been good to have a discussion on why its irrelevant for your users or staff to not have the graphical view. So that we can understand what we missed or if it really falls into the category of something that needs to be configurable.

I guess it is a question for @epanepucci and the UI/UX working group in general. @matyson and I only execute the orders. :D

marcus-oscarsson:
> I guess it is a question for @epanepucci and the UI/UX working group in general. @matyson and I only execute the orders. :D

Sure no problem, we just need to keep it mind for the future.

epanepucci:
The way I see it, the user might be interested in selecting their samples based on:
- sample name
- protein name/acronym
- project priority

At MAXIV, staff is in charge of loading pucks into the sample changer dewar and the user never knows (or cares) in which position their pucks are located.

The graphical views were useful when little database integration was available and dewar capacities were small; nowadays, for most use cases, the sample containers (pucks, plates, chips) are/should be entered in advance into the database and their samples should be made available in mxcube via their assigned (user given) labels.

I think we could agree that a table view should always be present (not configurable) and additional views be made as opt-in.

matyson:
> > Whats the reason for hiding the graphical view ?.
> 
> It's a choice. Beamline staff at MAX IV does not want users to see this graphical view (it is irrelevant to the users where pucks and samples are in the dewar).
> 
> > What happens of there is no sample_list_view_modes section what would the default behavior be ?
> 
> @matyson?

The schema validator enforces the section to be present as it is required. Also, it ensures that at least one view is shown in the components.

marcus-oscarsson:
@matyson: Yes, but what happens if the section is missing from the configuration file, I guess the application wont start ?

marcus-oscarsson:
@epanepucci, thanks for the clarification. As you know there is a discussion pending on the design and UX of the sample list. Especially considering automation over several user groups, and also a general discussion regarding improving the overall UX.

I guess one might also want to hide show the graphical view, or other sample related views, depending on which user you are.

epanepucci:
Marcus, yes, I guess Danielle and I could start the UX working group :-D

The automation over several groups is very important IMO.
Regarding what to show or not depending on user, I assume you mean user-roles as in, regular user, staff, etc? I think the kind of views provided to a **regular user** should be decided by the site. Other user roles could have access to other views but most importantly a staff-user could have access to samples from more than one proposal for the case in which automation is to run samples from several groups. But this is kind of another, although related, topic.

marcus-oscarsson:
> Marcus, yes, I guess Danielle and I could start the UX working group :-D
> 
:), Yes, that would be awesome !. 

> The automation over several groups is very important IMO. 

Yes, Id say that its one of the main improvement points for us.

>Regarding what to show or not depending on user, I assume you mean user-roles as in, regular user, staff, etc? I think the kind of views provided to a **regular user** should be decided by the site. Other user roles could have access to other views but most importantly a staff-user could have access to samples from more than one proposal for the case in which automation is to run samples from several groups. But this is kind of another, although related, topic.

Yes I was thinking of user-roles. We have been trying to keep that sort of functionality to a minimum for the time being because it quickly becomes quite complicated. However; its perhaps needed and there is support for it if needed.

Until now we have simply relied on the access to a certain set of samples and that access is given via LIMS. A regular user simply have access to a restricted set of samples but the views available to staff and "regular user" are for the moment the same.

matyson:
> @matyson: Yes, but what happens if the section is missing from the configuration file, I guess the application wont start ?

Yes. you're right. I got a bit confused on how these schemas could enforce different setups without breaking the app, e.g., with default values and special validations but in the end i assumed it was ok not to start for it was a required view of the app that would go missing if not properly configured.

marcus-oscarsson:
I see, I guess the best would be if the app starts with only list view or alternatively both views if not configured at all. The change would in this way be "backward compatible" and don't require any configuration changes.

axelboc:
> I see, I guess the best would be if the app starts with only list view or alternatively both views if not configured at all. The change would in this way be "backward compatible" and don't require any configuration changes.

This is also my long-term vision, which I suggested in #1725:

> I would also be in favor of having a near-zero config experience out of the box, which means having a default config that gets merged/overridden by the content of ui.yaml.

However, I think this zero-config experience needs a more thought-out solution that covers the whole config. I would not recommend doing it piece-meal. The config has been evolving organically for a long time and it's a bit all over the place at the moment. We need to pick a direction, document it, and aim for it.

That's why I suggested making `sample_list_view_modes` mandatory and preventing start-up of the server if it's missing (i.e. "fail fast"): I think it's safer to make new options mandatory until we have a more robust way to declare a global fallback config.

That being said, all that really matters to me is that front-end components don't have to decide for themselves what to do if, for instance, `sample_list_view_modes` is not configured. We want all this fallback logic in one place in the back-end, ideally described in a declarative format. Front-end components should be able to assume that every config option is always specified. The goal is to make the front-end more predictable and less prone to regressions.

marcus-oscarsson:
Its true that the configuration format and handling needs to be looked at again and its a good initiative. A near zero config solution would really be ideal. 

I would for the sake of the other developers argue that we should try to make all new configuration options optional by adding a sensible default. It would be nice with a message if a default is used, i.e. "Option xyz is missing in configuration file .../path... using default key, value.

Like this the change becomes transparent and we don't need to change the current configuration files too much for production systems. Which will probably make the update to a new configuration format easier as well. 

Failing fast can be very help full in some cases when an option is critical and there is no really good default value. In this case I would argue that it would make sense to have a default value.

I think in general the PyDantic schema validation exception could be clearer and formatted better. There is not much we can do about that, but we could perhaps add a link to the documentation, so that one would know what to add. For instance. However; IMO in this case the best would be with a default.

```
...

pydantic.v1.error_wrappers.ValidationError: 1 validation error for UIPropertiesListModel
sample_list_view_modes
  field required (type=value_error.missing)

See https://mxcubeweb.readthedocs.io/en/stable/config/ui.html

```

matyson:
Hi, i am willing to work on this issue again towards a solution that better attend the points of the discussion. So, what's the conclusion here? How can i adjust these changes to address the project's needs?

marcus-oscarsson:
Thanks, that super nice of you. I think the only thing missing, possibly, for the moment is to make `sample_list_view_modes` optional. If its not too complicated.

Then we have to continue the discussion on configuration in https://github.com/mxcube/mxcubeweb/issues/1725. I think a robust configuration system would be very beneficial for what we are doing. So its a great initiative !

jbflo:
Sorry for my late reply
@matyson @fabcor-maxiv 
As a follow up you might as well want to remove the UI Component NewSampleIsaraView
"mxcubeweb/ui/src/containers/SampleIsaraView.jsx "
This was a specific Component made by/for MaxIV to Handle ISARA Graphical View. 

And i think it is not being maintain and might be broken.

matyson:
@jbflo as far as i could investigate this is not being used by MAXIV anymore, but still could be used by other facilities since it derives from [this](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/HardwareObjects/Cats90.py#L144) and it's called in [here](https://github.com/mxcube/mxcubeweb/blob/develop/ui/src/containers/SampleGridTableContainer.jsx#L966C2-L973C6). Im glad to remove it, but are you sure it's not being used by any other facility listed in the `Cats90` file?

PR #1821, opened 2025-08-12
Remove the `--no-fix` option of the `ruff` hook

fabcor-maxiv:
When `ruff check --fix` makes modifications to files, it is not immediately clear where and why the files are modified. When running locally, one can see the changes with `git diff`, but it still does not clarify why (which Ruff rule was broken).

Without the `--fix` option, Ruff clearly shows where the violations are. Then the developer is free to decide how to fix the violations:
* The developer can choose to do the fixes manually.
* The developer can choose to run `poetry run ruff check --fix`.
* The developer can choose to add exceptions.
* The developer can choose to do a combination of the above.

GitHub: https://github.com/mxcube/mxcubecore/issues/1220
GitHub: https://github.com/mxcube/mxcubecore/pull/1243

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,10 +32,12 @@ repos:
 
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.9.6
+    rev: v0.9.10  # Keep in sync with `poetry.lock` and `pyproject.toml`
     hooks:
+      # For the `ruff` hook (which runs `ruff check`),
+      # we purposefully skip the `--fix` option,
+      # and recommend running a command like `poetry run ruff check --fix` instead.
       - id: ruff
-        args: [--fix]
       - id: ruff-format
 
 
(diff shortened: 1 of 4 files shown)


fabcor-maxiv:
This is intended to be the exact same change that we did in MXCuBE-Core already: <https://github.com/mxcube/mxcubecore/pull/1243>.

Since we remove the `--fix` option for Ruff linter in Pre-commit hooks, we need to offer a straightforward way to run the fixes. And this is what this pull request offers.

fabcor-maxiv:
Ruff has no dependencies of its own, so adding Ruff to the dependencies should have a very low cost, it should never cause version conflicts or cause stress on the dependency resolution.

marcus-oscarsson:
Thanks looks good,I don't know why but I thought we already removed `--no-fix`. Maybe that was on `mxcubecore` and I thought we did it on web as well. Anyhow, thanks for having a look.

fabcor-maxiv:
> Thanks looks good,I don't know why but I thought we already removed `--no-fix`. Maybe that was on `mxcubecore` and I thought we did it on web as well. Anyhow, thanks for having a look.

Yes, as I [commented above](https://github.com/mxcube/mxcubeweb/pull/1821#issuecomment-3179769269), this was done for MXCuBE-Core only. And recently I noticed that I had forgotten to do it for MXCuBE-Web as well.

PR #1824, opened 2025-08-18
Show SampleChanger maintenance info in a tooltip on hover

alessandro-olivo:
This feature allows displaying enhanced information about the status of the sample changer in the tooltip of the related widget. In particular, this includes the `global_state` and the `message` returned by the sample changer maintenance HWO.

The intended behavior is similar to how the "Ring Current" tooltip already works when you hover with the cursor.

### Source of information
The values are retrieved from the endpoint sample_changer/get_initial_state (see [component/samplechanger.py](https://github.com/mxcube/mxcubeweb/blob/46cbe7b0e37717e5a77ef5f0bb1fd9067ab0d11d/mxcubeweb/core/components/samplechanger.py#L365C1-L377C1)) and from the websocket channel (see [routes/signals.py](https://github.com/mxcube/mxcubeweb/blob/e9efe2af38fccf5b11666ea4d6214ca87addcd12/mxcubeweb/routes/signals.py#L198C1-L205C10)).

For the latter source, I had to restore part of commit 337efd37 in order to send updates of the `global_state` dictionary to the frontend:
```diff
def sc_maintenance_update(*args):
    if len(args) == 3:
        # ...
-       _, cmd_state, message = args
+       global_state, cmd_state, message = args
    else:
        # ...
+       global_state = {}
        cmd_state, message = args

    try:
        server.emit(
            "sc_maintenance_update",
            {
+               "global_state": global_state,
                "commands_state": json.dumps(cmd_state),
                "message": message,
            },
            namespace="/hwr",
        )
    except Exception as msg:
        logging.getLogger("HWR").error("error sending message: %s" + str(msg))
```
I think this is fine, because as far as I understood, the issue that the commit fixed was that changes in the SC state on the frontend were triggered by two different signals, one emitted from the SC HWO and the other from the SC maintenance HWO. With this feature, the SC `state` changes remain under the responsibility of the SC HWO, while `global_state` should be considered a dictionary containing additional information about the SC.

Maybe we could rename it to something like `detailed_status` or `global_info`.

### Examples
1) The SC HWO is loaded and the `global_state` and the `message` parameters are defined 
   image

2) The `message` parameter is defined
   image

3) The SC maintenance HWO is not loaded
   image

### Notes:
This sample changer widget is no longer an InOutSwitch. There is no Power Off/On command popping up when you click on the state. Since the SC commands are already available in the Equipment tab, I thought it wasn’t strictly necessary, but I can eventually add this functionality if needed.

### DIFF
--- a/ui/src/containers/BeamlineSetupContainer.jsx
+++ b/ui/src/containers/BeamlineSetupContainer.jsx
@@ -5,13 +5,13 @@ import { useDispatch, useSelector } from 'react-redux';
 
 import { setAttribute } from '../actions/beamline';
 import { stopBeamlineAction } from '../actions/beamlineActions';
-import { sendCommand } from '../actions/sampleChanger';
 import BeamlineAttribute from '../components/BeamlineAttribute/BeamlineAttribute';
 import BeamlineCamera from '../components/BeamlineCamera/BeamlineCamera';
 import DeviceState from '../components/DeviceState/DeviceState';
 import InOutSwitch from '../components/InOutSwitch/InOutSwitch';
 import MachInfo from '../components/MachInfo/MachInfo';
 import OneAxisTranslationControl from '../components/MotorInput/OneAxisTranslationControl';
+import SampleChanger from '../components/SampleChanger/SampleChanger';
 import BeamlineActions from './BeamlineActionsContainer';
 
 function BeamlineSetupContainer() {
@@ -23,6 +23,9 @@ function BeamlineSetupContainer() {
     (state) => state.beamline.hardwareObjects,
   );
   const sampleChangerState = useSelector((state) => state.sampleChanger.state);
+  const sampleChangerMaint = useSelector(
+    (state) => state.sampleChangerMaintenance,
+  );
 
   function renderBeamstopAlignmentOverlay() {
     const motor = hardwareObjects['diffractometer.beamstop_distance'];
@@ -212,14 +215,9 @@ function BeamlineSetupContainer() {
         </Nav>
         <Nav className="me-3">
           <Nav.Item>
-            <InOutSwitch
-              openText="Power On"
-              offText="Power Off"
-              openValue="PowerOn"
-              offValue="PowerOff"
-              labelText="Sample Changer"
-              value={sampleChangerState}
-              onSave={(cmdparts, args) => dispatch(sendCommand(cmdparts, args))}
+            <SampleChanger
+              state={sampleChangerState}
+              maintenance={sampleChangerMaint}
             />
           </Nav.Item>
           {renderActuatorComponent()}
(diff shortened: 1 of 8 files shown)


ui/src/containers/BeamlineSetupContainer.jsx
```diff
@@ -212,14 +215,9 @@ function BeamlineSetupContainer() {
...
           <Nav.Item>
-            <InOutSwitch
-              openText="Power On"
-              offText="Power Off"
-              openValue="PowerOn"
-              offValue="PowerOff"
-              labelText="Sample Changer"
-              value={sampleChangerState}
-              onSave={(cmdparts, args) => dispatch(sendCommand(cmdparts, args))}
+            <SampleChanger
```
marcus-oscarsson: How about calling this component `SampleChangerStatus` or similair, `SampleChanger` feels a bit too general in my opinion ?

marcus-oscarsson:
Looks good and as always a nice contribution :). I'm not certain that the additional information in `global_state` tells the user alot. But that depends on the content of `global_state` and perhaps also who the user is. 

I guess it could possibly contain some additional information that is useful for the user and its hidden away in a overlay so I think this would be ok.

The `Power On` and `Power Off` commands was added/requested by MaxIV. We are not using those commands and if we where to use them we would use the `equipment` tab/view. So it would be fine for use to remove them, but we should ask the guys at MaxIV if them commands are still used from the status label/switch, @epanepucci, @elmjag, @fabcor-maxiv

elmjag:
> The `Power On` and `Power Off` commands was added/requested by MaxIV. We are not using those commands and if we where to use them we would use the `equipment` tab/view. So it would be fine for use to remove them, but we should ask the guys at MaxIV if them commands are still used from the status label/switch, @epanepucci, @elmjag, @fabcor-maxiv

This was requested before my time. I think dropping them from the main (`sample view`) tab is a good idea.

I don't see a reason why normal beamline user would want to power on and off the robot during routine data collections. And if for some special reason they need it, it is still available on the `equipment` tab.

ui/src/containers/BeamlineSetupContainer.jsx
epanepucci: agree on this

epanepucci:
I agree with Elmir regarding the sample changer. Also, UX wise I'd rather have those widgets be information only and move actions to *beamline actions* instead for things more regularly needed by every user.

marcus-oscarsson:
:+1: Thanks guys. So in principle fine.

marcus-oscarsson:
Your proposed changes seems to be fine @alessandro-olivo, you can rebase and merge (just have a look at the name of the `SampleChanger` component name)

alessandro-olivo:
I'm back on this PR, thanks everyone for the contributions :smile:
I'll change the name of the component to something more meaningful

> I'm not certain that the additional information in `global_state` tells the user alot. But that depends on the content of `global_state` and perhaps also who the user is.
 
You're right. Actually, this feature was requested by the staff, and they’ll be most likely the only ones who care about this information.

alessandro-olivo:
I decided to name it `SampleChangerInfo`, similar to `MachineInfo` since they have a similar purpose

PR #1827, opened 2025-08-20
Disable Run button on RJSF validation error

mockoocy:
Previously only redux-form validation using `validate` function was used to mark form data as "invalid".

Now the RJSF validation is also tracked as a part of component state. The "Run now" button will be disabled if either RJSF or `validate` function complain about any of the fields.

image


axelboc (approved):

marcus-oscarsson:
Nice, thanks alot @mockoocy

Issue #1832, opened 2025-08-26
"Problem with available workflows" after sign in


matyson:
In the demo version, after signing in we get this logged exception:

<img width="801" height="184" alt="Image" src="https://github.com/user-attachments/assets/b17e6e68-96b1-47c3-abe4-bd5f42559a3e" />

Which, after some inspection in the `beamline_config.yml` and in the `ednaparams.yml` i noticed that the section has no attribute `name`, only `path`, `requires` and `title`. I wonder if this example is outdated or the code section itself, as it doesn't seem to broke anything other than this exception message.

Also, if we omit the `workflow` key from the `beamline_config.yml` file we get a different error message due to the workflow now being `None` but, again, only an exception message:

```
2025-07-31 07:28:20,800 |MX3.HWR|ERROR  | Problem with available workflows or naming/pathing: {wf.name}
Traceback (most recent call last):
 File "/app/web/mxcubeweb/core/components/workflow.py", line 20, in get_available_workflows
   for wf in beamline.workflow.get_available_workflows():
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get_available_workflows'
```
My first attempt towards a solution in the [workflow.py](https://github.com/mxcube/mxcubeweb/blob/19316471f6d2c64419a44090da9f1be3ea45dfbc/mxcubeweb/core/components/workflow.py#L14) was to check for `None` attributes and completely skip the try block, but i don't really know if it's being used or breaking something if not defined. And the other problem is regarding the `name` key in the the `wf` dictionary, should we get rid of it or just update the demo to include this parameter?

A 1st draft towards the solution:

```python
def get_available_workflows(self):
    workflows = {}
    beamline = HWR.beamline

    if getattr(beamline, "workflow", None):
        try:
            for wf in beamline.workflow.get_available_workflows():
                # Rename name and path to wfname and wfpath in order to avoid name
                # clashes
                wf["wfname"] = wf.pop("name")  # <-- Should i remove this?
                wf["wfpath"] = wf.pop("path")

                workflows[wf["wfname"]] = wf
        except Exception:
            msg = "Problem with available workflows or naming/pathing: {wf.name}"
            hwr_logger.exception(msg)

    if getattr(beamline, "gphl_workflow", None):
        # Add Global Phasing workflows if available
        workflows.update(beamline.gphl_workflow.get_available_workflows())

[...]

PR #1833, opened 2025-08-26
Fix "Problem with available workflows" for demo

matyson:
Update `ednaparams.[yml/xml]` to fix "Problem with available workflows" exception in sign in. Relates to #1832


marcus-oscarsson:
Thanks @matyson

walesch-yan (approved):

PR #1834, opened 2025-08-26
Make workflow optional in beamline configuration

matyson:
Check if `beamline.workflow` is None before attempting to get available workflows, this avoids misleading exception messages when the workflow section is undefined on purpose in `beamline_config.yml`. Related to #1832.


marcus-oscarsson:
Thanks !

PR #1837, opened 2025-08-28
Retrieving information about the beam from `beamadapter`

marcus-oscarsson:
Retrieving information about the beam from the `beamadapter` instead of using the beamline route.

Moved `get_beam_info` to `get_value` of the `BeamAdapter` object

### DIFF
--- a/ui/src/actions/login.js
+++ b/ui/src/actions/login.js
@@ -1,7 +1,8 @@
 /* eslint-disable promise/prefer-await-to-then */
-import { fetchBeamInfo, fetchBeamlineSetup } from '../api/beamline';
+import { fetchBeamlineSetup } from '../api/beamline';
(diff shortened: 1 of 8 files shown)


ui/src/actions/login.js
```diff
@@ -94,8 +95,8 @@ export function getInitialState() {
       fetchQueueState()
         .then((queue) => ({ queue }))
         .catch(notify),
-      fetchBeamInfo()
-        .then((beamInfo) => ({ beamInfo }))
+      sendGetValue('beam', 'beam')
+        .then((beamInfo) => ({ beamInfo: beamInfo.value }))
```
marcus-oscarsson, line 99: Ill try to come up with a way to get rid of this extra '.value' that indirectly comes from the use of PyDantic models for the adapter `get_value`.

axelboc (approved):

PR #1838, opened 2025-08-28
Call to `get_beam_info` in `get_viewport_info` not needed anymore

marcus-oscarsson:
For some reason the route that retrieves information about the video stream also retrieves information about the beam. 

I assume it used to be the case that the beam information was not available via a separate call or that the relevant information was not available when it was needed (not yet fetched). The information about the beam does no longer need to be part of the camera route as we are, now, fetching all the relevant routes before initial render.


dominikatrojanowska:
LGTM :)

PR #1839, opened 2025-08-29
Made get_value as a ResourceHandler attribute instead of command

marcus-oscarsson:
The `ResourceHandler` prevents the use of `commands` for observer users. Retrieving an adapter value via `get_value` needs to be accessible to observers, thus made it a `ResourceHandler` attribute instead of command.

The method name `get_value` could simply be changed to `value` as well. I might do that change in a future PR.

### DIFF
--- a/ui/src/api/hardware-object.js
+++ b/ui/src/api/hardware-object.js
@@ -8,7 +8,7 @@ export function sendExecuteCommand(objectType, objectId, command, value) {
     .safeJson();
 }
 
-export function sendGetAttribute(objectType, objectId, attributeName) {
+export function fetchAttribute(objectType, objectId, attributeName) {
   return endpoint.get(`/${objectType}/${objectId}/${attributeName}`).safeJson();
 }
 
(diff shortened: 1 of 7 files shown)


ui/src/api/hardware-object.js
```diff
@@ -17,7 +17,7 @@ export function sendSetValue(objectId, objectType, value) {
 }
 
 export function sendGetValue(objectId, objectType) {
-  return sendExecuteCommand(objectType, objectId, 'get_value', {});
+  return sendGetAttribute(objectType, objectId, 'get_value');
```
axelboc: A suggestion for `ui/src/api/hardware-object.js`:

```js
// It's a GET method, so the prefix should have been changed to `fetch`
export function fetchAttribute(objectType, objectId, attributeName) {
  return endpoint.get(`/${objectType}/${objectId}/${attributeName}`).safeJson();
}

// Very thin abstraction to help with readability
export function fetchValue(objectType, objectId) {
  return fetchAttribute(objectType, objectId, 'get_value');
}
```

marcus-oscarsson: Thanks, well spotted

axelboc (approved):
Nice :+1: — sorry meant to approve earlier

marcus-oscarsson:
Thanks !, just struggling with typing `fetchAttribute` :rofl:

PR #1840, opened 2025-08-29
Setting aperture via BeamAdapter instead of diffractometer route

marcus-oscarsson:
The change of "aperture" is done via the "Beam" `HardwareObject` but was handled in the diffractometer route. This PR moves the setting of beam size to the `BeamAdapter`. 

We are in reality dealing with beam size in general and not aperture size.

### DIFF
--- a/mxcubeweb/core/adapter/beam_adapter.py
+++ b/mxcubeweb/core/adapter/beam_adapter.py
@@ -1,3 +1,4 @@
+import logging
 from typing import ClassVar
 
 from mxcubecore.HardwareObjects.abstract import AbstractBeam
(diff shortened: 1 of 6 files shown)


mxcubeweb/core/adapter/beam_adapter.py
```diff
@@ -51,6 +52,12 @@ def _get_aperture(self) -> tuple:
 
         return aperture_list, current_aperture
 
+    def set_value(self, value: int) -> HOBeamModel:
+        self._ho.set_value(value)
+        msg = f"Changing aperture size to: {value}"
```
beteva: Maybe replace aperture by beam ?

beteva:
Great initiative!

mxcubeweb/core/adapter/beam_adapter.py
marcus-oscarsson: I simply kept the log as it was, but your right

axelboc (approved):
Loving your work :blush:

PR #1841, opened 2025-08-29
Fix issue with running beamline actions

elmjag:
Right now if you run beamline action, the 'run_action' request fails with:

```
501 NOT IMPLEMENTED
{"error":"Server error"}
```

and the beamline action does not run.

These change fixes that issue.

### DIFF
--- a/mxcubeweb/core/server/resource_handler.py
+++ b/mxcubeweb/core/server/resource_handler.py
@@ -346,7 +346,7 @@ class ResourceHandler:
                 # We only handle a single argument when using Pydantic models
                 # We handle complex structures by checking if param_data
                 # contains a value or a dictionary.
-                if isinstance(param_data[param_name], dict):
+                if isinstance(param_data.get(param_name), dict):
                     return param_type.parse_obj(param_data[param_name])
 
                 # If its a single value and Pydantic model, pass the entire


mxcubeweb/core/server/resource_handler.py
```diff
@@ -346,7 +346,7 @@ def _validate_param_data(
                 # We only handle a single argument when using Pydantic models
                 # We handle complex structures by checking if param_data
                 # contains a value or a dictionary.
-                if isinstance(param_data[param_name], dict):
+                if isinstance(param_data.get(param_name), dict):
```
marcus-oscarsson, line 349: Hm, not sure I see what the difference is, are you sure this solves the issue ?

marcus-oscarsson: Ill have a look as well

elmjag: `param_data.get(param_name)` returns `None` if `param_name` key does not exist. 

The `isinstance(None, dict)` evaluates to false, and will fall through to do the validation `param_type.parse_obj(param_data)`. Which seems to be the correct thing to do for beamline actions.

elmjag: > Ill have a look as well

Yes, pretty please :) I don't quite understand this code, perhaps there is a better way to deal with this.

marcus-oscarsson: Ok, now I see. so with `.get` you avoid the `KeyError` and you don't get the issue. So that would fix the problem (or avoid it). However, we can remove the entire if statement. It was added as a convenience/guard for when several dictionaries are passed (so that only the first one is used). But its up to us to make that it does not happen

```
                # We only handle a single argument when using Pydantic models
                # We handle complex structures by checking if param_data
                # contains a value or a dictionary.
                if isinstance(param_data[param_name], dict):
                    return param_type.parse_obj(param_data[param_name])
```

marcus-oscarsson:
Thanks @elmjag, I would just remove that entire `if` statement

mxcubeweb/core/server/resource_handler.py
```diff
@@ -346,7 +346,7 @@ def _validate_param_data(
...
                     return param_type.parse_obj(param_data[param_name])
```
dominikatrojanowska, line 350: Assign to a variable instead of repeating operations:
```suggestion
                some_fancy_name = param_data.get(param_name)
                if isinstance(some_fancy_name, dict):
                    return param_type.parse_obj(some_fancy_name)
```

mxcubeweb/core/server/resource_handler.py
marcus-oscarsson: Hm, actually @elmjag ... wait. Your fix is the good one. However I agree that the comment is a bit confusing. Ill try to explain

marcus-oscarsson: So my comment above is correct but there is a special case for the `set_value` of an adapter, that can either take N arguments (of native python types) or a single PyDantic model. So this `if` statement takes care of the case when we receive a dictionary containing a pydantic model. ie `{value: { .... data ...}}` instead if simply `{....data ...}` which is the case for `set_value`.

PR #1847, opened 2025-09-01
Moved all diffractometer routes in /routes to the diffractometer adapter

marcus-oscarsson:
Moved all the remaining "diffractometer routes" in `/routes` to the `DiffractometerAdapter`


axelboc (approved):
:heart_eyes:

ui/src/actions/sampleview.js
```diff
@@ -355,7 +354,9 @@ export function changeAperture(size) {
 
 export function changeCurrentPhase(phase) {
   return async (dispatch) => {
-    await sendUpdateCurrentPhase(phase);
+    await sendExecuteCommand('diffractometer', 'diffractometer', 'set_phase', {
+      phase,
+    });
     dispatch(setCurrentPhase(phase));
```
axelboc, line 360: Oh just to double check: does `sendExecuteCommand` resolve to a value — i.e. does the server respond with the updated phase value? If so, it would be best to use that value to update the Redux store.

Or maybe the updated value is sent soon after via a signal and the `dispatch` is not needed?

marcus-oscarsson: Oh, good point. I overlooked the dispatch. `sendExecuteCommand` can return a value, in the case of `set_phase` its done via an event (because its a action that can take some time). We should perhaps when you mention it distinguish between the two types, i.e have a  `sendAsync` variant or something ?

marcus-oscarsson: At the moment, I think I will keep things as it is. But Ill think about a pattern or how to better handle/specify values updated by events

axelboc: > We should perhaps when you mention it distinguish between the two types, i.e have a sendAsync variant or something ?

I see, thanks for the clarification! Yeah in that case it might be good, eventually, to have two distinct endpoints for predictability. :100:

PR #1849, opened 2025-09-01
Moved detector routes to `DetectorAdapter`

marcus-oscarsson:
Moved detector routes to DetectorAdapter, to be merged after #1846


ruff.toml
```diff
@@ -10,6 +10,7 @@ ignore = [
     "COM812",  # Ruff recommends to ignore this linter rule when using its formatter
     "D",  # pydocstyle: ignored, too many hits (958)
     "PLR2004",  # magic-value-comparison: ignored, too many hits
+    "N814", # To avoid, Camelcase `HardwareRepository` imported as constant `HWR`
```
marcus-oscarsson: Ignoring rule N814, I don't know why this rule has not been triggered before ?

axelboc (approved):

ruff.toml
fabcor-maxiv: I think it would be better to add `HWR` to a white-list of allowed names:

* <https://docs.astral.sh/ruff/settings/#lint_pep8-naming_ignore-names>

marcus-oscarsson: Not sure, I think that If one does that sort of thing its by purpose and one would be aware of that its not standard or the most pretty thing. I think that HWR would be the only case for mxcube, it seems unnecessary to have a white list for that

fabcor-maxiv: I do not understand the logic of what you wrote... :/

We can add `HWR` to the `ignore-names` list, and `HWR` will be allowed to appear anywhere -- more precisely it will not trigger for the `pep-naming`/`N*` rules. Isn't that what we want?

marcus-oscarsson: It depends what the other `pep-naming` rules imply I guess. I guess a white list would work if its only to do with how the name is used. It otherwise seems quite unnecessary to add a white list for a rule that we wont use.

fabcor-maxiv: I guess there is a misunderstanding somewhere... :D

We want to be able to use the name `HWR` anywhere in the Python code, right? So we tell the linter that `HWR` is always a valid name (for the `pep-naming`/`N*` rules). That's it. Isn't that what we want?

I am confused where the problem is... :D

fabcor-maxiv: If we go with the `ignore-names` solution, then we can likely remove all occurrences of `N814` in `ruff.toml`.

marcus-oscarsson: No, thats only part of the problem. We also want to maintain a minimal set of configuration and rules and have a tool that is in the way as little as possible (provides a good developer experience). Setting up a white list for a rule that we wont use, or have extremely small impact on our code seems unnecessary. In that case its perhaps better to ignore the rule. 

Anyhow Ill setup the whitelist and we will see how that works.

fabcor-maxiv: OK, let's ask a 3rd party for a tie breaker... I really do not understand your logic on this... :D

It is probably me getting confused as to what you have in mind :D

marcus-oscarsson: What we have in mind is to use the white list and remove the other N814. My argument was that it somehow seems unnecessary considering time/effort vs benefit but its alright in this case, the white list will work.

axelboc: If I understand, @marcus-oscarsson's point is that the rule is so silly that it's not even worth configuring to get rid of this false positive (since the rule is supposed to forbid renaming an import to constant case, which is clearly not the case here).

As a general rule, I tend to agree with the practice of disabling rules with false positives. I try to look out for improvements to those rules in release notes and renable them when they no longer get in the way. If I have enough time and courage, I might even report the false positive to the maintainers of the rule.

In this case, adding `HWR` to a white list is kind of like agreeing with Ruff in saying that `HWR` is the constant-case version of `HardwareRepository`, which it isn't... Anyway, probably getting too philosophical :joy:

fabcor-maxiv: I do not get your point, at all, but it is fine, it is not a big deal either way.

marcus-oscarsson: So, exactly, that is more or less how I saw it :). I think most cases when this rule would get triggered, by code written by any of us, would be intentional and similair to the case above. I can accept the few cases that slips threw over editing the white list every time.

Anyhow, lets try the white list, the cases when this happens should be quite few

fabcor-maxiv: > the cases when this happens should be quite few

I hope so :)

I wish `mxcubecore/HardwareRepository.py` was named `mxcubecore/hardware_repository.py` instead. I wish the code was `from mxcubecore import hardware_repository as hwr`.

If we were to disable `N814` globally, we would run the risk of letting more of this inconsistent, unexpected, and confusing casing usage enter the code base.

marcus-oscarsson: We can of course do that change, or something similair :).

PR #1850, opened 2025-09-02
Hide the "Unknown" phase button.

mockoocy:
Currently in the phase selection UI element, there is "Unknown" option, which just does not do anything (the ChangeEvent is ignored). It seems preferable to just have "Unknown" appear only as the current value (when set from the hardware object), as it currently is, but not as an option in the select.

### DIFF
--- a/ui/src/components/SampleView/PhaseInput.jsx
+++ b/ui/src/components/SampleView/PhaseInput.jsx
@@ -20,13 +20,19 @@ function PhaseInput() {
       value={value}
       data-busy={isBusy || undefined}
       onChange={(evt) => {
-        if (evt.target.value !== 'Unknown') {
-          dispatch(changeCurrentPhase(evt.target.value));
-        }
+        dispatch(changeCurrentPhase(evt.target.value));
       }}
     >
+      <option hidden key="Unknown">
+        Unknown
+      </option>
+
       {options.map((option) => (
-        <option key={option}>{option}</option>
+        // The "Unknown" option is hidden, so that it only appears as the selected value
+        // when the diffractometer hardware object sets it.
+        <option key={option} hidden={option === 'Unknown'}>
+          {option}
+        </option>
       ))}
     </Form.Select>
   );


ui/src/components/SampleView/PhaseInput.jsx
```diff
@@ -20,14 +20,21 @@ function PhaseInput() {
...
+        the current phase, but may not be selected.
+       */}
+      <option hidden key="Unknown">
+        Unknown
+      </option>
+      {options
+        .filter((option) => option !== 'Unknown')
+        .map((option) => (
+          <option key={option}>{option}</option>
+        ))}
```
axelboc: Makes total sense, but I think this would work too:

```suggestion
      {options.map((option) => (
        <option key={option} hidden={option === 'Unknown')>{option}</option>
      ))}
```

axelboc (approved):

marcus-oscarsson:
The phase dropwdown also displays in what phase diffractometer is. If we remove `Unknown` we don't know if the diffractometer is in an `Unknown` state/position/invalid

ui/src/components/SampleView/PhaseInput.jsx
beteva: If you want to remove unknown, you could do it in a specific diffractometer HO.

beteva:
If you want to remove unknown, you could do it in a specific diffractometer HO.

mockoocy:
@marcus-oscarsson, @beteva 
[Here's a recording of what this PR is trying to achieve](https://github.com/user-attachments/assets/1fa46045-22df-4f88-8863-2e26368fb960)

We don't want to get rid of the "Unknown" phase entirely (as we could using diffractometer HO config).
We just want to remove it from the select input, so it does not show up to user.

axelboc:
Now that I see the recording, I'm wondering why that "Unknown" value even exists and ends up in the front-end. Seems more like a loading state to me. The value should not change to "Unknown", it should just go from "Transfer" directly to "BeamLocation" and the drop-down should have an orange background and be disabled while that change is happening in the back-end.

marcus-oscarsson:
I think its a great idea to remove "unknown" from the list of phase that the user can select, but we need to be able to display unknown or a value with the same meaning. Like in your video, very nice !

We can easily remove the "Unknown" phase from the phase list. I agree that this would make alot of sense in most cases. However sometimes for some strange reasons the MD phase is really unknown because the motors are in position that don't correspond to a known position. That position could also be called invalid or whatever but its been named unknown in the dawn of mxcube.

So the unknown label signals something to the user in this sense we can not really remove it.

axelboc:
Alright, so "Unknown" is a sort of error/invalid value, which is fine.

But here it is used while moving from one valid value to another valid value. I feel like there should be a way to either:

- not update the value to "Unknown" until the motor has stopped moving (and hasn't reached a "known" value)
- or not show "Unknown" in the UI while the motor is moving (while in "RUNNING" state).

beteva:
> Now that I see the recording, I'm wondering why that "Unknown" value even exists and ends up in the front-end. Seems more like a loading state to me. The value should not change to "Unknown", it should just go from "Transfer" directly to "BeamLocation" and the drop-down should have an orange background and be disabled while that change is happening in the back-end.

Actually what we see at ESRF is exactly this. The phase menu turns to yellow when the phase changes and until it is done, shows the phase you've started from.

marcus-oscarsson:
Ok, so seems like this is cleared out now. Thanks alot for the PR. Ill merge it ASAP ! :)

axelboc:
Indeed, the PR prevents "Unknown" from being a selectable option, which is good regardless :+1:

PR #1852, opened 2025-09-03
Drop version number from the start-up log message

elmjag:
The logged version is outdated. Let's skip including version number in the start-up log message.


fabcor-maxiv (approved):
This breaks my workflow!

:D

fabcor-maxiv:
We could print the version strings of MXCuBE-Core and MXCuBE-Web, though... that would be useful.

PR #1853, opened 2025-09-03
remove N814 exceptions

walesch-yan:
This PR removes the `N814` rule exceptions from `ruff.toml`. The rule was ignored in many files since it triggers on 

```python
from mxcubecore import HardwareRepository as HWR
```

Since #1849, `HWR` is whitelisted, hence the file-wide exceptions are not needed anymore

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -30,7 +30,6 @@ extend-ignore-names = ["HWR"] # To avoid "N814": Camelcase `HardwareRepository` 
     "E501",
     "FBT002",
     "N813",
-    "N814",
     "PTH109",
 ]
 "mxcubeweb/__version__.py" = [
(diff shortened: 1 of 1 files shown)


fabcor-maxiv (approved):
Yep

fabcor-maxiv:
We could do the same for MXCuBE-Core.

walesch-yan:
> We could do the same for MXCuBE-Core.

Yes, to stay consistent with the configurations across both repos I agree

fabcor-maxiv:
> > We could do the same for MXCuBE-Core.
> 
> Yes, to stay consistent with the configurations across both repos I agree

Here: <https://github.com/mxcube/mxcubecore/pull/1363>.

PR #1858, opened 2025-09-03
remove unused actions in sampleview

walesch-yan:
This PR removes actions in `sampleView,` that are never called in the code. For some I will give a short explanation below

### DIFF
--- a/ui/src/actions/sampleview.js
+++ b/ui/src/actions/sampleview.js
@@ -13,10 +13,6 @@ import {
 } from '../api/sampleview';
 import { showErrorPanel } from './general';
 
-export function setMotorMoving(name, status) {
-  return { type: 'SET_MOTOR_MOVING', name, status };
-}
-
 export function setBeamInfo(info) {
   return { type: 'SET_BEAM_INFO', info };
 }
@@ -84,10 +80,6 @@ export function clearSelectedShapes() {
   return { type: 'CLEAR_SELECTED_SHAPES' };
 }
 
-export function addCentringPoint(x, y) {
-  return { type: 'ADD_CENTRING_POINT', x, y };
-}
-
 export function addShapeAction(shape) {
   return { type: 'ADD_SHAPE', shape };
 }
@@ -135,10 +127,6 @@ export function updateMotorState(name, value) {
 
 export function setShapes(shapes) {
   return { type: 'SET_SHAPES', shapes };
-}
-
-export function setGridOverlay(level) {
-  return { type: 'SET_GRID_OVERLAY', level };
 }
 
 export function toggleDrawGrid() {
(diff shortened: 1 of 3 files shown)


ui/src/actions/sampleview.js
```diff
@@ -137,10 +129,6 @@ export function setShapes(shapes) {
   return { type: 'SET_SHAPES', shapes };
 }
 
-export function setGridOverlay(level) {
-  return { type: 'SET_GRID_OVERLAY', level };
-}
-
```
walesch-yan, line 143: setOverlay is used instead

ui/src/reducers/beamline.js
```diff
@@ -221,20 +221,6 @@ function beamlineReducer(state = INITIAL_STATE, action = {}) {
...
-        motors: {
-          ...state.motors,
-          [action.name.toLowerCase()]: {
-            ...state.motors[action.name.toLowerCase()],
-            state: action.status,
-          },
-        },
-      };
-    }
-
```
walesch-yan, line 237: Weirdly enough an action (that was never called) defined in `sampleview` would trigger this function in the beamline reducer

ui/src/reducers/sampleview.js
```diff
@@ -34,25 +33,14 @@ function sampleViewReducer(state = INITIAL_STATE, action = {}) {
...
-      return state.clickCentringPoints.length === 2
-        ? { ...state, clickCentring: false, clickCentringPoints: [] }
-        : {
-            ...state,
-            clickCentringPoints: [
-              ...state.clickCentringPoints,
-              { x: action.x, y: action.y },
-            ],
-          };
-    }
```
walesch-yan, line 55: After removing this, the `clickCentringPoints`  variable from the state would always be an empty list. The variable was never used in the code, hence I also completely removed it from the state

ui/src/reducers/sampleview.js
```diff
@@ -91,9 +79,6 @@ function sampleViewReducer(state = INITIAL_STATE, action = {}) {
     case 'SET_IMAGE_RATIO': {
       return { ...state, imageRatio: action.clientWidth / state.width };
     }
-    case 'SET_VIDEO_SIZE': {
-      return { ...state, videoSize: action.width };
-    }
```
walesch-yan, line 96: Obsolete due to `SET_IMAGE_SIZE`

ui/src/reducers/sampleview.js
```diff
@@ -115,9 +100,6 @@ function sampleViewReducer(state = INITIAL_STATE, action = {}) {
         videoMessageOverlay: { show: action.show, msg: action.msg },
       };
     }
-    case 'SET_CENTRING_METHOD': {
-      return { ...state, centringMethod: action.centringMethod };
-    }
```
walesch-yan, line 120: There is an action in `actions/queue.js` that would trigger this. However, the centringMethod variable is exclusively called from `state.queue`, not `state.sampleView`. Hence, I removed it (the exact same code snippet exists in `reducers/queue.js`)

axelboc (approved):
Haha so much dead code, awesome :grin:

mockoocy (approved):

Issue #1845, opened 2025-09-01
Prepare beamline route


marcus-oscarsson:
There is a route/resource called `/prepare_beamline` that was requested by MAXIV back in the day, https://github.com/mxcube/mxcubeweb/blob/0b0225efadf763f1ddebd850cede0c0af6344c92/mxcubeweb/routes/beamline.py#L29.

@elmjag, @fabcor-maxiv, @matyson Is this route still in use, and what is it used for ?

elmjag:
Yes, we use it. This route calls `prepare_for_new_sample()` method on the collect hardware object.

It prepares beamline hardware for manually(?) mounting a new sample.

That code does things like:

- close detector cover
- put MD3 into transfer phase
- move back detector (to a 'safe' position)
- close safety shutter

As far as I understand, the idea is to put beamline in a state where a user can enter the hutch and exchange the sample on the diffractometer.

marcus-oscarsson:
I see, I'm wondering if we really need a route for this. I think this can be better done in a another place and not directly triggered from the UI. We at least know in the back end when a new sample is mounted manual or from the sample changer so it could be done in the `SamleChanger` or `Lims` components of `mxcubeweb`. The ideal would be if we could find a way to handle this in `mxcubecore`. 

Perhaps, you already have some ideas where this can be done better ?

beteva:
@elmjag we have exactly the same sequence when we change the sample, as it is part of the interlock for the sample changer robot. We call the procedure it in the local sample changer hardware object (see prepare_load in the FlexHCD.py), As @marcus-oscarsson said, it is more practical and natural to attach it to the sample changer and and the UI only triggers load/unload, no matter if it is automatic or manual, if is done with the sample changer.
Otherwise, we also have procedures which executes actions when we open/close the hutch. It is automatic and connected to the PSS (personal safety system), so can be triggered with the hutch door opening/closing. We also have in the hutch actions a button to choose if you want to use this or not.
I see that prepare_load is also implemented in other synchrotron’s sample changed HOs, so maybe it could be an idea to put a place holder in the AbstractSampleChanger.

elmjag:
I agree that it feels a bit weird that UI specifically triggers this 'prepare for manual sample' routine. If it's only used at MAX IV,  we should think about refactoring it.

I don't really know exactly sample mounting code works. I don't really have any suggestions. We, at MAX IV, will take a look at this, and see what suggestions we  come up with.

[...]

PR #1859, opened 2025-09-03
Moved all remaining beamline routes to BeamlineAdapter

marcus-oscarsson:
Moved all remaining beamline routes to the `BeamlineAdapter`. Ill add a few more details, and do some further cleanup ...


mxcubeweb/core/adapter/adapter_manager.py
```diff
@@ -124,32 +122,44 @@ def find_best_adapter(self, ho):
...
     def adapt_hardware_objects(self):
         _hwr = HWR.get_hardware_repository()
 
+        # Beamline is not added to the list of hardware objects
+        # returned by _hwr.hardware_objects as its considered a
+        # special object root object for all other hardware objects
+        # so we add it manually here. We give it the id 'beamline'
+        # so that it can be retrieved by the adapter manager but
+        # in reality has no id.
+        self.adapt_hardware_object(HWR.beamline, "beamline")
```
marcus-oscarsson, line 153: The `BeamlineAdapter` used to be a sort of special object. Now its a Adapter object like the others. However the `beamline` HardwareObject is considered special as its the container for all (most) other HardwareObjects. It does not necessarily have to be in this way, but It was a design decision made at the time the `Beamline` object was introduced. It also means that the `Beamline` object does **not have any id**. So we specifically add the `Beamline` object to the list adapted hardware objects and give the id "beamline" which is alright as far as adapters are concerned.

mxcubeweb/core/adapter/beamline_adapter.py
```diff
@@ -1,65 +1,72 @@
...
-    global BEAMLINE_ADAPTER
-
-    if BEAMLINE_ADAPTER is None:
-        BEAMLINE_ADAPTER = _BeamlineAdapter(*args)
-
-    return BEAMLINE_ADAPTER
-
-
-class _BeamlineAdapter:
+class BeamlineAdapter(ActuatorAdapterBase):
```
marcus-oscarsson, line 17: So the old some what strange custom `_BeamlineAdapter` object was made into a real `Adapter`

mxcubeweb/core/adapter/beamline_adapter.py
```diff
@@ -1,65 +1,72 @@
...
-                "GphlJsonParametersNeeded", self.gphl_json_wf_parameters_needed
+                "GphlJsonParametersNeeded", self._gphl_json_wf_parameters_needed
             )
             gphl_workflow.connect(
-                "GphlUpdateUiParameters", self.gphl_json_wf_update_ui_parameters
+                "GphlUpdateUiParameters", self._gphl_json_wf_update_ui_parameters
             )
 
-    def wf_parameters_needed(self, params):
+        from mxcubeweb.routes import signals
```
marcus-oscarsson, line 42: So this will still need some further cleanup, so that we can get rid of this import

mxcubeweb/core/adapter/beamline_adapter.py
```diff
@@ -76,13 +87,15 @@ def dict(self):
...
+        return {
+            "energyScanElements": self._get_available_elements(),
+            "path": HWR.beamline.session.get_base_image_directory(),
+            "actionsList": [],
+            "hardwareObjects": attributes,
+        }
 
-        if escan:
-            elements = escan.get_elements()
+    def prepare_beamline_for_sample(self) -> dict:
```
marcus-oscarsson, line 97: `prepare_beamline_for_sample` is now under the `BeamlineAdapter`, you guys (@fabcor-maxiv, @elmjag) have then to decide what to do with it.

ui/src/api/beamline.js
```diff
@@ -1,11 +1,17 @@
 import api from './api';
+import { fetchValue, sendExecuteCommand } from './hardware-object';
```
marcus-oscarsson: Ill remove this entire file and replace the calls to `fetchBeamlineSetup` and `sendPrepareBeamlineForNewSample` with `fetchValue` and `sendExecuteCommand`

mxcubeweb/core/adapter/beamline_adapter.py
fabcor-maxiv: Related to #1845 ?

marcus-oscarsson: yes, in a way. If you decide to do something with `prepare_beamline_for_sample`, keep in mind that it changed location, thats what I wanted to say.

PR #1861, opened 2025-09-04
Remove jest testing

walesch-yan:
This PR intends to remove code related to setting up `jest` tests. They are currently not used by the project, so I removed the dependency and an unused `setupTests.js` file.


axelboc (approved):
We would likely use something else like [Vitest](https://vitest.dev/) these days anyway.

PR #1862, opened 2025-09-04
clean-up beamline actions

walesch-yan:
This PR cleans up the beamline actions (as in redux actions). In principle, three things have been changed/removed:

- unused actions have been removed
- `actions/beamline.js` contained a list of action types as variables that should be supported, in practice only three were used, so removed them.
- There was an internal `STATE` enum used in the reducer that was missing possible states for the hwos, hence completely replaced the calls with the states defined in `constants.js`

### DIFF
--- a/ui/src/reducers/beamline.js
+++ b/ui/src/reducers/beamline.js
@@ -1,12 +1,5 @@
 import { HW_STATE, RUNNING } from '../constants';
 
-// The different states a beamline attribute can assume.
-export const STATE = {
-  IDLE: 'READY',
-  BUSY: 'BUSY',
-  ABORT: 'UNUSABLE',
-};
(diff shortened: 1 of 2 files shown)


ui/src/reducers/beamline.js
```diff
@@ -17,93 +10,93 @@ export const STATE = {
  *     state:  hardwareObjects current state, see STATE for more information
  *     msg:    arbitray message describing current state
  */
-export const INITIAL_STATE = {
+const INITIAL_STATE = {
```
walesch-yan, line 13: also removed this unecessary export

axelboc (approved):
<3

mockoocy (approved):

PR #1864, opened 2025-09-04
remove unecessary exports for harvester and login actions

walesch-yan:
This PR removes the `export` of actions in `login` and `harvester`, that are not used outside of the corresponding files


axelboc (approved):

mockoocy (approved):

marcus-oscarsson:
On a streak !

Issue #1654, opened 2025-05-21
Clean up video streaming related configuration settings


fabcor-maxiv:
Clean up video streaming related configuration settings.

See comment below for details: <https://github.com/mxcube/mxcubeweb/issues/1654#issuecomment-3302513433>.

fabcor-maxiv:
Looks like I created this issue, but I have no idea what this is about. Was this created during the code camp at the latest MXCuBE meeting? @laispc, @walesch-yan, did you ask me to create this issue? Do you remember details about what this is supposed to be about?

marcus-oscarsson:
@walesch-yan Is on vacation for two weeks. I think the issue had to do with how the video options MJPEG and MPEG are handled in the server.yaml file. I think @laispc and @walesch-yan worked on it a bit after the meeting. Perhaps @laispc knows we otherwise have to wait for @walesch-yan to get back.

laispc:
Hi @fabcor-maxiv ,

Yes, that's correct, we asked you to create this issues during the last Code Camp (CC) in Hamburg. But it's indeed missing a better descrition, so I am adding it on this comment.
 
## The issue

As @marcus-oscarsson said, it's related to the video options in server.yaml file, but actually it is more focused on the `USE_EXTERNAL_STREMER` variable.

In summary, this variable is ignored: setting it to either true or false does not make difference on which streamer is used (external or internal). In practice, the other 2 variables are more important for this selection: `VIDEO_FORMAT` and `VIDEO_STREAM_URL`.

We realized that for using the external streamer, the `server.yaml` variables have to be set this way:
```
USE_EXTERNAL_STREAMER: true  # Ignored, so can be either true or false.
VIDEO_FORMAT: MPEG1  # Has to be MPEG1 if you want to use video-streamer
VIDEO_STREAM_URL: ws://localhost:8000
```
And if you want to change to the internal streamer:
```
USE_EXTERNAL_STREAMER: true  # Ignored, so can be either true or false.
VIDEO_FORMAT: HTTP  # Has to be any string different from "MPEG1"
VIDEO_STREAM_URL: '' # Has to be an empty string! Otherwise, it will change to external streamer.
```
The above settings were also discussed/investigated in this ticket: https://github.com/mxcube/mxcubeweb/issues/1495

After some conversation around this before/during the last Meeting with @walesch-yan and @michalgucwa, we agreed that this behavior is not so user-friendly. Some other ideas were raised, like:
[...]

PR #1865, opened 2025-09-04
Removed integrated MJPEG streamer - Part1

marcus-oscarsson:
This PR removes the integrated MJPEG streamer and the option `USE_EXTERNAL_STREAMER`. The later permitted to specify whether the internal streamer or an external streamer (most likely the mxcube [video-streamer](https://github.com/mxcube/video-streamer)) was to be used.

With this PR its thus no longer possible to use the internal streamer and one has to rely on i.e [video-streamer](https://github.com/mxcube/video-streamer)

As far as we know everybody is today already using the `video-streamer` mentioned above. 

In **Part-1** we focus on removing the backend bits, we will follow up with a **Part-2** for the front end and possibly a **Part-3** with some general improvements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,16 @@
+# Changelog
+
+`mxcubeweb` project change log. This file outlines significant and/or breaking changes that
+are important to be share between developers.
+
+## [4.521.0] - 2025-17-09 - PR1865
+
+This change was done in PR 1865: https://github.com/mxcube/mxcubeweb/pull/1865
+
+The internal MJPEG streamer was removed. Use the video-streamer project
+for handling video streams. See: https://github.com/mxcube/video-streamer
+
+### Removed
+
+- Internal MJPEG streamer
+- Option `USE_EXTERNAL_STREAMER` from `server.yaml`
--- a/demo.yaml/mxcube-web/server.yaml
+++ b/demo.yaml/mxcube-web/server.yaml
@@ -22,7 +22,6 @@ sso:
   CODE_CHALLANGE_METHOD: S256
 
 mxcube:
-  USE_EXTERNAL_STREAMER: true
   VIDEO_FORMAT: MPEG1
   VIDEO_STREAM_URL: ws://localhost:8000/ws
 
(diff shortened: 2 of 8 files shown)


fabcor-maxiv:
Maybe we should really start getting into the habit of maintaining a change log. I know we discussed this already in the past but never acted on it (I am as guilty as anyone else here, of course, not blaming anyone). It came back to my mind now, seeing all the backwards-incompatible removals (the clean ups are much appreciated of course).

I guess it is something that needs its own discussion thread rather than this comment. I will try to get to it tomorrow, unless someone else beats me to it.

marcus-oscarsson:
Actually, we've discussed change logs in the past and I've proposed that we should have one. The overall concuss has however been to not have one. I cant remember how the reasoning when at the moment. I suggest that we re-discuss the topic on our next developers meeting.

**Edit**: So Ill be all for having one, but as we discussed it before I just want to make sure that its discussed again before we go-ahead.

marcus-oscarsson:
Actually, I also wanted to bring up the topic of "planned releases" on our next developers meeting. We can perhaps find a good way to make that work for the project.

marcus-oscarsson:
Judging from our last developer meeting it then seems fine to proceed with this PR.

CHANGELOG.md
```diff
@@ -0,0 +1,16 @@
+# Changelog
```
marcus-oscarsson, line 1: I added a change-log as we discussed this last time. While doing this I had a feeling that it does not really add very much, but I'm perhaps wrong and its perhaps because I follow the development very closely in anyways

fabcor-maxiv: I think it is fine to start experimenting with this topic and see what we are comfortable with.

marcus-oscarsson: Ok, lets do that :)

laispc:
Hi @marcus-oscarsson , just a quick comment: yesterday in my meeting with Yan and Michal (about #1654 ), we quickly tested this PR (as it was related to our topic) and it looked fine for us too. :sparkles: 

We also tested it with `VIDEO_FORMAT` variable set to "MJPEG" (instead the original "MPEG1"), and the image got blank in the UI, but I think it's expected. The rest of the UI was fine.

marcus-oscarsson:
Thanks for the heads-up, we will clean up the front-end bit in "Part-2" and have a look at the MJPEG option.

PR #1867, opened 2025-09-05
cleanup remoteAccess actions

walesch-yan:
This is a straightforward cleanup, mainly the `setObservers` action is not used anymore, instead the `observers` state variable is updated during `SET_RA_STATE`, which is called every few seconds


axelboc (approved):

marcus-oscarsson:
Wonderful, thanks alot @walesch-yan  !

PR #1868, opened 2025-09-08
SSRF BL19U1 Beamline

Baokangwen:
SSRF BL19U1 Beamline mxcubeweb code


marcus-oscarsson:
Thanks, we are happy to see your PRs ! Do you think you can make the exercise as with https://github.com/mxcube/mxcubecore/pull/1369 ?

marcus-oscarsson:
Thanks we really appreciate the PR @Baokangwen, do you think you can re-base this one and cleanup the commit history a bit ? Would it be possible to use latin characters ? I'd love to understand Mandarin but I think it would be complicated.

PR #1872, opened 2025-09-09
Fixing Rotate to option of mesh grid

dominikatrojanowska:
Related discussion: https://github.com/mxcube/mxcubecore/issues/1360

Max IV Diffractometer objects inheriting from `GenericDiffractometer`, do not introduce `centringPhi` attribute, instead they have `phi_motor_hwobj`.

### DIFF
--- a/mxcubeweb/core/components/sampleview.py
+++ b/mxcubeweb/core/components/sampleview.py
@@ -350,8 +350,20 @@ class SampleView(ComponentBase):
             phi_value = round(float(cp.as_dict().get("phi", None)), 3)
             if phi_value:
                 try:
-                    HWR.beamline.diffractometer.centringPhi.set_value(phi_value)
+                    logging.getLogger("user_level_log").info(
+                        "Rotating to position %s", phi_value
+                    )
+                    if hasattr(HWR.beamline.diffractometer, "centringPhi"):
+                        HWR.beamline.diffractometer.centringPhi.set_value(phi_value)
+                    else:
+                        # This is for backward compatibility - handle case where
+                        # `sample_centring` was not used in configuration files or
+                        # where it is set to false, in order to perform manual centring
+                        HWR.beamline.diffractometer.phi_motor_hwobj.set_value(phi_value)
                 except Exception:
+                    logging.getLogger("user_level_log").exception(
+                        "Error rotating to position"
+                    )
                     raise
 
     def start_auto_centring(self):


mxcubeweb/core/components/sampleview.py
```diff
@@ -350,8 +350,17 @@ def rotate_to(self, sid):
...
             if phi_value:
                 try:
-                    HWR.beamline.diffractometer.centringPhi.set_value(phi_value)
+                    logging.getLogger("user_level_log").info(
+                        "Rotating to position %s", phi_value
+                    )
+                    if hasattr(HWR.beamline.diffractometer, "centringPhi"):
+                        HWR.beamline.diffractometer.centringPhi.set_value(phi_value)
+                    elif hasattr(HWR.beamline.diffractometer, "phi_motor_hwobj"):
+                        HWR.beamline.diffractometer.phi_motor_hwobj.set_value(phi_value)
```
fabcor-maxiv: Maybe this to avoid the backwards-incompatible change where this would silently fail if there is no motor `phi_motor_hwobj` nor `centringPhi`:

```suggestion
                    if hasattr(HWR.beamline.diffractometer, "phi_motor_hwobj"):
                        HWR.beamline.diffractometer.phi_motor_hwobj.set_value(phi_value)
                    else:
                        HWR.beamline.diffractometer.centringPhi.set_value(phi_value)
```

mxcubeweb/core/components/sampleview.py
```diff
@@ -350,8 +350,17 @@ def rotate_to(self, sid):
             phi_value = round(float(cp.as_dict().get("phi", None)), 3)
             if phi_value:
                 try:
-                    HWR.beamline.diffractometer.centringPhi.set_value(phi_value)
+                    logging.getLogger("user_level_log").info(
+                        "Rotating to position %s", phi_value
+                    )
+                    if hasattr(HWR.beamline.diffractometer, "centringPhi"):
```
marcus-oscarsson, line 356: Hm, My suggestion would be that until we have the `AbstractDiffractometer` that we make sure that  `centringPhi` exists on the relevant Diffractometer objects. Alternatively, agree on another name that can be used for the future.

dominikatrojanowska: Cofiguration property `centering_sample` is responsible for presence of centeringPhi attribute of diffractometer, it is defined here in [GenericDiffractometer init method](https://github.com/mxcube/mxcubecore/blob/a72b7aeecb78b4f2c83c2cf1b0670d3cf2bdd1a8/mxcubecore/HardwareObjects/GenericDiffractometer.py#L454) (I was wrong previously that it is not introduced at all)
Some of the configuration files does not introduce this property. It's a case in the demo configs, maybe in some site-specific configs and in case of Max IV we do perform sample centring manually (@fabcor-maxiv please, confirm) so in our case the config property is set to false

dominikatrojanowska: in the future with AbstractDiffractometer deplyed it should be something like this:
```
HWR.beamline.diffractometer.motors_hwobj_dict["omega"].set_value(value)
// or
HWR.beamline.diffractometer.set_value_motors({"omega": value})
```

marcus-oscarsson: I think we will/should assume a certain well know geometry so that omega is always omega so to speak, meaning that it would be:
```
HWR.beamline.diffractometer.omega_motor.set_value(...)
```

beteva: The AbstractDiffractometer should solve this.The names of the diffractometer motors are fixed by their roles, so the rotation axis will always be omega.  You could  simply do HWR.beamline.diffractometer..omega.set_value(value)

dominikatrojanowska: cool, so what can we do **until** the https://github.com/mxcube/mxcubecore/pull/695 will be merged? seems like it is a long term, huge change.  
Can we merge the current PR? I will create a new one with your suggestion, but this will be for the future...

dominikatrojanowska: here it is: https://github.com/mxcube/mxcubeweb/pull/1877

marcus-oscarsson: I would then suggest to keep things as they already are, as this would otherwise risk to introduce a breaking change. We would need to verify that `omega` exists if you introduce `omega` like in https://github.com/mxcube/mxcubeweb/pull/1877.

You otherwise need to first make a PR to `mxcubecore` where we make sure `omega` is set correctly everywhere before merging  https://github.com/mxcube/mxcubeweb/pull/1877.

dominikatrojanowska: well, in that case the `Rotate to` function is broken at our site :disappointed: what is the breaking change I am introducing here? Sites without `sample_centring` set to `true`  should not use `Rotate to` function? :thinking:

marcus-oscarsson: Ok, that is possible, but we can't make a change that will break this feature for others without taking care. You just need to make sure that you have a `centringPhi` on your `HardwareObject` (for the time being). You could do, `self.centringPhi = self.omega` for instance. 

Or we need to make sure that `centringPhi` on all diffractometers is renamed to `omega` (if that is what we want to call it)

dominikatrojanowska:
overtaken by: https://github.com/mxcube/mxcubecore/pull/1385

PR #1873, opened 2025-09-11
split SampleView.css into multiple modules

walesch-yan:
This PR splits the `SampleView.css` into two smaller modules and removes selectors, that are no longer used in the code


ui/src/components/SampleView/SampleView.css
```diff
@@ -1,176 +0,0 @@
...
-table.table-striped > tbody > tr[data-selected='true'] {
-  background-color: #dff0d8;
-  color: #000;
-  font-weight: bold;
-}
-
-table.table-striped > tbody > tr[data-user-hidden='true'] {
-  background-color: #a3ceff;
-  font-style: italic;
-}
```
walesch-yan, line 176: These two were a little scary to remove, fortunately I checked and `data-selected`, resp. `data-user-hidden` are only set for elements in non-striped tables, except inside the `Gridform`, so I added it to the new module and created a specific selector for it instead of targeting all striped tables

axelboc (approved):
Awesome!

ui/src/components/SampleView/SampleView.css
```diff
@@ -1,176 +0,0 @@
...
-  width: 33.33333333%;
-}
-
-.divider {
-  margin-bottom: 0;
-}
-
-.canvas-container {
-  position: relative;
-}
```
walesch-yan, line 70: So this one was targeting a specific selector that is used inside `Fabric`'s  internal logic, but it was then again overwritten by the `coveringCanvas` selector

walesch-yan (commented):
Just two more small comments on particularities I discovered

ui/src/components/SampleView/SampleView.css
axelboc: Great, makes sense :+1:

PR #1874, opened 2025-09-11
split SampleGridTable.css into multiple modules

walesch-yan:
This was a big one :sweat_smile:. This PR splits `SampleGridTable.css` into multiple modules and removes unused `classNames` and selectors in all `SampleGridTable` related files. This leaves only a few more pure `css` files in the repo


ui/src/containers/SampleGridTableContainer.jsx
```diff
@@ -754,16 +750,12 @@ export default function SampleGridTableContainer(props) {
         >
           <Collapsible
             transitionTime={300}
-            className="sample-items-collapsible"
-            openedClassName="sample-items-collapsible"
+            className={styles.sampleItemsCollapsible}
```
walesch-yan, line 753: removed `openedClassName` here, because it is the same as `className`, which is always applied

ui/src/containers/SampleGridTableContainer.jsx
```diff
@@ -754,16 +750,12 @@ export default function SampleGridTableContainer(props) {
...
             open
             onClosing={forceVisible()}
             lazyget
-            trigger={getCollapsibleHeaderClose(cellID, 'collapsible-arrow-c')}
-            triggerWhenOpen={getCollapsibleHeaderOpen(
-              cellID,
-              'collapsible-arrow-c',
-            )}
+            trigger={getCollapsibleHeaderClose(cellID)}
+            triggerWhenOpen={getCollapsibleHeaderOpen(cellID)}
```
walesch-yan, line 758: passing the `css` class as a parameter to these functions was unnecessary, because they were only ever called once

ui/src/containers/SampleIsaraView.jsx
```diff
@@ -35,26 +27,6 @@ class NewSampleIsaraView extends React.Component {
...
-  // eslint-disable-next-line react/no-unused-class-component-methods
-  getCollapsibleHeaderClose(cssClass) {
-    return (
-      <div className="sample-items-collapsible-header">
-        <b>Isara</b>
-        <BsChevronDown className={cssClass} size="1em" />
-      </div>
-    );
-  }
-
```
walesch-yan, line 57: Might go around the scope of this PR, but I removed these functions, because they are not used and it saves me from importing styles, that are defined in another module

ui/src/containers/SampleIsaraView.jsx
```diff
@@ -19,14 +19,6 @@ class NewSampleIsaraView extends React.Component {
...
   }
 
-  // eslint-disable-next-line react/no-unused-class-component-methods
-  onClickCell(event, idx, disabled) {
-    if (!disabled) {
-      this.props.filter({ cellFilter: `${idx + 1}` });
-    }
-    event.stopPropagation();
-  }
-
```
walesch-yan, line 29: While at it, I also removed this unused function

ui/src/main.css
```diff
@@ -129,3 +129,7 @@ input[type='color']:focus,
 .nowrap-style {
   white-space: nowrap;
 }
+
+.btn-link:not([data-default-styles]) {
+  color: #333 !important;
+}
```
walesch-yan, line 135: finally, this one was scary, especially, because there are a few occasions of `<Button variant="link \>` elements in the code (outside of the `SampleGridTable` scope as well), to which this style might apply, hence I just decided to make it a global change by adding it to `main.css`, but we could change this

axelboc: Good call :100:

axelboc (approved):
Huge! Thanks for keeping the changes to the bare minimum for now, it made the review trivial :heart:

You've probably planned some more refactoring from here on, but the main two I noticed would be:

- to shorten some of the classnames now that they are properly scoped (e.g. in `SampleGridTableItemContainer`: `.sampleGridTableItem` => `.item`, that sort of thing)
- to replace some of the "state" classnames with `data` attributes to avoid using `cx` (e.g. `data-selected={selected[key] | undefined}`).

mockoocy (approved):

ui/src/containers/SampleIsaraView.jsx
jbflo: t i think the [SampleIsaraView.jsx] is not maintain/use , we could investigate if whether we could remove it.

walesch-yan:
> Huge! Thanks for keeping the changes to the bare minimum for now, it made the review trivial ❤️
> 
> You've probably planned some more refactoring from here on, but the main two I noticed would be:
> 
> * to shorten some of the classnames now that they are properly scoped (e.g. in `SampleGridTableItemContainer`: `.sampleGridTableItem` => `.item`, that sort of thing)
> * to replace some of the "state" classnames with `data` attributes to avoid using `cx` (e.g. `data-selected={selected[key] | undefined}`).

That's exactly what I had in mind to do later, + check if some styles could need an update/improvement, So for this PR it was really just about the splitting (and trying to not introduce any styling issues while at it :sweat_smile), so that further refactoring will be easier and it keeps the review simple

PR #1876, opened 2025-09-15
Remove unused NewSampleIsaraView

matyson:
Following the discussion in https://github.com/mxcube/mxcubeweb/pull/1819#issuecomment-3273866768. The component `NewSampleIsaraView` seems to be deprecated and no longer used in MAXIV. 

However, i am not sure if it's still being used by other facilities, so i opened this to expand the discussion and, perhaps, move towards even removing this [Cats90](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/HardwareObjects/Cats90.py#L144) hwo as it can be deprecated as well.


marcus-oscarsson:
Thanks, looks good !

marcus-oscarsson:
Well done !

marcus-oscarsson:
Also, thanks for the heads up. Its better to remove this component for now and then eventually re-introduce something similair if needed in the future, seeing that its not been maintained for some time.

PR #1877, opened 2025-09-15
Fixing `Rotate to` option of mesh grid with AbstractDiffractometer introduced

dominikatrojanowska:
According to [discussion](https://github.com/mxcube/mxcubeweb/pull/1872#discussion_r2341048310) adjusted for mxcubecore containing AbstractDiffractometer. To be tested and merged after: https://github.com/mxcube/mxcubecore/pull/695


marcus-oscarsson:
Please let us know how you would like to do here, I guess there are plenty of options. The easiest would be to introduce `centringPhi` on your diffractometer. Its otherwise possible to introduce `omega_motor` or similar but this has to be done with some care and comes with some risk, or some sort of configuration attribute on the diffractometer that denotes rotation axis.

PR #1879, opened 2025-09-16
Sync number of snapshots between HWR and frontend on startup

matyson:
When not explicitly setting `num_snapshots` in `collect.yaml` file and loading this property in `MYCollect`(for [example](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/HardwareObjects/ESRF/ESRFMultiCollect.py#L40)) hwo the `number_of_snapshots` on the frontend may be different than the one set on the backend due to the fact that we are keeping two sources of truth for the same data ( in the reducer store and in the HWO) and we are also keeping at least three aliases for the same thing, e.g., `take_snapshots`, `num_snapshots` and `number_of_snapshots`.

TBH, im not happy with this solution as it feels hacky, but i didn't felt comfortable with diving into `AbstractCollect` and `AbstractMultiCollect` objects and trying to unify these properties. Im willing to try it, though, just need some guidance (in particular with the data collection parameters which its use isn't clear for me).

### DIFF
--- a/mxcubeweb/core/components/queue.py
+++ b/mxcubeweb/core/components/queue.py
@@ -280,6 +280,10 @@ class Queue(ComponentBase):
             "centringMethod": HWR.beamline.queue_manager.centring_method,
         }
 
+        self.app.queue.set_num_snapshots(
+            res.get("numSnapshots", self.app.DEFAULT_NUM_SNAPSHOTS)
+        )
+
         res.update(settings)
         return res
 


marcus-oscarsson:
Well done ! I think you did well to not dig too deep into this issue immediately. This code is very convoluted and is partly used by the workflow engine so its not that easy to debug. We will try to help out with cleaning this one up, we would also like to make this much simpler. 

I think your fix is fine for the time being.

PR #1880, opened 2025-09-17
Validate beamline attributes and handle server errors

### LINKED ISSUE #1842: BeamlineAttributeForm does not display errors or handle limits
BeamlineAttributeForm, seen below, does not handle limits or display any possible error that occurs wile setting the value

Image

axelboc:
Fix #1842

The form is now managed with `react-hook-form`, with the following validation rules:

- required and valid number
- value with beamline attribute `limits`

For UX, the errors are of course not shown until the form becomes "dirty", that is until the user first tries to submit a value. After that, the validation happens as the user types — this way, they can see quickly if they go over/under the limits.

[Screencast from 2025-09-17 11-44-56.webm](https://github.com/user-attachments/assets/dbff1179-d01b-44d8-b6ce-9e66bd5bca41)

If a server error occurs during submission, the form shows "Unable to set value". In the recording below, I disabled the max validation to demonstrate.

> Side note: the server currently responds with a 500 internal server error, which isn't ideal.

[Screencast from 2025-09-17 11-33-23.webm](https://github.com/user-attachments/assets/abc93f86-4153-491d-8855-d91b2e9e9850)

### DIFF
--- a/ui/src/components/BeamlineAttribute/BeamlineAttributeForm.jsx
+++ b/ui/src/components/BeamlineAttribute/BeamlineAttributeForm.jsx
@@ -1,60 +1,73 @@
-import { useEffect, useRef } from 'react';
+import { useEffect } from 'react';
 import { Button, ButtonToolbar, Form } from 'react-bootstrap';
+import { useForm } from 'react-hook-form';
 import { TiTick, TiTimes } from 'react-icons/ti';
 
 import styles from './BeamlineAttribute.module.css';
 
 function BeamlineAttributeForm(props) {
-  const { value, isBusy, step, precision, onSave, onCancel } = props;
-  const inputRef = useRef(null);
+  const { value, isBusy, step, precision, limits, onSave, onCancel } = props;
+
+  const {
+    register,
+    setFocus,
+    setError,
+    handleSubmit: makeOnSubmit,
+    formState: { isDirty, errors },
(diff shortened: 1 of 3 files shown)


ui/src/components/BeamlineAttribute/BeamlineAttributeForm.jsx
```diff
@@ -1,60 +1,73 @@
...
-            <TiTimes size="1.5em" />
-          </Button>
-        ) : (
-          <Button type="submit" variant="success" size="sm">
-            <TiTick size="1.5em" />
-          </Button>
-        )}
-      </ButtonToolbar>
+    <Form noValidate onSubmit={makeOnSubmit(handleSubmit)}>
+      <div className="d-flex">
```
axelboc, line 43: Adding a `div` here so the error message appears underneath the field and button. The relevant changes can be seen by hiding whitespace changes in GitHub.

marcus-oscarsson:
Wouhha, awesome !

marcus-oscarsson:
> Side note: the server currently responds with a 500 internal server error, which isn't ideal.
What error code should we use, I guess 400 (Bad Request) ?

marcus-oscarsson:
Thanks again, very nice !

axelboc:
No worries, yeah 400 with a clear error message about the limits or whatever else goes wrong.

PR #1881, opened 2025-09-17
[WIP] AbstractDiffractometer and AbstartSampleView

beteva:
Changes due to AbstractDiffractometer and AbstartSampleView new implementation. Dependant on
mxcube/mxcubecore/pull/1386


mxcubeweb/core/adapter/diffractometer_adapter.py
```diff
@@ -53,8 +55,8 @@ def state(self):
         return "READY" if self._ho.is_ready() else "BUSY"
 
     def head_configuration(self) -> dict:
-        data = self._ho.get_head_configuration()
-        return data.dict() if data else {}
+        data = self._ho.get_head_type.name  # noqa: F841
+        return {}
```
walesch-yan, line 59: It seems that this functionality is now missing, so `get_head_configuration` does not exist on `AbstractDiffractometer`, but (only) needed for SSXChip

PR #1884, opened 2025-09-19
Processing fields are required if at least one of them is filled in

dominikatrojanowska:
Max IV processing pipelines can't deal with partially filled up processing parameters. In our case such data needs to be reprocessed and that can even lead to lost beamtime.
If any site found this partial fill-up useful, we'll try to solve it another way.

Data Collection Processing parameters at current PR stage:

image

`Space group` as `SelectField` is not indicating the error value due to some bug, fixed here https://github.com/mxcube/mxcubeweb/pull/1885


marcus-oscarsson:
Thanks !

PR #1885, opened 2025-09-19
Fixing validation of `SelectField`

dominikatrojanowska:
In case of error in `SelectField` value the style is consist wih the `InputField`: indicator, red border line, error information.
Previously, the only effect of inappropriate value were inactive `Run` and `Add to Queue` buttons.

Screenshot from 2025-09-19 14-39-36


marcus-oscarsson:
Thanks @dominikatrojanowska

PR #1888, opened 2025-09-23
Disable unavailable elements in PeriodicTable + refactor

mockoocy:
There are two things, kinda packed together. The reason is that I wanted originally to prevent user from being able to select an element that is not available.
When working with it, it was really annoying that data was defined alongside the layout in the JSX, so I've separated these. 
The diff is quite lengthy, since I've removed ~1300 lines of JSX and added a ~1000 lines long json, hence I've commented more interesting bits over there.

For convenience, copy-pasted commit messages:
Move element data to a json file

Currently the JSX for the `PeriodicTable` component is very long, as it
both defines the data and renders it.
It is ~800 lines of mostly repeated code.
I've added a json blob, with selected fields from
[there](https://github.com/Bowserinator/Periodic-Table-JSON/blob/master/PeriodicTableJSON.json).

There are also some very minor changes in variable names, etc.

Disable unavailable elements

Currently elements which are not set as available (via `energy_scan`
config file) have lower opacity.
They can still be selected (although selecting them will lead to an
invalid state of the form; the Element field will be filled, but Edge
field will not.
This change removes the possibility of selecting such fields.


ui/src/components/PeriodicTable/PeriodicTable.jsx
```diff
@@ -10,1398 +11,61 @@ export default class PeriodicTable extends React.Component {
...
+                styles.element,
+                availableElements.includes(element.symbol) && styles.available,
+                selectedElement === element.symbol && styles.selected,
+              ]
+                .filter(Boolean)
+                .join(' ')}
+            >
+              <div className={styles.atomicNumber}>{element.number}</div>
+              <div className={styles.symbol}>{element.symbol}</div>
+              <div className={styles.atomDetails}>
```
mockoocy, line 52: In reality, when operating with screen width of over 800 pixels, then only the element's symbol is displayed.
That is a very odd behaviour (I'd expect extra text to appear on larger screen sizes). 
I suspect these 2 additional divs could be removed in the future.

ui/src/components/PeriodicTable/PeriodicTable.jsx
```diff
@@ -10,1398 +11,61 @@ export default class PeriodicTable extends React.Component {
...
+              data-category={element.category}
+              data-column={element.column}
+              className={[
+                styles.element,
+                availableElements.includes(element.symbol) && styles.available,
+                selectedElement === element.symbol && styles.selected,
+              ]
+                .filter(Boolean)
+                .join(' ')}
+            >
```
mockoocy: It's there, so that it's possible to get rid of logic regarding dynamically adding classnames that was defined in the other methods of the component.

ui/src/components/PeriodicTable/PeriodicTable.module.css
```diff
@@ -73,79 +71,73 @@
...
-.periodicRow:nth-child(-n + 6) .cell:nth-child(18):hover .element {
+
+.element[data-category='noble gas']:hover {
   background-color: rgba(64, 192, 0, 0.9);
 }
-.periodicRow:nth-child(n + 9) .cell:hover .element {
+.element[data-category='lanthanide']:hover,
+.element[data-category='actinide']:hover {
   background-color: rgba(0, 96, 160, 0.9);
 }
```
mockoocy: This creates the same colouring of the elements.
Used data-properties, as it seemed like the most convenient way to achieve the same effect; could not use the `nth-child` since the table has been rendered as a css grid.

ui/src/components/PeriodicTable/PeriodicTable.module.css
```diff
@@ -45,18 +44,17 @@
   border-color: rgba(0, 0, 0, 0.1);
   transform: scale(1.5, 1.5);
   z-index: 1;
-  background-color: #008080;
```
mockoocy, line 48: background colour is now set by the `div.element`, not `div.cell`. Since cell will expand to match it's content, the visual effect is the same anyway.
Also, each possiblle cell is coloured; the css rules below cover all the elements anyway.

ui/src/components/PeriodicTable/elements.json
axelboc, line 1: Just `elements.json` for the filename would be best I think. It's already in the `PeriodicTable` folder.

ui/src/components/PeriodicTable/PeriodicTable.jsx
axelboc: I would also use `data` attributes for these:

```jsx
data-available={availableElements.includes(element.symbol) || undefined}
data-selected={selectedElement === element.symbol || undefined}
```

So that means that `tableCellNode.children[0].className = styles.element;` in `onClickHandler` can be removed, right? No worries if you need to do it in this PR.

ui/src/components/PeriodicTable/PeriodicTable.module.css
```diff
@@ -73,79 +71,73 @@
...
-
-.cell:nth-child(-n + 2) .element,
-.cell:nth-child(n + 13) .element {
+.element[data-column='2'],
+.element[data-column='13'],
+.element[data-column='14'],
+.element[data-column='15'],
+.element[data-column='16'],
+.element[data-column='17'],
+.element[data-column='18'] {
```
axelboc: It might be simpler to add an `"alt": true` field to all the elements in those columns in the JSON file, and a `data-alt={element.alt || undefined}` attribute. (Or `altBg` or whatever makes the most sense.)

ui/src/components/PeriodicTable/PeriodicTable.module.css
```diff
@@ -5,20 +5,19 @@
...
-  height: 8%;
-}
-
 .cell {
   float: left;
   position: relative;
-  width: 5.55%;
-  height: 100%;
+  width: auto;
+  height: auto;
```
axelboc: These two properties may not be needeed?

ui/src/components/PeriodicTable/PeriodicTable.module.css
```diff
@@ -73,79 +71,73 @@
...
-}
-.periodicRow:nth-child(10) .cell:nth-child(18) .element {
-  transform-origin: 100% 100%;
-}
-
 @media (min-width: 600px) {
   .periodic {
     height: 460px;
     margin-right: -2px;
+    display: grid;
```
axelboc: What does it look like below 600px? I see there is still `float: left`, so I assume it just flows based on the available space? Might still be possible to achieve the same thing with `display:grid` and `grid-auto-columns` or something (just to remove `float: left`... :joy: )?

ui/src/components/PeriodicTable/PeriodicTable.module.css
```diff
@@ -5,20 +5,19 @@
...
-  height: 100%;
+  width: auto;
+  height: auto;
   opacity: 0.4;
+  /* overrode when element is available */
+  pointer-events: none;
 }
 
-.cell.available {
+.cell:has(.element.available) {
```
axelboc: Why not move the `available` class (or `data-available` attribute) to `div.cell`?

All in all, this should do what you want in fewer lines:

```
.cell {
  float: left;
  position: relative;
  cursor: pointer;
}

.cell:not([data-available]) {
  pointer-events: none;
  opacity: 0.4;
}
```

axelboc (commented):
Great call moving the elements data to JSON. Glad to see those dynamic `className` assignments go away! :clap: 

I just have a few suggestions with `data-` attributes to clean up the CSS a bit further.

elmjag:
FIY @axelboc and everybody else, [mockoocy](https://github.com/mockoocy) is on leave right now. He will be back in a week or so.

ui/src/components/PeriodicTable/PeriodicTable.module.css
mockoocy: I can't recall why I've put `display: grid` in the media query, not in outer scope tbh. Probably did not notice the query :D
Also I've removed the `float: left`, so now it will be just a smaller grid. Good catch!

axelboc (approved):
Nicely done!

PR #1890, opened 2025-09-23
fix a few dev dependency vulnerabilities

walesch-yan:
This PR should fix a few vulnerabilities coming from subdependencies of our dev dependencies. In general the things that have been updated to newer versions are:

- vite
- cypress
- eslint
- wait-on (only used for CI workflows)

Quickly tested and seems that the version updates did not break anything

### DIFF
--- a/ui/src/serverIO.js
+++ b/ui/src/serverIO.js
@@ -62,11 +62,9 @@ import { store } from './store';
 const { dispatch } = store;
 
 class ServerIO {
-  constructor() {
-    this.hwrSocket = null;
-    this.loggingSocket = null;
-    this.connectionLostTimeout = undefined;
-  }
+  hwrSocket = null;
+  loggingSocket = null;
+  connectionLostTimeout;
 
   listen() {
     clearTimeout(this.connectionLostTimeout);
--- a/ui/package.json
+++ b/ui/package.json
@@ -53,19 +53,19 @@
     "wretch": "2.7.0"
   },
   "devDependencies": {
-    "@esrf/eslint-config": "1.1.5",
+    "@esrf/eslint-config": "1.5.1",
     "@testing-library/cypress": "10.0.3",
     "@types/react": "17.0.60",
     "@types/react-dom": "17.0.20",
     "@types/react-redux": "7.1.33",
     "@vitejs/plugin-react-swc": "3.8.0",
-    "cypress": "14.5.2",
+    "cypress": "14.5.4",
     "cypress-terminal-report": "6.1.2",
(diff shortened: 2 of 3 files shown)


ui/src/serverIO.js
```diff
@@ -62,11 +62,9 @@ import { store } from './store';
...
 
 class ServerIO {
-  constructor() {
-    this.hwrSocket = null;
-    this.loggingSocket = null;
-    this.connectionLostTimeout = undefined;
-  }
+  hwrSocket = null;
+  loggingSocket = null;
+  connectionLostTimeout;
```
walesch-yan, line 67: Added this change as well, as the new version of the linter was complaining about it

axelboc (approved):

fabcor-maxiv:
I know it is out of scope for this merge request but still I want to mention it since I am thinking about it now and we can open a separate issue for discussion if you think it is worth it...

The usage of things like `wait-on` (and `&` to put the server process in the background) always seemed suspicious to me. I had a quick look and there seems to be an "official" Cypress GitHub Action that seems to be able to handle starting (and stopping!) the server in the background in a more controlled manner: <https://github.com/marketplace/actions/cypress-io#start-server>. Is it something worth considering? On one hand it would be more elegant. On the other hand we do not know what that Cypress Action really does and if we can trust it.

fabcor-maxiv:
Ha! And now I see that the `wait-on` solution is suggested in the Cypress documentation itself: <https://docs.cypress.io/app/continuous-integration/overview#Solutions>. So, I guess it is fine. Probably that is what everyone else does anyway... :D

marcus-oscarsson:
@fabcor-maxiv: Why do you think its suspicious, because we don't know when the actual process has really started or for some other reason, just curious ?

fabcor-maxiv:
> Why do you think its suspicious, because we don't know when the actual process has really started or for some other reason, just curious ?

I guess it just feels odd to me and kind of brittle as well. But it is a pattern that I see in many projects (particularly in container based projects which is nonsense when we have built-in features like [`HEALTHCHECK`](https://docs.docker.com/reference/dockerfile/#healthcheck) that way too few people use).

* Who takes care of stopping the server? (I guess it is run in a container so the container gets killed anyway)
* Shouldn't it be the responsibility of the client itself (Cypress in this case) to wait for the server and retry until the server is ready?
* `wait-on` is yet another dependency (these days we should really think about those things), but here it would be trading a dependency for another (the Cypress GitHub Action)

But really it should be its own discussion now. Sorry for the ramblings... :D

PR #1900, opened 2025-09-26
Connect `SSXChipControl` to Redux store

axelboc:
I'm trying to simplify `SampleViewContainer` as much as possible to make it easier to refactor it to a function component.

In this PR, I start by connecting `SSXChipControl` to the Redux store directly so that I can remove a lot of props. I'll do the same with the other components in separate PRs.

---

FYI, I'm considering to refactor `SSXChipControl` further but it seems to be broken at the moment in the demo. Maybe I'm missing some configuration?  :shrug: 

**EDIT**: there's apparently ongoing work on this feature, so I'll wait until it's done

### DIFF
--- a/ui/src/containers/SampleViewContainer.jsx
+++ b/ui/src/containers/SampleViewContainer.jsx
@@ -4,11 +4,7 @@ import { Col, Container, Row } from 'react-bootstrap';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 
-import {
-  executeCommand,
-  logFrontEndTraceBack,
-  setAttribute,
-} from '../actions/beamline';
+import { logFrontEndTraceBack, setAttribute } from '../actions/beamline';
 import { displayImage, showErrorPanel } from '../actions/general';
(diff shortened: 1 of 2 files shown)


ui/src/containers/SampleViewContainer.jsx
```diff
@@ -142,20 +136,7 @@ class SampleViewContainer extends Component {
...
-                  showForm={this.props.showForm}
-                  currentSampleID={currentSampleID}
-                  sampleData={this.props.sampleList[currentSampleID]}
-                  defaultParameters={this.props.defaultParameters}
-                  groupFolder={this.props.groupFolder}
-                  hardwareObjects={this.props.hardwareObjects}
-                  uiproperties={uiproperties.sample_view_motors}
-                  sampleViewActions={this.props.sampleViewActions}
-                  grids={grids}
-                  selectedGrids={selectedGrids}
```
axelboc, line 155: `selectedGrids` was not used by `SSXChipControl`

PR #1901, opened 2025-09-26
Connect `SampleImage` to Redux store directly and bind sample view actions

axelboc:
As with #1900, I'm trying to simplify `SampleViewContainer` as much as possible so I can refactor it to a function component. (I've marked #1900 as draft for now as there is apparently incoming work on the SSX chip feature, so best not to risk conflicts at this stage.)

In this PR, I connect `SampleImage` to the Redux store directly so that I can move all the sample view actions bindings down one level.

### DIFF
--- a/ui/src/actions/sampleview.js
+++ b/ui/src/actions/sampleview.js
@@ -29,7 +29,7 @@ export function setOverlay(level) {
   return { type: 'SET_OVERLAY', level };
 }
 
-export function setAperture(size) {
+function setAperture(size) {
   return { type: 'SET_APERTURE', size };
 }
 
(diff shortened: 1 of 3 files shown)


ui/src/containers/SampleViewContainer.jsx
```diff
@@ -19,7 +19,6 @@ import {
   setPlate,
 } from '../actions/sampleChanger';
 import { syncWithCrims } from '../actions/sampleGrid';
-import * as sampleViewActions from '../actions/sampleview'; // eslint-disable-line import/no-namespace
```
axelboc, line 22: This was the last remaining namespace import! It helped `knip` identify more unused exports in `actions/sampleview.js`.

ui/src/actions/sampleview.js
```diff
@@ -29,7 +29,7 @@ export function setOverlay(level) {
   return { type: 'SET_OVERLAY', level };
 }
 
-export function setAperture(size) {
```
axelboc, line 32: Unused exports identified by `knip` now that we no longer `import *` this file in `SampleViewContainer`.

marcus-oscarsson:
Thanks alot !

PR #1903, opened 2025-09-26
Refactor derived state computations in `SampleViewContainer`

axelboc:
Refactoring some imperative code to a more functional style.

### DIFF
--- a/ui/src/containers/SampleViewContainer.jsx
+++ b/ui/src/containers/SampleViewContainer.jsx
@@ -40,23 +40,21 @@ class SampleViewContainer extends Component {
   }
 
   getControlAvailability(name) {
-    const available =
-      this.props.uiproperties.sample_view_video_controls.components.find(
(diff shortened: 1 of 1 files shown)


ui/src/containers/SampleViewContainer.jsx
```diff
@@ -40,62 +40,41 @@ class SampleViewContainer extends Component {
...
-
-            break;
-          }
-          // No default
-        }
-      });
-    }
+    const points = Object.fromEntries(
+      Object.entries(shapes).filter(([_, shape]) => shape.t === 'P'),
+    );
```
axelboc, line 67: Once we start using `useSelector` in `SampleImage`, we'll abstract this into state selector functions, like 

```js
const selectPoints = makeShapesSelector('P');

...
const points = useSelector(selectPoints)
```

walesch-yan (approved):

PR #1904, opened 2025-09-26
Return new shape from `addShape` and `add2DPoint` thunks to avoid callbacks

axelboc:
By returning the newly created shape from the thunk, we can just await it as the result of the `dispatch`.


marcus-oscarsson:
Nice !, thanks !

PR #1902, opened 2025-09-26
Map state to props in `SampleImage`

axelboc:
I've made two separate commits:

1. one to remove most of the explicit props passed down from `SampleViewContainer`,
2. and another to remove the spread of the `sampleview` state.

`<SampleImage {...this.props.sampleViewState} />` was a huge pain in the butt. Removing it will help refactor `SampleImage` in the future.

PR #1905, opened 2025-10-01
Access Redux store from `PlateManipulator` to remove props

axelboc:
Follows up on #1901 and #1902 but with the `PlateManipulator` component.

To ease the review, I've made a separate commit to make store access more specific where possible in `PlateManipulator`.

### DIFF
--- a/ui/src/containers/SampleViewContainer.jsx
+++ b/ui/src/containers/SampleViewContainer.jsx
@@ -9,16 +9,6 @@ import {
   logFrontEndTraceBack,
   setAttribute,
 } from '../actions/beamline';
-import { showErrorPanel } from '../actions/general';
-import {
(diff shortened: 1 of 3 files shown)


ui/src/containers/SampleViewContainer.jsx
```diff
@@ -134,28 +124,7 @@ class SampleViewContainer extends Component {
                 />
               )}
               {this.props.sampleChangerContents.name === 'PlateManipulator' && (
-                <PlateManipulator
-                  contents={this.props.sampleChangerContents}
-                  loadedSample={this.props.loadedSample}
-                  select={this.props.select}
```
axelboc, line 140: This prop was unused.

walesch-yan (approved):

PR #1907, opened 2025-10-01
Fix crash when field with ui schema has no group

### LINKED ISSUE #1895: Ui crashes if no {group: } attribute is specified for a field that has other `"ui:"` configuration.
The UI crashes if no {group: } attribute is specified for a field that has other `"ui:"` configuration.

```
"field3": {
    "ui:readonly": "true",
    "ui:options": {"group": "SomeGroup"},             # Omitting this line would crash the UI
    },
```

Omitting the ` "ui:options": {"group": "SomeGroup"}` while also having i.e  `"ui:readonly": "true",` in the configuration results in the following error:

Image

It comes from this section in the code:
Image

mockoocy:
Closes #1895
Currently code assumes that if field has an entry in uiSchema JSON, then it has an entry for group - due to a missing nullish coalescence.

### DIFF
--- a/ui/src/components/Tasks/CustomObjectFieldTemplate.jsx
+++ b/ui/src/components/Tasks/CustomObjectFieldTemplate.jsx
@@ -11,7 +11,7 @@ export default function CustomObjectFieldTemplate({
 }) {
   const groups = groupBy(
     properties,
-    (field) => uiSchema[field.name]?.['ui:options'].group,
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
Well done !, thanks !

axelboc (approved):

walesch-yan (approved):

marcus-oscarsson:
I'm not going to wait for the docs this time, seems like there is some issue with the doc build

PR #1908, opened 2025-10-01
Deactivate Sonar and Upgrade Bandit workflow

Phantominumm:
This pull request updates the Bandit security scanning workflow to improve automation, reporting, and integration with GitHub code scanning. The main changes include replacing the Bandit scan step with a custom workflow that generates SARIF output, introducing a new script to convert Bandit results to SARIF format, and updating dependencies and configuration for better maintainability.

**SonarCloud deactivate for the moment:**

* Comments out the SonarCloud workflow steps, waiting for the return of the audit with CGI to see which code analysis we will use, SonarCloud (sonarQube) has limits with free plan and sometimes fails a PR because we can't skip the rule.

**Bandit workflow improvements:**

* Replaces the previous Bandit GitHub Action with a custom workflow that installs Bandit and dependencies, runs Bandit with dynamically generated skip codes from `ruff.toml`, and saves results as JSON.
* Uploads both Bandit JSON and SARIF results as workflow artifacts and to GitHub code scanning, improving visibility of security issues.
*  **JSON file with all the scan of bandit can be download from actions / Bandit and SonarQube then click on the commit you want to know.**
* **SARIF results (only the warnings) will be shown in the security / code scanning tab** 

**New script for Bandit-to-SARIF conversion:**

* Introduces `.github/scripts/bandit_to_sarif.py`, a Python script that reads Bandit JSON results and outputs SARIF-compliant files for code scanning tools.

**Configuration updates:**

* Updates `ruff.toml` to ignore `INP001` for Python scripts in `.github/scripts/`, preventing irrelevant linting errors for those files.

### DIFF
--- a/.github/workflows/bandit_sonar.yaml
+++ b/.github/workflows/bandit_sonar.yaml
@@ -18,61 +18,49 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Generate Bandit skips from Ruff
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
+
+      - name: Install dependencies
+        run: pip install bandit[sarif]
+
+      - name: Extract Bandit skips from ruff.toml
         id: bandit_skips
         run: |
-          pip install ruff jq
-          SKIPS=$(ruff rule --all --output-format json \
-            | jq -r '.[] | select(.code | test("^S[0-9]{3}$")) | .code' \
-            | sed 's/S/B/' \
-            | paste -sd, -)
+          SKIPS=$(python3 - <<'PY'
+          import re
+          import tomllib as toml
+          with open('ruff.toml', 'rb') as f:
+              config = toml.load(f)
+          all_codes = config.get('lint', {}).get('ignore', [])
+          for codes in config.get('lint', {}).get('per-file-ignores', {}).values():
+              all_codes.extend(codes)
+          bandit_codes = [c for c in set(all_codes) if re.match(r'^S\d{3}$', c)]
+          print(','.join(sorted(bandit_codes)))
+          PY
+          )
           echo "skips=$SKIPS" >> $GITHUB_OUTPUT
 
-      - name: Bandit Scan
-        uses: shundor/python-bandit-scan@ab1d87dfccc5a0ffab88be3aaac6ffe35c10d6cd
-        with:
-          exit_zero: true
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          excluded_paths: ./test/
-          level: MEDIUM
-          confidence: HIGH
-          skips: ${{ steps.bandit_skips.outputs.skips }}
+      - name: Run Bandit Scan
+        run: |
+          SKIP_PARAM="${{ steps.bandit_skips.outputs.skips }}"
+          bandit -r mxcubeweb/ -f sarif -o results.sarif --exclude ./test/ \
+            --severity-level medium --confidence-level medium \
+            ${SKIP_PARAM:+--skip $SKIP_PARAM}
+        continue-on-error: false
 
-      - name: Upload SARIF file to GitHub
+      - name: Upload Bandit results as artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: bandit-results
+          path: results.sarif
+        if: always()
+
+      - name: Upload SARIF to Code Scanning
+        if: always()
         uses: github/codeql-action/upload-sarif@v3
         with:
           sarif_file: results.sarif
-          category: bandit
-
-  sonarcloud:
-    name: SonarCloud Scan
-    runs-on: ubuntu-latest
-    if: >
-      github.event.pull_request.head.repo.full_name == github.repository ||
-      github.event_name != 'pull_request'
-    steps:
-      - uses: actions/checkout@v4
-      - name: Set up Java
-        uses: actions/setup-java@v3
-        with:
-          distribution: 'temurin'
-          java-version: '17'
-
-      - name: Cache SonarCloud packages
-        uses: actions/cache@v3
-        with:
-          path: ~/.sonar/cache
-          key: ${{ runner.os }}-sonar
-          restore-keys: ${{ runner.os }}-sonar
-
-      - name: SonarCloud Scan
-        uses: SonarSource/sonarcloud-github-action@v2
-        with:
-          args: >
-            -Dsonar.organization=mxcubeweb
-            -Dsonar.projectKey=mxcube_mxcubeweb
-            -Dsonar.coverage.exclusions=**
-            -Dsonar.cpd.exclusions=**
-            -Dsonar.cpd.skip=true
-        env:
-          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+        continue-on-error: true


Phantominumm:
forgot to say that the parameters are  **--severity-level medium**   and **--confidence-level medium**  which can be changed as needed

ruff.toml
```diff
@@ -391,6 +391,9 @@ extend-ignore-names = ["HWR"] # To avoid "N814": Camelcase `HardwareRepository`
     "E501",
     "UP032",
 ]
+".github/scripts/*.py" = [
+    "INP001"
```
fabcor-maxiv: Add a trailing comma:

```suggestion
    "INP001",
```

.github/workflows/bandit_sonar.yaml
```diff
@@ -18,61 +18,84 @@ jobs:
...
+#         uses: SonarSource/sonarcloud-github-action@v2
+#         with:
+#           args: >
+#             -Dsonar.organization=mxcubeweb
+#             -Dsonar.projectKey=mxcube_mxcubeweb
+#             -Dsonar.coverage.exclusions=**
+#             -Dsonar.cpd.exclusions=**
+#             -Dsonar.cpd.skip=true
+#         env:
+#           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
```
fabcor-maxiv: Why comment out? It is better to delete. You can fish it back from the git history if/when it is needed again.

fabcor-maxiv:
Do I understand it correctly that we can not make use of <https://bandit.readthedocs.io/en/latest/formatters/sarif.html>?

And what about this <https://github.com/marketplace/actions/bandit-by-pycqa>, can't we use it? It seems to be the "official" Bandit GitHub Action. I recall that at some point I argued against using a GitHub action, but if I remember correctly it was not the "official" one, while this one seems to be "official".

Phantominumm:
> Do I understand it correctly that we can not make use of https://bandit.readthedocs.io/en/latest/formatters/sarif.html?
> 
> And what about this https://github.com/marketplace/actions/bandit-by-pycqa, can't we use it? It seems to be the "official" Bandit GitHub Action. I recall that at some point I argued against using a GitHub action, but if I remember correctly it was not the "official" one, while this one seems to be "official".

- The command used to scan ( bandit -r mxcubeweb/ ... ) can't create sarif format so I choose json and upload the json file as it gives the same informations than a sarif.
Code scanning in security ( github repo) needed a sarif file to work so afterwards I did the conversion.

- I will take a look for the GitHub Action but the problem was the limitations for the scan, it's easier and we have more liberty for parameters without the Action but I will see if it's a good solution now.

.github/workflows/bandit_sonar.yaml
```diff
@@ -18,61 +18,51 @@ jobs:
...
       - uses: actions/checkout@v4
 
-      - name: Generate Bandit skips from Ruff
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
+
+      - name: Install dependencies
+        run: pip install bandit[toml] tomli
```
fabcor-maxiv: Do we need `tomli`? Looks like this workflow is running with Python 3.11 which has [`tomllib`](https://docs.python.org/3/library/tomllib.html) already. Isn't `tomllib` good enough for what we need to do here?

```suggestion
        run: pip install bandit[toml]
```

fabcor-maxiv:
> The command used to scan ( bandit -r mxcubeweb/ ... ) can't create sarif format

How so? I seem to be able to run `bandit --format=sarif --recursive mxcubeweb` just fine. Now I do not know how Sarif is supposed to look like, but it delivers something that does look like the [documentation](https://bandit.readthedocs.io/en/latest/formatters/sarif.html).

Phantominumm:
> > The command used to scan ( bandit -r mxcubeweb/ ... ) can't create sarif format
> 
> How so? I seem to be able to run `bandit --format=sarif --recursive mxcubeweb` just fine. Now I do not know how Sarif is supposed to look like, but it delivers something that does look like the [documentation](https://bandit.readthedocs.io/en/latest/formatters/sarif.html).

bandit: error: argument -f/--format: invalid choice: 'sarif' (choose from 'csv', 'custom', 'html', 'json', 'screen', 'txt', 'xml', 'yaml') 
I've got this error the first time I tried with sarif, I changed it for this reason

fabcor-maxiv:
> > > The command used to scan ( bandit -r mxcubeweb/ ... ) can't create sarif format
> > 
> > 
> > How so? I seem to be able to run `bandit --format=sarif --recursive mxcubeweb` just fine. Now I do not know how Sarif is supposed to look like, but it delivers something that does look like the [documentation](https://bandit.readthedocs.io/en/latest/formatters/sarif.html).
> 
> bandit: error: argument -f/--format: invalid choice: 'sarif' (choose from 'csv', 'custom', 'html', 'json', 'screen', 'txt', 'xml', 'yaml') I've got this error the first time I tried with sarif, I changed it for this reason

Ah, OK, then it is probably because the workflow needs to install `bandit[sarif]` (in other words, it needs the `sarif` extra optional group of dependencies): <https://bandit.readthedocs.io/en/latest/start.html#installation>.

Phantominumm:
> > > > The command used to scan ( bandit -r mxcubeweb/ ... ) can't create sarif format
> > > 
> > > 
> > > How so? I seem to be able to run `bandit --format=sarif --recursive mxcubeweb` just fine. Now I do not know how Sarif is supposed to look like, but it delivers something that does look like the [documentation](https://bandit.readthedocs.io/en/latest/formatters/sarif.html).
> > 
> > 
> > bandit: error: argument -f/--format: invalid choice: 'sarif' (choose from 'csv', 'custom', 'html', 'json', 'screen', 'txt', 'xml', 'yaml') I've got this error the first time I tried with sarif, I changed it for this reason
> 
> Ah, OK, then it is probably because the workflow needs to install `bandit[sarif]` (in other words, it needs the `sarif` extra optional group of dependencies): https://bandit.readthedocs.io/en/latest/start.html#installation.

Thanks it's better this way, I think it's fine now, will take a look about the github actions anyway

ruff.toml
```diff
@@ -392,6 +392,5 @@ extend-ignore-names = ["HWR"] # To avoid "N814": Camelcase `HardwareRepository`
     "UP032",
 ]
 
-
```
fabcor-maxiv, line 395: We do not need this change.

fabcor-maxiv:
@Phantominumm, can you squash all the commits into a single commit? Otherwise, I guess it is fine.

fabcor-maxiv:
The unnecessary change in `ruff.toml` is back :/

fabcor-maxiv (approved):
OK, I guess it is fine like this.

~I hope you get time to investigate if we can use <https://github.com/marketplace/actions/bandit-by-pycqa> to make things simpler. Or maybe you already did investigate, I have not fully understood if you have already. It seems like it has support for "skips", Sarif format, confidence level, and severity level.~ Actually, maybe it does not matter, the action does not seem to offer much that we can not do with a simple command line, I am not sure.

Phantominumm:
> OK, I guess it is fine like this.
> 
> ~I hope you get time to investigate if we can use https://github.com/marketplace/actions/bandit-by-pycqa to make things simpler. Or maybe you already did investigate, I have not fully understood if you have already. It seems like it has support for "skips", Sarif format, confidence level, and severity level.~ Actually, maybe it does not matter, the action does not seem to offer much that we can not do with a simple command line, I am not sure.

Yes, we talk about that with marcus and for the moment we think this way is fine and the github actions is not necessary

fabcor-maxiv:
> Yes, we talk about that with marcus and for the moment we think this way is fine and the github actions is not necessary

OK, fair enough :)

Thanks!

PR #1909, opened 2025-10-01
Add CheckboxWidget

### LINKED ISSUE #1896: Render checkboxes on one line (without label duplication)
As seen on the screen shot below the check boxes are rendered with a "Title" and with a "Label", this looks a bit odd. It would be nicer to render the checkbox on one line with the the label (in place of the title)

Image

mockoocy:
Closes #1896
The custom checkbox widget is used to provide the description tooltip for the description fields.
By default RJSF renders the checkbox with the description and label provided next to the checkbox, as text paragraph.
To make it more consistent, and reduce visual clutter, the description will be rendered in a tooltip, as in other fields.

There is also a small CSS change, which makes the label title vertically aligned with the description tooltip icon - it's also rendered directly after the label.

Result (for a `bool`-valued field) 
image

### DIFF
--- a/ui/src/components/Tasks/CustomFieldTemplate.jsx
+++ b/ui/src/components/Tasks/CustomFieldTemplate.jsx
@@ -1,7 +1,5 @@
-import { BsInfoCircleFill } from 'react-icons/bs';
-
-import TooltipTrigger from '../TooltipTrigger';
 import styles from './CustomFieldTemplate.module.css';
+import { FieldDescriptionTooltip } from './Widgets/FieldDescriptionTooltip/FieldDescriptionTooltip';
 
 /**
  * CustomFieldTemplate is a custom field template for React JSON Schema Form (RJSF).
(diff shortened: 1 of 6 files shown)


ui/src/components/Tasks/Widgets/FieldDescriptionTooltip/FieldDescriptionTooltip.jsx
```diff
@@ -0,0 +1,29 @@
...
+ * @typedef
+ * {Object} FieldDescriptionTooltipProps
+ * @property {string} description - The description text to display in the tooltip.
+ */
+
+/**
+ * FieldDescriptionTooltip is a component that displays a tooltip
+ * with a description when hovering over an info icon.
+ * @param {FieldDescriptionTooltipProps} props
+ * @return {JSX.Element} The rendered tooltip component.
```
axelboc: No need to type the return value for React components; it's better/easier to let the type system infer it. See the React documentation, which never uses `JSX.Element`: https://react.dev/learn/typescript#useful-types

ui/src/components/Tasks/CustomFieldTemplate.jsx
```diff
@@ -30,6 +28,19 @@ export default function CustomFieldTemplate(props) {
   const span = Number(uiSchema['ui:options']?.col) || 6;
   const gridClass = `col-${span}`;
 
+  if (schema.type === 'boolean') {
+    return (
+      <div className={`${gridClass} ${classNames}`.trim()}>
```
axelboc: Maybe move this class name logic to the `gridClass` variable (and rename it if needed)?

axelboc (approved):
Awesome!

marcus-oscarsson:
Well done !

Issue #1894, opened 2025-09-26
Generic Task Form - Layout


marcus-oscarsson:
The PR: #1678, by @mockoocy, introduced a really nice concept that made it possible to group the various inputs into sections. We had the occasion to try to put this to use the other day and we noticed a few things that would be nice to take a closer look at.

Ill try to make use of sub-issues for this issue, its a larger topic that in the end consists of a few smaller issues.

**Documentation**
First of all the documentation was actually quite hard to find, I knew that it was somewhere and we finally found it in the [queue documentation](https://mxcubecore.readthedocs.io/en/stable/dev/queue.html), which also makes sense. But there is also this [json -schema UI documentation](https://mxcubecore.readthedocs.io/en/stable/dev/json-schema-generated-user-interface.html). So one is easily a bit lost, it would be nice with a link or explaining section in the later.

**Main Issues**
 - The first issue we had is a bug that crashes the front end if no `{group: }` attribute is specified for a field that has other `"ui:` configuration. (#1895)
 
**Improvement on layout**
 - We think that the layout and UX can be improved by adding the unit right after the label instead of in the information circle (#1896)
 - Render checkboxes one line (without duplicating its label) (or all components with their labels on one line) (#1897)
 - Default for show-more (#1898)
 - [Readonly fields are not displayed the same #1899](https://github.com/mxcube/mxcubeweb/issues/1899)

marcus-oscarsson:
Done, well done ;)

PR #1910, opened 2025-10-01
Add NumericWithUnit and SmartText Widgets

### LINKED ISSUE #1897: Rendring of unit (short-description)
It would be nice if the unit (or short description) could be rendered after the label within brackets. Alternatively have separately handing of units and short description. Units being displayed in brackets and the description in the little i circle.

Image

mockoocy:
Closes #1897
This provides a way to display unit for numeric fields. By default RJSF renders the numeric fields with a built-in `TextWidget`. Thus it would be required for user to always specify in the `uiSchema` widget to be new `NumericWithUnitWidget`.
That wouldn't be convenient, so the `TextWidget` is being overriden with `SmartTextWidget`, which just renders `NumericWithUnitWidget` if the field is numeric.

This is how it looks
image

It requires setting a `unit` field in the pydantic model, like so:
```python
class TestUserCollectionParameters(BaseModel):
    num_images: int = Field(0, description="")
    exp_time: float = Field(100e-6, gt=0, lt=1, unit="s", description="s")
```

Once it were to be merged, I'd add documentation section for it, along making documentation changes described in the #1894.

Note: feel free to propose alternative names for these widgets, the current names are not really that good... :D

### DIFF
--- a/ui/src/components/Tasks/CustomFieldTemplate.jsx
+++ b/ui/src/components/Tasks/CustomFieldTemplate.jsx
@@ -42,12 +42,15 @@ export default function CustomFieldTemplate(props) {
     );
   }
 
+  const unit = schema.unit ? ` [${schema.unit}]` : '';
+
   return (
     <div className={fieldClassNames}>
       <div className={styles.fieldTitle}>
         {label && (
           <label htmlFor={id} className={styles.fieldLabel}>
             {label}
+            <span className={styles.unitLabel}>{unit}</span>
             {required && <span className="text-danger">*</span>}
           </label>
         )}
--- a/ui/src/components/Tasks/CustomFieldTemplate.module.css
+++ b/ui/src/components/Tasks/CustomFieldTemplate.module.css
@@ -38,3 +38,10 @@
 .fieldHelp p {
   margin-bottom: 0;
 }
+
+.unitLabel {
+  /* font-weight is inherited from the label by default */
+  font-weight: normal;
+  font-family: monospace;
+  color: var(--bs-secondary);
+}


mockoocy:
Note: It is a different approach from the one described in the linked issue. It comes from some "experimental" branch of our fork, as we were also trying to make some more complex forms with it :)
This one seems nicer anyway ;)

ui/src/components/Tasks/Widgets/NumericWithUnitWidget/NumericWithUnitWidget.jsx
```diff
@@ -0,0 +1,42 @@
...
+  schema = {},
+}) {
+  const unit = schema.unit ?? '';
+
+  return (
+    <div className={styles.wrapper}>
+      <Form.Control
+        id={id}
+        type="number"
+        className={styles.input} /* ← gives padding-right */
```
axelboc:
```suggestion
        className={styles.input} // gives padding-right
```

ui/src/components/Tasks/Widgets/NumericWithUnitWidget/NumericWithUnitWidget.module.css
```diff
@@ -0,0 +1,20 @@
...
+.unit {
+  position: absolute;
+  top: 50%;
+  transform: translateY(-50%);
+  /* sit right before the spin buttons */
+  right: 2.25rem;
+  white-space: nowrap;
+  font-size: 0.875rem;
+  color: var(--bs-secondary);
+  pointer-events: none; /* don’t block clicks on the input */
```
axelboc: Best to avoid special characters :wink: 

```suggestion
  pointer-events: none; /* don't block clicks on the input */
```

ui/src/components/Tasks/Widgets/NumericWithUnitWidget/NumericWithUnitWidget.jsx
```diff
@@ -0,0 +1,42 @@
+import { Form } from 'react-bootstrap';
+
+import styles from './NumericWithUnitWidget.module.css';
+/**
+ *
+ * @param {import('@rjsf/utils').WidgetProps} props
+ * @returns
+ */
+export default function NumericWithUnitWidget({
```
axelboc: This is used for all numeric inputs, not just those with units, so I would rename it `NumericWidget`

ui/src/components/Tasks/GenericTaskForm.jsx
```diff
@@ -313,6 +314,9 @@ class GenericTaskForm extends React.Component {
                 FieldTemplate: CustomFieldTemplate,
                 FieldErrorTemplate: CustomFieldErrorTemplate,
               }}
+              widgets={{
+                TextWidget: SmartTextWidget,
```
axelboc: Maybe `TextOrNumericWidget`?

axelboc (approved):
Looks great!

ui/src/components/Tasks/GenericTaskForm.jsx
marcus-oscarsson: yes, that reads nicer

marcus-oscarsson:
Well done @mockoocy and big thumps up for the reactivity super fast !

marcus-oscarsson:
I have one minor comment, I also think it already looks great but I think the unit is easy to miss, would it help to make the text a bit bigger or bold (or maybe both) ?

marcus-oscarsson:
I like the use of "unit" in the pydantic model very good !

ui/src/components/Tasks/Widgets/NumericWithUnitWidget/NumericWithUnitWidget.module.css
mockoocy: Sorry for these. I did not take a good enough look at these

mockoocy:
> I have one minor comment, I also think it already looks great but I think the unit is easy to miss, would it help to make the text a bit bigger or bold (or maybe both) ?

I made the text a bit bigger (uses the same font size as the input text). 
I see that there is a problem for wide inputs, where the unit is far away from the text. For now I've decided to use some lighter colour to make it standout more. 
image

I also see two ways it could be improved
1. there could be a different background around the unit
2. the unit could be displayed right after the input text

But for now this looks alright IMO

marcus-oscarsson:
Great for the font size, I would probably no use cyan for the color. I think black or very dark gray is fine. You could maybe try to add it after the label with parentheses or brackets around it. i.e **Exp. Time _[s]_** or **Exp. Time [s]**

walesch-yan:
I agree with Marcus on the unit being easy to miss and I also do not like the cyan color, I used a plugin for colorblindness and especially when having a weakness to the color green, the cyan color becomes a very light gray, difficult to see in that case in my opinion. Otherwise, great work on the PR!

mockoocy:
Okay, I went with the unit in the square brackets.
image
So there will be less code added, so that's also nice I suppose :D

marcus-oscarsson:
Excellent well done !

marcus-oscarsson:
Thanks !

PR #1912, opened 2025-10-02
Improve mesh grid outlines visibily

matyson:
Currently, when drawing a new mesh grid it's hard to see it due to it's only outlined with a thin border. So, i am adding an intermediate background color to the currently being drawn grid. The proposed change:

image

Also, i took the liberty to add some utility functions to handle the hardcoded colors (but i didn't refactor all of them):

```js
/**
 * @typedef {Object} Color
 * @property {number} r - red channel [0-255]
 * @property {number} g - green channel [0-255]
 * @property {number} b - blue channel [0-255]
 */

/**
 * Return an RGBA string from Color object
 *
 * @param {Color} color
 * @param {number} alpha - alpha channel [0-1]
 * @return {string} RGBA string
 */
function toRGBA(color, alpha) {
  return `rgba(${color.r}, ${color.g}, ${color.b}, ${alpha})`;
}
```

But i can remove them or refactor to a better suggestion. Also, i was wondering if there was an opinionated choice of colors here, i tried to stick within the yellow-green spectra. 

And one more thing: i didn't fully understood [this](https://github.com/mxcube/mxcubeweb/blob/develop/ui/src/components/SampleView/DrawGridPlugin.jsx#L523) line, so its probably misinterpreted in my changes as well.

### DIFF
--- a/ui/src/components/SampleView/DrawGridPlugin.jsx
+++ b/ui/src/components/SampleView/DrawGridPlugin.jsx
@@ -147,6 +147,24 @@ function _GridData() {
   };
 }
 
+/**
+ * @typedef {Object} Color
+ * @property {number} r - red channel [0-255]
+ * @property {number} g - green channel [0-255]
+ * @property {number} b - blue channel [0-255]
+ */
+
+/**
+ * Return an RGBA string from Color object
+ *
+ * @param {Color} color
+ * @param {number} alpha - alpha channel [0-1]
+ * @return {string} RGBA string
+ */
+function toRGBA(color, alpha) {
+  return `rgba(${color.r}, ${color.g}, ${color.b}, ${alpha})`;
+}
+
 export default class DrawGridPlugin {
   constructor() {
     this.startDrawing = this.startDrawing.bind(this);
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
How does it look like when the grid is selected ?

Another thing we noticed is that the mesh does not behave as points when it comes to right click. When we right click on a "unsaved" point we can do data collections directly but not from a mesh. Is it something you where planing to look at ?

matyson:
> How does it look like when the grid is selected ?

Selected grids should stay the same, these changes only affect the currently being drawn mesh.

image

image

> Another thing we noticed is that the mesh does not behave as points when it comes to right click. When we right click on a "unsaved" point we can do data collections directly but not from a mesh. Is it something you where planing to look at ?

RN, we're only working with these "visibility" changes and we aim to add some new features to the grids in the future, but this right click behaviour is not a priority at the moment. I will include it in our list of tasks, though, as it is a good consistency addition.

matyson:
So, i also wanted to addInnerLines to the grid currently being drawn with this latest change, but it felt like spaghetti conditioning:

image

matyson:
I'm closing this, will wait for the upcoming changes in the sample view. Also, my ideas diverged a lot since then, might come up with something better designed later.

Issue #1906, opened 2025-10-01
Review remaining global styles


axelboc:
I had a quick look and there doesn't seem to be much global CSS left in the project (Bootstrap and FontAwesome excluded), which is awesome! 🎉  So I think it's time to do a bit of an audit to identify remaining work.

---

- [ ] `ssxchipcontrol.css` imported from `SSXChipControl.jsx` and `SSXChip.jsx`

  Ongoing work on SSX.

- [x] `rachat.css` imported from `Main.jsx`, relating to the chat widget:

  ```js
  import 'react-chat-widget/lib/styles.css';
  import './rachat.css';
  ```

  Work is planned to replace this unmaintained library.

- [ ] `components/Tasks/style.css` imported from `fields.jsx` and `GenericTaskForm.jsx`

  This relates to the generic forms and `json-schema-form`.

  - [ ] Double check that all the declarations in the file are used
  - [ ] Investigate a way to modularise them, or if not possible, at least rename the CSS file and import it once from `index.jsx`

- [x] `main.css` imported from `index.jsx`

  This file is meant to stay but:

  - [x] Double check that all the declarations in the file are used and that none of them can be modularised #1914 
  - Remaining styles: are required; override Bootstrap colors; or are too difficult to remove and can be opted out of with `data-default-styles` attribute.

- [x] Two global CSS imports of `slick-carousel` in `SampleGridTableContainer.jsx`:

  ```js
  import 'slick-carousel/slick/slick.css';
  import 'slick-carousel/slick/slick-theme.css';
  ```

  This is required for the `Slider` component of the `react-slick` package.

  - ~Check that `react-slick` is still maintained and an essential dependency~
  - ~Move global CSS imports to `main.jsx`~
  - [x] Remove `react-slick` completely

- [ ] `react-contexify/ReactContexify.css` in `index.jsx`

  There's an issue to replace this dependency with Bootstrap: #1844

- [ ] `'@fortawesome/fontawesome-free/css/all.min.css'` imported from `index.jsx`

  - [ ] Consider moving before `main.css` import.

walesch-yan:
Thanks for the nice overview @axelboc 👍 

[...]

PR #1914, opened 2025-10-02
Clean up `main.css`

axelboc:
Part of #1906 

Most of the remaining "questionable" styles either override Bootstrap colours (which is fine), or are more complicated to get rid of as they affect Bootstrap components.

A while back I introduced a `data-default-styles` attribute to opt out of some of the most problematic styles when needed (`.modal-dialog`, `.btn` and `.dropdown-item`). I think it's a decent solution for the time being.

### DIFF
--- a/ui/src/main.css
+++ b/ui/src/main.css
@@ -22,20 +22,6 @@ body {
   display: flex;
 }
 
-.popover {
-  width: auto !important;
-  z-index: 1040;
-}
-
-#test {
-  min-width: fit-content;
-}
-
-.device-status-container > div {
-  margin-right: 1em;
-  width: 12%;
-}
-
 button[disabled] {
   opacity: 0.2;
   pointer-events: none;
@@ -55,8 +41,7 @@ input[type='email']:focus,
 input[type='url']:focus,
 input[type='search']:focus,
 input[type='tel']:focus,
-input[type='color']:focus,
-.uneditable-input:focus {
+input[type='color']:focus {
   border-color: rgba(126, 161, 185, 0.8);
   box-shadow:
     0 1px 1px rgba(0, 0, 0, 0.075) inset,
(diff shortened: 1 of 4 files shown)


ui/src/main.css
```diff
@@ -22,20 +22,6 @@ body {
...
 }
 
-.popover {
-  width: auto !important;
-  z-index: 1040;
-}
-
-#test {
-  min-width: fit-content;
-}
```
axelboc, line 32: Moved to `ssxchipcontrol.css`, which is where it's used, for the time being.

ui/src/main.css
```diff
@@ -22,20 +22,6 @@ body {
   display: flex;
 }
 
-.popover {
-  width: auto !important;
-  z-index: 1040;
-}
```
axelboc, line 28: Merged with duplicate declaration below.

ui/src/main.css
```diff
@@ -22,20 +22,6 @@ body {
...
-
-#test {
-  min-width: fit-content;
-}
-
-.device-status-container > div {
-  margin-right: 1em;
-  width: 12%;
-}
```
axelboc, line 37: Unused

ui/src/main.css
```diff
@@ -55,8 +41,7 @@ input[type='email']:focus,
 input[type='url']:focus,
 input[type='search']:focus,
 input[type='tel']:focus,
-input[type='color']:focus,
-.uneditable-input:focus {
```
axelboc, line 59: Unused I think

ui/src/main.css
```diff
@@ -126,10 +113,6 @@ input[type='color']:focus,
   color: #000000;
 }
 
-.nowrap-style {
-  white-space: nowrap;
-}
```
axelboc, line 131: Modularized in `SampleListViewContainer.module.css`

marcus-oscarsson:
Thanks @axelboc !

PR #1915, opened 2025-10-02
Moved sample-changer signal handling to sample-changer component

marcus-oscarsson:
Simply moving all signal handlers for the sample-changer defined in `signals.py`  to the sample-changer component. This will make the migration to an adapter easier.

### DIFF
--- a/mxcubeweb/routes/signals.py
+++ b/mxcubeweb/routes/signals.py
@@ -1,10 +1,8 @@
 import contextlib
-import json
 import logging
 
 from mxcubecore import HardwareRepository as HWR
 from mxcubecore import queue_entry as qe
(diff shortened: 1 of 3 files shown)


mxcubeweb/routes/signals.py
```diff
@@ -242,16 +131,7 @@ def queue_execution_entry_finished(entry, message):
     if not mxcube.queue.is_interleaved(entry.get_data_model()):
         server.emit("task", get_task_state(entry), namespace="/hwr")
 
-    queue_toggle_sample(entry)
-
-
-def queue_toggle_sample(entry):
```
marcus-oscarsson, line 248: I had to move one of the queue signal handlers to the queue component as its used by both the queue and sample-changer components

PR #1917, opened 2025-10-03
Rename / bug fix for get_full_path function

rhfogh:
Clean-up / rename, changing get_full_path to get_full_paths. The function returns two paths, so the name ought to be plural.
Previous code had an if statement that was always true

Coupled with parallel PR in mxcubecore

The autotest fails, but it does not look like the failure has anything to do with the changes in the PR.

### DIFF
--- a/mxcubeweb/core/components/queue.py
+++ b/mxcubeweb/core/components/queue.py
@@ -1029,7 +1029,7 @@ class Queue(ComponentBase):
                 params.get("subdir", "").strip("/").split("/")[0:-1]
             )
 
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
Hey @rhfogh thanks for doing this PR. I'm sorry I missed it and I made the same. I think your PR failed because it was done before the actual change was merged on `mxcubecore` right ?

rhfogh:
@marcus-oscarsson Could be, I cannot remember, but it would explain it.

rhfogh:
Superseded by PR #1918

PR #1920, opened 2025-10-07
remove hardcoded braggy-link

walesch-yan:
A previous PR #1846 was made to replace the hard coded braggy link, with an entry in the config files. While the Entry was still there, it seems that #1849 ignored this change by mistake. Meaning the config entry etc still existed, but the code was still refering to the hardcoded link instead of using the one from the configuration.

This PR should fix this behaviour


marcus-oscarsson:
Thanks @walesch-yan

PR #1923, opened 2025-10-10
Change react-chat-widget by a version compatible with react 18

Phantominumm:
This pull request replaces the `react-chat-widget` dependency with a custom in-house chat widget implementation. The new widget is implemented in `LegacyChatWidgetAdapter.jsx` and styled with `LegacyChatWidget.css`, providing similar functionality while removing reliance on the external library. 

**Chat Widget Refactor and Replacement**

* Removed the `react-chat-widget` package from `package.json`, eliminating the external dependency.
* Implemented a custom chat widget in `LegacyChatWidgetAdapter.jsx`, replicating the previous widget's features (message handling, badge count, UI, etc.).
* Added new styles for the custom widget in `LegacyChatWidget.css` to mimic the old chat and updated imports in `Main.jsx`.


ui/src/components/LegacyChatWidget.css
```diff
@@ -0,0 +1,124 @@
+.rcw-widget-container {
```
marcus-oscarsson: It would be neat if you could convert this css to `moudle.css` like this https://create-react-app.dev/docs/adding-a-css-modules-stylesheet/

ui/src/components/ChatComponent/ChatComponent.jsx
```diff
@@ -14,6 +8,12 @@ import {
...
 import { fetchChatMessages, sendChatMessage } from '../api/remoteAccess';
 import { store } from '../store';
+import {
+  addResponseMessage,
+  addUserMessage,
+  setBadgeCount,
+  Widget,
+} from './LegacyChatWidgetAdapter';
 
 function ChatWidget() {
```
marcus-oscarsson, line 30: We could probably move this logic to the `LegacyChatWidget` you created and then remove this `ChatWidget` and simply rename `LegacyChatWidget` to `ChatComponent` or `Chat` ....

ui/src/components/LegacyChatWidgetAdapter.jsx
```diff
@@ -0,0 +1,267 @@
...
+    return `${d.getHours().toString().padStart(2, '0')}:${d
+      .getMinutes()
+      .toString()
+      .padStart(2, '0')}`;
+  } catch {
+    return '';
+  }
+}
+
+export function Widget({
```
marcus-oscarsson: So you could either add the logic from `ChatWidget` in `Widget` or create a new function i.e `DragableReduxChat` or similair.

ui/src/components/LegacyChatWidgetAdapter.jsx
```diff
@@ -0,0 +1,267 @@
...
+              backgroundColor: 'red',
+              color: 'white',
+              width: 24,
+              height: 24,
+              borderRadius: '50%',
+              display: 'flex',
+              alignItems: 'center',
+              justifyContent: 'center',
+              fontSize: 12,
+              fontWeight: 700,
```
walesch-yan: You can move these styles into `rcw-bdage` in your style sheet, since this is the only component that uses the class

ui/src/components/LegacyChatWidgetAdapter.jsx
```diff
@@ -0,0 +1,267 @@
+import { useEffect, useRef, useState } from 'react';
+
+import { resetChatMessageCount } from '../actions/remoteAccess';
+import { store } from '../store';
+
+const _messages = [];
+let _badgeCount = 0;
+let _isOpen = false;
+const _subs = new Set();
```
axelboc: This pub/sub pattern is not very Reacty :grin: — I also don't like all those local state variables being outside of the component.

axelboc (commented):
I see, so you basically replace `react-chat-widget` "in place". I like the approach! The implementation needs some work, though. I feel like you're fighting against React :sweat_smile:

Why do you need so much local state inside the widget? Why do you need the pub/sub pattern (and with multiple subscriptions)? The messages are already in the global Redux store, take advantage of it! Dispatch actions from your event handlers and avoid `useEffect` at all cost: https://react.dev/learn/you-might-not-need-an-effect

I recommend converting this PR to draft for now :wink: — good luck :muscle: and don't hesitate to come see me! :open_hands:

marcus-oscarsson:
Thanks alot @axelboc. @Phantominumm will be back on Wednesday so we will have another look then. Ill WIP this at the moment because its true that @Phantominumm still working on it.

ui/src/components/LegacyChatWidgetAdapter.jsx
```diff
@@ -0,0 +1,267 @@
...
+  subtitle = '',
+  handleNewUserMessage,
+  _launcherLabel = '',
+  _autofocus = false,
+}) {
+  const [state, setState] = useState({
+    messages: [..._messages],
+    badge: _badgeCount,
+    open: false,
+  });
```
mockoocy: These should be three different pieces of state I believe.

ui/src/components/LegacyChatWidgetAdapter.jsx
```diff
@@ -0,0 +1,267 @@
...
+    const v =
+      inputRef.current && inputRef.current.value
+        ? inputRef.current.value.trim()
+        : '';
+    if (!v) {
+      return;
+    }
+    if (typeof handleNewUserMessage === 'function') {
+      handleNewUserMessage(v);
+    }
```
mockoocy: Maybe I don't get something, but [this](https://github.com/mxcube/mxcubeweb/blob/eaeaadb2ee572ba6a7211c3fdeb3ed44fe843e2e/ui/src/components/ChatWidget.jsx) piece of code suggests that it'd be always a function?

ui/src/components/ChatComponent/ChatComponent.module.css
mockoocy, line 1: Are these styles imported from the `react-chat-widget` library? If so, I'd put information about it in some comment on the top of the file. It makes it a bit easier to distinguish if these are deliberate choices, or kept for backwards compatibility :)

mockoocy (changes requested):

ui/src/components/LegacyChatWidgetAdapter.jsx
Phantominumm: It was due to a Lint error that was a temp solution to push something without errors but yes this part was useless

Phantominumm:
Some changes but there is more to do

ui/src/components/ChatComponent/ChatComponent.module.css
Phantominumm: Yes that was inspired from it, I adapt it for my needs, the goal is to have something close to the old chat

ui/src/components/ChatWidget.jsx
```diff
@@ -24,49 +37,197 @@ function ChatWidget() {
...
+    const newMessage = {
+      id: `u-${Date.now()}-${Math.random()}`,
+      type: 'user',
+      text,
+      date: new Date().toISOString(),
+    };
+    setMessages((prev) => [...prev, newMessage]);
+  }, []);
+
   useEffect(() => {
```
Phantominumm: I didn't find a way to not use Effect here (or maybe really big changes in the logic) but I'm all ears for a solution.

axelboc: Chat messages are a server-side state, so to me they need to be fetched with other server states when logging in — cf. `getInitialState()` in `login.js` — and kept in the Redux store. This means adding a store slice with its own reducer, actions (`ADD_MESSAGE`, `MARK_ALL_READ`) and action creators/thunks (`addMessage`, `sendMessage`, `markAllAsRead`).

This `useEffect` should only dispatch the `markAllAsRead` action on mount. Most of the logic (notably the creation of message objects with `id`, `date`, etc.) should live in the action creators/thunks.

By having the messages in the Redux store, you can also get rid of the `externalAddResponseMessage` "hack": `serverIO` can simply dispatch the `addMessage` action when a new message is received.

Does that make sense?

axelboc: In my initial review, I thought that the messages were already stored in Redux but I now realise this is not the case, sorry for the confusion. I think it's the way to go though — it will simplify quite a few things as explained, and it will be more consistent with the rest of the application.

Phantominumm: Yes, I think I got most of it, I will work on, it will be clear and if didn't I will ask for help

Phantominumm:
So I think I’m getting close to a good solution, there are probably still a few mistakes and some cleanup to do.

axelboc:
Yeah that looks a lot better already, nicely done!

I think now it's about finding a good place to put the message processing logic so it's not duplicated. I have a feeling that you'll need to put it in a reusable utility function and call that utility from 1) the `fetchChatMessages` function in the `api/remoteAccess.js` module, and 2) in the `ra_chat_message` handler in `socketIO.js`.

It'd be good to find a way also, when logging in, to fetch the chat messages in parallel with everything else (as part of the `Promise.all` promise).

Finally, but perhaps for another PR, I think that `chatMessageCount` is actually a computed state and that it can be recomputed during render inside the components that need it (instead of being kept in sync in the Redux store). The logic can be abstracted in a reusable selector function (i.e. to be passed to `useSelector`) if needed.

marcus-oscarsson:
Wonderful @Phantominumm, really well done for a first react component :). There is just one conflict to be solved in `login.js` then we can go head and merge this. We will probably make some minor adjustments later but this is a great start !

ui/src/actions/login.js
```diff
@@ -158,6 +162,20 @@ export function getInitialState() {
...
+    );
+    if (chatMessagesData?.chatMessages) {
+      const { user } = getState().login;
+      dispatch(
+        processFetchedChatMessages(
+          chatMessagesData.chatMessages.messages,
+          user.username,
+        ),
+      );
+    }
```
axelboc, line 177: There's already a `setInitialState` action being dispatched with all the data, so it would be simpler I think to process the initial chat messages with `processChatMessageRecord` in the `SET_INITIAL_STATE` case block of the `remoteAccess` reducer.

ui/src/actions/remoteAccess.js
```diff
@@ -108,13 +110,50 @@ export function updateTimeoutGivesControl(timeoutGivesControl) {
...
+      date: new Date().toISOString(),
+    };
+
+    dispatch(addChatMessage(newMessage));
   };
 }
 
-export function incChatMessageCount(count = 1) {
-  return { type: 'INC_CHAT_MESSAGE_COUNT', count };
+export function markAllAsRead() {
```
axelboc, line 149: Arguably, this could dispatch an empty action called `MARK_ALL_READ`, and the state update logic could be moved to the `remoteAccess` reducer. This, together with my other comment about `SET_INITIAL_STATE` would maybe remove the need for a `SET_CHAT_MESSAGES` action.

ui/src/serverIO.js
```diff
@@ -122,11 +122,12 @@ class ServerIO {
 
     this.hwrSocket.on('ra_chat_message', (record) => {
       const { username } = store.getState().login.user;
-      if (record.username !== username && !record.read) {
-        addResponseMessage(
-          `${record.date} **${record.nickname}:** \n\n ${record.message}`,
+      if (record.username !== username) {
+        const message = processChatMessageRecord(
+          { ...record, read: false },
```
axelboc, line 127: Doesn't `processChatMessageRecord` already take care of adding the `read` flag? If I'm not mistaken, the `read` flag is entirely a front-end state, right? So the `record` received here should never have it.

ui/src/components/ChatComponent/selector.js
axelboc, line 1: `selectors.js`?

ui/src/components/ChatComponent/chatMessages.js
```diff
@@ -0,0 +1,29 @@
...
+    } catch {
+      // Ignore
+    }
+  }
+  return normalizedDate;
+}
+
+export function processChatMessageRecord(record, currentUsername) {
+  const isSelf = record.username === currentUsername;
+  const normalizedDate = normalizeDateToISO(record.date);
```
axelboc, line 18: I don't think you need that `normalizeDateToISO` function. I would also throw an error if the date is invalid (instead of ignoring the problem), as it could help uncover a bug.

```js
const date = record.date ? new Date(record.date) : new Date();
if (Number.isNaN(date.getTime()) {
  throw new Error('Invalid date');
}

// ...
date: date.toISOString(),
```

ui/src/components/ChatComponent/ChatComponent.jsx
```diff
@@ -0,0 +1,184 @@
...
+      return '';
+    }
+    return `${d.getHours().toString().padStart(2, '0')}:${d
+      .getMinutes()
+      .toString()
+      .padStart(2, '0')}`;
+  } catch {
+    return '';
+  }
+}
```
axelboc, line 28: The chat widget should assume that the dates in the Redux store are defined and in the right format (ISO). If they are not, there's a problem higher up in the chain. I would therefore move the relevant line to the JSX and remove this function entirely:

```js
`${d.getHours().toString().padStart(2, '0')}:${d.getMinutes().toString().padStart(2, '0')}`
```

ui/src/components/ChatComponent/ChatComponent.jsx
```diff
@@ -0,0 +1,184 @@
...
+                  aria-label="Type a message"
+                  value={messageText}
+                  onChange={(e) => setMessageText(e.target.value)}
+                  onKeyDown={handleKeyDown}
+                />
+                <button
+                  type="button"
+                  className={`${styles.send} btn btn-primary`}
+                  onClick={submit}
+                  aria-label="Send"
```
axelboc, line 154: No need for this label since it's the same as the text of the button

ui/src/components/ChatComponent/ChatComponent.jsx
```diff
@@ -0,0 +1,184 @@
...
+                <textarea
+                  className={styles.newMessage}
+                  placeholder="Type a message..."
+                  rows={2}
+                  aria-label="Type a message"
+                  value={messageText}
+                  onChange={(e) => setMessageText(e.target.value)}
+                  onKeyDown={handleKeyDown}
+                />
+                <button
```
axelboc, line 150: Would have been good to use Bootstrap components instead of Bootstrap classnames directly.

ui/src/components/ChatComponent/ChatComponent.jsx
```diff
@@ -0,0 +1,184 @@
...
+            className={styles.launcher}
+            aria-label="Toggle chat"
+            onClick={toggleOpen}
+          >
+            <div
+              className={`${styles.launcherIcon} ${
+                isOpen ? styles.launcherIconOpen : ''
+              }`}
+              aria-hidden="true"
+            />
```
axelboc, line 173: Could you use a FontAwesome icon here?

ui/src/components/ChatComponent/ChatComponent.jsx
```diff
@@ -0,0 +1,184 @@
...
+                  >
+                    <div className={styles.messageText}>
+                      <div>
+                        <div>
+                          <strong>{m.name}</strong>
+                          <span> : </span>
+                        </div>
+                        <div>
+                          {(() => {
+                            const raw = String(m.message || '');
```
axelboc, line 119: Same as the date, the widget should assume that the message is defined and is a string; otherwise, there's a problem somewhere else.

ui/src/components/ChatComponent/ChatComponent.jsx
```diff
@@ -0,0 +1,184 @@
...
+                            const lines = raw
+                              .split('\n')
+                              .map((l) =>
+                                l.endsWith('\r') ? l.slice(0, -1) : l,
+                              );
+                            return lines.map((line, idx) => (
+                              <span key={`${m.id}-${line}`}>
```
axelboc, line 126: No need to prefix the `key` with the message ID — keys just need to be unique within their parent.

ui/src/components/ChatComponent/ChatComponent.jsx
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  return (
+    <div className={styles.chatWidgetDragable}>
+      <Draggable>
+        <div>
+          <div
+            className={`${styles.widgetContainer} ${
+              isOpen ? styles.widgetContainerOpen : styles.widgetContainerClosed
+            } ${isOpen ? '' : styles.hidden}`}
```
axelboc, line 88: Instead of toggling classnames, you could use a `data` attribute and refactor the styles accordingly:

```jsx
data-open={isOpen || undefined}
```

ui/src/components/ChatComponent/ChatComponent.jsx
```diff
@@ -0,0 +1,184 @@
...
+      messagesContainerRef.current.scrollHeight;
+  }, [isOpen, messages.length]);
+
+  function toggleOpen() {
+    setIsOpen((prev) => {
+      if (prev) {
+        dispatch(markAllAsRead());
+      }
+      return !prev;
+    });
```
axelboc, line 58: I think it would work to just check `isOpen` directly:

```js
if (!isOpen) {
  dispatch(markAllAsRead());
}

setIsOpen(!isOpen);
```

axelboc (commented):
Would have appreciated a ping @marcus-oscarsson :sweat_smile: 

Anyway, no worries, it's looking great now that the server state is in Redux and the one-way data flow principle is respected. My remaining feedback is just for polishing.

Thanks and congrats @Phantominumm, it wasn't an easy piece of work :trophy:

marcus-oscarsson:
Thanks for the input @axelboc we will have a look at your suggestions together with @Phantominumm in a follow up PR.

PR #1926, opened 2025-10-16
[Sample List] make sure we always display the Flex view even for single Cell

jbflo:
This ensure a consistent view for Beamline using Flex Sample changer

and should change nothing for those not using it.

Note : when Flex Sample changer as only single cell loaded we still want the View for multiple cell being Display

### DIFF
--- a/ui/src/containers/SampleGridTableContainer.jsx
+++ b/ui/src/containers/SampleGridTableContainer.jsx
@@ -130,8 +130,8 @@ function checkForOverlap(el1, el2) {
 }
 
 // Helper function to determine puck colsm value
-function getColsm(isSingleCell, puckCount) {
-  if (isSingleCell) {
+function getColsm(isSingleCellAndNotFlex, puckCount) {
+  if (isSingleCellAndNotFlex) {
     return puckCount <= 4 ? 3 : true;
   }
   return puckCount === 1 ? 2 : puckCount === 2 ? 4 : puckCount === 3 ? 6 : true;
@@ -176,6 +176,7 @@ export default function SampleGridTableContainer(props) {
   const isSingleCell = Object.values(sampleList).every(
     (sample) => sample.cell_no === 1 || sample.cell_no === 0,
   );
+  const isSingleCellAndNotFlex = isSingleCell && !type.includes('FLEX');
 
   // this supose to replace old shouldComponentUpdate , not sure we need it
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


jbflo:
@marcus-oscarsson that would  solve the concern of id23-1

hope this clean enough  😊

ui/src/containers/SampleGridTableContainer.jsx
```diff
@@ -966,7 +973,7 @@ export default function SampleGridTableContainer(props) {
     const isFLEX = type.includes('FLEX');
     const isMOCK = type.includes('Mock');
 
-    if (isFLEX) {
+    if (isFLEX && !isSingleCell) {
```
axelboc, line 976: Just to match the order of other conditions:

```suggestion
    if (!isSingleCell && isFLEX) {
```

ui/src/containers/SampleGridTableContainer.jsx
```diff
@@ -976,7 +983,8 @@ export default function SampleGridTableContainer(props) {
         />
       );
     }
-    if (isSingleCell || isMOCK) {
+
+    if (isSingleCell || isMOCK || (isFLEX && isSingleCell)) {
```
axelboc: I don't think `&& isSingleCell` does anything: :thinking:

- if it's `true` then `isSingleCell ||` at the start already passes the condition;
- if it's `false` then it doesn't matter what `isFLEX` is; it never passes the condition (i.e. so adding `|| (isFLEX && isSingleCell)` doesn't change the behaviour compared to what was there before).

jbflo: correct, the extra "|| (isFLEX && isSingleCell)" is redundant. thanks.
i forgot that i added the  && !isSingleCel in the first condition.

marcus-oscarsson:
Do we at all need the `isSingleCell` now, or how do we use it ?

marcus-oscarsson:
Thanks JB :)

jbflo:
> isSingleCell

I think there are a few beamlines using CATS, this boolean help having a different view (The Mockup view).
it could be replace by something else. a more conventional name maybe.

marcus-oscarsson:
I see, its perhaps for a different PR. I thought as we are no longer treating the single cell case differently we could simply remove it. So that All sample changers are displayed as a sample changer with at-least one cell. I thought that would simplify the logic somewhat.

ui/src/containers/SampleGridTableContainer.jsx
marcus-oscarsson: Is the Flex sample changer ever single cell (with what we want to do ?)

jbflo: we have this : const isSingleCellAndNotFlex = isSingleCell && !type.includes('FLEX');
this could also  be 
const isFlexAndIsSingleCell =  type.includes('FLEX') && isSingleCell ;

i found it simpler to check  isSingleCellAndNotFlex

marcus-oscarsson: Ah, I was thinking is it at all needed but lets not make things more complicated at the moment. I mean that we could simply say that all sample changers have at least one cell and that would simplify the logic a bit but lets keep it like this for the time being.

PR #1927, opened 2025-10-16
Update `ruff.toml`'s `lint.per-file-ignores` table

fabcor-maxiv:
These exceptions are not relevant any longer.
Either the code has been fixed or the rule is ignored globally.


marcus-oscarsson:
Thanks

PR #1929, opened 2025-10-17
Remove the `/docker` directory

### LINKED ISSUE #1928: Should we delete the Docker stuff?
I do not recall ever hearing about anyone using the Docker stuff from the `/docker` directory. Is anyone using it? Is anyone maintaining it? Should we delete the whole directory?

fabcor-maxiv:
This does not seem to be used.
This does not seem to be maintained.

GitHub: fix https://github.com/mxcube/mxcubeweb/issues/1928


marcus-oscarsson:
Not sure about this one, I think these docker images would work, or at least with little modification. It somehow seems unnecessary to remove them. We are currently not using docker, but we may for some purposes in the future and then this code can become handy.

fabcor-maxiv:
> we may for some purposes in the future and then this code can become handy.

We can always fish stuff back out of old git commits, it would not be completely gone.

marcus-oscarsson:
Sure but you need to remember that it was there and look for it, its not the same.

PR #1932, opened 2025-10-17
Add package initializer in `/test`

fabcor-maxiv:
[Pytest recommends to arrange tests in importable packages](https://docs.pytest.org/en/stable/explanation/pythonpath.html#import-modes).


marcus-oscarsson:
Thanks, looks good

fabcor-maxiv:
> Thanks, looks good

I take this as an official approval and will merge now

PR #1933, opened 2025-10-17
Migrated samplecentring route and component to SampleView adapter

marcus-oscarsson:
Migrated `samplecentring` route and component to `SampleViewAdapter` object.

 - Removing the routes statically defined in `routes/samplecentring.py`
 - Moved signal handlers from `signals.py` to `SampleViewAdapter`, like this we can hopefully relatively soon remove `signals.py`
 - The previous `SampleCentring` component  was turned into an adapter `SampleViewAdapter` mapping to the `SampleView` HardwareObject


ui/src/api/sampleview.js
```diff
@@ -1,59 +1,91 @@
...
 }
 
 export function sendAbortCentring() {
-  return endpoint.put(undefined, '/centring/abort').res();
+  return sendExecuteCommand('sample_view', 'sample_view', 'abort_centring', {});
 }
 
 export function sendMoveToPoint(id) {
   return endpoint.put(undefined, `/centring/${id}/moveto`).res();
 }
```
walesch-yan: this endpoint was removed with `samplecentring.py` it seems that there is currently no cmmand in the adapter to replace it

mxcubeweb/core/adapter/sample_view_adapter.py
```diff
@@ -1,67 +1,77 @@
...
-    def _emit_shapes_updated(self):
-        shape_dict = {}
+    SUPPORTED_TYPES: ClassVar[list[object]] = [AbstractSampleView]
 
-        for shape in HWR.beamline.sample_view.get_shapes():
-            _s = to_camel(shape.as_dict())
-            shape_dict.update({shape.id: _s})
+    def __init__(self, ho, role, app):
+        super().__init__(ho, role, app, resource_handler_config)
+        self.app = app
```
walesch-yan: The app initialization is already done in the parent class `__init__`. Not urgent, just a small comment:)

ui/src/api/sampleview.js
marcus-oscarsson: Well spotted, thanks alot for having a look :+1: . Ill think about a nice way of testing adapters and components given the spec we give ...

ui/src/api/sampleview.js
```diff
@@ -1,59 +1,93 @@
-import api from './api';
-
-const endpoint = api.url('/sampleview');
+import { fetchAttribute, sendExecuteCommand } from './hardware-object';
 
 export function fetchImageData() {
-  return endpoint.get('/camera').safeJson();
+  return fetchAttribute('sample_view', 'sample_view', 'sample_image_meta_data');
```
axelboc: Might be worth adding two local functions: `fetchSampleViewAttribute` and `sendSampleViewCommand` to reduce repetition :shrug:

axelboc (approved):

ui/src/api/sampleview.js
marcus-oscarsson: Yes its a good point, I was thinking about it but then I thought that the "helper that calls a helper chain" became a bit long if you see what I mean. It would make it easier to change in the future, ill see what I can come up with.

axelboc: yeah, maybe it's the end helpers that aren't needed (`fetchImageData` and the like)

marcus-oscarsson: Also a good point and perhaps a long discussion. It makes the code easier to read but a bit harder to debug as you need to follow all the calls. 

I think the helpers such as `fetchImageData` is not really needed once the underlying API is clear and well defined (and easy enough to understand by reading the call). It has not really been the case until now :rofl:.  

We are slowly improving I guess :). I think a helper for fetchAttribute`(sample_view', 'sample_view' ...)` would indeed be reasonable. We only have one `sample_view` (and it will probably remain like that) but the API allows for several ones you could i theory have `fetchAttribute(sample_view', 'sample_view_one' ...)` `fetchAttribute(sample_view', 'sample_view_two' ...)`.

marcus-oscarsson: The first `sample_view` corresponding to the "type" in `fetchAttribute(sample_view', 'sample_view_two' ...). is in theory not needed but we kept it like this to avoid id clashes (for now). One can perhaps see it as a sort of object namespace.

axelboc: No worries, it's only cosmetic anyway.

marcus-oscarsson: I finally made the helpers, and then Ill follow up with a PR the uses the helpers `fetchSampleViewAttribute` and `sendSampleViewCommand` directly.

PR #1934, opened 2025-10-17
Improve Python docstrings

fabcor-maxiv:
The intention of this commit is to improve the Python docstrings so that tools like Sphinx can generate a more useful output.

For this purpose, the Ruff `D` rules are enabled except `D1`, which means all the rules from the `pydocstyle` group except the rules about missing docstrings.

This does not attempt to unify type hints between docstrings and annotations. This does not attempt to add missing docstrings, or fix incomplete or incorrect docstrings.

Most of this was done automatically by Ruff with `ruff check --fix`. Then some manual editing was done to fix remaining issues:

* Some light rephrasing, in particular to extract a title line.
* Some formatting of code embedded in docstrings.
* Some typo fixes here and there.

Still, a lot of docstrings do not follow the Google style, although that is our chosen style according to our style guidelines.

### DIFF
--- a/mxcubeweb/app.py
+++ b/mxcubeweb/app.py
@@ -1,6 +1,7 @@
-"""
-Module that contains application wide settings and state as well as functions
-for accessing and manipulating those.
+"""Application-wide settings and state.
+
+Module that contains application-wide settings and state
+as well as functions for accessing and manipulating them.
 """
 
 import logging
@@ -118,14 +119,11 @@ class MXCUBEApplication:
         enabled_logger_list,
         cfg,
     ):
-        """
-        Initializes application wide variables, sample video stream, and applies
+        """Initializes application wide variables, sample video stream, and applies.
 
-        :param hwr: HardwareRepository module
-        :param bool allow_remote: Allow remote usage, True else False
-        :param bool ra_timeout: Timeout gives control, True else False
-
-        :return None:
+        Params:
+            allow_remote(bool): Allow remote usage, ``True`` else ``False``.
+            ra_timeout(bool): Timeout gives control, ``True`` else ``False``.
         """
         # The routes created by the AdapterResourceHandler
         # via the factory are kept between calls to init as they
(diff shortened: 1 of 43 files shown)


marcus-oscarsson:
It would be useful to know what improvement these are, it looks to be mostly formatting but also some other things. A quick summary would help, so that one knows what to look for before reviewing.

fabcor-maxiv:
> It would be useful to know what improvement these are, it looks to be mostly formatting but also some other things. A quick summary would help, so that one knows what to look for before reviewing.

Right, I need to add that. Will do...

fabcor-maxiv:
> It would be useful to know what improvement these are, it looks to be mostly formatting but also some other things. A quick summary would help, so that one knows what to look for before reviewing.

Done

marcus-oscarsson (approved):

marcus-oscarsson:
Ok, looks good as far as I can see. W are indeed using Google doc strings so its great if the documentation is formatted correctly especially if it makes the generated documentation nicer. I have to say that it looks strange to have the title on the same line as `"""`, it does not read as well when one read the source code.

fabcor-maxiv:
> it looks strange to have the title on the same line as `"""`, it does not read as well when one read the source code

I have a slight preference for this style, but I do not mind much either way. If it is possible to enforce this with the tools, we can change.

beteva:
> > it looks strange to have the title on the same line as `"""`, it does not read as well when one read the source code
> 
> I have a slight preference for this style, but I do not mind much either way. If it is possible to enforce this with the tools, we can change.

Is the blank line between the two lines of the title really part of the style. It looks very strange to read in the code, especially if the sentence is split between the two lines.

fabcor-maxiv:
> Is the blank line between the two lines of the title really part of the style.

Yes.

* https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html
* https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings

fabcor-maxiv:
> Is the blank line between the two lines of the title really part of the style.

Ah, I misread. The title should fit on one line. So obviously there should never be such a thing as a "blank line between the two lines of the title". One should rephrase the title to make it fit one one line. Whatever does not fit in the title can be described in the body of the docstring.

In this pull request, I had to do some manual edits to figure out a title that fits on one line. Maybe I did a poor job at this, but honestly stuff was not great before, I doubt I made things much worse.

fabcor-maxiv:
I have not looked at this into details, but I was under the impression that maybe under some conditions Ruff allowed a docstring to go further than the 88th column. Do not take my word for granted, though. I really have not investigated this, maybe I hallucinated this.

PR #1939, opened 2025-10-20
[SampleGridTableItem ] Refactor to function comp

jbflo:
Converted SampleGridTableItem from a class component to  functional component and some minor cleanup

🚀 Changes Made
- Converted class component to functional component
- Streamlined prop handling with default values

Props Cleanup/ Removed:
itemKey (unused)
selected (unused)
allowedDirections (unused)
onClick (unused handler)
data-attributes (unused)

Renamed/Simplified method names for better readability:
seqId -> getsequenceId
sampleInformation - > getSampleInformation
sampleDisplayName - > getSampleName

export component as default.

### DIFF
--- a/ui/src/components/SampleGrid/SampleGridTableItem.jsx
+++ b/ui/src/components/SampleGrid/SampleGridTableItem.jsx
@@ -16,7 +16,7 @@ import styles from './SampleGridTableItem.module.css';
 
 export default function SampleGridTableItem({
   sampleData = {},
-  queueOrder = [],
+  queueOrder = '',
   current = false,
   picked = false,
   pickButtonOnClickHandler,
@@ -156,13 +156,10 @@ export default function SampleGridTableItem({
               <Badge
                 href={limsLink}
                 target="_blank"
+                rel="noopener noreferrer"
                 bg="light"
                 text="primary"
                 className={`${styles.samplesGridTableItemNameProteinAcronym} ms-1 mt-2`}
-                data-type="text"
-                data-pk="1"
-                data-url="/post"
-                data-title="Enter protein acronym"
               >
                 {sampleName}
               </Badge>
--- a/ui/src/components/SampleGrid/SampleGridTableItem.jsx
+++ b/ui/src/components/SampleGrid/SampleGridTableItem.jsx
@@ -1,6 +1,4 @@
-/* eslint-disable react/destructuring-assignment */
 import cx from 'classnames';
-import React from 'react';
 import {
   Badge,
   Button,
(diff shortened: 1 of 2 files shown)


ui/src/components/SampleGrid/SampleGridTableItem.jsx
```diff
@@ -136,91 +124,71 @@ export class SampleGridTableItem extends React.Component {
...
+            >
+              <Badge
+                href={limsLink}
+                target="_blank"
+                bg="light"
+                text="primary"
+                className={`${styles.samplesGridTableItemNameProteinAcronym} ms-1 mt-2`}
+                data-type="text"
+                data-pk="1"
+                data-url="/post"
```
marcus-oscarsson: What does these data attributes do and whats the `data-url` for ?

marcus-oscarsson: Not sure that its an issue in this case as its our own link but we should add rel="noopener noreferrer" to links using target="_blank"

jbflo: @axelboc might explain better. he knows the data-attributes very-well

but i just check and i think those data-attributes : data-pk, data-type, data-title are probably not needed in this case . as this is not an editable text field. 
and the backend does not waiting for any POST request.

I leave them but i was not sure why they were there : 😃

marcus-oscarsson: You can remove data attributes, they do not seem to be used

ui/src/components/SampleGrid/SampleGridTableItem.jsx
```diff
@@ -16,32 +15,24 @@ import containerStyles from '../../containers/SampleGridTableContainer.module.cs
...
+  current = false,
+  picked = false,
+  pickButtonOnClickHandler,
+  children,
+}) {
+  const pickButtonOnClick = useCallback(
+    (e) => {
+      if (pickButtonOnClickHandler) {
+        pickButtonOnClickHandler(e, sampleData.sampleID);
+      }
```
axelboc: I quite like this syntax personally, but it's a matter of taste:

```suggestion
      pickButtonOnClickHandler?.(e, sampleData.sampleID);
```

axelboc: I don't think you need to memoise this handler with `useCallback`. You can even just inline it in the JSX I think.

axelboc (approved):
Nice! Can you please follow it up with a quick PR to either inline (or extract into components if relevant) the three functions that return JSX? I think `getSampleName` could also be moved into a `util.ts` file.

mockoocy (approved):

marcus-oscarsson:
Thanks JB !

PR #1940, opened 2025-10-20
fix slick styles

walesch-yan:
Currently there is a visual bug, where the left arrow of the slider for Sample list items is not visible:
Screenshot from 2025-10-20 17-35-47
This is caused by a css style, that is not applied to slick elements. This PR adds a `:global` field to make sure that the style is used. The result:
Screenshot from 2025-10-20 17-35-33

### DIFF
--- a/ui/src/containers/SampleGridTableContainer.module.css
+++ b/ui/src/containers/SampleGridTableContainer.module.css
@@ -81,18 +81,7 @@
   margin-left: 0 !important;
 }
 
-.samplesGridTableItemTasks > .slick-next::before,
-.samplesGridTableItemTasks > .slick-prev:before {
-  color: #9e9c9c;
-  opacity: 100%;
-}
-
-.samplesGridTableItemTasks > .slick-prev.slick-disabled:before,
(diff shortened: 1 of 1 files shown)


ui/src/containers/SampleGridTableContainer.module.css
```diff
@@ -81,18 +81,7 @@
...
-.samplesGridTableItemTasks > .slick-next::before,
-.samplesGridTableItemTasks > .slick-prev:before {
-  color: #9e9c9c;
-  opacity: 100%;
-}
-
-.samplesGridTableItemTasks > .slick-prev.slick-disabled:before,
-.samplesGridTableItemTasks > .slick-next.slick-disabled:before {
-  opacity: 25%;
-}
```
walesch-yan, line 93: I removed these styles as well, as this is already the default behavior in `slick-theme.css` and they did not work due to a missing `:`

jbflo (approved):

axelboc (approved):

mockoocy (approved):

PR #1941, opened 2025-10-21
Remove `.gitmodules`

fabcor-maxiv:
`.gitmodules` is not in use currently.


fabcor-maxiv:
I think it is not in use. But maybe it is. I do not know. I do not know what this file is for. It should be for git submodules and we do not have git submodules, right?

mockoocy (approved):

marcus-oscarsson:
Thanks, yes you can remove the `.gitmodules` file

PR #1942, opened 2025-10-21
Remove file `/test/.babelrc`

fabcor-maxiv:
The file `/test/.babelrc` is unused.


fabcor-maxiv:
I guess it is the Babel from the JavaScript ecosystem, but there is no JavaScript in `/test`. Seems like this file has no reason to be here.

dominikatrojanowska (approved):

axelboc (approved):

PR #1944, opened 2025-10-21
Allow f-strings in logging

dominikatrojanowska:


marcus-oscarsson:
Please no, I think we had this conversation before. We would like to be able to use fstrings

dominikatrojanowska:
I closed this already, but I wanted to share two thoughts:
- With this PR I did not disabled f-strings completely. I only removed few of them from logging methods and also removed corresponding ruff rule exclusions. Which are always possible to be added in the future. 
- I've just found the [discussion](https://github.com/mxcube/mxcubeweb/issues/1567#issuecomment-2677691943) you probably meant. But as far as I understand the conclusion was to change it bit by bit to follow these ruff rules some day. If I am mistaken, maybe a good solution would be to add the general rule exclusion [on top of the ruff.toml file](https://github.com/mxcube/mxcubeweb/blob/8f80c9a8dd3a72805ba7700cd84dd4e53db42719/ruff.toml#L14).

marcus-oscarsson:
Its true that we did not add all exceptions regarding f-strings yet. But we should be able to use f-strings basically everywhere including logging and exceptions. 

It would be good if we could add this rule to the ignore list instead.

marcus-oscarsson:
So I would instead add "G004" to the ignore list like: https://github.com/mxcube/mxcubecore/blob/9f39ccd5e00d7542d1c6723d0f2027324e49904f/ruff.toml#L16

fabcor-maxiv (approved):
Seems fine to me.

One should feel free to make a pull request to disable `G004` globally.

As far as I know, there is no way to tell Ruff to enforce f-strings for logging, but I have not tried or investigated much.

marcus-oscarsson:
Ok, I did not mean that we should enforcing f-strings. What I meant is that we would like to keep the code as it was, ignoring the rule `G004`. I guess this PR was made partly to comply with `G004`, a rule we don't really want to apply.

My suggestion was hence to remove the changes related to `G004` and add it to the ignore list.

marcus-oscarsson:
Excellent, thanks alot !

PR #1945, opened 2025-10-22
Fix file mode bits

fabcor-maxiv:
Some files had mode bits different from `644` for no reason.


marcus-oscarsson:
Ok, thanks. Could you rebease and merge please ?

PR #1947, opened 2025-10-29
OAV video: make `start_streaming()` hook optional

elmjag:
Let's make `start_streaming()` method of the camera HWO optional.

We don't use at MAX IV. We start the video streaming process using external software (supervisitor). I suspect other facilities want to do it this way as well.

### DIFF
--- a/mxcubeweb/app.py
+++ b/mxcubeweb/app.py
@@ -143,10 +143,11 @@ class MXCUBEApplication:
 
         MXCUBEApplication.mxcubecore.init()
 
-        MXCUBEApplication.init_sample_video(
-            _format=cfg.app.VIDEO_FORMAT,
-            port=cfg.app.VIDEO_STREAM_PORT,
-        )
+        if cfg.app.MXCUBE_STARTS_VIDEO_STREAM:
+            MXCUBEApplication.init_sample_video(
+                _format=cfg.app.VIDEO_FORMAT,
+                port=cfg.app.VIDEO_STREAM_PORT,
+            )
 
         MXCUBEApplication.init_logging(log_fpath, log_level, enabled_logger_list)
 
--- a/demo.yaml/mxcube-web/server.yaml
+++ b/demo.yaml/mxcube-web/server.yaml
@@ -27,6 +27,10 @@ mxcube:
 
   # At which port to stream from
   VIDEO_STREAM_PORT: 8000
+
+  # Start OAV stream from MXCuBE process
+  MXCUBE_STARTS_VIDEO_STREAM: true
+
   # Mode, SSX-CHIP, SSX-INJECTOR, OSC
   mode: OSC
 
--- a/demo/mxcube-web/server.yaml
+++ b/demo/mxcube-web/server.yaml
@@ -64,6 +64,10 @@ mxcube:
 
   # At which port to stream from
   VIDEO_STREAM_PORT: 8000
+
+  # Start OAV stream from MXCuBE process
+  MXCUBE_STARTS_VIDEO_STREAM: true
+
   # Mode, SSX-CHIP, SSX-INJECTOR, OSC
   mode: OSC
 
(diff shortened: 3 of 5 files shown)


mxcubeweb/app.py
```diff
@@ -178,8 +178,18 @@ def init(
     @staticmethod
     def init_sample_video(_format, port) -> None:
         """Initialize video streaming."""
+        camera = HWR.beamline.sample_view.camera
```
marcus-oscarsson: It would in my opinion nicer to make this configurable with a configuration option or possibly a command line switch/option.

mockoocy (approved):

marcus-oscarsson (changes requested):
It would be nice if we could avoid this sort of "duck typing" and make this configurable instead.

mxcubeweb/app.py
elmjag: There used to be a `USE_EXTERNAL_STREAMER` configurable, but it was removed in this PR https://github.com/mxcube/mxcubeweb/pull/1865

Should we restore it? Or do we want to introduce some other configurable?

Perhaps something like `START_OAV_STREAM` ? It could default to `True`. The logic would be:

If `START_OAV_STREAM` is `True`, run the `start_streaming()` method, otherwise skip it.

fabcor-maxiv: Seems to me like there should be an abstract class that declares a `start_streaming` method. So that in our implementation we can implement as a "no-op". This seems to me like it would be the cleaner solution.

marcus-oscarsson: @fabcor-maxiv thats another solution but then you don't have the option to not use it without changing implementation

marcus-oscarsson: @elmjag the `USE_EXTERNAL_STREAMER` was used to differ between the "exernal" `video-streamer` software or the internal ("built-in") mjpeg streamer. The process was always started by mxcube, so external did not mean externally managed/started.

If that is what you want, and why not, then I would add a configuration option to achieve this. You could keep the name `USE_EXTERNAL_STREAMER` but perhaps something like `MXCUBE_STARTS_VIDEO_STREAM` or `MXCUBE_STARTS_OAV_STREAM` or similar is more intuitive.

elmjag: I have added the `MXCUBE_STARTS_VIDEO_STREAM` configuration option now.

fabcor-maxiv: Maybe `MXCUBE_MANAGES_OAV_STREAM`.

I think it is important that MXCuBE also stops the process. We had in the past an issue where the Video Streamer was not terminated properly and we ended up having multiple Video Streamer processes running. Has this been taken into consideration here?

But for the record, I still think this is not a good solution. At the very least there should be an abstract class that declares this API.

@elmjag, I think we might be better off adding our own `start_streaming` method that does nothing. But obviously I do not mind if you want to go further with this solution. Your call.

marcus-oscarsson:
Thanks @elmjag, looks great !

marcus-oscarsson (approved):

mxcubeweb/app.py
elmjag: > I think it is important that MXCuBE also stops the process. We had in the past an issue where the Video Streamer was not terminated properly and we ended up having multiple Video Streamer processes running. Has this been taken into consideration here?
> 

I think MXCuBE tries to do that. When you stop demo beamline, following is logged:

`2025-11-07 08:36:25,877 |HWR    |INFO   | Shutting down video_stream...`

Don't know if that actually works. Implementing correct termination of processes is surprising complicated. This is why I think sticking to dedicated software, like Supervisor, is preferable.

elmjag: I'll merge this.

I like this solution. I see some situations where it could be useful to switch between MXCuBE or external software starts streaming, without having to change the Camera HWO code.

marcus-oscarsson: > 2025-11-07 08:36:25,877 |HWR |INFO | Shutting down video_stream...

It does and this works well so far :)

marcus-oscarsson: > I like this solution. I see some situations where it could be useful to switch between MXCuBE or external software starts streaming, without having to change the Camera HWO code.

For sure, very nice. Thanks !

PR #1948, opened 2025-10-29
Fix QueueSettings' labels behaviour

matyson:
I am refactoring the `QueueSettings` dropdown menu and the `GroupFolderInput` form to fix the labels expected behaviour, right now the click on the labels don't check their boxes neither the tab navigation in the dropdown menu check the boxes on space click, so i basically turn the DropdownItems into divs so that they respect their childs standard behaviour. Also, refactored the `GroupFolderInput` form so that the label focus the input box and the enter key triggers the form submission along with the button click.

Quick demo:

https://github.com/user-attachments/assets/26e4e467-d352-4c49-a5ab-3d9461c03a4c

### DIFF
--- a/ui/src/containers/GroupFolderInput.jsx
+++ b/ui/src/containers/GroupFolderInput.jsx
@@ -29,25 +29,25 @@ function GroupFolderInput() {
   }
 
   return (
-    <Form as="div">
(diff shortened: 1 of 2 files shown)


axelboc (approved):
Brilliant :+1:

PR #1949, opened 2025-10-30
demo beamlines: use `beamline.yaml` config file name

elmjag:
For demo beamlines rename beamline configuration files from `beamline_config.yml` to `beamline.yaml`.

The `beamline_config.yml` file name is now deprecated in favor of `beamline.yaml`.


marcus-oscarsson:
Thanks !

PR #1946, opened 2025-10-24
Reworking on redux store: queue, sample order, sample list...

dominikatrojanowska:
This PR is an outcome of: https://github.com/mxcube/mxcubeweb/issues/1943

Changes within this PR:
1. Uploading samples from LIMS / SC:
	* resets the **queue.queue** - which is not correct, it should be empty as nothing was added to the queue. It may be emptied by refreshing the page. WIP
	* resets **sampleGrid.order**
	* resets **sampleGrid.sampleList**

2. Adding **new** sample to queue (`QueueGUI` > `Create Sample` > `Queue`) :
	* appends **queue.queue** with this new sample, 
	
        Screenshot from 2025-10-24 16-40-04

	* appends **sampleGrid.order**, what's important does not clear the previous content, which afaik may lead to the similar issues https://github.com/mxcube/mxcubeweb/pull/1814#issuecomment-3430846458
	
	Screenshot from 2025-10-24 16-42-21

	* updates **sampleGrid.sampleList**, adds new sample at some (alphabetical) point, this new sample has property: ``queueID`` and this is important in order to have `Next Sample` button and be listed in `Queued Samples` of  `Queue GUI` (screenshot 4)

	Screenshot from 2025-10-24 16-43-56

3. Adding **exisitng** sample to the queue (`Sample View` > `RMB` > `Add to Queue`):
	* appends **queue.queue** with this sample
	
	Screenshot from 2025-10-24 16-45-05

	* **sampleGrid.order** remains as it was before
	
	Screenshot from 2025-10-24 16-45-24
	
	* updates **sampleGrid.sampleList**, enqueued sample has new property: queueID
	
	Screenshot from 2025-10-24 16-46-06

PR #1950, opened 2025-10-30
[Sample Gid] Fix Mount Sample issue

jbflo:
This PR Fix the mount sample issue from the **sample grid side**

- Replace sample Order list by the  sample list
Issue occur after this change : https://github.com/mxcube/mxcubeweb/pull/1814

I  also add a commit to :
- Use destructuring for sampleGrid state selection

"Follow-up PR: Remove sampleGrid.order dependency/state globally

### DIFF
--- a/ui/src/containers/SampleGridTableContainer.jsx
+++ b/ui/src/containers/SampleGridTableContainer.jsx
@@ -140,11 +140,10 @@ function getColsm(isSingleCellAndNotFlex, puckCount) {
 export default function SampleGridTableContainer(props) {
   const navigate = useNavigate();
   const dispatch = useDispatch();
-  const sampleList = useSelector((state) => state.sampleGrid.sampleList);
-  const filterOptions = useSelector((state) => state.sampleGrid.filterOptions);
-  const selected = useSelector((state) => state.sampleGrid.selected);
+  const { selected, sampleList, order, filterOptions } = useSelector(
+    (state) => state.sampleGrid,
+  );
   const queue = useSelector((state) => state.queue);
-  const order = useSelector((state) => state.sampleGrid.order);
   const sampleChanger = useSelector((state) => state.sampleChanger);
   const contextMenu = useSelector(
     (state) => state.contextMenu.genericContextMenu,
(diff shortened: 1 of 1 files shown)


ui/src/containers/SampleGridTableContainer.jsx
```diff
@@ -828,18 +820,12 @@ export default function SampleGridTableContainer(props) {
   }
 
   function mountAndCollect() {
-    let sampleData = null;
-
     // If several samples selected mount the first one and add the others to the queue
```
jbflo, line 816: This comment was not taking into account , now it is.

mockoocy (approved):
Thanks for looking into that, your fix seems to work fine.

I wonder if perhaps `order` state could be removed from the `sampleGrid.js` reducer then? I am not sure if it is being used anywhere now.

jbflo:
> I wonder if perhaps `order` state could be removed from the `sampleGrid.js` reducer then? I am not sure if it is being used anywhere now.

it actually used to display the order of the samples in the Queue, [1, 2 , 3 , 4 .....]

image

mockoocy:
> it actually used to display the order of the samples in the Queue, [1, 2 , 3 , 4 .....]

Alright then, I've thought that these come from `queue.queue` and did not double check that :)
Now I see it, 
https://github.com/mxcube/mxcubeweb/blob/a618b1424d95f8d801c4a4d1192ad477d0575361/ui/src/containers/SampleGridTableContainer.jsx#L577-L581

Anyway `sampleGrid.order` and `queue.queue` seem to be always populated with the same values. It should not be that way, but it's a different discussion :D I guess it could be a left for future investigations. 

Nonetheless, the above does not seem to matter with your changes, which is great 🎉!


Btw, it seems worth checking out for @dominikatrojanowska, perhaps #1946 is no longer needed once this one were to be merged.

jbflo:
> sampleGrid.order

"Now I see what you meant - yes we can remove the sample order completely and just use queue.queue directly. This approach is much  cleaner."

marcus-oscarsson:
There is a conflict, could you have a quick look ?

marcus-oscarsson:
On a streak :) :rocket:

dominikatrojanowska:
> I wonder if perhaps `order` state could be removed from the `sampleGrid.js` reducer then? I am not sure if it is being used anywhere now.

Yeees, perhaps the `order` state may be deleted. It is used in few places right now but I suspect it may be exchanged with `queue` state. This duality that we achieved with these two states leads to the confusion and as far as I'm aware it does not provide any extra functionality that wouldn't be achieved with one only.

PR #1953, opened 2025-11-04
Calling SampleChangerAdapter instead of component

marcus-oscarsson:
The sample changer component no longer exists the consequence of the current code is that no sample is mounted when calling `execute_entry_with_id`. 

Calling the  `SampleChangerAdapter` instead to maintain the current behavior, a future development would be to push back this behavior to the `mxcubecore` queue so that even an "empty" sample gets mounted.

### DIFF
--- a/mxcubeweb/core/components/queue.py
+++ b/mxcubeweb/core/components/queue.py
@@ -16,6 +16,9 @@ from mxcubecore.model.queue_model_enumerables import CENTRING_METHOD
 from mxcubecore.queue_entry.base_queue_entry import QUEUE_ENTRY_STATUS
 
 from mxcubeweb.core.components.component_base import ComponentBase
+from mxcubeweb.core.models.adaptermodels import (
+    SampleInputModel,
+)
 from mxcubeweb.core.models.generic import SimpleNameValue
 from mxcubeweb.core.util.convertutils import (
     str_to_camel,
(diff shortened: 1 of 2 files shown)


mxcubeweb/core/components/queue.py
```diff
@@ -1821,8 +1821,11 @@ def execute_entry_with_id(self, sid, tindex=None):
                 not len(current_queue[sid]["tasks"])
             ) and sid != self.app.lims.get_current_sample().get("sampleID", ""):
                 try:
-                    self.app.sample_changer.mount_sample(current_queue[sid], wait=True)
+                    self.app.mxcubecore.get_adapter("sample_changer").mount_sample(
+                        current_queue[sid], wait=True
+                    )
```
fabcor-maxiv, line 1829: I recommend having as little code as possible in the `try` block, so that we are more precise about what is acceptable to catch and ignore at runtime, as opposed to what should straight up crash MXCuBE.

```suggestion
                sample_changer = self.app.mxcubecore.get_adapter("sample_changer")
                try:
                    sample_changer.mount_sample(
                        current_queue[sid], wait=True
                    )
```

marcus-oscarsson: Not sure I think thats a good idea at this stage. I think its perfectly fine to fail the queue if for some reason `get_adapter("sample_changer")` would return None. If this would happen we don't want it to be raised further. Logging it is completely fine.

elmjag:
This solves issue https://github.com/mxcube/mxcubeweb/issues/1957 \o/

elmjag (approved):

PR #1954, opened 2025-11-04
[Sample State Part 2]  update Sample Grid Table Item to better display Sample information like State

jbflo:
This PR UPDATE SampleGridTableItem 

See Part 1 First  https://github.com/mxcube/mxcubeweb/pull/1951

Goal:  display Sample State  (from the Sample Changer)

What changed

- Extracted sample information rendering into its own component (improves readability & reuse).
- Moved sampleStateBackground and getSampleName helpers to utils.js.
- Inlined getSequenceId() & getItemControls() 
- Simplified the “Copy to Clipboard” component by removing the button node.
- Added a Sample State badge to each item.
- adjusted styles accordingly.

image

 

Testing
- Verified that the Sample State badge reflects states returned by the Sample Changer.

The component need a little refactoring to adjust things properly.
Rather than opening 4 separate PRs, this is a single PR with clearly separated commits so each change can be reviewed in isolation.

Follow-ups
confirm  if the sample state change / sample_list is update when mount or unmounted sample

### DIFF
--- a/ui/src/components/CopyToClipboard/CopyToClipboard.jsx
+++ b/ui/src/components/CopyToClipboard/CopyToClipboard.jsx
@@ -1,5 +1,4 @@
 import { useState } from 'react';
-import { Button } from 'react-bootstrap';
 import { CopyToClipboard as TriggerCopyToClipboard } from 'react-copy-to-clipboard';
 import { MdContentCopy } from 'react-icons/md';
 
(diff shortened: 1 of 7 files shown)


axelboc (approved):
Thanks!

axelboc:
Sorry, the diff wasn't showing me everything when I first opened it. Having another look :sweat_smile:

jbflo:
> Sorry, the diff wasn't showing me everything when I first opened it. Having another look 😅
 
 Avoid "Files changed 7" Tab for this one.
 preferable " Commits 5 " Tab. It should be easier if you go commit by commit 😅

axelboc:
Sorry, got pulled into a meeting. Looks good to me (for real this time)!

marcus-oscarsson:
In what version of `mxcubecore` did you add the backend bit for this ?, we should update `pyproject.toml`. I guess this would not work otherwise ?

jbflo:
> In what version of `mxcubecore` did you add the backend bit for this ?, we should update `pyproject.toml`. I guess this would not work otherwise ?

Yes Thanks for noticing.
FOR User of the EMBLFlexHCD : This should be merge first : [1] https://github.com/mxcube/mxcubecore/pull/1432
[2]https://github.com/mxcube/mxcubecore/pull/1433/

but technically this PR does not contain Breaking change.
if the information is not in Hardware object level,  the sample grid should behave as before.

Then it's up to every body to update their Sample Changer if they want to see the Sample State information.
That is why i am hesitating to bind this to a  mxcubecore version.

But yes specifically for us "ESRF" for this to test properly we should merge [1]  [2]

PR #1956, opened 2025-11-05
Fix error when retrieving shapes list

mockoocy:
Currently when retrieving a `Grid` object from `shapes` attribute of the sample view adapter, there is a json-serialization error.

Bug reproduction:
1. Add some shapes
2. Refresh the page
3. There will be an error in the network tab for /shapes, the shapes won't appear in the canvas (nor the redux store)

It just makes sure that all shapes are properly serialized before sending them to the backend.


axelboc (approved):
Nice!

marcus-oscarsson:
Thanks, well done. I was about to do the same :)

PR #1958, opened 2025-11-05
Replace react-slick slider with flexbox for sample task item

matyson:
Following discussion in #1906 this works aims to replace the current Slider component in sample TaskItem with a simpler overflow-scroll css utility.  Since it simplified a lot the layout, perhaps it needs some more UI indicators that the field of tasks is scrollable. Any thoughts?

https://github.com/user-attachments/assets/14253331-e1fd-4244-a7cb-358394ab76ff

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -42,12 +42,10 @@
     "react-lazyload": "^3.2.0",
     "react-redux": "7.2.9",
     "react-router-dom": "^6.19.0",
-    "react-slick": "^0.29.0",
     "react-zoom-pan-pinch": "3.3.0",
     "redux": "4.2.1",
     "redux-form": "^8.3.10",
     "redux-logger": "3.0.6",
-    "slick-carousel": "^1.8.1",
     "socket.io-client": "^4.8.1",
     "wretch": "2.7.0"
   },
--- a/ui/pnpm-lock.yaml
+++ b/ui/pnpm-lock.yaml
@@ -77,9 +77,6 @@ importers:
       react-router-dom:
         specifier: ^6.19.0
         version: 6.30.1(react-dom@17.0.2(react@17.0.2))(react@17.0.2)
-      react-slick:
-        specifier: ^0.29.0
-        version: 0.29.0(react-dom@17.0.2(react@17.0.2))(react@17.0.2)
       react-zoom-pan-pinch:
         specifier: 3.3.0
         version: 3.3.0(react-dom@17.0.2(react@17.0.2))(react@17.0.2)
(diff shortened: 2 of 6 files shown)


axelboc (commented):
Great!

I guess you could perhaps use `overflow-x: scroll`  so the scrollbar is always visible, and use `scrollbar-width: thin` to make it a bit more discrete?

Otherwise, there are techniques to show a fading gradient on the scrollable edge of the container. I actually found this CSS-only implementation which looks very short: https://css-scroll-shadows.vercel.app/?bgColor=fcc88d&shadowColor=222222&pxSize=15

marcus-oscarsson:
Nice, well done !. I think some sort of indicator that its scrollable would be nice. Its not a very used feature so I think its alright as it is now especially because its a simple solution. Its a good compromise. But yes a scroll indicator of sorts would be nice. I guess you have to click on it to scroll as well ?

axelboc:
In the recording, it looks like you have to click on the sample card first before the scrollbar appears and before you can actually scroll. Do you know why that is by any chance?

marcus-oscarsson:
Thats a good idea, I like the idea with a gradient at the edge I think that most people understands that

matyson:
I don't have to click it to scroll, it was a missclick while recording sorry. But i think my mac does mask this scroll behaviour a bit, it hiddens it by default even when forcing the `overflow: scroll`. It should look like this in linux:

https://github.com/user-attachments/assets/488abd3d-808a-4fbe-bd49-1505be84efc3

I will try to cook something with the scrolling shadows, good idea.

matyson:
Okay, i've been experimenting some things and i think the "always on" scrollbar + `scroll-snap-align: center;` combined resemble the old carousel behaviour:

https://github.com/user-attachments/assets/6aaa6093-8f2b-40a7-9946-e27ecb885fd5

I tried the horizontal scroll shadow but the css-only didn't felt as good with different background colors for the selected/unselected sample states ( tbh im not very good with css gradients so could be a skill issue situation as well)

axelboc (approved):
Great!

marcus-oscarsson:
ok, thanks :)

jbflo:
is The styles being display consistently ???
It seems the Css changes look a bit odd in Chrome , Firefox and opera.

Chrome
Screenshot from 2025-11-24 15-48-35

opera
image


firefox
image

matyson:
> is The styles being display consistently ??? It seems the Css changes look a bit odd in Chrome , Firefox and opera.
> 
> Chrome Screenshot from 2025-11-24 15-48-35
> 
> opera image
> 
> firefox image

Well, i tried this extensively in my mac, but the way each browser implements the feature can be different. I even spin up a ubuntu VM to try it and it was indeed different from the mac one, but still similar. I don't know, perhaps natively on linux x86 can still be different. Also, mac seem to deal with the `overflow: scroll` its own way.

I appreciate that you noticed this, maybe i screwed it up :/

PR #1959, opened 2025-11-06
Fixing `set_image_size` request

marcus-oscarsson:
Changing the image size currently gives an error because the response is missing beam position and size. Big thanks to @Phantominumm for the hint :).

This PR adds image size and position back to the response. This is how it was used to be before moving to the `sample_view_adapter`.

I will try to follow up with a PR that decouples the beam size from the image size in the front end, so that this data is properly decoupled.

### DIFF
--- a/mxcubeweb/core/adapter/sample_view_adapter.py
+++ b/mxcubeweb/core/adapter/sample_view_adapter.py
@@ -209,6 +209,12 @@ class SampleViewAdapter(AdapterBase):
 
         pixels_per_mm = HWR.beamline.diffractometer.get_pixels_per_mm()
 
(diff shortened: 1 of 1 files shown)


mxcubeweb/core/adapter/sample_view_adapter.py
```diff
@@ -209,6 +209,12 @@ def sample_image_meta_data(self) -> dict:
 
         pixels_per_mm = HWR.beamline.diffractometer.get_pixels_per_mm()
 
+        # We need to get the beam size when changing the image
```
marcus-oscarsson, line 212: So this dependency is as you understand not necessary and we should remove it from the front end

PR #1960, opened 2025-11-06
Fixing take snapshot

marcus-oscarsson:
This fixes an issue introduced in the "take snapshot" function when switching to  `SampleViewAdapter`. The return data was handled as `json` instead of `blob` and the date returned was "empty" because of the missing `seek(0)`

### DIFF
--- a/ui/src/api/hardware-object.js
+++ b/ui/src/api/hardware-object.js
@@ -8,6 +8,10 @@ export function sendExecuteCommand(objectType, objectId, command, value) {
     .safeJson();
 }
 
+export function sendExecuteCommandRaw(objectType, objectId, command, value) {
+  return endpoint.put(value, `/${objectType}/${objectId}/${command}`);
+}
+
 export function fetchAttribute(objectType, objectId, attributeName) {
   return endpoint.get(`/${objectType}/${objectId}/${attributeName}`).safeJson();
 }
--- a/mxcubeweb/core/adapter/sample_view_adapter.py
+++ b/mxcubeweb/core/adapter/sample_view_adapter.py
@@ -251,10 +251,12 @@ class SampleViewAdapter(AdapterBase):
 
         try:
             image = self._ho.take_snapshot(overlay_data=overlay_data)
-            b = io.BytesIO()
-            image.save(b, "JPEG")
+            data = io.BytesIO()
+            image.save(data, "JPEG")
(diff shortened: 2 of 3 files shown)


ui/src/api/hardware-object.js
```diff
@@ -8,6 +8,10 @@ export function sendExecuteCommand(objectType, objectId, command, value) {
     .safeJson();
 }
 
+export function sendExecuteCommandRaw(objectType, objectId, command, value) {
```
marcus-oscarsson, line 11: I'm not super happy with the name `sendExecuteCommandRaw` but I did not come up with something better at the moment, please feel free to suggest :) ?

axelboc (commented):
I see, that's interesting. In my view, if `sendExecuteCommand` is no longer guaranteed to respond with JSON, then it shouldn't call `safeJson()`. This call should be moved out, for instance:

```js
export function sendSetValue(objectId, objectType, value) {
  return sendExecuteCommand(objectType, objectId, 'set_value', { value }).safeJson();
}

// or maybe even higher up, in specific hardware-object API files:
export function sendSetVideoSize(width, height) {
  return sendSampleViewCommand('set_image_size', {
    width,
    height,
  }).safeJson();
}
```

Then you can just call `blob()` inside `sendTakeSnapshot`:

```js
export function sendTakeSnapshot(canvasData) {
  return sendExecuteCommand('sample_view', 'sample_view', 'snapshot', {
    value: canvasData,
  }).blob();
}

// or even better:
export function sendTakeSnapshot(canvasData) {
  return sendSampleViewCommand('snapshot', {
    value: canvasData,
  }).blob();
}
```

marcus-oscarsson:
Hm, returning something else than `json` is quite uncommon, but it does happen (like in this case). I was thinking of renaming/structuring `sendExecuteCommand` but it would then also mean that we need to change everything that is using `sendExecuteCommand` to either use another function, add `.json()` or add a set/layer of helper functions for each "category" of API calls. 

I have to think about exactly how to do this. Ill follow up with a PR that does this re-structuring.

axelboc (approved):
:+1:

marcus-oscarsson:
I noticed that I miss placed the call to `.blob()` so I fixed that. I will merge this while thinking about how to better structure/abstract `sendExecuteCommand`.

PR #1961, opened 2025-11-06
Beam position not needed when image size changes

marcus-oscarsson:
The actual beam position does not change when the size of the video stream changes. Removed beam position from "SET_IMAGE_SIZE" action.


walesch-yan (approved):

marcus-oscarsson:
Thanks, Welcome back :)

PR #1962, opened 2025-11-10
introduced grid cache and mesh grid shape

mockoocy:
Currently grids are being redrawn when any property of `SampleImage` component changes. It leads to serious performance issues when very large grids are being drawn (over 100k cells) which is legitimate concern for MicroMAX users.

This change makes it so that the grids are being redrawn when any property of the `gridData` object changes, which greatly limits the amount of rerenders.
Each grid is being stored in a `gridCache` object which centralizes logic regarding adding and remoing grids.
Also logic regarding drawing grid cells has been moved to a custom fabricJS shape - `MeshGrid`. It is responsible for rendering the grid itself (without extra data like heatmap or cell labels).

It is pretty evident that additional improvements could be done to this code; both in the performance and code quality fronts. The shapes could still be reorganized, so that there is one that renders both the grid cells and, heatmap and so on to make things clearer. Also, moving motors still issues lots of costly re-renders. Nonetheless, it seems like a good starting point for some future changes.

### DIFF
--- a/ui/src/components/SampleView/DrawGridPlugin.jsx
+++ b/ui/src/components/SampleView/DrawGridPlugin.jsx
@@ -1,8 +1,228 @@
+/* eslint-disable max-classes-per-file */
+
 import 'fabric';
+
+import { isEqual } from 'lodash';
 
 const TAU = Math.PI * 2;
 
 const { fabric } = globalThis;
+
+/**
+ * @typedef {Object} CachedGrid
+ * @property {GridGroup} gridGroup - The Fabric.js group representing the grid.
+ * @property {GridData} gridData - The data used to define the grid.
+ */
+
+/**
+ * This class is responsible for caching grids and rendering them on a canvas.
+ * Drawing big grids can be slow, so we cache them to improve performance.
+ */
+export class GridCache {
+  /**
+   * @type {Map<string, CachedGrid>} - Map of grid id to CachedGrid.
+   */
+  grids = new Map();
+
+  /**
+   * Adds a grid to the cache and renders it on cache.
+   * If the grid has not changed, it will not be re-added.
+   *
+   * @param {GridData} gridData - The grid to add.
+   * @param {DrawGridPlugin} drawGridPlugin - A DrawGridPlugin instance.
+   * @param {number} pixelsPerMM - The number of pixels per millimeter.
+   * @param {number} scale - The scale of the grid.
+   * @returns {CachedGrid} the added grid.
+   */
+  addGrid(gridData, drawGridPlugin, pixelsPerMM, scale) {
+    const { id } = gridData;
+    const cachedGrid = this.getGrid(id);
+    const newGridData = drawGridPlugin.setPixelsPerMM(pixelsPerMM, gridData);
+    // We don't recreate the grid if it's already in the cache
+    // and the grid data is the same
+    const gridDataChanged =
+      cachedGrid && !isEqual(cachedGrid.gridData, newGridData);
+
+    if (cachedGrid && !gridDataChanged) {
+      return cachedGrid;
+    }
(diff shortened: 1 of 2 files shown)


ui/src/components/SampleView/DrawGridPlugin.jsx
```diff
@@ -634,8 +795,9 @@ export default class DrawGridPlugin {
       visible: true,
       hoverCursor: 'pointer',
       id: gd.id,
+      top,
+      left,
```
mockoocy, line 799: I've had encountered some bugs with positioning of `GridGroup`, `MeshGrid` and `MeshGridResult` shapes.
So there are some seemingly random changes like these.
The idea is that the `GridGroup` sets the `top, left` properties, instead having its children set them independently. It has led to need of some other position-related changes above.

axelboc (commented):
I really like the new `MeshGrid` object. It really cleans things up nicely. :clap: 

I have to say that I'm a bit less convinced about the `GridCache` class. I feel like there should be a more Reacty way to avoid the unnecessary re-renders/redraws you mention, maybe with memoisation, [refs](https://react.dev/learn/referencing-values-with-refs) or other escape hatches. Given the amount of code `GridCache` adds, I think it's worth a second look, which I'm happy to do if you like?

Would you consider splitting your `MeshGrid` work into a dedicated PR in the meantime? That part is good to go for sure.

marcus-oscarsson:
Thanks @mockoocy. That's a nice initiative. I also like the idea of the `MeshGrid`. I'm as @axelboc not entirely convinced the the `GridCache` is the right way to go. I also think that this is better dealt with by using what React already provides. It may mean splitting things into smaller components and moving away from the `DrawGridPlugin`.

I would also recommend renaming `MeshGrid` to either simply `Mesh` or `Grid` or similair.

marcus-oscarsson:
I also wanted to take the opportunity to tell you that I think that there are now several people working on the sample view and "shapes". So its extra important that you create an issue before starting some work related to the sample view. @mockoocy, @matyson, @walesch-yan

mockoocy:
> I really like the new `MeshGrid` object. It really cleans things up nicely. :clap: 
> 
> I have to say that I'm a bit less convinced about the `GridCache` class. I feel like there should be a more Reacty way to avoid the unnecessary re-renders/redraws you mention, maybe with memoisation, [refs](https://react.dev/learn/referencing-values-with-refs) or other escape hatches. Given the amount of code `GridCache` adds, I think it's worth a second look, which I'm happy to do if you like?
> 
> Would you consider splitting your `MeshGrid` work into a dedicated PR in the meantime? That part is good to go for sure.



I'm actually not super happy with the `GridCache` either, but I was not sure about how to do it in a React-friendly way without touching too much stuff. Any suggestions are really welcome.

For splitting of `MeshGrid` - I'll do it on Wednesday. I'm leaving work for today, and tomorrow's a holiday in Poland :)

marcus-oscarsson:
Great thanks @mockoocy, and an important holiday as well (same as in France I imagine, "Armstice" :) )

mockoocy:
The `MeshGrid` (renamed to just `Grid`) has been moved to a separate [PR](https://github.com/mxcube/mxcubeweb/pull/1963)

I won't close this PR, as it will serve to actually add some optimization for the grid lines / rewrite of the `GridCache` along the way.

mockoocy:
One solution that I am wondering about is that if we could fabric's [object caching](https://fabricjs.com/docs/fabric-object-caching/) somehow.
Currently all the shapes are being cleared and then added again in `componentDidUpdate`, which is not ideal.

mockoocy:
This one got stale, so I am closing it. 
Perhaps I'll come back to it on some later occasion :)

PR #1963, opened 2025-11-12
Add Grid shape

mockoocy:
Logic regarding grids has been moved to a separate shape. Also adds some type declarations.

Derived from https://github.com/mxcube/mxcubeweb/pull/1962

### DIFF
--- a/ui/src/components/SampleView/DrawGridPlugin.jsx
+++ b/ui/src/components/SampleView/DrawGridPlugin.jsx
@@ -5,6 +5,110 @@ const TAU = Math.PI * 2;
 const { fabric } = globalThis;
 
 /**
+ * @typedef {Object} GridOptions
+ * @property {number} cellRows - The number of rows in the grid.
+ * @property {number} cellCols - The number of columns in the grid.
+ * @property {number} cellWidth - The width of each cell in the grid.
+ * @property {number} cellHeight - The height of each cell in the grid.
+ * @property {number} cellHSpace - The horizontal space between cells.
+ * @property {number} cellVSpace - The vertical space between cells.
+ * @property {string} color - Color of the grid lines.
+ * @property {number} cellTW - The total width of each cell, including the horizontal space.
+ * @property {number} cellTH - The total height of each cell, including the vertical space.
+ * @property {boolean} [selected = false] - Whether the grid is selected.
+ */
+/**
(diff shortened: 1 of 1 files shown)


axelboc (approved):
Brilliant, thanks a lot! :100:

ui/src/components/SampleView/DrawGridPlugin.jsx
```diff
@@ -41,10 +145,14 @@ const GridGroup = fabric.util.createClass(fabric.Group, {
  *
  */
 const MeshGridResult = fabric.util.createClass(fabric.Object, {
-  type: 'GridGroup',
+  type: 'MeshGridResult',
```
marcus-oscarsson: If we call the object `Grid` then this should perhaps be `GridResult` ?

ui/src/components/SampleView/DrawGridPlugin.jsx
```diff
@@ -521,100 +585,81 @@ export default class DrawGridPlugin {
...
-        top,
-        height,
-        width,
-        cellTH,
-        cellTW,
-      );
-      if (!this.drawing) {
-        if (this.gridResultFormat === 'RGB') {
-          const fillingMatrix = this.cellFillingFromData(
+    const meshGrid = new Grid({
```
marcus-oscarsson: `const grid` ?

marcus-oscarsson:
Looks really nice thanks alot !. 

I would be a bit careful with the wording `MeshGrid` and `Grid`. It seems like we are using `Grid` quite consistently i.e `GridData` and `Grid` but there is sometimes still `MeshGrid`. Its perhaps worth looking at. Perhaps most importantly `MeshGridResult` but I spotted `MeshGrid` also in the comments.

mockoocy:
> Looks really nice thanks alot !.
> 
> I would be a bit careful with the wording `MeshGrid` and `Grid`. It seems like we are using `Grid` quite consistently i.e `GridData` and `Grid` but there is sometimes still `MeshGrid`. Its perhaps worth looking at. Perhaps most importantly `MeshGridResult` but I spotted `MeshGrid` also in the comments.

I've replaced all mentions of a "mesh grid" with "grid" now. Should be more consistent now :)

marcus-oscarsson:
That was fast, very nice :)

marcus-oscarsson (approved):

PR #1972, opened 2025-11-27
`move_to_beam` is on Diffratcomter not on SampleView

marcus-oscarsson:
The method `move_to_beam` is on the Diffratctomter hardware object not on `SampleView`


beteva:
For time being. It will be in SampleView HO with the new diffractometer.

PR #1973, opened 2025-11-27
Switch to Redux hooks in `EquipementContainer` and children

axelboc:
Two separate commits: 

1. First I switch to Redux hooks in `EquipmentContainer`
2. Then I move the `useSelector` and `dispatch` calls down into the `Harvester` and `SampleChanger` components to avoid passing props.

### DIFF
--- a/ui/src/components/Equipment/Harvester.jsx
+++ b/ui/src/components/Equipment/Harvester.jsx
@@ -1,52 +1,41 @@
-/* eslint-disable react/destructuring-assignment */
 import React from 'react';
 import { Badge, Button, Card } from 'react-bootstrap';
 import { contextMenu, Item, Menu, Separator } from 'react-contexify';
 import { FcCollect, FcRefresh, FcUpload } from 'react-icons/fc';
+import { useDispatch, useSelector } from 'react-redux';
 
+import {
+  harvestAndLoadCrystal,
+  harvestCrystal,
+  refresh,
+} from '../../actions/harvester.js';
 import CopyToClipboard from '../CopyToClipboard/CopyToClipboard.jsx';
 import ImageViewer from '../ImageViewer/ImageViewer.jsx';
 import { sampleStateBackground } from '../SampleGrid/util.js';
 import styles from './equipment.module.css';
 
-export default function Harvester(props) {
-  function showContextMenu(event, id) {
-    let position = {
-      x: event.clientX,
-      y: event.clientY,
-    };
-    if (props.inPopover) {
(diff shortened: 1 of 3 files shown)


ui/src/components/Equipment/Harvester.jsx
```diff
@@ -1,52 +1,41 @@
...
+  harvestCrystal,
+  refresh,
+} from '../../actions/harvester.js';
 import CopyToClipboard from '../CopyToClipboard/CopyToClipboard.jsx';
 import ImageViewer from '../ImageViewer/ImageViewer.jsx';
 import { sampleStateBackground } from '../SampleGrid/util.js';
 import styles from './equipment.module.css';
 
-export default function Harvester(props) {
-  function showContextMenu(event, id) {
```
axelboc, line 13: I've also inlined a few callbacks.

ui/src/components/Equipment/SampleChanger.jsx
```diff
@@ -76,19 +85,22 @@ function renderSampleChangerTreeNode(props) {
...
+      mountSample({
+        sampleID: props.label,
+        location: props.label,
+      }),
+    );
   }
 
   function handleUnmountClick() {
     toggleDropdown();
-    props.unload(props.label);
```
axelboc, line 91: Note that the label was not actually passed to the `unmountSample` action — cf. https://github.com/mxcube/mxcubeweb/pull/1973/files#diff-1f863941734e9265496f0d863d76db18f63b7fb4e0cda13756d54fffbdb697ddL138

If this was a mistake and it should be passed, please let me know.

ui/src/components/Equipment/SampleChanger.jsx
```diff
@@ -131,21 +143,12 @@ function SampleChangerTreeItem(props) {
...
 // eslint-disable-next-line react/no-multi-comp
-export default function SampleChanger(props) {
-  function scan() {
-    props.scan('');
-  }
+export default function SampleChanger() {
+  const dispatch = useDispatch();
 
-  function unload() {
-    props.unload('');
```
axelboc, line 140: Same here, this empty string was not actually passed to the `unmountSample` action.

ui/src/components/Equipment/SampleChanger.jsx
marcus-oscarsson: It should not be passed :)

marcus-oscarsson:
Nice, Thanks !

PR #1977, opened 2025-12-01
Inline JSX in `Harvester` and `SampleChanger` components

axelboc:
Plus a couple other refactorings:

- I turn the `renderSampleChangerTreeNode` function into a proper component.
- I destructure a few more variables, inline a few event handlers, etc.

I'm planning one or two more refactorings in `SampleChanger` specifically.

Issue #1844, opened 2025-09-01
Replace `react-contextify` with Bootstrap for context menus


marcus-oscarsson:
The library `react-contextify` is used for some context menus. This library can be replaced by using only `Bootstrap`. See #1430 for more information (original issue).

PR #1978, opened 2025-12-01
Make better use of Bootstrap in `SampleChanger` equipment

axelboc:
This is based off #1977 so I'll rebase manually and mark it as ready once #1977 is merged.

### BEFORE

- The code of `SampleChangerTreeNode` was particularly outdated. It made use of a hidden checkbox for toggling the tree items, used `react-contextify` to implement a very hard-to-discover context menu, etc.
- While comparing the code and the UI, I also noticed a few bugs, like the "<- Mounted" text and the "Abort" button not appearing when they should.
- The UI was not particularly pretty or usable (layout shift while mounting a different sample, link-like text for dropdown toggles, etc.)

[Screencast from 2025-12-01 15-05-43.webm](https://github.com/user-attachments/assets/11834965-0f91-487a-9a8d-8aa353907c50)

### AFTER

- I make better use of Bootstrap overall to improve the UI, remove a bunch of custom/inline styles, and remove one occurrence of `react-contextify` (cf. #1844 #1906).
- I use a proper local React state to manage the expanded state of each tree node.
- I fix the bug with the "<- Mounted" text and "Abort" button.
- I improve the UX slightly by:
  - disabling the drop-downs while mounting a sample;
  - removing the "Mount" option for an already-mounted sample;
  - showing a fallback text when no sample is mounted.

[Screencast from 2025-12-01 15-03-43.webm](https://github.com/user-attachments/assets/c08b0649-649a-4ddb-aadd-967d3f0cba77)

### DIFF
--- a/ui/src/components/Equipment/SampleChanger.jsx
+++ b/ui/src/components/Equipment/SampleChanger.jsx
@@ -1,7 +1,5 @@
-/* eslint-disable jsx-a11y/control-has-associated-label */
 import { useState } from 'react';
-import { Button, Card, Dropdown, DropdownButton, Form } from 'react-bootstrap';
-import { contextMenu, Item, Menu, Separator } from 'react-contexify';
+import { Alert, Button, ButtonGroup, Card, Dropdown } from 'react-bootstrap';
 import { useDispatch, useSelector } from 'react-redux';
 
 import {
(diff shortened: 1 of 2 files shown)


ui/src/components/Equipment/SampleChanger.jsx
```diff
@@ -14,127 +12,107 @@ import {
...
-  }
-}
-
 function SampleChangerTreeNode(props) {
-  const { label, children } = props;
+  const { label, disabled, children } = props;
 
   const dispatch = useDispatch();
-  const inputId = getUniqueId();
+  const [expanded, setExpanded] = useState(false);
```
axelboc, line 19: The new local state that replaces the hidden checkbox, the `.treeNode:checked ~ ul` style, the `treeNodeCbxClick` function above, and more.

ui/src/components/Equipment/SampleChanger.jsx
```diff
@@ -14,127 +12,107 @@ import {
...
-  function handleUnmountClick() {
-    toggleDropdown();
-    dispatch(unmountSample());
-  }
-
-  function toggleDropdown() {
-    setDropdownIsOpen(!dropdownIsOpen);
-  }
-
-  const ls = status === 'Loaded' ? { display: 'inline' } : { display: 'none' };
```
axelboc, line 99: `status === 'Loaded'` is basically replaced with `loadedSample?.address === label` (cf. `isMounted`)

ui/src/components/Equipment/SampleChanger.jsx
```diff
@@ -14,127 +12,107 @@ import {
...
   );
 }
 
 // eslint-disable-next-line react/no-multi-comp
 function SampleChangerTreeItem(props) {
-  const { label, status, dm } = props;
+  const { label, dm, disabled } = props;
   const dispatch = useDispatch();
 
-  const [dropdownIsOpen, setDropdownIsOpen] = useState(false);
```
axelboc, line 79: Bootstrap's `Dropdown` actually manages its own state very well so there's no need to control it.

ui/src/components/Equipment/SampleChanger.jsx
```diff
@@ -14,127 +12,107 @@ import {
...
-              <Dropdown.Item onClick={handleUnmountClick}>
-                Umount this position
-              </Dropdown.Item>
-            )}
-          </DropdownButton>
-          <span style={ls}>
-            &nbsp;
-            <i className="fas fa-arrow-left" /> <b>(Mounted)</b>
+    <li className={styles.treeItem}>
+      <Dropdown className="d-inline-flex align-items-center mt-1">
```
axelboc, line 71: I replace `DropdownButton` with the expanded `Dropdown` syntax because I wanted to put the "<- Mounted" text inside the drop-down's `div` wrapper.

ui/src/components/Equipment/SampleChanger.jsx
```diff
@@ -14,127 +12,107 @@ import {
...
-      </Menu>
-    </div>
+        </Button>
+
+        <Dropdown.Toggle
+          id={`sample_changer_tree_node_${label}`}
+          split
+          variant="secondary"
+          disabled={disabled}
+        />
```
axelboc, line 39: I replace the context menu with a drop-down using Bootstrap's `ButtonGroup` and this "split" drop-down toggle.

ui/src/components/Equipment/SampleChanger.jsx
```diff
@@ -169,60 +147,53 @@ function SampleChanger() {
...
-          onClick={() => {
-            dispatch(scan(''));
-          }}
+          onClick={() => dispatch(scan(''))}
         >
           <i className="fas fa-qrcode" /> Scan all containers
         </Button>
 
-        <span style={{ marginLeft: '1em' }}>
-          {scState === 'MOVING' && (
```
axelboc, line 191: Now checking for `'LOADING'` instead of `'MOVING'`. I hope that's right?

mockoocy (approved):
Great stuff, even the keyboard-based navigation on the tree works well there :D

marcus-oscarsson:
Thanks, very nice. Its a nice improvement. Aesthetically speaking it was already not very nice but I think this aspect can still be improved. Its functional as it is so I think we can go a head as this includes some nice updates :)

PR #1979, opened 2025-12-02
Turn sample changer tree into a component

axelboc:
Last but not least, I convert the `renderTree` function into a recursive React component.

### DIFF
--- a/ui/src/components/Equipment/SampleChanger.jsx
+++ b/ui/src/components/Equipment/SampleChanger.jsx
@@ -107,6 +107,35 @@ function SampleChangerTreeItem(props) {
 }
 
 // eslint-disable-next-line react/no-multi-comp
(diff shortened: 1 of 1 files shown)


ui/src/components/Equipment/SampleChanger.jsx
```diff
@@ -107,41 +107,44 @@ function SampleChangerTreeItem(props) {
...
-          disabled={isLoading}
-        >
-          {node.children.map(renderTree)}
-        </SampleChangerTreeNode>
-      );
-    }
+function SampleChangerTree(props) {
+  const { root = false, node, isLoading } = props;
 
+  if (!node.children) {
```
axelboc, line 113: Note that I invert the logic so that the early return has the least code.

marcus-oscarsson:
Thanks !

PR #1982, opened 2025-12-02
Improve queue and logs layout

axelboc:
### BEFORE

- The motors and queue sidebars are constrained by the height of the sample image.
-  The logs are flush against the queue so look like they are part of the queue tabs' contents.
- The active queue tab doesn't look quite right because of the bottom border.
- There's an annoying white gap at the bottom of the logs above a certain screen size.
- The general design of the queue/logs sidebar is a bit heavy (dark borders, heavy separators between logs, etc.)
- The CSS of the queue/logs sidebar has a bunch of fixed percentage-based heights and absolutely positioned elements.

image

The annoying white gap at the bottom of the logs:

image

### AFTER

- The motors and queue sidebars grow to fill the available vertical space.
- The logs are spaced out from the queue.
- The active queue tab looks right.
- The white gap at the bottom of the logs is no longer present.
- The general design is a bit cleaner and lighter. Note, however, that I didn't go as far as to rethink the design of the logs and queue items. I focused more on the layout.
- The CSS of the queue/logs sidebar is a lot simpler and more straightforward.

[Screencast from 2025-12-02 15-03-43.webm](https://github.com/user-attachments/assets/49bfa500-1a2f-4081-8f59-ab2a670b85e8)

### DIFF
--- a/ui/src/components/Main.module.css
+++ b/ui/src/components/Main.module.css
@@ -7,7 +7,7 @@
 
 .main {
   flex: 1;
-  display: grid;
+  display: flex;
 }
 
 .readOnly {
--- a/ui/src/components/Main.module.css
+++ b/ui/src/components/Main.module.css
@@ -1,6 +1,13 @@
-.main {
-  flex: 1 1 0%;
+.root {
+  flex: 1;
   min-width: 0;
+  display: flex;
+  flex-direction: column;
+}
+
+.main {
+  flex: 1;
+  display: grid;
 }
 
 .readOnly {
--- a/ui/src/components/Notify/UserMessage.module.css
+++ b/ui/src/components/Notify/UserMessage.module.css
@@ -1,35 +1,39 @@
 .message {
-  position: relative;
-  padding: 0.4em;
-  opacity: 1;
+  padding: 0.25rem 0.75rem;
+  border-color: rgb(var(--user-message-color));
+  background-color: rgba(var(--user-message-color), 0.2);
 }
 
-.message:not(:first-child) {
-  border-top: 2px solid #333;
+.message:first-child {
+  padding-top: 0.5rem;
 }
 
-.INFO {
-  border-color: #4e677b;
-  background-color: rgba(91, 141, 180, 0.8);
+.message:last-child {
+  padding-bottom: 0.5rem;
 }
 
-.WARNING {
-  border-color: #2196f3;
-  background-color: rgba(240, 173, 78, 0.8);
+.message[data-severity='INFO'] {
+  --user-message-color: 91, 141, 180;
 }
 
-.ERROR {
-  border-color: #d9534f;
-  background-color: rgba(217, 83, 79, 0.8);
+.message[data-severity='WARNING'] {
+  --user-message-color: 240, 173, 78;
 }
 
-.DEBUG {
-  border-color: #d9534f;
-  background-color: rgba(217, 83, 79, 0.8);
+.message[data-severity='ERROR'] {
+  --user-message-color: 217, 83, 79;
 }
 
-.messageText {
-  color: #fff;
-  margin-left: 7px;
-  font-weight: bold;
+.message[data-severity='DEBUG'] {
+  --user-message-color: 217, 83, 79;
+}
+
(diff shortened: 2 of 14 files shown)


ui/src/components/Main.module.css
```diff
@@ -1,6 +1,13 @@
...
+.root {
+  flex: 1;
   min-width: 0;
+  display: flex;
+  flex-direction: column;
+}
+
+.main {
+  flex: 1;
+  display: flex;
```
axelboc, line 10: All pages have a Bootstrap fluid `Container` as unique child. The code above ensures that those containers fill the entire space available to them in both width and height. Note that fluid containers have `width: 100%`, so we don't need to set `flex: 1` on them.

ui/src/components/Notify/UserMessage.module.css
```diff
@@ -1,37 +1,30 @@
...
 .message {
   position: relative;
   padding: 0.4em;
-  border-color: #333;
   opacity: 1;
 }
 
+.message:not(:first-child) {
+  border-top: 2px solid #333;
+}
```
axelboc: Change border from 5 to 2px, and remove top border from the first log message.

marcus-oscarsson:
Awesome, as always super nice :). I'm wondering if we should allow the content to grow below the video. It looks sort of weird and it would perhaps be nice to use that entire space for something else (not sure at the moment). I do think that having "longer" content on the sides of the video looks strange

axelboc:
> I'm wondering if we should allow the content to grow below the video. It looks sort of weird and it would perhaps be nice to use that entire space for something else (not sure at the moment). I do think that having "longer" content on the sides of the video looks strange

To me it looks weirder when the sidebar is cropped to the height of the video. :sweat_smile: I think it's a matter of getting used to it. Either way, it's a shame not to use the available space, since the queue can be content-heavy in real life. Note that the space below the sample image depends on the width/height ratio of the browser window, so it may not be that big on some wide monitors.

axelboc:
I've pushed one more commit to improve the design of the logs:

- smaller icons, lighter background color and black text for the log messages;
- fallback text when there are no logs;
- subtle border on container;
- container doesn't take up space until there are logs (up to a specified max height of `25vh`).

[Screencast from 2025-12-03 15-27-33.webm](https://github.com/user-attachments/assets/ea41e89a-b6aa-462d-99c1-0576afc230c9)

@marcus-oscarsson regarding the sidebar going past the bottom of the sample image, if you really don't like it I can easily revert it. Just let me know :wink:

marcus-oscarsson:
I really love the new log look ! :1st_place_medal: Super nice. I think these sort of details make a huge impact :) 

I personally prefer when the layout is symmetric so that that log ends at the same level as the rest of the interface. I agree that it looks strange when it gets cut.  In general I think we can do better, how about moving the entire log under the video ? We could think about having a nice widget which can be expanded or toggled one/multi line ?

marcus-oscarsson:
Its perhaps not for this PR but I think its worth thinking about and perhaps doing relatively soon. What do you think @ddesanctis

marcus-oscarsson:
Ok. So ill merge this as it is for the moment as it contains lots of nice improvements. We should then take some time to work on how the logs are displayed

DdSID29:
hi @axelboc it looks great, I like how much lighter it looks and less packed. I have two comments:
- it seems to me that the space between queue entries is a bit too much, and maybe we can also make the queue entries a bit slimmer?
- I wonder what happens when you resize the window, in the first video it does not rescale and in fact a scroll bar appears and it is something we would like to avoid IMO

what do you think?

axelboc:
Thanks! Yeah, there's more to do in the queue itself, I agree. :+1:

The minimum height of the content (more specifically of the sample image) depends on the _width_ of the window. When you reduce the height of the window, at some point you reach this minimum content height and that's when the vertical scrollbar appears. Changing this beahviour would be quite tricky and not necessarily a good idea.

To avoid the vertical scrollbar, you can either reduce the width of the window proportionately or zoom out in your browser. Of course, there's still work to do to make the interface responsive in width.

PR #1983, opened 2025-12-02
Convert `SampleQueueContainer` and `CurrentTree` to function components

axelboc:
... and use Redux hooks API as always and destructure props. No logic or UI changes here. I'm not even moving props down or anything like that. I only renamed a couple of functions/variables to avoid conflicts and for clarity.

As always I'll rebase and mark as ready once #1982 is merged, and I recommend hiding white-space changes when reviewing :grin:

### DIFF
--- a/ui/src/components/SampleQueue/CurrentTree.jsx
+++ b/ui/src/components/SampleQueue/CurrentTree.jsx
@@ -1,5 +1,3 @@
-/* eslint-disable react/destructuring-assignment */
-import React from 'react';
 import { Item, Menu } from 'react-contexify';
 
 import CharacterisationTaskItem from './CharacterisationTaskItem';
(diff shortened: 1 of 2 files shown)


ui/src/containers/SampleQueueContainer.jsx
```diff
@@ -16,164 +13,116 @@ import TodoTree from '../components/SampleQueue/TodoTree';
...
+          Log messages:
         </div>
-      </Stack>
-    );
-  }
-}
-
-function mapStateToProps(state) {
-  return {
-    searchString: state.queueGUI.searchString,
```
axelboc, line 139: `searchString` and `selectedShapes` were unused.

ui/src/components/SampleQueue/CurrentTree.jsx
```diff
@@ -59,173 +59,165 @@ export default class CurrentTree extends React.Component {
...
-                    sampleId={sampleData.sampleID}
-                    checked={this.props.checked}
-                    showForm={this.props.showForm}
-                    shapes={this.props.shapes}
-                    showDialog={this.props.showDialog}
-                  />
-                );
-              }
-            }
+  if (sampleId) {
```
marcus-oscarsson, line 98: I guess this in principle could go below the `if` but It does not seem like it was like that before

ui/src/containers/SampleQueueContainer.jsx
marcus-oscarsson: Yes, there used to be a search (it no longer exists :) )

marcus-oscarsson:
Looks, good. Thanks !

ui/src/components/SampleQueue/CurrentTree.jsx
axelboc: Ha, yes. I'm getting rid of the `show` prop anyway in #1984

marcus-oscarsson: Good point

marcus-oscarsson:
:+1: Thanks !

PR #1984, opened 2025-12-03
Move props down and refactor `CurrentTree`

axelboc:
Apart from moving `useSelector`/`dispatch()` down and renaming/destructuring a few variables, I simplify the code of `SampleQueueContainer`, `CurrentTree` and `TodoTree` a little as well (I'll explain in comments).

I'll stop here for the time being.

### DIFF
--- a/ui/src/containers/SampleQueueContainer.jsx
+++ b/ui/src/containers/SampleQueueContainer.jsx
@@ -1,11 +1,7 @@
 import { Nav, Stack } from 'react-bootstrap';
 import { useDispatch, useSelector } from 'react-redux';
 
-import { showDialog } from '../actions/general';
-import { addTask } from '../actions/queue';
 import { showList } from '../actions/queueGUI';
-import { showTaskForm } from '../actions/taskForm';
-import { showWorkflowParametersDialog } from '../actions/workflow';
 import UserMessage from '../components/Notify/UserMessage';
 import CurrentTree from '../components/SampleQueue/CurrentTree';
 import QueueControl from '../components/SampleQueue/QueueControl';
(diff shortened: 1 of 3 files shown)


ui/src/containers/SampleQueueContainer.jsx
```diff
@@ -16,42 +12,22 @@ import styles from './SampleQueueContainer.module.css';
...
-  if (currentSampleID) {
-    const sampleData = sampleList[currentSampleID] || {};
-    sampleName = sampleData.sampleName || '';
-    proteinAcronym = sampleData.proteinAcronym
-      ? `${sampleData.proteinAcronym} -`
-      : '';
-  }
+  const currentSample = currentSampleID
+    ? sampleList[currentSampleID]
+    : undefined;
```
axelboc, line 30: Previously, we were passing down `currentSampleID` to `CurrentTree` and accessing the current sample's object/data there (with `sampleList[currentSampleID`). Now, I access the current sample's object here and pass it down. This allows me to:

- simplify the rendering logic of the "Current" tab (`sampleName` and `proteinAcronym`);
- render `CurrentTree` only if there is a sample mounted, ensuring that I have an always-defined `currentSample` object in `CurrentTree` (which in turns simplifies a bunch of logic).

ui/src/containers/SampleQueueContainer.jsx
```diff
@@ -16,42 +12,22 @@ import styles from './SampleQueueContainer.module.css';
...
-  const shapes = useSelector((state) => state.shapes);
 
-  // go through the queue, check if sample has been collected or not
-  // to make todo lists
-  const todo = [];
+  // Find samples in the queue that have not yet been collected
+  const todo = sampleOrder
+    .filter((id) => queue.includes(id))
+    .map((id) => sampleList[id])
+    .filter((sample) => sample.sampleID !== currentSampleID && sample.checked);
```
axelboc, line 26: I now store sample objects instead of IDs in the `todo` array that is passed down to `TodoTree`. This avoids having to retrieve `sampleList` from the Redux store in `TodoTree`.

axelboc:
NPM was down for a few minutes. All good now but not sure how to retrigger the readthedocs build.

ui/src/containers/SampleQueueContainer.jsx
marcus-oscarsson: Good idea 👍

ui/src/containers/SampleQueueContainer.jsx
marcus-oscarsson: Very nice !

PR #1989, opened 2025-12-08
Sync samples with Lims on proposal selection

### LINKED ISSUE #1981: Sync Samples with Lims on proposal selection
Can we sync samples from Lims right after selecting proposal? 

From the discussions in #1663 the sync from Lims was implemented but we still get the all samples from SC on starting up the application and have to manually sync it in the Sample page, which can expose sensitive data to unauthorized users. I think the ideal would be to prefetch the samples from lims on selecting proposal or, if not possible, on mounting/navigating to Sample page.

I can do it, but want some feedback if it make sense for you as well. What do you think?

matyson:
Closes #1981 . I chose to handle the sync directly in SelectProposal dialog since both `selectProposal`and `getLimsSamples` actions were already implemented.

### DIFF
--- a/ui/src/components/LoginForm/SelectProposal.jsx
+++ b/ui/src/components/LoginForm/SelectProposal.jsx
@@ -7,6 +7,7 @@ import {
   selectProposal,
   signOut,
 } from '../../actions/login';
+import { getLimsSamples } from '../../actions/sampleGrid';
 import ActionButton from './ActionButton';
 import SessionTable from './SessionTable';
 import styles from './SessionTable.module.css';
@@ -21,8 +22,13 @@ function SelectProposal() {
   const dispatch = useDispatch();
 
   const login = useSelector((state) => state.login);
-  const { loginType, proposalList, selectedProposalID, showProposalsForm } =
-    login;
+  const {
+    loginType,
+    proposalList,
+    selectedProposalID,
+    showProposalsForm,
+    limsName,
+  } = login;
 
   const show =
     showProposalsForm || (loginType === 'User' && selectedProposalID === null);
@@ -43,6 +49,11 @@ function SelectProposal() {
 
   const defaultActiveTab =
     TAB_PRIORITY.find((tabId) => showTab[tabId]) ?? TAB_PRIORITY[0];
+
+  /**
+   * @type {boolean}
+   */
+  const autosyncLims = useSelector((state) => state.general.autosyncLims);
 
   const filteredSessions = proposalList.filter(
     ({ title, number, code }) =>
@@ -69,6 +80,14 @@ function SelectProposal() {
       dispatch(signOut());
     } else {
       dispatch(hideProposalsForm());
+    }
+  }
+
+  function handleSelectProposal(sessionId) {
+    dispatch(selectProposal(sessionId));
+
+    if (autosyncLims) {
+      dispatch(getLimsSamples(limsName[0].name));
     }
   }
 
@@ -159,7 +178,7 @@ function SelectProposal() {
       <Modal.Footer>
         <ActionButton
           selectedSession={selectedSession}
-          onClick={() => dispatch(selectProposal(selectedSessionId))}
+          onClick={() => handleSelectProposal(selectedSessionId)}
         />
         <Button
           variant="outline-secondary"
--- a/demo.yaml/mxcube-web/server.yaml
+++ b/demo.yaml/mxcube-web/server.yaml
@@ -38,6 +38,13 @@ mxcube:
   # by the client application.
   SESSION_REFRESH_INTERVAL: 9000
 
+  # This is used to determine whether we can get samples from the sample changer
+  # if false, Only LIMS samples will be available to fetch for the sample list
+  USE_GET_SAMPLES_FROM_SC: true
+
+  # Sync samples with LIMS on proposal selection
+  AUTOSYNC_LIMS: false
+
   usermanager:
     class: UserManager
     inhouse_is_staff: true
--- a/demo/mxcube-web/server.yaml
+++ b/demo/mxcube-web/server.yaml
@@ -71,9 +71,12 @@ mxcube:
   # Mode, SSX-CHIP, SSX-INJECTOR, OSC
   mode: OSC
 
-  USE_GET_SAMPLES_FROM_SC: true
   # This is used to determine whether we can get samples from the sample changer
-  # if false, Only LIMS samples will be availble to fetch for the sample list
+  # if false, Only LIMS samples will be available to fetch for the sample list
+  USE_GET_SAMPLES_FROM_SC: true
+
+  # Sync samples with LIMS on proposal selection
+  AUTOSYNC_LIMS: false
 
   # Interval with which request to refresh token is made
   # by the client application.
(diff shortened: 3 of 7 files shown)


axelboc (approved):

marcus-oscarsson:
Yes, I guess that makes sense. The only sort of problem I see is that the user never chooses which LIMS to synchronize with. I guess is not really an issue, (@antolinos what do you think ?). In the case of ISPyB the user can simply make another specific sync

mockoocy (approved):

antolinos:
I believe it is a good idea and will work for many of you, but I wonder whether it should be configurable. In the ESRF case, I see a couple of drawbacks:
- People belonging to a proposal use multiple LIMS, often simultaneously. Some pucks are registered in ISPyB and others in DRAC, and users then end up synchronizing twice. This happens especially in BAGs, where some people are moving to DRAC while others are still using ISPyB. One could argue that once we phase out ISPyB this will no longer be a problem, which I agree with. However, at some point everybody will abandon ISPyB (or whatever lims is currently used), and this dual mechanism will be very useful to keep.
- Synchronization is fast… but not always. Now we can associate search models and reference files with crystals, and this data needs to be downloaded during synchronization. It may be slow (a few seconds). I think this is acceptable during sample synchronization but not when selecting the proposals.
- You might want to collect samples manually or perform XRF or energy scans; in these cases, you don’t need to synchronize pucks. So why force it?

For the future, these are ~requirements~  wishes hat I think will need to be implemented in a near future and having a specific button for synch might help:
- You might want to synchronize only certain samples. Example: synchronize only uncollected samples 
- You might want to put in the SC pucks from multiple proposals and collect them all together

These are just some ideas from an external point of view and are very biased by the way I see things

matyson:
I could add a configuration to enable autosync on proposal selection to keep the current behaviour as is, but i think even this won't attend the use cases above for multiple synching. In particular, we also have interest in multiple proposal selection, but i think would need to understand better how the synching between proposals/lims should be performed because even now the current "Sync button" in samples page only reproduces the same function call as listed in this PR. 

So, should we have more than one `limsName` available it should be selectable somewhere in the UI.

antolinos:
Hi again,

We currently have the option to choose the LIMS in the syncing button:
Screenshot from 2025-12-10 15-50-51

Do you refer to this?

marcus-oscarsson:
It would be nice if it was configurable to sync on session selection for the time being. I think its something we would use but our current use cases are quite complicated and it would be nice to be able to opt out if needed. 

It would be nice if it could be something like: `None`, `ispyb`, `drac` 

Its true that this operation may take some time, so it would be really useful with a busy/loading screen for the proposal selection and sync process.

matyson:
> Hi again,
> 
> We currently have the option to choose the LIMS in the syncing button: Screenshot from 2025-12-10 15-50-51
> 
> Do you refer to this?

Ohh, you're right! I just misread it, my bad. So multiple lims samples as a config could be something like:

```yaml
autosync_lims:
  - ispyb
  - drac
```

Not sure about the proper naming conventions for these kind of configs. Also, for multiple lims are we merging the sample list? I only glanced at this [ISPyBAbstractLims](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/HardwareObjects/abstract/ISPyBAbstractLims.py#L182) and since it just gets from one proposal at a time i'm assuming the facility specific ones merge the sample list..

marcus-oscarsson:
That would work, at the moment we only sync with one at the time but we can keep the option to have two if ever needed

marcus-oscarsson:
Just a question, does this really solve #1981 completely. The users can still retrieve the samples from the SC on the sample list view, see screen shot from @antolinos above. In that case it wont matter if the lims samples are retrieved at proposal selection, right ?

Would it not then make more sense to either remove or change the behavior of `Get samples from sc`. In your case you actually only need one option in the sample list view `Get samples` right ?

Just some thoughts after have talked to @antolinos

jbflo:
> Just a question, does this really solve #1981 completely. The users can still retrieve the samples from the SC on the sample list view, see screen shot from @antolinos above. In that case it wont matter if the lims samples are retrieved at proposal selection, right ?
> 
> Just some thoughts after have talked to @antolinos

@marcus-oscarsson
There is a config flag that hides the “Get samples from SC” option in the UI: USE_GET_SAMPLES_FROM_SC.

When it is set to False, users no longer have the option to retrieve raw SC samples, only the synchronized ones.

You can maybe use this boolean to control whether mxcube auto-syncs with a default LIMS on proposal selection or not.

> 
> Would it not then make more sense to either remove or change the behavior of `Get samples from sc`. In your case you actually only need one option in the sample list view `Get samples` right ?
> 
I think the current behavior of “Get samples from SC” is fine as it is.
We just need to better understand its use case and clearly define where/how and when it should be called.

Hope that helps!

ui/src/components/LoginForm/SelectProposal.jsx
```diff
@@ -72,6 +78,11 @@ function SelectProposal() {
     }
   }
 
+  function handleSelectProposal(sessionId) {
+    dispatch(selectProposal(sessionId));
+    dispatch(getLimsSamples(limsName[0].name));
+  }
+
```
jbflo, line 93: @matyson I think it’s worth checking the backend side for this.
With this approach, the backend does something and then you revert it in the UI. It’s probably cleaner to fix the behavior directly in the backend instead of compensating in the frontend.

In [mxcubeweb/core/components/lims.py around L411:](https://github.com/mxcube/mxcubeweb/blob/dc5753586b977360baca37b77c68fc5a35e0f28c/mxcubeweb/core/components/lims.py#L411)

`self.get_sample_list_from_sc()  # changing this one line might solve your issue
`

You could try replacing that with:

`self.synch_with_lims(limsname)
`

matyson: So, what you're suggesting is to use this `USE_GET_SAMPLES_FROM_SC` directly in the `lims.py` to choose whether to sync from sc or the lims, right?

marcus-oscarsson: I don't think its a good idea to mix these two functionalities, its two separate features. I think we need to take a step back and look at the use cases that you are trying to solve. To me it seems like you would like to prevent users from seeing/using samples that do not belong to them but are in the sample changer (for understandable reasons)

We should perhaps meet quickly to discuss the details a bit more closer ?

matyson: > To me it seems like you would like to prevent users from seeing/using samples that do not belong to them but are in the sample changer 

Yes, that's my goal, but it seems to me that showing all the samples is a very used feature and therefore difficult to find a good solution for all. 

Absolutely, we can meet. What's your availability?

marcus-oscarsson: I sent an email, this afternoon after 14h or tomorrow between 9h and 11h or after 13h

matyson:
So, as we've discussed i added the configuration `AUTOSYNC_LIMS` as a boolean to app general configuration in the`server.yaml` file's `mxcube` section (not sure if it makes sense there though, i can change it if you see fit). 

Also, i took the initiative to add the `USE_GET_SAMPLES_FROM_SC` missing templates and docs as they seemed related enough.

PR #1990, opened 2025-12-10
update vulnerable packages

walesch-yan:
Similar to mxcube/mxcubecore#1471, this PR updates packages with vulnerabilities and `poetry`.

The most notable changes here are for `authlib`, `werkzeug`, and `vite`. I tested those specifically to see if the update had any side effects, but everything seems to be working fine.


marcus-oscarsson:
Thanks

PR #1992, opened 2025-12-11
Make sample name consistent

axelboc:
Sometimes the sample name was displayed as `<proteinAcronym> - <sampleName>`, sometimes as `<proteinAcronym> <sampleName>` (no dash), sometimes as `<sampleName> - <proteinAcronym>`, ...

I'm now making sure it's the same everywhere. I went with `<proteinAcronym> - <sampleName>` (or just `<sampleName>` if `proteinAcronym` is not defined), which was used in the current sample tab of the queue.

### DIFF
--- a/ui/src/components/ConfirmCollectDialog/TaskTable.jsx
+++ b/ui/src/components/ConfirmCollectDialog/TaskTable.jsx
@@ -2,6 +2,7 @@ import { useState } from 'react';
 import { OverlayTrigger, Popover, Table } from 'react-bootstrap';
 import { useSelector } from 'react-redux';
 
+import { getSampleName } from '../../utils';
 import TaskOverlayTable from './TaskOverlayTable';
 import styles from './TaskTable.module.css';
 
@@ -47,7 +48,6 @@ export default function TaskTable(props) {
                 : task.parameters;
 
             const sample = sampleList[task.sampleID];
-            const sampleName = `${sample.sampleName} - ${sample.proteinAcronym}`;
             const relativePath = parameters.fullPath.split(rootPath).pop();
 
             return (
@@ -70,7 +70,7 @@ export default function TaskTable(props) {
                 <tr id={task.queueID}>
                   <td>{task.label}</td>
                   <td>
-                    {sampleName} ({sample.location})
+                    {getSampleName(sample)} ({sample.location})
                   </td>
                   <td>
                     <b style={{ color: '#337ab7' }}>...{relativePath}</b>
--- a/ui/src/components/SampleGrid/util.js
+++ b/ui/src/components/SampleGrid/util.js
@@ -16,15 +16,3 @@ export function sampleStateBackground(key) {
 
   return map[key] || 'secondary';
 }
-
-export function getSampleName(sampleData) {
-  let name = sampleData?.proteinAcronym || '';
-
-  if (sampleData?.sampleName && name) {
-    name += ` - ${sampleData?.sampleName}`;
-  } else {
-    name = sampleData?.sampleName || '';
-  }
-
-  return name;
-}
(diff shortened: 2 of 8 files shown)


ui/src/components/ConfirmCollectDialog/TaskTable.jsx
```diff
@@ -47,7 +48,6 @@ export default function TaskTable(props) {
                 : task.parameters;
 
             const sample = sampleList[task.sampleID];
-            const sampleName = `${sample.sampleName} - ${sample.proteinAcronym}`;
```
axelboc, line 50: So now it's swapped, and it works better if `proteinAcronym` is not defined.

ui/src/components/SampleGrid/util.js
```diff
@@ -16,15 +16,3 @@ export function sampleStateBackground(key) {
 
   return map[key] || 'secondary';
 }
-
-export function getSampleName(sampleData) {
```
axelboc, line 20: Moved to a new `utils.js` file higher up and simplified.

ui/src/components/SampleQueue/TodoItem.jsx
```diff
@@ -17,15 +18,15 @@ export default function TodoItem({ sampleData }) {
...
-  const { sampleID, sampleName = '', proteinAcronym = '' } = sampleData;
+  const { sampleID } = sampleData;
 
   return (
     <div className={styles.nodeSample}>
       <div className={styles.taskHead}>
         <div className={styles.nodeName}>
           <p className="pt-1 me-auto">
             <b>{`${sampleID} `}</b>
-            {`${proteinAcronym} ${sampleName}`}
```
axelboc, line 28: no dash here

ui/src/containers/SampleListViewContainer.jsx
```diff
@@ -316,10 +317,8 @@ export default function SampleListViewContainer() {
     const sample = sampleList[key];
     let fi = false;
     if (sample) {
-      const sampleFilter =
-        `${sample.sampleName} ${sample.proteinAcronym}`.toLowerCase();
```
axelboc, line 320: When filtering the sample list by text, this is the sample name that was used to match the search text. This was inconsistent with the sample names displayed.

There was also the issue that spaces were being trimming out as the user was typing a search text, so one could effectively never type a space and therefore never search for the exact name of a sample as displayed (apart from copy/pasting it). To fix this, the search text is no longer trimmed before being stored in the Redux state.

ui/src/utils.js
```diff
@@ -0,0 +1,4 @@
+export function getSampleName(sample) {
+  const { sampleName, proteinAcronym } = sample;
```
axelboc, line 2: I now make the assumption that `sampleName` is always defined. If it's not the case, please let me know.

marcus-oscarsson: `sampleName` should always exist `proteinAcronym` may not exist which should mean "" or None

ui/src/components/ConfirmCollectDialog/TaskTable.jsx
marcus-oscarsson: Swapped ? You mean that whats displayed now is `proteinAcronym`-`sampleName` ?

axelboc: yep exactly, like it is in other places

marcus-oscarsson: Ok, I see. So it was swapped in the `TaskTable` compared to the other places

marcus-oscarsson:
Thanks !

marcus-oscarsson:
Looks good !

PR #1994, opened 2025-12-11
Refactor queue task item components

axelboc:
The first three commits are straightforward: 1) convert to function components, 2) destructure props, 3) inline JSX-returning functions.

In the fourth commit, [9b6cd6badcbe2c86f9d6a259fcdaa602aa1c202d](https://github.com/mxcube/mxcubeweb/pull/1994/commits/9b6cd6badcbe2c86f9d6a259fcdaa602aa1c202d), I clean up the props, move the dispatch calls down, etc. This is where things become a little more interesting. I'll leave some comments as usual.

### DIFF
--- a/ui/src/components/SampleQueue/CharacterisationTaskItem.jsx
+++ b/ui/src/components/SampleQueue/CharacterisationTaskItem.jsx
@@ -1,4 +1,3 @@
-/* eslint-disable react/destructuring-assignment */
 /* eslint-disable jsx-a11y/anchor-is-valid */
 /* eslint-disable react/no-array-index-key */
 import { Button, Table } from 'react-bootstrap';
@@ -9,15 +8,18 @@ import styles from './Item.module.css';
 import TaskItemContainer from './TaskItemContainer';
 
 function TaskItem(props) {
-  function getResult(_state, data) {
-    if (props.data.state !== TASK_COLLECTED) {
+  const { index, data, sampleId, pointID, shapes, showForm, addTask } = props;
+  const wedges = data.type === 'Interleaved' ? data.parameters.wedges : [data];
+
+  function getResult() {
+    if (data.state !== TASK_COLLECTED) {
       return <span />;
     }
 
-    return <div className={styles.resultBody}>{getDiffPlan(data)}</div>;
+    return <div className={styles.resultBody}>{getDiffPlan()}</div>;
   }
 
-  function getDiffPlan(data) {
+  function getDiffPlan() {
     let diffPlan = [];
     if (
       'diffractionPlan' in data &&
@@ -41,7 +43,6 @@ function TaskItem(props) {
   }
 
   function showDiffPlan() {
-    const { data, sampleId } = props;
     const tasks = data.diffractionPlan;
 
     // if there is a single wedge, display the form, otherwise, add all wedges as differente dc-s
@@ -50,29 +51,28 @@ function TaskItem(props) {
       delete data.diffractionPlan[0].sampleID;
       const [{ type, parameters }] = data.diffractionPlan;
 
-      props.showForm(type, sampleId, data.diffractionPlan[0], parameters.shape);
+      showForm(type, sampleId, data.diffractionPlan[0], parameters.shape);
     } else {
       tasks.forEach((t) => {
         const pars = {
           type: 'DataCollection',
           label: 'Data Collection',
           helical: false,
-          shape: props.pointID,
+          shape: pointID,
           ...t.parameters,
         };
 
-        props.addTask([sampleId], pars, false);
+        addTask([sampleId], pars, false);
       });
     }
   }
 
-  function showForm() {
-    const { data, sampleId } = props;
+  function handleParamsTableClick() {
     const { type, parameters } = data;
-    props.showForm(type, sampleId, data, parameters.shape);
+    showForm(type, sampleId, data, parameters.shape);
   }
 
-  function pointIDString(wedges) {
+  function pointIDString() {
     let res = '';
 
     wedges.forEach((wedge) => {
@@ -81,7 +81,7 @@ function TaskItem(props) {
         !res.includes(`${wedge.parameters.shape}`)
       ) {
         try {
-          res += `${props.shapes.shapes[wedge.parameters.shape].name} :`;
+          res += `${shapes.shapes[wedge.parameters.shape].name} :`;
         } catch {
           res = String(res);
         }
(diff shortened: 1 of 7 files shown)


ui/src/components/SampleQueue/CharacterisationTaskItem.jsx
```diff
@@ -1,16 +1,22 @@
...
+import { addTask } from '../../actions/queue';
+import { showTaskForm } from '../../actions/taskForm';
 import { TASK_COLLECTED } from '../../constants';
 import TooltipTrigger from '../TooltipTrigger';
 import styles from './Item.module.css';
 import TaskItemContainer from './TaskItemContainer';
 
 function CharacterisationTaskItem(props) {
-  const { index, data, sampleId, pointID, shapes, showForm, addTask } = props;
+  const { index, data, sampleId, pointID } = props;
```
axelboc: Prop `pointID` was and is still not passed to this component. It's used in the `showDiffPlan` function below. Please advise.

ui/src/components/SampleQueue/CurrentTree.jsx
```diff
@@ -147,14 +120,8 @@ function CurrentTree(props) {
                 <CharacterisationTaskItem
                   key={taskData.queueID}
                   index={i}
-                  id={`${taskData.queueID}`}
                   data={taskData}
                   sampleId={sampleId}
-                  checked={checked}
-                  showForm={(...args) => dispatch(showTaskForm(...args))}
-                  addTask={addTask}
```
axelboc, line 164: The `addTask` action was not being dispatched. This is a regression I introduced last week in #1984. It is now fixed (cf. `CharacterisationTaskItem`)

ui/src/components/SampleQueue/CurrentTree.jsx
```diff
@@ -116,14 +100,8 @@ function CurrentTree(props) {
                 <XRFTaskItem
                   key={taskData.queueID}
                   index={i}
-                  id={`${taskData.queueID}`}
                   data={taskData}
                   sampleId={sampleId}
-                  checked={checked}
-                  showForm={(...args) => dispatch(showTaskForm(...args))}
-                  plotsData={plotsData}
-                  plotsInfo={plotsInfo}
```
axelboc, line 129: A few props were unused throughout like `id`, `checked` and `showDialog`, but it's worth pointing out the two unused props here: `plotsData` and` plotsInfo`, which were passed only to `XRFTaskItem`, as it could hint at a regression.

ui/src/components/SampleQueue/XRFTaskItem.jsx
```diff
@@ -1,19 +1,24 @@
...
 /* eslint-disable jsx-a11y/anchor-is-valid */
 import { OverlayTrigger, Popover } from 'react-bootstrap';
+import { useDispatch, useSelector } from 'react-redux';
 
+import { showTaskForm } from '../../actions/taskForm';
 import styles from './Item.module.css';
 import TaskItemContainer from './TaskItemContainer';
 
 function XRFTaskItem(props) {
-  const { index, data, sampleId, shapes, showForm } = props;
```
axelboc: The `shapes` prop was not actually passed to `XRFTaskItem`. Now it is. It's used in `pointIDString` below. This might fix a bug with the header text of XRF task items.

ui/src/components/SampleQueue/CurrentTree.jsx
marcus-oscarsson: Plotting is no longer used so `plotsData` and `plotsInfo` are indeed not used

ui/src/components/SampleQueue/CharacterisationTaskItem.jsx
marcus-oscarsson: It should be used and passed

marcus-oscarsson: Its probably not very common that `showDiffPlan` is used so perhaps no one noticed

ui/src/components/SampleQueue/CurrentTree.jsx
marcus-oscarsson: Thanks, I dont think we had the time to notice ... :)

marcus-oscarsson:
Thanks well done, the last commit is indeed quite tricky and there a few traps :). I think you got it right !

ui/src/components/SampleQueue/CharacterisationTaskItem.jsx
marcus-oscarsson: Ah now i see, the key is `if (tasks.length <= 1)`. In 99.999 of the cases `length` is 1 thats why we did not notice !

axelboc: I'm not sure where that `pointID` is supposed to come from though. I've just tried going back in the history of `CurrentTree` with git blame (up to 8 years ago) and it looks like `pointID` has never been passed. See this commit which introduces the reference to `this.props.pointID`: https://github.com/mxcube/mxcubeweb/commit/e3ad7fc655a501c78da9b8ddb17c413353d97f6a, and the `CurrentTree` file at the point in time: https://github.com/mxcube/mxcubeweb/blob/e3ad7fc655a501c78da9b8ddb17c413353d97f6a/mxcube3/ui/components/SampleQueue/CurrentTree.js#L187.

axelboc: Is it maybe supposed to be `shape: t.parameters.shape`, and it works because of the spread of `...t.parameters` on the next line?

marcus-oscarsson: My my what a mess,`t.parameters.shape` seems correct

marcus-oscarsson: I guess we did not see the issue as the `if` branch with `length > 1` executes rarely. But we would otherwise have had a "undefined variable `pointID `" error. But `t.parameters.shape` should be correct like above:

https://github.com/mxcube/mxcubeweb/blob/9b6cd6badcbe2c86f9d6a259fcdaa602aa1c202d/ui/src/components/SampleQueue/CharacterisationTaskItem.jsx#L30

marcus-oscarsson: Wow thats a mess !

ui/src/components/SampleQueue/CharacterisationTaskItem.jsx
```diff
@@ -68,34 +26,29 @@ export default class TaskItem extends Component {
...
+        showTaskForm(type, sampleId, data.diffractionPlan[0], parameters.shape),
       );
     } else {
       tasks.forEach((t) => {
         const pars = {
           type: 'DataCollection',
           label: 'Data Collection',
           helical: false,
-          shape: this.props.pointID,
           ...t.parameters,
```
axelboc, line 38: I've just remove the `shape` key — if `t.parameters` has a `shape`, then it will be spread into the object on the next line.

marcus-oscarsson: Perfect, that should do the trick !

PR #1995, opened 2025-12-12
Implement environment variable reading at yaml files

pedrobenetton:
As discussed in the last developer's meeting, me and another developer of LNLS were having trouble having to share the same variables at our configuration repository. Specifically the variable CLIENT_ID from server.yaml cannot be the same for us, since it is used for our SSO authentication and must be different for each IP address (we test our codes in different machines).

This implementation replaces any string that starts with "_ENV_" with the environment variable of that string. This is done after the yaml file has been read.

### DIFF
--- a/mxcubeweb/config.py
+++ b/mxcubeweb/config.py
@@ -20,9 +20,34 @@ from mxcubeweb.core.models.configmodels import (
 
 class ConfigLoader:
     @staticmethod
+    def resolve_env(value):
+        if isinstance(value, str) and value.startswith("_ENV_"):
+            env_value = os.getenv(value)
+            if env_value is None:
+                error_msg = f"{value} starts with _ENV_ but was not set in environment"
+                raise RuntimeError(error_msg)
+            logging.getLogger("HWR").debug(f"Environment variable {value} has been set")
+            return env_value
+        return value
+
+    @staticmethod
+    def walk_and_resolve(data):
+        if isinstance(data, dict):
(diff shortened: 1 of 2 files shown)


marcus-oscarsson:
Well done !, could you also add some developers doc in the documentation section about how this works and the syntax ?

marcus-oscarsson:
Hello @pedrobenetton and happy new year ! :). How did it go with the documentation ? :)

pedrobenetton:
Hello @marcus-oscarsson  happy new year! Last days were a little busy, thanks for the comment.
I added a new file in the docs folder, let me know if it needs any  change.

mxcubeweb/config.py
```diff
@@ -19,10 +19,35 @@
 
 
 class ConfigLoader:
+    @staticmethod
+    def resolve_env(value):
+        if isinstance(value, str) and value.startswith("_ENV_"):
+            env_value = os.getenv(value)
+            if env_value is None:
+                error_msg = f"{value} starts with _ENV_ but was not set in environment"
```
marcus-oscarsson, line 27: Ok, thats nice :), I guess that one could discuss weather the `_ENV_ ` should be included in the actual environment variable name or not. However, lets keep it like this for now at least its clear from where the variable comes.

marcus-oscarsson (approved):

marcus-oscarsson:
Thanks well done ! :)

PR #1997, opened 2025-12-17
Remove unused logic from `CharacterisationTaskItem`

axelboc:
In `CurrentTree`, `CharacterisationTaskItem` is rendered when the task's `type` is `'Characterisation'`, yet in `CharacterisationTaskItem`, there's logic that checks if the task's `type` is `'Interleaved'`.

So I simplify `CharacterisationTaskItem` accordingly.

### DIFF
--- a/ui/src/components/SampleQueue/CharacterisationTaskItem.jsx
+++ b/ui/src/components/SampleQueue/CharacterisationTaskItem.jsx
@@ -1,5 +1,5 @@
 /* eslint-disable jsx-a11y/anchor-is-valid */
-/* eslint-disable react/no-array-index-key */
+
 import { Button, Table } from 'react-bootstrap';
 import { useDispatch, useSelector } from 'react-redux';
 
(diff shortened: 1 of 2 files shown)


ui/src/components/SampleQueue/CharacterisationTaskItem.jsx
```diff
@@ -12,30 +12,29 @@ import TaskItemContainer from './TaskItemContainer';
...
-    const tasks = data.diffractionPlan;
-
-    // if there is a single wedge, display the form, otherwise, add all wedges as differente dc-s
-    if (tasks.length <= 1) {
-      delete data.diffractionPlan[0].run_number;
-      delete data.diffractionPlan[0].sampleID;
-      const [{ type, parameters }] = data.diffractionPlan;
-
+    // If there is a single wedge, display the form; otherwise, add all wedges as different data collections
+    if (diffractionPlan.length === 1) {
```
axelboc, line 25: `<= 1` replaced with `=== 1` since we already check for `> 0` in the JSX.

ui/src/components/SampleQueue/CharacterisationTaskItem.jsx
```diff
@@ -12,30 +12,29 @@ import TaskItemContainer from './TaskItemContainer';
...
+      const [task] = diffractionPlan;
+      const { run_number, sampleID, ...taskData } = task;
       dispatch(
-        showTaskForm(type, sampleId, data.diffractionPlan[0], parameters.shape),
```
axelboc, line 30: I replace the `delete` operators (which mutate the object) with object destructuring.

ui/src/components/SampleQueue/CharacterisationTaskItem.jsx
```diff
@@ -44,27 +43,21 @@ function CharacterisationTaskItem(props) {
...
   function handleParamsTableClick() {
-    const { type, parameters } = data;
-    dispatch(showTaskForm(type, sampleId, data, parameters.shape));
+    dispatch(showTaskForm(type, sampleId, data, shape));
   }
 
   function pointIDString() {
     let res = '';
 
-    wedges.forEach((wedge) => {
```
axelboc, line 54: `wedges` was always initialised to `[data]`, so I simplify the code accordingly here and in the JSX.

ui/src/components/SampleQueue/CharacterisationTaskItem.jsx
```diff
@@ -74,121 +67,107 @@ function CharacterisationTaskItem(props) {
       pointIDString={pointIDString()}
     >
       <div className={styles.taskBody}>
-        {wedges.map((wedge, i) => {
-          const { parameters } = wedge;
-          const { fileName, path = '' } = parameters;
-          const pathEndPart = path.slice(-40);
-
-          return (
-            <div key={`wedge-${i}`}>
```
axelboc, line 83: With whitespace hidden in the diff, you can see that this `div` goes away and the `style` attribute below (`0.5em` is already the default padding in `styles.dataPath`).

ui/src/components/SampleQueue/CharacterisationTaskItem.jsx
```diff
@@ -74,121 +67,107 @@ function CharacterisationTaskItem(props) {
...
-                      </td>
-                    )}
-                  </tr>
-                </tbody>
-              </Table>
-
-              {data.state === TASK_COLLECTED && (
-                <div className={styles.resultBody}>
-                  {'diffractionPlan' in data &&
-                    Object.keys(data.diffractionPlan).length > 0 && (
```
axelboc, line 175: If defined, `diffractionPlan` is an array, not an object, so I remove `Object.keys()`.

marcus-oscarsson:
Thanks, nice !

PR #1998, opened 2025-12-18
Pre-fill the data collection for with sample data

mockoocy:
Unit cell parameter data and space group for a particular sample are already included for a particular sample when using ISPyB and possible other LIMS.

> [!NOTE]
> I am not sure how generic are these sampeData paremeters; e.g. if using different LIMS could lead to `cellA` being available under different name. In that case it would be undefined and nothing would be prefilled.

### DIFF
--- a/ui/src/components/Tasks/DataCollection.jsx
+++ b/ui/src/components/Tasks/DataCollection.jsx
@@ -386,6 +386,16 @@ export default connect((state) => {
       ].acq_parameters.osc_range;
   }
 
+  const {
+    cellA,
+    cellAlpha,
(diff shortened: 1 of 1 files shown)


marcus-oscarsson (approved):

marcus-oscarsson:
> I am not sure how generic are these sampeData paremeters; e.g. if using different LIMS could lead to `cellA` being available under different name. In that case it would be undefined and nothing would be prefilled.

This should be standardized and it more or less is, but the names are subject to change. Lets use this for the moment It will work for the time being.

PR #1999, opened 2025-12-18
Simplify interleaved data collection availability logic in queue UI

axelboc:
- Simplify some code in `CurrentTree` related to creating interleaved data collections
- Remove a couple of bitwise operators in the `queueGUI` reducer to avoid ending up with `0`/`1` values in the `displayData.collapsed/selected` state instead of booleans.

### DIFF
--- a/ui/src/components/SampleQueue/TaskItemContainer.jsx
+++ b/ui/src/components/SampleQueue/TaskItemContainer.jsx
@@ -65,10 +65,10 @@ export default function TaskItemContainer({
   }
 
   function taskHeaderOnClick(e) {
-    if (!e.ctrlKey) {
-      dispatch(collapseItem(data.queueID));
-    } else {
+    if (e.ctrlKey) {
       dispatch(selectItem(data.queueID));
+    } else {
+      dispatch(collapseItem(data.queueID));
     }
   }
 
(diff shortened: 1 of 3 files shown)


ui/src/components/SampleQueue/TaskItemContainer.jsx
```diff
@@ -140,10 +140,7 @@ export default function TaskItemContainer({
             />
           )}
         </div>
-        <Collapse
-          id={`collapse-${data.queueID}`}
-          in={Boolean(displayData.collapsed)}
```
axelboc, line 145: Removing this `Boolean` type cast now that it's guaranteed to remain a boolean in the state.

ui/src/reducers/queueGUI.js
```diff
@@ -88,16 +88,34 @@ function queueGUIReducer(state = INITIAL_STATE, action = {}) {
       return { ...state, showConfirmCollectDialog: action.show };
     }
     case 'COLLAPSE_ITEM': {
-      const displayData = { ...state.displayData };
-      displayData[action.queueID].collapsed ^= 1; // eslint-disable-line no-bitwise
```
axelboc, line 92: In addition to turning the `collapsed` state into a number, this line also mutates the Redux state, which is not recommended.

walesch-yan (approved):

marcus-oscarsson:
Thanks !

PR #2001, opened 2025-12-19
split sampleImage and video

walesch-yan:
This PR intends a first refactoring of the `SampleImage` component, by extracting the video stream logic and putting it into its own functional component (`VideoPlayer`). This component handles the creation and management of the main stream.

See comments below for details on changes.

### DIFF
--- a/ui/src/actions/sampleview.js
+++ b/ui/src/actions/sampleview.js
@@ -111,8 +111,6 @@ export function setVideoSize(width, height) {
       pixelsPerMm: json.pixelsPerMm,
       sourceScale: json.scale,
     });
-
-    globalThis.initJSMpeg();
   };
 }
 
--- a/ui/src/components/SampleView/SampleImage.jsx
+++ b/ui/src/components/SampleView/SampleImage.jsx
@@ -1,5 +1,4 @@
 /* eslint-disable react/destructuring-assignment */
-/* eslint-disable jsx-a11y/control-has-associated-label */
 import 'fabric';
 
 import React from 'react';
@@ -26,7 +25,6 @@ import { HW_STATE, QUEUE_RUNNING } from '../../constants';
 import SampleControls from '../SampleControls/SampleControls';
 import DrawGridPlugin from './DrawGridPlugin';
 import GridForm from './GridForm';
-import { JSMpeg } from './jsmpeg.min.js';
 import styles from './SampleImage.module.css';
 import {
   makeCentringHorizontalLine,
@@ -36,6 +34,7 @@ import {
   makePoints,
   makeTwoDPoints,
 } from './shapes';
+import VideoPlayer from './VideoPlayer';
 
 const { fabric } = globalThis;
 fabric.Group.prototype.hasControls = false;
@@ -61,7 +60,6 @@ class SampleImage extends React.Component {
     this.configureGrid = this.configureGrid.bind(this);
     this.updateGridResults = this.updateGridResults.bind(this);
     this.selectedGrid = this.selectedGrid.bind(this);
-    this.initJSMpeg = this.initJSMpeg.bind(this);
     this.centringMessage = this.centringMessage.bind(this);
     this.selectShape = this.selectShape.bind(this);
     this.deSelectShape = this.deSelectShape.bind(this);
@@ -73,7 +71,6 @@ class SampleImage extends React.Component {
     this.drawGridPlugin = new DrawGridPlugin();
     this.drawGridPlugin.setGridResultFormat(props.meshResultFormat);
     this._keyPressed = null;
-    this.player = null;
     this.removeShapes = this.removeShapes.bind(this);
   }
 
@@ -112,20 +109,13 @@ class SampleImage extends React.Component {
     document.addEventListener('keydown', this.keyDown, false);
     document.addEventListener('keyup', this.keyUp, false);
 
-    globalThis.initJSMpeg = this.initJSMpeg;
-    this.initJSMpeg();
     this.renderSampleView();
   }
 
   componentDidUpdate(prevProps) {
-    const { imageRatio, width } = this.props; // #NOSONAR
+    const { imageRatio, width } = this.props;
     if (imageRatio !== prevProps.imageRatio || width !== prevProps.width) {
-      // #NOSONAR
       this.setImageRatio();
-    }
-    // Initialize JSMpeg for decoding the MPEG1 stream
-    if (prevProps.videoFormat !== 'MPEG1') {
-      this.initJSMpeg();
     }
 
     this.renderSampleView();
@@ -134,15 +124,6 @@ class SampleImage extends React.Component {
   componentWillUnmount() {
     if (this.canvas) {
       this.canvas.dispose();
-    }
-
-    if (this.player) {
-      try {
-        this.player.destroy();
-        this.player = null;
-      } catch {
-        this.player = null;
-      }
     }
 
     this.canvas.off('mouse:down', this.leftClick);
@@ -160,8 +141,6 @@ class SampleImage extends React.Component {
     imageOverlay.removeEventListener('contextmenu', this.rightClick);
     imageOverlay.removeEventListener('wheel', this.wheel);
     imageOverlay.removeEventListener('dblclick', this.goToBeam);
-
-    globalThis.initJSMpeg = null;
   }
 
   onMouseMove(options) {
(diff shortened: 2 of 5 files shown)


ui/src/actions/sampleview.js
```diff
@@ -111,8 +111,6 @@ export function setVideoSize(width, height) {
       pixelsPerMm: json.pixelsPerMm,
       sourceScale: json.scale,
     });
-
-    globalThis.initJSMpeg();
```
walesch-yan, line 115: No longer necessary, the new hook automatically restarts the player when the image size changes, by keeping up to date with the `width` and `height` values from the state

ui/src/components/SampleView/jsmpeg.min.js
```diff
@@ -320,7 +320,6 @@ JSMpeg.Player = (function () {
     this.pause();
     this.source.destroy();
     this.video && this.video.destroy();
-    this.renderer && this.renderer.destroy();
```
walesch-yan, line 323: This was an interesting one, the JSMpeg player was destroying it underlying canvas element, when the .destroy() function of the player was called, which worked before, because we always recreated the element. The new hook now only destroys the player and keeps the canvas when sizes change. For more info see [this Issue](https://github.com/phoboslab/jsmpeg/issues/407)

ui/src/components/SampleView/SampleImage.jsx
```diff
@@ -955,12 +882,17 @@ class SampleImage extends React.Component {
               selectGrid={this.selectShape}
               selectedGrids={this.props.selectedGrids.map((grid) => grid.id)}
             />
-            {this.createVideoPlayerContainer(this.props.videoFormat)}
+            <div className={styles.videoCanvasWrapper}>
+              <VideoPlayer />
+              <canvas
+                id="canvas"
+                className={styles.coveringCanvas}
+                aria-label="SampleViewCoveringCanvas"
```
walesch-yan: while at it, added the label to remove the eslint exception

ui/src/components/SampleView/VideoPlayer.jsx
```diff
@@ -0,0 +1,81 @@
...
+          ref={canvasRef}
+          id="sample-img"
+          className={styles.video}
+          aria-label="SampleVideo"
+        />
+      );
+    }
+    default: {
+      return (
+        <p className={styles.video}>Unsupported video format: {videoFormat}</p>
```
walesch-yan: I used a switch here, for the case, that we would like to add other formats later on. 
Note: At the moment the default is only a placeholder, since MXCuBE internally always sets the `videoFormat` to MJPEG, if it is not `MPEG1`

ui/src/components/SampleView/SampleImage.jsx
```diff
@@ -779,56 +756,6 @@ class SampleImage extends React.Component {
     return result;
   }
 
-  createVideoPlayerContainer(format) {
-    let source = '/mxcube/api/v0.1/sampleview/camera/subscribe';
```
walesch-yan, line 783: The `HTTPStreamer` has already been removed, so this default value cannot be used

ui/src/components/SampleView/SampleImage.jsx
```diff
@@ -36,6 +34,7 @@ import {
   makePoints,
   makeTwoDPoints,
 } from './shapes';
+import VideoPlayer from './VideoPlayer.jsx';
```
axelboc:
```suggestion
import VideoPlayer from './VideoPlayer';
```

ui/src/components/SampleView/SampleImage.jsx
axelboc: The label is for humans (typically screen reader users) so it would have to be something like "Sample view".

ui/src/components/SampleView/VideoPlayer.jsx
```diff
@@ -0,0 +1,81 @@
...
+        />
+      );
+    }
+    case 'MPEG1': {
+      return (
+        <canvas
+          ref={canvasRef}
+          id="sample-img"
+          className={styles.video}
+          aria-label="SampleVideo"
```
axelboc: If you use "Sample view" as label for the interactive canvas, adding a label here doesn't serve any purpose for screen reader users, so I would instead hide this canvas from screen readers entirely with `aria-hidden="true"` (note the string value). Same with the `img` above, but via an empty `alt` attribute: `alt=""`

ui/src/components/SampleView/VideoPlayer.jsx
axelboc: Might be best to throw an error after the switch?

ui/src/components/SampleView/VideoPlayer.jsx
```diff
@@ -0,0 +1,81 @@
...
+        playerRef.current = null;
+      }
+    };
+  }, [videoFormat, videoSource, videoWidth, videoHeight]);
+
+  if (!videoURL) {
+    return null;
+  }
+
+  switch (videoFormat) {
```
axelboc: I tend to prefer multiple consecutive `if () { return }` blocks to remove one level of indentation.

ui/src/components/SampleView/VideoPlayer.jsx
```diff
@@ -0,0 +1,81 @@
...
+import styles from './VideoPlayer.module.css';
+
+export default function VideoPlayer() {
+  const videoWidth = useSelector((state) => state.sampleview.width);
+  const videoHeight = useSelector((state) => state.sampleview.height);
+  const videoFormat = useSelector((state) => state.sampleview.videoFormat);
+  const videoURL = useSelector((state) => state.sampleview.videoURL);
+  const videoHash = useSelector((state) => state.sampleview.videoHash);
+
+  const videoSource = `${videoURL}/${videoHash}`;
```
axelboc: You could probably remove the `video` prefix from all (or at least some of) these variables.

axelboc: Also for `videoSource`, you could use a single `useSelector` and add a ternary to check for truthiness:

```js
const source = useSelector(state => {
  const { videoURL, videoHash } = state.sampleview;
  return videoURL ? `${videoURL}/${videoHash}` : undefined;
});

// then further down check if `source` is truthy instead of `videoURL`
```

ui/src/components/SampleView/VideoPlayer.jsx
```diff
@@ -0,0 +1,81 @@
...
+
+  const canvasRef = useRef(null);
+  const playerRef = useRef(null);
```
axelboc: I don't think you need this ref. You can just store the `player` instance in a constant inside the `useEffect` and reference it directly in the clean-up function.

ui/src/components/SampleView/VideoPlayer.jsx
```diff
@@ -0,0 +1,81 @@
...
+      return () => {};
+    }
+
+    canvas.width = videoWidth;
+    canvas.height = videoHeight;
+
+    if (playerRef.current) {
+      playerRef.current.destroy();
+      playerRef.current = null;
+    }
```
axelboc: Don't think this is needed, especially once you switch to a local `player` variable.

ui/src/components/SampleView/jsmpeg.min.js
axelboc: Is it possible to avoid modifying the source code of jsmpeg maybe? Otherwise we risk forgetting to re-apply this change next time we upgrade jsmpeg (if ever).

What about calling all three other methods independently in the `useEffect`'s clean-up function?

```
  player.pause();
  player.source.destroy();
  player.video?.destroy();
```

axelboc (commented):
Love it! A great step for cleaning up `SampleView`. I just a few touch-ups to suggest.

ui/src/components/SampleView/jsmpeg.min.js
walesch-yan: This indeed is a great suggestion! Have not thought about it

walesch-yan:
@axelboc, thanks for the overall suggestions. I applied them to this newer version

ui/src/components/SampleView/VideoPlayer.jsx
walesch-yan: This was indeed there from a previous test I made and forgot to remove :sweat_smile:

ui/src/components/SampleView/VideoPlayer.jsx
walesch-yan: It's a good point, I used `aria-hidden: "true"`, in addition, I set `tabindex={-1}` to remove it from tab focus (the interactive canvas still is focusable from keyboard.

axelboc (approved):
Looks great, thanks!

ui/src/components/SampleView/VideoPlayer.jsx
axelboc: Good call! :100:

marcus-oscarsson:
Awesome, thanks !

PR #2003, opened 2026-01-08
Show all sections by default for the custom forms

### LINKED ISSUE #1898: Defaullt to show-more
It would be great if there the default could be to display all sections, or make it possible to pass a display option for show-more for a group. We could either decide to enable/disable the show-more feature for some groups and then default it to hidden when enabled (that would make sense). But its quite limiting to hide all other sections by default apart from "Acquisition"

Image

mockoocy:
This PR introduces two distinct changes in fact (I hope it's okay, both are small :))
The `<CollapsableRows>` component has been rewrote in functional style. 

Additionally, the default behaviour of said component, when used in generic forms, is to be expanded. For this purpose `defaultCollapsed` prop has been added there.
Closes #1898 

Happy new year! :tada:

### DIFF
--- a/ui/src/components/Tasks/CustomObjectFieldTemplate.jsx
+++ b/ui/src/components/Tasks/CustomObjectFieldTemplate.jsx
@@ -32,7 +32,7 @@ export default function CustomObjectFieldTemplate({
         return (
           <div key={group}>
             <FieldsHeader title={group} />
-            <CollapsableRows>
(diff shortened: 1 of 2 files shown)


axelboc (approved):
Neat! (and happy New Year to you too and everyone else! :grin: )

marcus-oscarsson:
No entirely sure that we have agreed and that the default should be to show then contents. Lets try and see how it behaves.

PR #2004, opened 2026-01-08
Update dependencies from dependabot

Phantominumm:
This pull request updates dependencies from the dependabot

Dependency updates:

* Added `storybook` as a devDependency at version 9.1.17.
* Update `qs` to 6.14.1 and `js-yaml` to 4.1.1 
* Upgraded runtime dependencies: `classnames` to 2.5.1, `react-draggable` to 4.5.0, `react-icons` to 4.12.0, `react-lazyload` to 3.2.1, `react-router-dom` to 6.30.3, and `socket.io-client` to 4.8.3.


walesch-yan:
Since we are not directly using `storybook`, I would not add it as devdependency. Instead you can update `@esrf/eslint` and probably also `eslint`, which should solve this as well:)

marcus-oscarsson:
Thanks !

PR #2005, opened 2026-01-08
Make readonly text inputs gray in generic forms

### LINKED ISSUE #1899: Readonly fields are not displayed the same
Read only fields are not displayed the same (greyed out/disabled). In the screen shot below, `sub_sampling` and `frequencey` are both `readonly` however only `sub_sampling` is the only one being displayed as "disabled". It seems to be related to the "description" display  

Image

mockoocy:
Currently when setting `ui:readonly` property in uiSchema, it behaves differently based on the input type.
For dropdowns `<select>` tag receives `disabled` property, which in turn triggers `.form-control:disabled` global boostrap css rule. It causes the input to have a gray background.

For text inputs it works different. `readonly` property is being set, which would require `:read-only` pseudo selector. Bootstrap does not define `.form-control:read-only`, so readonly input fields don't appear any different than "normal ones".

This change applies same styling `.form-control:disabled` for readonly inputs

Closes #1899

### DIFF
--- a/ui/src/components/Tasks/CustomFieldTemplate.module.css
+++ b/ui/src/components/Tasks/CustomFieldTemplate.module.css
@@ -45,3 +45,8 @@
   font-family: monospace;
   color: var(--bs-secondary);
 }
+
+.fieldInput :global(.form-control):read-only {
+  background-color: var(--bs-secondary-bg);
+  opacity: 1;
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
Thanks !

PR #2008, opened 2026-01-09
Update remix-run

Phantominumm:
* Added `@remix-run/router` version 1.23.2 to `dependencies` in `package.json` to force secure versions of transitive dependencies (from dependabot)

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -21,6 +21,7 @@
   "dependencies": {
     "@fortawesome/fontawesome-free": "^5.15.4",
     "@reduxjs/toolkit": "2.6.1",
+    "@remix-run/router": "^1.23.2",
     "@rjsf/core": "5.13.1",
     "@rjsf/utils": "5.13.1",
     "@rjsf/validator-ajv8": "5.13.1",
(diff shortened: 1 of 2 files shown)


ui/package.json
```diff
@@ -21,6 +21,7 @@
   "dependencies": {
     "@fortawesome/fontawesome-free": "^5.15.4",
     "@reduxjs/toolkit": "2.6.1",
+    "@remix-run/router": "^1.23.2",
```
marcus-oscarsson, line 24: This is as far as I know a peer dependency. It would be worth checking if the package that has this dependency can or should be updated. It would otherwise be useful to note with a comment that this change was made because of some security flaw :)

I don't know if there is a established practice for where to list this dependency. There used to be a `peerDependencies` section, `devDependecies` somehow feels wrong. Anyhow as I understand it `pnpm install` installs all dependencies regardless of category right ? I'm sure @axelboc as some good advice ?

axelboc:
This doesn't seem relevant after the rebase — another PR must have already updated this dependency in the lock file.

ui/package.json
axelboc: Putting it in `peerDependencies` would not have worked, as it specifies the dependencies that a potential consumer package would be required to install.

However, this would definitely not have been the way to go. In order of preference I would have:

- upgraded `react-router-dom` — this is what was done in #2004, so great! :tada: 
- replaced `react-router-dom` with a lighter solution like `wouter`
- [configured pnpm](https://pnpm.io/settings#overrides) to override the version of `@remix-run/router`

marcus-oscarsson:
@Phantominumm, @axelboc, yes that was done by the `react-router-dom` update I presume in #2004.  

@axelboc Is it worth looking at using `wouter` instead then ?

marcus-oscarsson:
No longer needed

PR #2013, opened 2026-01-16
Properly initialize SocketIO Test Client

mockoocy:
`flask_socketio.test_client` expects to receive a Flask app object, but `Server` is being passed instead.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -108,7 +108,7 @@ def client():
 
     assert resp.status_code == 200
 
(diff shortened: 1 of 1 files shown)


mockoocy:
The motivation is that our internal pipeline broke, because from `Flask-SocketIO=5.6.0` it tries to read `.config` property of the object passed there.

dominikatrojanowska:
LGTM :)

PR #2015, opened 2026-01-22
Ml react chat

Phantominumm:
Here the changes that was proposed in the previous PR #1923 after the merge:  

- Now processes initial chat messages in the `SET_INITIAL_STATE` case of the `remoteAccess` reducer.
- Moved chat state update logic (e.g., mark all as read) to the reducer; `markAllAsRead` now dispatches a simple action.
- Removed unnecessary read flag from server IO, as it is handled in the frontend.
- Removed `normalizeDateToISO`, now throws on invalid dates.
- Uses a FontAwesome icon for the launcher.
- Change key generation for message lines.
- Uses a data attribute (data-open) for open/close state instead of toggling class names.


marcus-oscarsson:
Really nice cleanup ! :)

marcus-oscarsson:
@axelboc, would you like to have a look before I merge ?

ui/src/components/ChatComponent/chatMessages.js
```diff
@@ -1,21 +1,12 @@
...
-
 export function processChatMessageRecord(record, currentUsername) {
   const isSelf = record.username === currentUsername;
-  const normalizedDate = normalizeDateToISO(record.date);
+
+  let date = record.date ? new Date(record.date) : new Date();
+  if (Number.isNaN(date.getTime())) {
+    // eslint-disable-next-line no-console
+    console.error('Invalid date provided:', record.date);
+    date = new Date(); // Use current date as fallback
```
axelboc: Did you consider throwing here? If the date is wrong, something is wrong somewhere else, as users don't input the date themselves ever.

```js
  throw new Error('Invalid date');
```

axelboc (approved):
It's great, thank you!

ui/src/components/ChatComponent/chatMessages.js
marcus-oscarsson: I think it would be nice to replace the `console.log` with a `throw` as well :)

Phantominumm: @marcus-oscarsson It's done.

marcus-oscarsson:
Thanks alot @Phantominumm, there seems to be some conflicts. Maybe you can do a re-base next week and then have a look with @walesch-yan for merging. I you have any questions don't hesitate to send me a message :)

marcus-oscarsson:
Thanks !

PR #2016, opened 2026-01-23
[Flask-Security] - Fixed handling of invalid authentication

marcus-oscarsson:
Fixed handling of invalid authentication in Flask-Security and general Flask exception handling. 

 - Fixed then registration of invalid authentication handler. The configuration of the invalid authentication handler changed in a previous version of Flask-Security. The user is now redirected to /login when accessing a none API resource with invalid authentication

 - The flask exception handler was set twice with two different functions, stack trace logged to `server-access` log and a 404 error is returned (again for none API resources)

### DIFF
--- a/mxcubeweb/server.py
+++ b/mxcubeweb/server.py
@@ -1,17 +1,14 @@
 import logging
 import os
-import traceback
 from pathlib import Path
 
 import flask_security
 import gevent
 import werkzeug
-from flask import (
-    Flask,
-    request,
-)
+from flask import Flask, redirect, request
 from flask_login import current_user
 from flask_socketio import SocketIO
+from werkzeug.exceptions import HTTPException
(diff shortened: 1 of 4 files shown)


mxcubeweb/server.py
```diff
@@ -45,9 +42,40 @@ def __init__(self):
...
     @staticmethod
     def exception_handler(e):
-        err_msg = "Uncaught exception while calling %s" % request.path
-        logging.getLogger("exceptions").exception(err_msg)
-        return err_msg + ": " + traceback.format_exc(), 409
+        err_msg = "Error while calling %s" % request.path
+
+        if e.code == 404:
+            logging.getLogger("server_access").info(
+                f"URL {request.path} not found on servrer"
```
walesch-yan, line 49: Small typo here (servrer), otherwise a very nice PR!:)

PR #2019, opened 2026-01-26
Replaced the static Server object with an "ordinary" Server object

marcus-oscarsson:
Replaced the static Server object with an ordinary Server object that is created via create/factory function. This cleans up the Server creation and setup logic. Parts of the init logic for middle-ware and add-ons have been moved to separate files, perhaps dividing things a bit too much, but its at-least consistent and makes things "hopefully" easier to work with.

This PR requires #2016 and #2018 to be merged first.


marcus-oscarsson:
@walesch-yan had a look and made review of this PR. 

To all: Feel free to look at our PR's and review them. Thanks. I will merge this one.

PR #2021, opened 2026-01-26
Enable adding large grid from two 2D points

dominikatrojanowska:
Make it possible to create grids that are larger then the field of view of the on-axis-video. This PR adds a feature, where it's possible to define mesh grid by:
* creating a top-left 2D point
* create a bottom-right 2D point
* create a mesh with these two 2D points inside. 

Note, the size of mesh overlaps with those points, as in the calculation of numbers of columns and rows the closest higher integers are taken. Of course I tested with these large grids, just to show the size better I attached the screenshots of smaller one.

New option in context menu:
Screenshot from 2026-01-26 15-14-45

Drawn grid:
Screenshot from 2026-01-26 15-15-12

I know there is some [development](https://github.com/mxcube/mxcubeweb/tree/esrf-develop-id29-chip) inluding large grids going on, but at Max IV we needed easier and quicker solution, so sharing that one :) let me know if you are interested in adding that to this repository!

### DIFF
--- a/ui/src/components/SampleView/ContextMenu.jsx
+++ b/ui/src/components/SampleView/ContextMenu.jsx
@@ -43,9 +43,12 @@ export default function ContextMenu() {
   const sampleData = useSelector(
     (state) => state.sampleGrid.sampleList[sampleID],
   );
-  const { clickCentring } = useSelector((state) => state.sampleview);
+  const { clickCentring, imageRatio, pixelsPerMm, beamSize } = useSelector(
+    (state) => state.sampleview,
+  );
   const groupFolder = useSelector((state) => state.queue.groupFolder);
   const contextMenu = useSelector((state) => state.contextMenu);
+  const allShapes = useSelector((state) => state.shapes.shapes);
   const { sampleViewX, sampleViewY, shape, pageX, pageY, show } = contextMenu;
 
   const isDrawGridAvailable = useSelector(
@@ -289,6 +292,11 @@ export default function ContextMenu() {
           action: () => createLineOnCanvas(shape.id),
           key: 'create_line',
         },
+        {
+          text: 'Add Grid',
+          action: () => draw2DPointsGrid(shape.id),
+          key: 'create_grid',
+        },
         ...genericTasks.line,
       ],
       LINE: [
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
Thanks @dominikatrojanowska, its a very nice feature and quite a practical way of doing things. I'm not sure that using 2D points is a good starting point even if thats an easy solution. I would perhaps be more in favor of adding a separate POI for this. I'm not entirely sure at the moment, I will need to discuss it with my colleagues.

marcus-oscarsson:
A more practical question is, how do you select the points if one of them are not in the screen ? (If its a large mesh ?)

dominikatrojanowska:
> A more practical question is, how do you select the points if one of them are not in the screen ? (If its a large mesh ?)

At demo version moving motors doesn't work to show it up but at our fork I just clicked on 1st 2D point, move motors, click the 2nd one with CTRL pressed and as I've chosen two points the `Add grid` shows up in the context menu. Screenshot with a grid exceeding the camera image boundaries:
image

marcus-oscarsson:
I see. Me and many others use the SHIFT+click short cut (move to beam) to move around. Would this interfere with the point selection.

This is a very nice feature, but I think we do need to re-consider the use of 2D points.  Do you think that it would be worth extending the "Grid dialog" with a list of all shapes (like in a separate tab, like this all shapes could be selectable and
removed from there ?

I'm partly wondering because creating these large grids will also "hide" other shapes so it would be nice with easy way to access the shapes.

For the 2D-Points, its a cleaver way of handling this. However its not very clear that one can create a big mesh/grid from two 2D points (from a users perspective). One would really need to know that this feature exists. I'm wondering if it would not be more intuitive to have sort of dedicated two step process. Where the user either by using the context menu or the Grid dialog starts the process, i.e by selecting "draw grid from points" and then use a sort of "pick" button to pick/set the corners ?

dominikatrojanowska:
> I see. Me and many others use the SHIFT+click short cut (move to beam) to move around. Would this interfere with the point selection.

well, I meant CTRL+click, not SHIFT. So no, it should not interfere. Btw clicking on 2D point with SHIFT pressed works similarly to clicking without any other button - just changing the selection.

> I'm partly wondering because creating these large grids will also "hide" other shapes so it would be nice with easy way to access the shapes.

the grids could be hidden, there is a `Hide` button in grid context menu

> This is a very nice feature, but I think we do need to re-consider the use of 2D points. Do you think that it would be worth extending the "Grid dialog" with a list of all shapes (like in a separate tab, like this all shapes could be selectable and removed from there ?

well, with this approach the solution is quite straightforward and simple, I'm not sure if it's worthy to complicate this... But if more developers would say so, I could do it - no problem - there might be some other advantages of creating shapes tab. 

> For the 2D-Points, its a cleaver way of handling this. However its not very clear that one can create a big mesh/grid from two 2D points (from a users perspective). One would really need to know that this feature exists. I'm wondering if it would not be more intuitive to have sort of dedicated two step process. Where the user either by using the context menu or the Grid dialog starts the process, i.e by selecting "draw grid from points" and then use a sort of "pick" button to pick/set the corners ?

That makes sense to me as well... Maybe when `Draw Grid` is active the additional option should be visible in the Context Menu: `Start grid ` and after the starting point is picked, the next right click should show `Finish grid` instead. What do you think?

Started a [discussion](https://github.com/mxcube/mxcubeweb/discussions/2025) as this PR seems to have few debts

Issue #1627, opened 2025-04-15
Creating Characterization from Samplelist crashes the UI


walesch-yan:
When selecting the characterization button from the context menu of a sample in the sample list, the UI will crash as it can be seen below:

![Image](https://github.com/user-attachments/assets/65cfb305-4a62-4e4b-b89e-23780925d126)

The problem is due to the `positionID` being set to -1 while adding the task from the samplelist, while the Charcterization component expects it to be a string. 

Since I assume that this feature is not really used (these tasks should be added on the data collection page) I would voucher to remove add datacollection and characterization from the contextmenu of the samples in the sample list tab. Alternatively, we can work on a fix for it

gabadinho:
Hello :)

Just noticed this bug, still in the current version as of this writing.
Although I don't understand why we should remove the context menu entries, pretty sure advanced users use them...

<img width="1639" height="1021" alt="Image" src="https://github.com/user-attachments/assets/08e5504d-f1b7-46b6-82f8-77d41d4ac49f" />
<img width="1639" height="1021" alt="Image" src="https://github.com/user-attachments/assets/82c49d87-ab51-4fd5-8885-0852e9a3903d" />

Actually the Data collection entry seems also to be broken, and so are these same entries but for the context menu of a puck, so these tasks entries should all be reviewed!

<img width="1639" height="1021" alt="Image" src="https://github.com/user-attachments/assets/5a24bb76-4fa3-4038-80aa-433d833791bf" />
<img width="1639" height="1021" alt="Image" src="https://github.com/user-attachments/assets/e7b26bbb-0b08-43bf-a026-c119dedcf56b" />

walesch-yan:
Hello and thanks for the feedback @gabadinho , I will take a closer look at solving this issue then. 
Regarding my comment about the removal of the buttons from the context menu in the samples tab: Users normally add the data collection and characterisation by right-clicking on the sample in the `onAxisViewer` (the video under the `DataCollection` tab), which opens another context menu with all the possible actions. 

As far as my understanding goes, the feature of directly adding it to a sample on the `Samples` tab is not really used, which is why there are almost no complaints about it.

gabadinho:
Hi @walesch-yan thanks for the fast reply! I also am a developer so I cannot really tell 100% what features are used, but I think it's worth it to try to figure out why and when this bug was introduced, specially since when you noticed this the Data collection entry did work, and I don't see any particular reason why wouldn't the Characterisation also work.

[...]

PR #2023, opened 2026-01-28
verify positionID as string before calling .includes()

walesch-yan:
There was an internal `.includes()` call on the `pointID` object of the `state.taskForm.pointID`. However, the default value of this parameter is `-1`. Since the expected type should be a `string`, any other value (like the default) would create a crash on the front-end. 

This scenario happened in #1627 . The `includes()` function call is only used inside an `if` condition, to set the default of images taken to 1 when using 2D points. This PR adds a type check in the condition to prevent the crash.


marcus-oscarsson:
Well done !, thanks @gabadinho for the report and @walesch-yan for the fix !,

PR #2027, opened 2026-02-02
fix datacollection ui crash on unmounted sample

### LINKED ISSUE #2026: Creating a Data collection from a sample/puck context menu (in Samples tab) crashes the UI
Somewhat similar to https://github.com/mxcube/mxcubeweb/issues/1627 but with the "Data collection" entry of the context menu (both sample and puck).

If you select this entry with no loaded sample (meaning: no Current sample on the Data collection tab) then the UI crashes.

Tested against mxcubeweb tag v4.629.0 with mxcubecore tag v1.449.0.

Image
Image

walesch-yan:
This PR closes #2026.
The UI crash explained in #2026 happens, since `state.sampleGrid.sampleList[state.queue.currentSampleID]` is `undefined` if no sample is mounted (`currentSampleID` is null in that case) by defaulting the call to an empty object in this case, the crash will be prevented.

This crash was introduced in #1998. This PR aimed at pre-filling some processing field parameters for a mounted sample. The suggested change would just not pre-fill the parameters in the dialog if the sample is not mounted

### DIFF
--- a/ui/src/components/Tasks/DataCollection.jsx
+++ b/ui/src/components/Tasks/DataCollection.jsx
@@ -394,7 +394,7 @@ export default connect((state) => {
     cellC,
     cellGamma,
     crystalSpaceGroup,
-  } = state.sampleGrid.sampleList[state.queue.currentSampleID];
+  } = state.sampleGrid.sampleList[state.queue.currentSampleID] ?? {};
 
   return {
     path: `${state.login.rootPath}/${subdir}`,


mockoocy (approved):
Did not know about such a case being possible 😅

axelboc (approved):

marcus-oscarsson:
> Did not know about such a case being possible 😅

The information can come prefilled from LIMS, something we might have to look at in the future when/if this information is made available. For now this is fine ! :)

marcus-oscarsson:
Thanks @walesch-yan

PR #2000, opened 2025-12-18
AbstractDiffractometer and AbstartSampleView

beteva:
This PR replaces #1881.

Changes due to AbstractDiffractometer and AbstartSampleView new implementation. Dependant on mxcube/mxcubecore/pull/1479

PR #2028, opened 2026-02-04
Fixing mxcubecore to < 2.x before PR #2000 is merged

marcus-oscarsson:
Fixing `mxcubecore` to < 2.x before PR #2000 is merged

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,7 @@ Flask-Security = { version = "5.7.1", extras = ["common", "fsqla"] }
 Flask-SocketIO = "^5.3.6"
 gevent = "^23.9.1"
 markupsafe = "^2.1.5"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -45,7 +45,7 @@ Flask-Security = { version = "5.7.1", extras = ["common", "fsqla"] }
 Flask-SocketIO = "^5.3.6"
 gevent = "^23.9.1"
 markupsafe = "^2.1.5"
-mxcubecore =  ">=1.419.0"
+mxcubecore =  "<2.0.0"
```
walesch-yan, line 48: Maybe I'm a little late for the discussion, but shouldn't it be `>=1.419.0,<2.0.0`

marcus-oscarsson: Yes, indeed !

PR #2033, opened 2026-03-03
remove dangerouslySetInnerHTML

walesch-yan:
This PR removes a `dangerouslySetInnerHTML` call within the centring message box. It was used to make it possible to add a line break with an html tag (`<br/>`). The PR replaces the possible line-break html tags with `\n`. Then css will handle the line break. 

Currently, this is not a security vulnerability in our code, since the messages cannot be broadcasted to other users nor the back-end but generally an improvement of the usage of best -practices (separates data in state from presentation (HTML) ) and reduces possible future risks, which may come from a code refactoring.

### DIFF
--- a/ui/src/components/SampleView/SampleImage.jsx
+++ b/ui/src/components/SampleView/SampleImage.jsx
@@ -60,7 +60,6 @@ class SampleImage extends React.Component {
     this.configureGrid = this.configureGrid.bind(this);
     this.updateGridResults = this.updateGridResults.bind(this);
     this.selectedGrid = this.selectedGrid.bind(this);
-    this.centringMessage = this.centringMessage.bind(this);
     this.selectShape = this.selectShape.bind(this);
     this.deSelectShape = this.deSelectShape.bind(this);
     this.clearSelection = this.clearSelection.bind(this);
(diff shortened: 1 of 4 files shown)


ui/src/components/SampleView/SampleImage.jsx
```diff
@@ -892,7 +872,15 @@ class SampleImage extends React.Component {
...
-            <div>{this.centringMessage()}</div>
+            {this.props.videoMessageOverlay.show && (
+              <div
+                key={this.props.clickCentringClicksLeft}
+                id="video-message-overlay"
+                className={styles.videoMessageOverlay}
+              >
+                {this.props.videoMessageOverlay.msg}
+              </div>
+            )}
```
walesch-yan, line 883: I also took the liberty to remove the `centringMessage` function which simply rendered this `div`

ui/src/components/SampleView/SampleImage.module.css
```diff
@@ -42,6 +42,7 @@
   animation: fade 2s 2s forwards;
   opacity: 1;
   font-weight: bold;
+  white-space: pre-line;
```
walesch-yan, line 45: This change allows the line-break with `\n` in text

axelboc (approved):
Clever!

marcus-oscarsson:
Thanks, well done !

PR #2036, opened 2026-03-03
Fix cell counting related SampleView filling

matyson:
While trying to update the heatmap results after a mesh scan we noticed that the cell filling was not considering the cell counting method (e.g. zig-zag, inverse-zig-zag) to fill the grid's result. This was due to the `SampleImage` component not being initialized correctly from config params from `state.taskForm.defaultParameters.mesh.cell_counting` -> `state.taskForm.defaultParameters.mesh.acq_parameters.cell_counting`. 

Then, we included the countCellFun call inside the cell filling function to populate the correct matrix position according to the cellCount order.


marcus-oscarsson:
Thanks @matyson, well done !

PR #2037, opened 2026-03-03
fix displayImage parameters in SampleImage

matyson:
Fixing `displayImage` call in `SampleImage`, this function expects two args to be sent separately and not a single string.

### DIFF
--- a/ui/src/components/SampleView/SampleImage.jsx
+++ b/ui/src/components/SampleView/SampleImage.jsx
@@ -533,20 +533,20 @@ class SampleImage extends React.Component {
       const [cellIdxX, cellIdxY] = this.drawGridPlugin.getClickedCellIndex(
         shapeData,
         option.target,
-        option.pointer,
+        option.e,
       );
 
       const imgNum = this.drawGridPlugin.countCells(
         shapeData.cellCountFun,
-        cellIdxY,
         cellIdxX,
+        cellIdxY,
         shapeData.numRows,
         shapeData.numCols,
       );
 
       const { resultDataPath } = shapeData;
       if (resultDataPath !== undefined) {
-        this.props.displayImage(`${resultDataPath}&img_num=${imgNum}`);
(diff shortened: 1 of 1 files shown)


matyson:
Also, this [section](https://github.com/mxcube/mxcubeweb/blob/4c957f4e4ecb3d5497e6b201f2b795d3b537c58e/ui/src/components/SampleView/SampleImage.jsx#L533C1-L537C9) of the code seems to return `undefined` values and as far as i could investigate `option.pointer` does not have `.offsetX` and `.offsetY` so i thought maybe replace it with `option.e` but i still don't get the right indexes. Any ideas on how to fix it?

marcus-oscarsson:
> Fixing `displayImage` call in `SampleImage`, this function expects two args to be sent separately and not a single string.

Thanks :)

marcus-oscarsson:
> Also, this [section](https://github.com/mxcube/mxcubeweb/blob/4c957f4e4ecb3d5497e6b201f2b795d3b537c58e/ui/src/components/SampleView/SampleImage.jsx#L533C1-L537C9) of the code seems to return `undefined` values and as far as i could investigate `option.pointer` does not have `.offsetX` and `.offsetY` so i thought maybe replace it with `option.e` but i still don't get the right indexes. Any ideas on how to fix it?

Hm that call looks strange, I looks like the function `getClickedCellIndex` has this signature `getClickedCell(gd, shapeGroup, e)`. So I think `option.pointer` should simply be `e`. Would that make sense ?

matyson:
> Hm that call looks strange, I looks like the function `getClickedCellIndex` has this signature `getClickedCell(gd, shapeGroup, e)`. So I think `option.pointer` should simply be `e`. Would that make sense ?

Okay, i think i found the error, should be `option.e` and the indexes were inverted in `countCells` (though im not entirely sure because row and col names are getting me a bit confused, but now `imgNum` make sense here)

PR #2038, opened 2026-03-04
Draft: validate prefix instead of filename and stricter regexp

epanepucci:
- the current validation regexp allows characters that are not UNIX friendly because it validates against the filename prop that is used to indicate the *template* of the filenames that will be generated:

DataCollection.jsx +370
```javascript
  if (state.taskForm.sampleID) {
    fname = state.taskForm.taskData.parameters.fileName;
  } else {
    const prefix = selector(state, 'prefix');
    fname = `${prefix}_[RUN#]_[IMG#]`;
  }
```

- this simplified regexp validates only against valid UNIX characters except the **'dot'** character

- it also validates the **prefix** prop which is editable by the user instead of the **filename**

### DIFF
--- a/ui/src/components/Tasks/validate.js
+++ b/ui/src/components/Tasks/validate.js
@@ -13,7 +13,7 @@ function validate(values, props) {
   // here we update the resolution limits based on the energy the typed in the form,
   // the limits come from a table sent by the client
 
-  const validFname = /^[\w#\-[\]{}]+$/u.test(props.filename);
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
It seems like a reasonable fix, why did you close it ?

epanepucci:
> It seems like a reasonable fix, why did you close it ?

I felt it needed more testing. But it is definetely a fix to be applied; I managed to collect data with a '#' symbol in it (which the Eiger/SIMPLON allows) and all hell broke loose :-D

PR #2040, opened 2026-03-05
Fix displayImage command and validation

matyson:
More on `displayImage` executeCommand. I've noticed that the adapter `display_image` is expecting the param `img_num` instead of `imgNum` and that it did'nt had a type annotation so it was never getting validated. Also, for the `path` arg the `validate_input_str` does not handle the case if the string is a valid path string. So, for the changes:

- make input string validation include path chars ( "/", "-") but this IMO should be handled in a dedicated validator for path strings, which i don't know if it is in development along with the pydantic schemas that could be great addition to the resource handler validation. Please, share thoughts on this one.
- camelCase to snake_case object call in reducer: i wonder if this one could be automatically converted by the `resource_handler` ??
- include type hint in img_num param to get it validated by resource handler

### DIFF
--- a/mxcubeweb/core/adapter/detector_adapter.py
+++ b/mxcubeweb/core/adapter/detector_adapter.py
@@ -2,6 +2,7 @@ from typing import ClassVar
 
 from mxcubecore import HardwareRepository as HWR
 from mxcubecore.HardwareObjects.abstract import AbstractDetector
+from pydantic import BaseModel, Field, FilePath
 
 from mxcubeweb.core.adapter.adapter_base import AdapterBase
 from mxcubeweb.core.models.configmodels import ResourceHandlerConfigModel
@@ -10,6 +11,11 @@ resource_handler_config = ResourceHandlerConfigModel(
     commands=["display_image"],
     attributes=["data", "get_value"],
 )
+
+
+class DisplayImageParams(BaseModel):
+    path: FilePath = Field(description="Path to images' masterfile")
+    img_num: int = Field(description="Index of the image to display")
 
 
 class DetectorAdapter(AdapterBase):
@@ -36,9 +42,10 @@ class DetectorAdapter(AdapterBase):
     def state(self):
         return self._ho.get_state().name.upper()
 
-    def display_image(self, path: str, img_num) -> dict:
+    def display_image(self, params: DisplayImageParams) -> dict:
         """Notify ADXV and/or Braggy of the image to display."""
         res = {"image_url": ""}
+        path, img_num = str(params.path), params.img_num
 
         if path:
             fpath, img = HWR.beamline.detector.get_actual_file_path(path, img_num)
--- a/ui/src/actions/general.js
+++ b/ui/src/actions/general.js
@@ -42,7 +42,7 @@ export function displayImage(path, imgNum) {
       'detector',
       'detector',
       'display_image',
-      { path, imgNum },
+      { path, img_num: imgNum },
     );
     window.open(data.image_url, 'braggy');
   };


axelboc:
> camelCase to snake_case object call in reducer: i wonder if this one could be automatically converted by the `resource_handler` ??

Just my two cents on this: I think it's better to be explicit as it helps when searching through the codebase for instance. I've had experiences with frameworks that do this sort of casing conversions in the past and it's always been a nightmare.

Hopefully one day we can generate TypeScript types from the Pydantic models to avoid this sort of issues.

matyson:
> Hopefully one day we can generate TypeScript types from the Pydantic models to avoid this sort of issues.

Oh yeah, that would be *the dream*! ( life would already be better just having TS 😭)

mxcubeweb/core/server/resource_handler.py
```diff
@@ -28,15 +28,18 @@ def valid_object_id(object_id: str) -> bool:
...
+    posix path characters such as dot (.), backslash (/) and dash (-).
 
     Args:
         input_string (str): The string to validate.
 
     Returns:
         bool: True if the string is valid, False otherwise.
     """
-    pattern = r"^[a-zA-Z0-9._]*$"
+    pattern = r"^[a-zA-Z0-9._/-]*$"
```
marcus-oscarsson: I have a doubt about `/`, where do you need that, I imagine for a path ?

We should perhaps have a specific validation made for paths, to avoid i.e `../../` and other sort of possibly "nasty" things ?

matyson: The `display_image(self, path: str, img_num: int)` function expects a path to the images masterfile and the validator blocks it. I also think that it should be a different type of validation for this specific type, how do you propose to include this custom type?

mxcubeweb/core/server/resource_handler.py
```diff
@@ -28,15 +28,18 @@ def valid_object_id(object_id: str) -> bool:
 
 
 def validate_input_str(input_string: str) -> bool:
-    """Validate that input string contains only alphanumeric characters and/or dot (.).
+    """Validate input string.
+
+    Ensure that input string contains only alphanumeric characters and/or
+    posix path characters such as dot (.), backslash (/) and dash (-).
```
mockoocy: I think it deprecates this comment :D 


https://github.com/matyson/mxcubeweb/blob/c46c895dc20379b83b5b80c606bb1aa50582c994/mxcubeweb/core/server/resource_handler.py#L383-L387

mxcubeweb/core/server/resource_handler.py
marcus-oscarsson: If you give `display_image` a Pydantic Model for the path it will be validated against the PyDantic model. That model could be a tailored to a string. The downside/limitation, at the moment, is that we only allow one pydantic model per function so that would need to also model img_num. Something like `DisplayImageInput`. 

Example: https://github.com/matyson/mxcubeweb/blob/c46c895dc20379b83b5b80c606bb1aa50582c994/mxcubeweb/core/adapter/actuator_adapter.py#L63

marcus-oscarsson: So the mode would be something like this:
```
class DisplayImageInput(BaseModel):
    path: str = Field("", description="")      # Or perhaps `PathModel` with `field_validator` return instead of  str`
    img_num: int = Field("", description=")
```

mockoocy: AFAIK You can do something like
```python
class Params(BaseModel):
    path: Path
```

It would coerce strings into `Path` objects too. 
Still there needs to be some kind of mitigation against different types of nastiness (e..g resolving paths), but that's more general to working with paths overall :)

marcus-oscarsson: If we could start already with this that would be nice, I think using the `field_validator` is more explicit and clear compared to a complex regexp or simply using `Path`.

marcus-oscarsson: Anyhow, I'm sure you will come up with something nice if you go for validating paths.

marcus-oscarsson: > If you give `display_image` a Pydantic Model for the path it will be validated against the PyDantic model. That model could be a tailored to a string. The downside/limitation, at the moment, is that we only allow one pydantic model per function so that would need to also model img_num. Something like `DisplayImageInput`.
> 
> Example: https://github.com/matyson/mxcubeweb/blob/c46c895dc20379b83b5b80c606bb1aa50582c994/mxcubeweb/core/adapter/actuator_adapter.py#L63

@matyson: By the way I realized that the limit of one `PyDantic` model no longer exists, we can use any number of `PyDantic` arguments. You should be able to do: `display_image(self, path: PathModel, img_num: int)`

matyson: ok, in order to preserve the openapi spec generation i chose to turn all the params in a single pydantic model, and use the builtin `FilePath` type to validate if the path exists and if it's a file ( maybe this is too much validation, you tell me)

marcus-oscarsson: Looks pretty good, but what paths does `FilePath` allow ?

matyson: according to [docs](https://docs.pydantic.dev/2.3/usage/types/file_types/) it allow any type of existing file. I could narrow it to valid file extensions like `.h5` but i don't know if every facility aims for the same file agreements..

marcus-oscarsson: Ok, hm I'm more worried about that it would be possible to move out of the data root using i.e `../` or other possibly nasty tricks. That is perhaps better checked elsewhere ?

mockoocy: I think that the data root, can be accessed from `HWR.beamline.session`, right? If so I would be in favor of validating the path here, so that we would need to worry less about the functions that use this path later on.

I think doing a check like `path.resolve().is_relative_to(Path(data_dir).resolve())` would be good

matyson: well, i believe the path is generated by the data collection form's prefix field and that also can be validated for weird characters on the form side before even hitting the server. We can also check with a post validation if it stays inside a "safe directory" like `/safe/data/` so we don't even bother executing `display_img`

marcus-oscarsson: Ok, hm to be explored. This is a great fix already, thanks alot @matyson  !

marcus-oscarsson (approved):

PR #2042, opened 2026-03-10
Tightened up motor definitions and adapted virtual-motor names

rhfogh:
Follow-up PR from mxcubecore PR #1518. These PRs must be applied together.


rhfogh:
Withdrawn., This was a companion to a misconceuived mxcubecore, now withdrawn.

PR #2047, opened 2026-03-13
Improved GphlWorkflowDialog popup

rhfogh:
Improved GPhLWorkflowDialog popup

- Compacted and removed excessive space
- Improved some widgets
- Made popup resizable and improved behaviour under popup resizing
Effects on GPhL interface only

### DIFF
--- a/ui/src/containers/GphlWorkflowParametersDialog.jsx
+++ b/ui/src/containers/GphlWorkflowParametersDialog.jsx
@@ -240,7 +240,7 @@ function GphlWorkflowParametersDialog(props) {
       <Form
         noValidate
         validated={validated}
-        className="m-3"
+        className="m-1"
         onSubmit={(e) => handleSubmit(e)}
       >
         {ui_schema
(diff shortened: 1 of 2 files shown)


marcus-oscarsson:
You are the only ones using this so I trust that you have tested it ?

marcus-oscarsson:
Its a bit hard to follow exactly whats done without knowing what you wanted to do ?

marcus-oscarsson:
Otherwise it looks alright

rhfogh:
@marcus-oscarsson Yeah, I have tested it, but it still needs looking at, like all PRs.

Sorry for the lack of documentation. The original version of the popup included far too much whitespace, so that the indexing table popup shown after characterisation only had space for a couple of rows. Also the popup could not be rescaled by hand (and when that was changed the rescaling behaviour was wrong). So I wanted to

- Reduce the margin around fields, particularly within the indexing table, to reduce wasted space.
- Put field labels in line with number fields, rather than above them, to save space.
- Set the starting space of the indexing table to a sensible value (I settled for 30% of the popup height)
- Make the popup resizable,and set the resizing behaviour so that the indexing table took up both increases and decreases in available space and no space was wasted on expanding whitespace (I was only partially successful here, and managed to prove that Claude is by no means infallible).
- Add colour coding for indexing table rows matching the predicted symmetry (which had always been in the Qt version but had not been added in the web version). 
- Make the highlighting of the selected row much more visible.

marcus-oscarsson:
Thanks @rhfogh, I had a second look at the code and I got most what you wrote above :). The extraction of the function that generates the inputs threw me of to start with ..

rhfogh:
@marcus-oscarsson It would probably have thrown me as well. This an example of Claude in action. Except that after asking five times for Claude to fix the scaling and a problem with labels being partially hidden behind whitespace, i had to do my own trial-and-error to get to the end. Thanks for looking it over.

marcus-oscarsson:
:), ok. I sort of suspected that there was a AI involved at some stage, the `useEffect` logic looked AI`ish. It seems to have done a good job ...

PR #2052, opened 2026-03-17
[AddSampleDialog] - Switched order between sample name and acronym

marcus-oscarsson:
Switched order between sample name and acronym, so that the acronym comes before the name

image


axelboc (approved):
Ah yes, nice one 😁

PR #2053, opened 2026-03-17
Generic dialog look and feel update/improvement 

marcus-oscarsson:
Added some more margin and changed the boolean input from a check box to a switch, which looks nicer if the title of the input and title are on a separate rows. Having the checkbox and the title on the same row makes, in our opinion, the UI feel a bit un-balanced when there are a few controls in the dialog

**Before**
image

**After**
image

 - Removed the need of having a shape which is not necessarily a requirement for these forms
 - Added missing beam size data


beteva:
Looks nicer indeed.

axelboc (approved):
Makes sense

PR #2055, opened 2026-03-17
Attempt to make the click centering more robust

marcus-oscarsson:
Attempt to make the click centering more robust by adding a lock to avoid race conditions produced by clicks while a "centering step/click" is already being handled.


walesch-yan (approved):
Awesome!

PR #2056, opened 2026-03-17
Handling update of a sample state for specific samples

marcus-oscarsson:
This PR makes it possible to handle state changes of specific samples by listening to `infoChanged`. The sample state such as `on-gonio` and `in-puck` or `blacklisted` is displayed in the sample list.

~**Requires corresponding PR [#1530](https://github.com/mxcube/mxcubecore/pull/1530) on `mxcubecore`**~


beteva (approved):
Looks good

PR #2058, opened 2026-03-17
[SampleList] - Not selecting samples when clicking Add Tasks to samples

marcus-oscarsson:
Using the `Add Tasks to samples` button also had the action of selecting all samples. The wanted behavior is that the current selection whatever it is, is maintained.


walesch-yan (approved):
Looks good to me!

PR #2065, opened 2026-03-31
[Queue] - Fix for Unused local variable `node_id`

marcus-oscarsson:
Removed unused local variable `node_id`

### DIFF
--- a/mxcubeweb/core/components/queue.py
+++ b/mxcubeweb/core/components/queue.py
@@ -1814,8 +1814,6 @@ class Queue(ComponentBase):
         HWR.beamline.queue_manager.set_pause(False)
 
         if tindex in ["undefined", "None", "null", None]:
-            node_id = current_queue[sid]["queueID"]
-
             # The queue does not run the mount defined by the sample entry if it has no
             # tasks, so in order function as expected; just mount the sample
             if (


mockoocy (approved):

mockoocy:
Well caught :) I find it quite odd that ruff does not complain about it

marcus-oscarsson:
:), Well I find RUFF quite odd most of the time. Jokes a side, perhaps RUFF never had any reason to check that line ...

walesch-yan:
> Well caught :) I find it quite odd that ruff does not complain about it

I think this rule is only checked at the function scope and not the branch-level. So I guess because the variable is assigned and used within the `else` block, ruff does not flag it, as it sees one usage of the variable. But very interesting to know :sweat_smile:

marcus-oscarsson:
> > Well caught :) I find it quite odd that ruff does not complain about it
> 
> I think this rule is only checked at the function scope and not the branch-level. So I guess because the variable is assigned and used within the `else` block, ruff does not flag it, as it sees one usage of the variable. But very interesting to know 😅

Oh is that the case perhaps, good idea !. However that would be quite a big flaw for a linter / static syntax checker.

walesch-yan:
> > > Well caught :) I find it quite odd that ruff does not complain about it
> > 
> > 
> > I think this rule is only checked at the function scope and not the branch-level. So I guess because the variable is assigned and used within the `else` block, ruff does not flag it, as it sees one usage of the variable. But very interesting to know 😅
> 
> Oh is that the case perhaps, good idea !. However that would be quite a big flaw for a linter / static syntax checker.

It definitely is! I quickly checked and I guess the rule that should trigger this is [F841](https://docs.astral.sh/ruff/rules/unused-variable/) and indeed the description states that it is only applied to the function scope

PR #2066, opened 2026-03-31
Fixed typos

marcus-oscarsson:
Fixed a few typos

### DIFF
--- a/ui/src/serverIO.js
+++ b/ui/src/serverIO.js
@@ -65,7 +65,7 @@ const { dispatch } = store;
 class ServerIO {
   hwrSocket = null;
   loggingSocket = null;
-  connectionLostTimeout;
+  connectionLostTimeout = null;
 
   listen() {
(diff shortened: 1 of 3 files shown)


ui/src/serverIO.js
```diff
@@ -294,7 +294,7 @@ class ServerIO {
         const msg = `${numClicks}-Click Centring: \nSelect centered position or center`;
         dispatch(videoMessageOverlay(true, msg));
       } else {
-        const msg = 'Auto loop centring: \n Save position or re-center';
+        const msg = 'Auto loop centering: \n Save position or re-center';
```
beteva: This one was correct - we said we prefer the English spelling, not the American one. 😄

marcus-oscarsson:
```suggestion
        const msg = 'Auto loop centring: \n Save position or re-center';
```

mockoocy (approved):

PR #2068, opened 2026-04-01
Added code that makes it possible to split the ui.yaml into modules

marcus-oscarsson:
This PR proposes to split the `ui.yaml` into "ui modules". The format of each section/module is the same as with the monolithic `ui.yaml`. There will be a follow up PR that discusses the format and suggests some simpler improvements.

There is a script in the script folder that takes a monolithic `ui.yaml` and splits into "ui module" files and creates the corresponding new `ui.yaml`

A schema version was also added to be able to better handle transitions, lets hope we don't need to rely on it too much.

```
config_metadata:
  version: 1.0.0
modules:
  sample_list_view_modes: ui-modules/sample_list_view_modes.yaml
  sample_view: ui-modules/sample_view.yaml
  sample_view_motors: ui-modules/sample_view_motors.yaml
  sample_view_video_controls: ui-modules/sample_view_video_controls.yaml
  beamline_setup: ui-modules/beamline_setup.yaml
  camera_setup: ui-modules/camera_setup.yaml
  session_picker: ui-modules/session_picker.yaml
```


marcus-oscarsson:
.... awhh `pathlib` .... whats the deal ! i really prefer `os.path` !

marcus-oscarsson:
I also made this backward compatible so we fallback to the monolithic (merged) file format if the there is no modules (meaning version 0).

mxcubeweb/config.py
```diff
@@ -57,6 +59,60 @@ def load(path: str, schema: BaseModel):
...
+                    data = yaml.load(mf)
+
+                    if isinstance(data, dict):
+                        merged.update(data)
+
+            return merged
+
+        # Legacy format, the ui.yaml already contains module directly in the
+        # ui.yaml file.
+        logging.getLogger("HWR").warning(
```
marcus-oscarsson, line 102: I was considering adding a command line switch instead something like `--strict-config` that could also be used to force the PyDantic models when they are used for HardwareObjects. In this case it would raise an exception if the configuration format was the old one. 

I chose to go with something more "soft" and just issue a warning, not sure about the timer ...

mxcubeweb/config.py
```diff
@@ -57,6 +59,60 @@ def load(path: str, schema: BaseModel):
...
+        if not isinstance(ui_index, dict):
+            return merged
+
+        # New format (version 1.0.0) uses a top-level 'modules' mapping
+        # while the legacy format (version 0) has no modules section
+        modules = ui_index.get("modules")
+
+        # If modules are defined (version 1.0.0 and greater), load and merge
+        # them into the ui_index.
+        if modules is not None:
```
marcus-oscarsson, line 80: We merge all the module files to one big file (the ui.yaml is then on the new format), otherwise we simply return the parsed file (the old format)

marcus-oscarsson:
This is ready to merged if someone wants to have another look ?

demo/mxcube-web/ui-modules/sample_view_motors.yaml
```diff
@@ -0,0 +1,77 @@
...
+      precision: 3
+      suffix: mm
+    - label: Samp-Y
+      attribute: diffractometer.sampy
+      role: sampy
+      step: 0.1
+      precision: 3
+      suffix: mm
+    - label: Sample Horizontal
+      attribute: diffractometer.phiz
```
beteva:
```suggestion
      attribute: diffractometer.sample_horizontal
```

demo/mxcube-web/ui-modules/sample_view_motors.yaml
```diff
@@ -0,0 +1,77 @@
...
+      role: sample_horizontal
+      step: 0.1
+      precision: 3
+      suffix: mm
+    - label: Sample Vertical
+      attribute: diffractometer.phiy
```
beteva:
```suggestion
      attribute: diffractometer.sample_vertical
```

demo/mxcube-web/ui-modules/sample_view_motors.yaml
```diff
@@ -0,0 +1,77 @@
...
+      role: sample_vertical
+      step: 0.1
+      precision: 3
+      suffix: mm
+    - label: Zoom
+      attribute: diffractometer.zoom
+      role: zoom
+      step: 1
+      precision: 1
+      suffix: mm
```
beteva: To be removed - there should be no more zoom motor.

demo/mxcube-web/ui-modules/sample_view_motors.yaml
```diff
@@ -0,0 +1,77 @@
...
+      suffix: °
+    - label: Kappa Phi
+      attribute: diffractometer.kappa_phi
+      role: kappa_phi
+      step: 0.1
+      precision: 1
+      suffix: °
+    - label: X
+      attribute: diffractometer.focus
+      role: phix
```
beteva:
```suggestion
      role: focus
```
More in line with the diffractometer motor roles

demo/mxcube-web/ui-modules/sample_view_motors.yaml
```diff
@@ -0,0 +1,77 @@
...
+      suffix: mm
+    - label: Zoom
+      attribute: diffractometer.zoom
+      role: zoom
+      step: 1
+      precision: 1
+      suffix: mm
+    - label: Focus
+      attribute: diffractometer.sampx
+      role: focus
```
beteva:
```suggestion
      attribute: diffractometer.sample_focus
      role: sample_focus
```

demo/mxcube-web/ui-modules/beamline_setup.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - label: Cryo
+      attribute: cryo
+      precision: 2
+      suffix: k
+    - label: Wavelength
+      attribute: energy.wavelength
+      step: 0.01
+      precision: 2
+      suffix: Å
+    - label: Detector
```
beteva:
```suggestion
    - label: Detector Distance
```

beteva (approved):
Very good idea. Apart from the cosmetic changes (for the sake of a cleanup), the rest is great.
Thanks

PR #2071, opened 2026-04-07
update to React18

walesch-yan:
This PR updates the React version of MXCuBE to 18. Due to some previous preparation, only minimal changes were necessary to have a functional application.


ui/package.json
```diff
@@ -29,15 +29,15 @@
...
     "react-bootstrap": "2.8.0",
     "react-collapsible": "^2.10.0",
     "react-contexify": "6.0.0",
     "react-copy-to-clipboard": "^5.1.0",
-    "react-dom": "^17.0.2",
+    "react-dom": "^18.3.1",
     "react-draggable": "^4.5.0",
     "react-error-boundary": "4.1.2",
-    "react-hook-form": "7.45.0",
+    "react-hook-form": "7.71.2",
```
walesch-yan, line 40: This package needed to be updated as well, as the `Cypress` tests would not pass the login form and fail all the time on the old version. (The Login form uses the library, and since the test types the username much faster than a normal user could, the incorrect batching in the previous version led to an incorrect username being sent to the backend from time to time)

ui/src/index.jsx
```diff
@@ -20,4 +20,5 @@ function Root() {
   );
 }
 
-ReactDOM.render(<Root />, document.querySelector('#root'));
+const root = createRoot(document.querySelector('#root'));
+root.render(<Root />);
```
walesch-yan, line 24: This was the only real change to do, as React would otherwise still run in React17 mode

axelboc (approved):
🎉

marcus-oscarsson:
Thanks @walesch-yan !

PR #2072, opened 2026-04-07
update react-redux

walesch-yan:
Of course shortly after #2071 was merged, I found an issue for it :sweat_smile: 

So it seems that when running `pnpm install` everything is fine, but when running `pnpm install --force` there are two version instances of redux running, one on 5, the other on version 4. This can introduce the following bug:

When doing any task like characterisation or datacollection, the application will crash. 
[Screencast from 2026-04-07 15-49-42.webm](https://github.com/user-attachments/assets/e1362cb6-39b6-4223-9dd1-3d1288848aac)

This PR solves this in the short-term by updating our `React-Redux` version to version 8. However, we are currently limited, as other conflicts appear when updating `Redux` to version 5 (and `React-Redux` to version 9). Especially with the `redux-form` package (which is not maintained anymore). Another note: all components that crashed directly used the `redux-form` dependency. Hence, I would advocate replacing this package in the future.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -40,7 +40,7 @@
     "react-hook-form": "7.71.2",
     "react-icons": "^4.12.0",
     "react-lazyload": "^3.2.1",
-    "react-redux": "7.2.9",
+    "react-redux": "8.1.3",
     "react-router-dom": "^6.30.3",
     "react-zoom-pan-pinch": "3.3.0",
     "redux": "4.2.1",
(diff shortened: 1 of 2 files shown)


axelboc:
What's the justification for using `pnpm install --force`? This doesn't seem like a good idea.

If it works when you delete your `node_modules` and run `pnpm install` again, then there's no issue in my opinion — the lock file is sane.

walesch-yan:
It was mainly me trying out different installation methods on a test branch to verify everything working fine. 

So it was mainly to make sure that both methods would leave you with a working application environment, but I see that removing the node_modules first, I should end up with the same if using `--force` or not so probably not to helpful in that case to use it

walesch-yan:
Thinking about it maybe we can close this without merging I was a little too fast making this PR :sweat_smile:

PR #2073, opened 2026-04-07
update supported pnpm and nodejs version

walesch-yan:
This PR updates the supported pnpm version to 10 and nodejs versions to 22 and 24.

Mainly pnpm v9 support will be removed at the end of the month. (see [active pnpm version](https://endoflife.date/pnpm).

While our current nodejs version either already reached their eol or will reach it at the end of the month (see [node versions](https://nodejs.org/en/about/previous-releases))

### DIFF
--- a/conda-environment.yml
+++ b/conda-environment.yml
@@ -28,8 +28,8 @@ dependencies:
   # Development dependencies
   # ========================
 
-  - nodejs ==18.*|20.*  # Make sure it matches `ui/package.json`
-  - pnpm ==9.*  # Make sure it matches `ui/package.json`
+  - nodejs ==22.*|24.*  # Make sure it matches `ui/package.json`
+  - pnpm ==10.*  # Make sure it matches `ui/package.json`
   - poetry ==2.2.1  # Make sure it matches `.pre-commit-config.yaml` and `poetry.lock`
 
 
--- a/ui/.npmrc
+++ b/ui/.npmrc
@@ -1,5 +1,0 @@
-# Enforce Node and pnpm versions as specified in `package.json`
-engine-strict=true
-
-# Save exact dependency versions in `package.json`
-save-prefix=""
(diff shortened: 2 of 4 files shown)


axelboc:
Might be good to also migrate to the [new config format](https://pnpm.io/settings) and remove `ui/.npmrc`.

Here's what I suggest for `ui/pnpm-workspace.yml`:

```yml
# Enforce Node and pnpm versions specified in `engines` field of `package.json`
engineStrict: true

# Use whichever pnpm version is available in the environment
managePackageManagerVersions: false

# Fail install if any dependency has unreviewed build scripts
strictDepBuilds: true
onlyBuiltDependencies:
  - <remove `node_modules` and run `pnpm install` to see which dependencies to add to this list>

# Save exact dependency versions in `package.json`
savePrefix: ''
```

walesch-yan:
That's a good point, I did not think about it yesterday. Looking through it, your suggestion sounds good to me, maybe we can also set the new `minimumReleaseAge` to a day. Might be a good security buffer against newly published supply chain attacks?

walesch-yan:
Also for the `managePackageManagerVersions` for the moment we are not setting the packageManager version but only enforce the major version through the engines field, so I guess we can add it, but I think for the moment it has no real effect

axelboc:
- Oh you're right about `managePackageManagerVersions`, no need to set it in this project, sorry.
- For `minimumReleaseAge`, I found it a bit frustrating as it would "hide" releases and refuse to install versions even when I explicitly wanted to install them... But it's definitely a good thing to have if you can work around this.

walesch-yan:
I think you can do something like 
```bash
pnpm add some-package@latest --config.minimum-release-age=0
```
If you really need to install a version that came out that day. But I understand your concern, since it is definitely a trade-off.

axelboc (approved):

PR #2076, opened 2026-04-09
Changing filename validation to prefix validation in the Tasks forms

### LINKED ISSUE #2044: Validate "prefix" instead of the "filename" in the tasks forms
As it is now, task forms validate the `filename` field, which is not editable by the user, while the `prefix` part of the filename is editable as a separate field. At the same time, the validation message is attached to the profile field. 

Moreover, due to the filename being built from wildcards for substitution, the validation allows for '#' in the `prefix`, which may lead to path issues. 

Then, it is worth validating only the `prefix` part of the filename and allowing only valid path characters.

This is to (re)do this PR: https://github.com/mxcube/mxcubeweb/pull/2038 :

```
The current validation regexp allows characters that are not UNIX friendly because it validates against the filename prop that is used to indicate the template of the filenames that will be generated:
DataCollection.jsx +370

  if (state.taskForm.sampleID) {
    fname = state.taskForm.taskData.parameters.fileName;
  } else {
    const prefix = selector(state, 'prefix');
    fname = `${prefix}_[RUN#]_[IMG#]`;
  }
this simplified regexp validates only against valid UNIX characters except the 'dot' character

it also validates the prefix prop which is editable by the user instead of the filename
```

piogor:
- adding a warning for an empty prefix
- fixes subdir validation,
- also removing some legacy comments.

Fixes #2044, follows this PR #2038:

```
The current validation regexp allows characters that are not UNIX friendly because it validates against the filename prop that is used to indicate the template of the filenames that will be generated:
DataCollection.jsx +370

  if (state.taskForm.sampleID) {
    fname = state.taskForm.taskData.parameters.fileName;
  } else {
    const prefix = selector(state, 'prefix');
    fname = `${prefix}_[RUN#]_[IMG#]`;
  }
this simplified regexp validates only against valid UNIX characters except the 'dot' character

it also validates the prefix prop which is editable by the user instead of the filename
```

### DIFF
--- a/ui/src/components/Tasks/validate.js
+++ b/ui/src/components/Tasks/validate.js
@@ -10,18 +10,16 @@ function validate(values, props) {
   }
   const currEnergy = Number.parseFloat(values.energy);
   const currTransmission = Number.parseFloat(values.transmission);
-  // here we update the resolution limits based on the energy the typed in the form,
-  // the limits come from a table sent by the client
-
-  const validFname = /^[\w#\-[\]{}]+$/u.test(props.filename);
 
   const emptyField = 'field is empty';
 
-  if (!validFname) {
+  const validPrefix = !values.prefix || /^[\w-]+$/u.test(values.prefix);
+
+  if (!validPrefix) {
     errors.prefix = INVALID_CHAR_MSG;
   }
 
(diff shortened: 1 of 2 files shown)


marcus-oscarsson:
Thanks !

PR #2077, opened 2026-04-10
update fabric

walesch-yan:
The fabric version we were using introduced a lot of vulnerabilities and was quite outdated.

This PR updates the package to the latest version. In between, a lot has changed within the library and our code needed updates in many places to be compatible. I added details as comments on this PR

### DIFF
--- a/ui/src/components/SSXChip/ChipCanvasUtils.js
+++ b/ui/src/components/SSXChip/ChipCanvasUtils.js
@@ -1,8 +1,6 @@
-import 'fabric';
+import { ActiveSelection, Canvas, FabricText, Rect } from 'fabric';
 
 import { showContextMenu } from './ChipContextMenu';
-
-const { fabric } = globalThis;
 
 export function renderChip(
   chipSizeX,
@@ -19,7 +17,7 @@ export function renderChip(
   const objects = [];
 
   objects.push(
-    new fabric.Rect({
+    new Rect({
       top: 0,
       left: 0,
       width: chipSizeX,
@@ -35,7 +33,7 @@ export function renderChip(
       lockSkewingY: true,
       lockRotation: true,
       hoverCursor: 'arrow',
-      type: 'CHIP',
+      customType: 'CHIP',
       objectIndex: [],
     }),
   );
@@ -50,7 +48,7 @@ export function renderChip(
     }
 
     objects.push(
-      new fabric.Text(label, {
+      new FabricText(label, {
         top: offset / 2,
         left:
           ci * (blockSizeX + spacing) + offset + blockSizeX + blockSizeX / 4,
@@ -69,7 +67,7 @@ export function renderChip(
         lockSkewingY: true,
         lockRotation: true,
         hoverCursor: 'pointer',
-        type: 'BLOCK',
+        customType: 'BLOCK',
       }),
     );
   }
@@ -82,7 +80,7 @@ export function renderChip(
     }
 
     objects.push(
-      new fabric.Text(label, {
+      new FabricText(label, {
         top: ri * (blockSizeY + spacing) + offset + blockSizeY,
         left: offset / 2,
         fontSize: blockSizeX * 0.7,
@@ -100,7 +98,7 @@ export function renderChip(
         lockSkewingY: true,
         lockRotation: true,
         hoverCursor: 'pointer',
-        type: 'BLOCK',
+        customType: 'BLOCK',
       }),
     );
   }
@@ -108,7 +106,7 @@ export function renderChip(
   for (let ri = 0; ri < rows; ri++) {
     for (let ci = 0; ci < cols; ci++) {
       objects.push(
-        new fabric.Rect({
+        new Rect({
           top: ri * (blockSizeY + spacing) + offset + blockSizeY,
           left: ci * (blockSizeX + spacing) + offset + blockSizeX,
           width: blockSizeX,
@@ -126,7 +124,7 @@ export function renderChip(
           lockSkewingY: true,
           lockRotation: true,
           hoverCursor: 'pointer',
-          type: 'BLOCK',
+          customType: 'BLOCK',
           objectIndex: [ri, ci],
         }),
       );
@@ -155,7 +153,7 @@ export function initChipCanvas(currentChipLayout) {
   const canvasWidth = numCols * (blockSizeX + spacing) + offset + blockSizeX;
   const canvasHeight = numRows * (blockSizeY + spacing) + offset + blockSizeY;
 
-  const chipCanvas = new fabric.Canvas('chip-canvas', {
+  const chipCanvas = new Canvas('chip-canvas', {
     width: canvasWidth,
     height: canvasHeight,
     backgroundColor: '#CCC',
@@ -167,7 +165,7 @@ export function initChipCanvas(currentChipLayout) {
     renderOnAddRemove: false,
   });
 
-  const detailCanvas = new fabric.Canvas('chip-detail-canvas', {
+  const detailCanvas = new Canvas('chip-detail-canvas', {
     width: canvasWidth,
     height: canvasHeight,
     backgroundColor: '#CCC',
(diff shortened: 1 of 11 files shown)


ui/src/components/SampleControls/SnapshotControl.jsx
```diff
@@ -6,8 +7,6 @@ import { sendTakeSnapshot } from '../../api/sampleview';
 import styles from './SampleControls.module.css';
 import { download } from './utils';
 
-const { fabric } = globalThis;
```
walesch-yan, line 9: First victory was to remove the `globalThis`

ui/src/components/SampleView/FabricObjects/grid.js
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * A custom Fabric.js object representing a grid.
+ * @typedef {fabric.IObjectOptions & GridOptions} Grid
+ * @extends FabricObject
+ */
+export class Grid extends FabricObject {
+  static type = 'Grid';
+  static ownDefaults = {
+    selected: false,
+  };
```
walesch-yan, line 27: For all of our custom fabric classes(`Grid`,`GridResult` and `GridGroup`), we have to extend `FabricObject`, as `fabric.utils.creatClass` is removed. Fabric follows now the native ES6 classes. This meant a few general changes for the classes, especially for the constructors and static properties.

Additionally, I moved every class into its own file (mainly due to linting

ui/src/components/SampleControls/SnapshotControl.jsx
```diff
@@ -1,3 +1,4 @@
+import { FabricImage } from 'fabric';
```
walesch-yan, line 1: One important note as well is that fabric's `Text` and `Image` have been renamed `FabricImage` and `FabricText`.

ui/src/components/SampleView/DrawGridPlugin.jsx
```diff
@@ -680,6 +478,8 @@ export default class DrawGridPlugin {
       id: gd.id,
       top,
       left,
+      originX: 'left',
+      originY: 'top',
```
walesch-yan, line 482: This is one of the biggest changes; the origin of any fabric object is now at `center, center,` while `top, left` was the previous default. For now, I set the `originX` and `originY` to avoid any unforeseen changes. In the future, we might have to adjust how we render them, as setting the origin becomes deprecated as well within fabric.

ui/src/components/SampleView/SampleImage.jsx
```diff
@@ -343,17 +346,37 @@ class SampleImage extends React.Component {
...
     const group = this.canvas.getActiveObject();
-    const clickPoint = new fabric.Point(e.offsetX, e.offsetY);
+    const clickPoint = new Point(e.offsetX, e.offsetY);
     let ctxMenuObj;
     let objectFound = false;
     // Existing selection clicked
     if (
       group &&
-      group.type === 'activeSelection' &&
+      group instanceof ActiveSelection &&
```
walesch-yan, line 379: getting the `type` property is not reliable anymore, here I use `instanceof` instead

ui/src/components/SampleView/SampleImage.jsx
```diff
@@ -437,7 +460,10 @@ class SampleImage extends React.Component {
         ) {
           objectFound = true;
           this.selectShape([obj], e.ctrlKey);
-          ctxMenuObj = obj;
+          ctxMenuObj = {
+            type: this.getShapeType(obj),
+            id: obj.id,
+          };
```
walesch-yan, line 466: another big change: within the `contextMenu`, we rely on the type to tell us which button options we render. For the default case, we just passed the fabric object. However, in this case, the type was not accessible. Since we used a custom object in all other cases, I created this `getShapeType`, which should return the type we need for the custom object to pass to the context menu

ui/src/components/SampleView/SampleImage.jsx
```diff
@@ -467,7 +493,10 @@ class SampleImage extends React.Component {
                 ctxMenuObj = { type: 'GridGroup', gridData, id: obj.id };
               }
             } else {
-              ctxMenuObj = obj;
+              ctxMenuObj = {
+                type: this.getShapeType(obj),
+                id: obj.id,
+              };
```
walesch-yan, line 499: Same as above

ui/src/components/SampleView/shapes.js
```diff
@@ -81,11 +79,11 @@ export function makeLine(
...
   id,
   text,
-  type,
+  customType,
   hover = 'crosshair',
 ) {
-  return new fabric.Line([x1, y1, x2, y2], {
-    fill: col,
+  return new Path(`M ${x1} ${y1} L ${x2} ${y2}`, {
+    fill: null,
```
walesch-yan, line 86: fabric's `Line` object is deprecated, so I switched to a `Path`

ui/src/components/SampleView/shapes.js
```diff
@@ -96,8 +94,7 @@ export function makeLine(
     lockScalingFlip: true,
     lockScalingX: true,
     lockScalingY: true,
-    hasRotatingPoint: false,
-    type,
+    customType,
```
walesch-yan, line 97: passing the `customType` as additional prop, as `type` is for internal use only

ui/src/components/SampleView/shapes.js
```diff
@@ -107,15 +104,21 @@ export function makeLine(
...
+  }
+  const [startPoint, ...rest] = commands;
+  const [, x1, y1] = startPoint;
+  const last = rest[rest.length - 1];
+  const [, x2, y2] = last;
+
+  const angle = (Math.atan2(y2 - y1, x2 - x1) * 180) / Math.PI;
+  return new Triangle({
+    left: x2,
+    top: y2,
```
walesch-yan, line 121: These changes were necessary, as we replaced the `Line` object by a `Path` object.

ui/src/components/SampleView/shapes.js
```diff
@@ -127,19 +130,28 @@ export function makeArrow(line, col, _select, _id, hover = 'crosshair') {
...
+  }
+  const [startPoint, ...rest] = commands;
+  const [, x1, y1] = startPoint;
+  const last = rest[rest.length - 1];
+  const [, x2, y2] = last;
+
+  const angle = (Math.atan2(y2 - y1, x2 - x1) * 180) / Math.PI;
+  return new Rect({
+    left: x1,
+    top: y1,
```
walesch-yan, line 154: Similar changes as for `makeArrow`

ui/src/components/SSXChip/ChipCanvasUtils.js
```diff
@@ -185,11 +183,11 @@ export function initChipCanvas(currentChipLayout) {
     if (event.button === 3) {
       let selection = [];
 
-      if (object.type === 'BLOCK') {
+      if (object.customType === 'BLOCK') {
```
walesch-yan, line 186: Here the switch from `type` to `customType` becomes clear. `type` is `static` in the new version and the prop is ignored during the creation of the element. However, here we have two types of rectangles, a `BLOCK` and a `CHIP`. Hence, I used the customType

mockoocy:
I will read into it later, but I am very glad someone took their time updating Fabric 🥳🥳
Was planning to do it one day, as it has been really painful to touch the Fabric v5 code.

marcus-oscarsson (approved):

walesch-yan:
> I will read into it later, but I am very glad someone took their time updating Fabric 🥳🥳 Was planning to do it one day, as it has been really painful to touch the Fabric v5 code.

I completely agree, it took some time as well and was quite the painful update. Maybe it makes sense to review the whole drawing logic at some point as well, the goal was to keep it to the original but using the newer version (and of course that everything should still be working;))

ui/src/components/SampleView/SampleImage.jsx
walesch-yan: For this @axelboc also suggested (and I agree), we should try to find a better designed context menu object to improve the communication between the context menu and the canvas. This would go over the scope of this PR, but good to keep in mind

PR #2078, opened 2026-04-10
Fixed link to test badge

marcus-oscarsson:
The link to the test badge was broken


mockoocy:
It seems to be already fixed? I guess this PR could be closed then.

PR #2082, opened 2026-04-15
Implement abort_centring option at hwrSocket

pedrobenetton:
This allows the 3-click centring to be aborted by mxcubecore, which will be useful for LNLS.

### DIFF
--- a/ui/src/serverIO.js
+++ b/ui/src/serverIO.js
@@ -39,6 +39,7 @@ import {
 } from './actions/sampleChanger';
 import { updateSampleState } from './actions/sampleGrid';
 import {
+  abortCentring,
   saveMotorPosition,
   setBeamInfo,
   setCurrentPhase,
@@ -139,6 +140,10 @@ class ServerIO {
 
     this.hwrSocket.on('update_shapes', (record) => {
       dispatch(setShapes(record.shapes));
+    });
+
+    this.hwrSocket.on('abort_centring', () => {
+      dispatch(abortCentring());
     });
 
     this.hwrSocket.on('update_pixels_per_mm', (record) => {


beteva:
Hi @pedrobenetton, Usually it is enough to click again on the 3-click centring button to cancel the centring.
Could you, please, explain us why do you need an extra abort?

pedrobenetton:
> Hi @pedrobenetton, Usually it is enough to click again on the 3-click centring button to cancel the centring. Could you, please, explain us why do you need an extra abort?

Hi! This is purely a UI/UX scenario. Sometimes users don't understand that they have to click on the icon one more time (or press esc) to exit the centring. All users don't want to save the centring point after it finishes or click on the screen to restart. This implementation would allow me to abort the centring (from mxcubecore) after the last movement is done, and this whole scenario would no longer be a concern for us.

beteva:
> This implementation would allow me to abort the centring (from mxcubecore) after the last movement is done, and this whole scenario would no longer be a concern for us

Sorry but I still do not understand what would be the criteria of aborting. I am afraid that will make mess in the treatment of the signals, sent after a centring. The point of having different scenarios is precisely that we know afterwards what we do with the centred point (discard, save,,,)

pedrobenetton:
> > This implementation would allow me to abort the centring (from mxcubecore) after the last movement is done, and this whole scenario would no longer be a concern for us
> 
> Sorry but I still do not understand what would be the criteria of aborting. I am afraid that will make mess in the treatment of the signals, sent after a centring. The point of having different scenarios is precisely that we know afterwards what we do with the centred point (discard, save,,,)

No problem, I will try to explain it better.
A feedback that I have received from beamline staff and also from all beamline users is that they would like the 3-click centring to only perform the movement part, where they click in a point of the screen, wait for it to move, click again... and so on.
They have asked me to **always** skip what happens afterwards, so that they don't have to press Esc to exit the centring. They also want to always discard the centring point that is generated.

After some studying, I have seen that some signals called at hwrSocket also abort centring (basically all the ones that run `stopQueue`), so I thought it would be ok to implement something that only aborts centring.
The idea is that, after the movement is done (via bluesky and at mxcubecore), I would always abort the centring. There is really no logic for "if something happens, abort centring".

Again, this is just something that has been asked to me and I think it doesn't interfere with the rest of the application. HardwareObjects from other beamlines will never call this signal and nothing changes for them. If you think it is not a good idea at all, then I will simply tell them to keep pressing Esc everytime the centring ends 😄

marcus-oscarsson:
So if I understand this correctly what they want is to simply perform one round of 3 click centring and then explicitly click the button again to restart.

If thats the case I would perhaps go for an option instead, i.e `click_centring_rounds=None|0...|4` or maybe `loop_click_centring=False`. Like this we can control the behavior ?

marcus-oscarsson:
@pedrobenetton: Did you have an opportunity to try what we suggested above, in that case did it work as you wanted ?

pedrobenetton:
> @pedrobenetton: Did you have an opportunity to try what we suggested above, in that case did it work as you wanted ?

@marcus-oscarsson I started looking at it but still need more time, I will push when I get something interesting

pedrobenetton:
> So if I understand this correctly what they want is to simply perform one round of 3 click centring and then explicitly click the button again to restart.
> 
> If thats the case I would perhaps go for an option instead, i.e `click_centring_rounds=None|0...|4` or maybe `loop_click_centring=False`. Like this we can control the behavior ?

Hi @marcus-oscarsson, after reading this message again, I think there is a difference: what we want is to always exit the 3-click centring after the 3rd click is done, without the option of starting a new 3-click centring round. Do you think there is a simpler way to do this?

From this PR, I would only need to have:

`emit("abort_centring", namespace="/hwr")`

at mxcubecore.

marcus-oscarsson:
Sorry again for taking some time to look at this. Looks good, its another solution that will work well, and might be useful in some other cases :)

marcus-oscarsson:
You can go head and rebase and merge :)

marcus-oscarsson (approved):

marcus-oscarsson:
Well done !

PR #2087, opened 2026-04-21
Fixed issue with prefix validation

marcus-oscarsson:
The validation was previously changed in https://github.com/mxcube/mxcubeweb/pull/2076. However the prefix also needs to be able to contain curly braces, hashes and brackets to function properly when using the sample list and collecting over several samples.


mockoocy (approved):
Seems okay to me. I guess the thumbs up from @piogor is some sort of approval too :D

PR #2089, opened 2026-04-28
Fix display_image braggy calls

matyson:
The `display_image` method inside the detector adapter checks for `self.app.config` attributes, which should be `self.app.CONFIG`.

Also, since `.braggy` attribute is Optional in the config model, i included an additional check for when it is None before checking for the `USE_BRAGGY` flag


walesch-yan (approved):
That's a good catch!

marcus-oscarsson:
Yes, indeed thanks !

PR #2091, opened 2026-04-28
Ensure attrs >=21.3.0 in the conda environment.

dominikatrojanowska:
`attr.setters.pipe` was introduced in `attrs` 21.3.0, so this ensures pips's vendored `rich` can import it before poetry starts managing packages.


marcus-oscarsson:
I don't think this is needed any longer, it seems to have been solved in the `attrs` side

PR #2092, opened 2026-04-29
update SnapshotControl, to work with latest Fabric

walesch-yan:
Since the last `Fabric` update, it seems that taking a snapshot using the button on the `SampleControls`, does not work anymore. The reason lies on an old function call to `setBackgroundImage` of the fabric canvas, which was removed, instead, we can set the backgroundImage directly and we also have to reset the origin of the `Image` we create


marcus-oscarsson:
Well done !

PR #2101, opened 2026-05-12
Allow defining controls for generic NState hardware objects.

mockoocy:
Makes it possible to create a sample view select widget via ui.yaml config, like so:
```yaml
sample_view:
  id: sample_view
  components:
    - attribute: beam.definer
      label: Beam Focus
```
works for NState HWOs only for now.

Example:
image

(front light is a weird choice, but went with whatever is available in the demo beamline :))

### DIFF
--- a/ui/src/components/SampleView/NStateSelect.jsx
+++ b/ui/src/components/SampleView/NStateSelect.jsx
@@ -1,0 +1,44 @@
+import { Form } from 'react-bootstrap';
+import { useDispatch, useSelector } from 'react-redux';
+
+import { setAttribute } from '../../actions/beamline';
+import styles from './NStateSelect.module.css';
+
+/**
+ *
+ * @typedef {Object} Props
+ * @property {string} name - The name of the hardware object to control.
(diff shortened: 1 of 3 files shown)


marcus-oscarsson:
That's nice :) Thanks !

marcus-oscarsson:
I guess the need from your side is beam focusing as hinted by the configuration. It would be interesting if we could use the same for `Phase Control` and `Beam Size`

marcus-oscarsson:
It would be worth mentioning which components that are supported in the documentation:

   https://mxcubeweb.readthedocs.io/en/stable/config/ui.html

mockoocy:
> I guess the need from your side is beam focusing as hinted by the configuration. It would be interesting if we could use the same for `Phase Control` and `Beam Size`

I took a surface-level look and these values don't come from `NStateAdapter` nor are they accessed simarily. I would like to make it more unified too :)

> It would be worth mentioning which components that are supported in the documentation:
Absolutely, will be there soon-ish :)

PR #2108, opened 2026-05-19
Add tooltip to Sample View select widgets

mockoocy:
Made NStateSelect more generic, allowing to share the same code for NState hardware objects along the existing inputs for phase and aperture.
Also adds a tooltip for these based off of description field in sample_view.components field inside ui.yaml config file.

The TooltipTrigger has been slightly modified to not render the tooltip in case no content is provided - it is helpful for the case where no tooltip has been provided for the sample view widget.

### DIFF
--- a/demo.yaml/mxcube-web/ui.yaml
+++ b/demo.yaml/mxcube-web/ui.yaml
@@ -13,8 +13,14 @@ sample_view:
   components:
     - attribute: phase_control
       label: Phase Control
+      tooltip: >
+        Use this widget to trigger a concerted change
+        in the diffractomer devices to reach the  desired phase.
     - attribute: beam_size
       label: Beam size
+      tooltip: >
+        Use this widget to trim the beam just before it
+        hits the sample.
 
 sample_view_motors:
   id: sample_view_motors
--- a/ui/src/components/SampleView/PhaseInput.jsx
+++ b/ui/src/components/SampleView/PhaseInput.jsx
@@ -1,40 +1,38 @@
-import { Form } from 'react-bootstrap';
 import { useDispatch, useSelector } from 'react-redux';
 
 import { changeCurrentPhase } from '../../actions/sampleview';
-import styles from './PhaseInput.module.css';
+import { NStateSelect } from './NStateSelect';
 
-function PhaseInput() {
+const READ_ONLY_OPTIONS = ['Unknown'];
+
+/**
+ * Source hook for the diffractometer phase.
+ *
+ * @returns {NStateSource}
+ */
+
+/**
+ * @typedef {Object} Props
+ * @property {string?} tooltip - Optional hover tooltip text.
+ *
+ * @param {Props} props
+ */
+function PhaseInput({ tooltip }) {
   const dispatch = useDispatch();
 
-  const value = useSelector((state) => state.sampleview.currentPhase);
-  const options = useSelector((state) => state.sampleview.phaseList);
-  const isBusy = useSelector(
-    (state) => state.beamline.hardwareObjects.diffractometer?.state === 'BUSY',
-  );
-
   return (
-    <Form.Select
+    <NStateSelect
       id="PhaseInput"
-      className={styles.select}
-      value={value}
-      data-busy={isBusy || undefined}
-      onChange={(evt) => {
-        dispatch(changeCurrentPhase(evt.target.value));
-      }}
-    >
-      <option hidden key="Unknown">
-        Unknown
-      </option>
-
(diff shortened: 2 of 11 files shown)


ui/src/components/SampleView/nStateSources.js
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * Source hook for an arbitrary NState hardware object exposed in
+ * `state.beamline.hardwareObjects`.
+ *
+ * @param {string} name - Hardware object name (e.g. "frontlight").
+ * @returns {NStateSource}
+ */
+export function useNStateHO(name) {
+  const dispatch = useDispatch();
+  const ho = useSelector((state) => state.beamline.hardwareObjects[name]);
```
axelboc: I would maybe assert that `ho` is defined here — i.e. throw an error if it's `undefined` — as it points to an error in the codebase or in the configuration and this needs to surface quickly.

ui/src/components/SampleView/PhaseInput.jsx
```diff
@@ -1,40 +1,22 @@
...
-        // when the diffractometer hardware object sets it.
-        <option key={option} hidden={option === 'Unknown'}>
-          {option}
-        </option>
-      ))}
-    </Form.Select>
+      {...usePhase()}
+      readOnlyOptions={READ_ONLY_OPTIONS}
+      description={description}
+    />
```
mockoocy, line 35: I've created these small components, so that I can safely use hooks without violating the "rules of hooks" as react needs to have some guarantee that each hook will be invoked same amount of times ecah render.

axelboc: Makes sense, though since you have those components, I'm not convinced that you need the `usePhase` and `useAperture` hooks. I would inline the code in the components and pass the props explicitly to `NStateSelect`.

axelboc (commented):
Very nice refactoring. 👏  The tooltip could have been added in a separate commit but not big deal. Just two comments from my end.

ui/src/components/SampleView/PhaseInput.jsx
mockoocy: That's right. These hooks were there since initially I did not want to have these small components like `PhaseInput`, `ApertureInput` and I did not take a step back to remove them :D I've inlined the props that are gathered from these now.

mockoocy:
I've removed the hooks and made it so that parameters for `NStateSelect` are being retrieved inside each component. An error is also being thrown now for lack of matching hardware object (now error boundary will show the name of the missing hardware object instead of rendering a select input with no options).

demo.yaml/mxcube-web/ui.yaml
```diff
@@ -13,8 +13,14 @@ sample_view:
   components:
     - attribute: phase_control
       label: Phase Control
+      description: >
```
marcus-oscarsson: I'm thinking out load here, is `description` a good name. I guess the intended use of this is for tool-tips and that would be the use in most cases as far as I can see. So maybe `tooltip` instead of `description` ? 

Could there for instance be a situation where we need both a tool-tip and a description , not sure what that would be exactly.

marcus-oscarsson:
Thanks, nice :+1:

demo.yaml/mxcube-web/ui.yaml
axelboc: `hint` or `help` to convey intent?

mockoocy: I've first wanted to go with `tooltip`, but `description` field already exists in the pydantic model, albeit it seems unused, https://github.com/mxcube/mxcubeweb/blob/f3cc89d7d0f7db51d0b38a2a1320f1330460a31f/mxcubeweb/core/models/configmodels.py#L106

marcus-oscarsson: Yes, it was added for the use in tooltips and similair contexts. But then I thought as we are going to be a bit more careful with how we deal with this UI configuration, versioning etc. We might as well take another think about the names we use, sometimes its better if the name is more specific so that its usage is more or less evident and specific to a certain function/usage. It prevents mis-use and -understanding and makes it easier to introduce something later that actually is a "description" or whatever.

mockoocy: Fair enough, I've renamed `description -> tooltip`

marcus-oscarsson:
Really well done, thanks !

axelboc (approved):
Approved of course, sorry 😊

PR #2111, opened 2026-05-20
Changed single Overlap to Offset in label

rhfogh:
There was one forgotten instance of 'Overlap' that neded changng.

NB 'overlap still occurs in XSData... objects, mxcubecore.HardwareObjects.abstract.ISPyBValueFactory, ESRFMetadataManagerClient, mxcubecore.HardwareObjects.,ICATLIMS, and EMBLCollect.

If any of this needs to be updated the owners of tegh relevant code need to step in.

### DIFF
--- a/ui/src/components/Tasks/Characterisation.jsx
+++ b/ui/src/components/Tasks/Characterisation.jsx
@@ -180,7 +180,7 @@ class Characterisation extends React.Component {
                     list={this.props.detector_mode_list}
                   />
                 )}
-                <InputField propName="offset" label="Overlap" />
+                <InputField propName="offset" label="Offset" />
               </FieldsRow>
             </CollapsableRows>
           </Form>


marcus-oscarsson:
`ESRFMetadataManagerClient` and `XSData` is not something we can easily change as its used by other software that may be using overlap internally. 

@olofsvensson: For the XSData do you see any issue with changing to offset ?

marcus-oscarsson:
Thanks  :)

olofsvensson:
@marcus-oscarsson I think XSData is a bit like ISPyB - we could change it but it would require some effort which is not very well spent as both XSData and ISPyB will be deprecated in a not so distant future. So for the moment I would recommend to keep "overlap" in XSData (and ISPyB).

marcus-oscarsson:
Ok, thats what I thought as well ...

rhfogh:
I thought as much, with XSData and ISPyB. May they rest in peace.

PR #2113, opened 2026-05-28
Enable site diffraction viewer

matyson:
Enabling [#1616](https://github.com/mxcube/mxcubecore/pull/1616) if accepted

- Change `hw.beamline.collect.adxv_notify` call to `.site_diffraction_viewer`
- prevent opening _blank page when braggy is not enabled

### DIFF
--- a/mxcubeweb/core/adapter/detector_adapter.py
+++ b/mxcubeweb/core/adapter/detector_adapter.py
@@ -47,7 +47,7 @@ class DetectorAdapter(AdapterBase):
 
         if path:
             fpath, img = HWR.beamline.detector.get_actual_file_path(path, img_num)
-            HWR.beamline.collect.adxv_notify(fpath, img)
+            HWR.beamline.collect.display_image(fpath, img)
             fpath = HWR.beamline.session.get_path_with_proposal_as_root(fpath)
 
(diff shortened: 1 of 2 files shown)


mxcubeweb/core/adapter/detector_adapter.py
```diff
@@ -47,7 +47,7 @@ def display_image(self, params: DisplayImageParams) -> dict:
 
         if path:
             fpath, img = HWR.beamline.detector.get_actual_file_path(path, img_num)
-            HWR.beamline.collect.adxv_notify(fpath, img)
+            HWR.beamline.collect.site_diffraction_viewer(fpath, img)
```
marcus-oscarsson: As I mentioned in the other PR, I would just call it it `display_image` or similair. Then each site can decide what to do in their collect sub class, its understood that it can be site specific.

ui/src/actions/general.js
```diff
@@ -44,6 +44,8 @@ export function displayImage(path, imgNum) {
       'display_image',
       { path, img_num: imgNum },
     );
-    window.open(data.image_url, 'braggy');
+    if (data.image_url) {
+      window.open(data.image_url, 'braggy');
```
marcus-oscarsson, line 48: 👍

PR #2122, opened 2026-06-15
Add TypeScript

mockoocy:
Adds a package.json script and a CI job.
There is also a.ts file to check whether
the CI Job works well.

had to update the @esrf/eslint-package (and it's peer dependecy `eslint`) to be able to use `typescript==6.0.3`.


axelboc:
Dammmmn ... well at least `tsc` succeeds :joy: 

image

axelboc:
As expected, it's now linting JS/JSX files with type-checking information. Adding this to `eslint.config.js` should help:

```js
{
      files: ['**/*.{js,jsx}'],
      languageOptions: {
        parserOptions: {
          projectService: false,
        },
      },
    },
```

ui/eslint.config.js
```diff
@@ -15,6 +16,14 @@ const config = defineConfig([
...
     },
   },
+  {
+    // The codebase is still being migrated to TypeScript, so run type-aware
+    // linting on `.ts`/`.tsx` files only. `@esrf/eslint-config` enables the
+    // type-checked rule sets on all JS/TS files.
+    // This results to uncomprehensible amount of errors related to lack of typing.
+    files: ['**/*.{js,jsx,cjs,mjs}'],
+    extends: [tseslint.configs.disableTypeChecked],
+  },
```
mockoocy, line 26: It's less broken now :)

Did not go with the proposed stuff to disable `parserOptions.projectServices` as it leads to some error* related to `typeChecked` rule that is on by default with `@esrf/eslint-config`.
I am not sure if these `projectServices` should stay on for now, I am not sure if they do anything at all given `disableTypeChecked`.




*:
> Error: Error while loading rule '@typescript-eslint/await-thenable': You have used a rule which requires type information, but don't have parserOptions set to generate type information for this file. See https://tseslint.com/typed-linting for enabling linting with type information.
Parser: typescript-eslint/parser

axelboc: Ah cool, the rest looks manageable.

ui/src/components/SampleView/NStateSelect.tsx
```diff
@@ -27,7 +21,7 @@ export function NStateSelect({
   onSelect,
   tooltip,
   readOnlyOptions = [],
-}) {
```
mockoocy, line 30: I've rewritten this small component to TypeScript as an example. It is really easy with a JSDoc ready :) 

I am not sure why the linter prefers interfaces over types, but I'll put my preference aside here.

mockoocy:
Overall got the CI to pass, so that's cool.
I see no obvious issue after playing with it for a while.

The PR got quite big huge in the process tho. It consists of many small commits, so I could just split it somehow if necessary.

ui/vite.config.js
```diff
@@ -16,6 +16,7 @@ export default defineConfig({
     },
   },
   build: {
+    target: 'es2020',
```
mockoocy: Cypress pipeline failed without a specified target. Went with `es2020` as it seems like a safe choice, but idk what is a valid option here tbh.

ui/package.json
```diff
@@ -50,17 +51,19 @@
...
     "@types/react-redux": "7.1.33",
     "@vitejs/plugin-react-swc": "3.8.0",
     "cypress": "14.5.4",
     "cypress-terminal-report": "6.1.2",
-    "eslint": "9.35.0",
+    "eslint": "9.39.4",
     "prettier": "3.0.0",
     "prop-types": "15.8.1",
+    "typescript": "6.0.3",
+    "typescript-eslint": "8.59.0",
```
mockoocy, line 66: Had to provide it as a direct dependency in order to silence one of it's rules.

ui/src/components/SampleView/NStateSelect.tsx
axelboc: Mostly because it's the most common style used in TypeScript projects: https://typescript-eslint.io/rules/consistent-type-definitions#why-is-the-default-interface

ui/src/global.d.ts
axelboc, line 1: You can replace this with:

```
/// <reference types="vite/client" />
```

ui/vite.config.js
axelboc: I think it's because our Cypress version is a bit old. TypeScript 6 was only officially supported from [v15.14](https://docs.cypress.io/app/references/changelog#15-14-0). Maybe try upgrading? It should be straightforward.

axelboc (approved):
Hooray! 🙌 🫶

CONTRIBUTING.md
```diff
@@ -154,7 +154,9 @@ code base:
 
 ### Type hints
 
-We strongly encourage the usage of type hints
+We strongly encourage the usage of type hints. For the Python codebase they can be realised using the built-in type hint system.
+For the web front-end, it is considered best practice to write new code using TypeScript.
```
marcus-oscarsson, line 158: Nice :)

ui/src/components/SampleView/NStateSelect.tsx
marcus-oscarsson: I would prefer types as well, but Ill adapt to whats widely used.

marcus-oscarsson:
Well done, thanks alot for taking care of this one !

ui/src/components/SampleView/NStateSelect.tsx
axelboc: It can easily be changed in `eslint.config.js` if you guys want, I have to strong opinion personally.

PR #2128, opened 2026-06-19
Fix caching dependencies.

mockoocy:
For now cache key used for storing pnpm store and Cypress binary is
computed of a common prefix and a hash of unexisting file.
When `hashFiles()` has non-existent path provided, it defaults to
empty string; thus each cache shares a common `cache-Linux` key.

This bug has been encountered when attempting to bump cypress version,
in that case it can't be resolved as Cypress changes it's exact path
depending on version and a new one is not installed.
New binary is notinstalled, because postinstall scripts are ignored in
the CI job, somehow omitting `pnpm-workspace.yml` specifications.
This is related to a regression fixed in `pnpm@10.34.0`.
https://github.com/pnpm/pnpm/releases/tag/v10.34.0
Since that version `--prefix` is an alias of `--dir`, which we will use
for now.

### DIFF
--- a/.github/workflows/ui.yaml
+++ b/.github/workflows/ui.yaml
@@ -39,16 +39,16 @@ jobs:
         uses: actions/cache@v4
         with:
           path: |
-            ~/setup-pnpm/node_modules/.bin/store
             ~/.cache/Cypress
-          key: cache-${{ runner.os }}-${{ hashFiles('pnpm-lock.yaml') }}
+            ~/setup-pnpm/node_modules/.bin/store
+          key: cache-${{ runner.os }}-${{ hashFiles('ui/pnpm-lock.yaml') }}
           restore-keys: ${{ runner.os }}-
 
       - name: Install UI dependencies
-        run: "${MAMBA_EXE} run --name mxcubeweb pnpm --prefix ui install"
+        run: "${MAMBA_EXE} run --name mxcubeweb pnpm --dir ui install"
 
       - name: Run Prettier
-        run: "${MAMBA_EXE} run --name mxcubeweb pnpm --prefix ui prettier"
+        run: "${MAMBA_EXE} run --name mxcubeweb pnpm --dir ui prettier"
 
   lint:
     name: JS-Lint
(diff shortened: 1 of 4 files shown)


mockoocy:
Of course it is worth also figuring out why `pnpm-workspace.yml` is ignored in the CI job.

axelboc (commented):
Not convinced with running `cypress install` separately. I think one of the problems comes from the caching of the pnpm store. Have you tried simply changing the path of the store to `~/.pnpm-store`, outside of `node_modules` without any other changes?

```
pnpm config set store-dir .pnpm-store
```

and then in the cache job:

```
        with:
          path: |
            ~/.pnpm-store
            ~/.cache/Cypress
```

axelboc:
Hmm never mind, the `pnpm-workspace.yml` issue is very strange. I use a similar set up in another project and it works fine: https://github.com/silx-kit/h5web/blob/main/.github/actions/install/action.yml

axelboc:
Ah hang on, are you certain that fixing the `hashFiles` call is not sufficient on its own to fix the problem? When the pnpm store cache is restored, `pnpm install` does not run the Cypress `postinstall` script again since it thinks that Cypress is already installed.

mockoocy:
> Ah hang on, are you certain that fixing the `hashFiles` call is not sufficient on its own to fix the problem? When the pnpm store cache is restored, `pnpm install` does not run the Cypress `postinstall` script again since it thinks that Cypress is already installed.

That's right. 
I did some more digging now and could not find documentation for `--prefix` flag in pnpm.
I used `--dir` instead, and it seems to properly resolve path for `pnpm-workspace.yml` file now (i.e. pnpm in CI behaves as specified there)

mockoocy:
As for adding the cypress install / cache steps: I personally prefer to do it explicitly and get rid of one postinstall, but I don't have strong feelings for that. 

And the thread there: https://github.com/cypress-io/cypress/issues/33980#issuecomment-4676640671 
makes it look like there's possibility that postinstall could be removed someday, so it would keep us on a "safe side" at least.

axelboc:
Looks like there were a few recent regressions with `--prefix`: https://github.com/pnpm/pnpm/releases/tag/v10.34.0 — I haven't seen any mention that it's been deprecated but you're right it's not mentioned in the docs any more and it looks like it's just an alias for `--dir`.

axelboc:
> As for adding the cypress install / cache steps: I personally prefer to do it explicitly and get rid of one postinstall, but I don't have strong feelings for that.

Yeah, the problem is that `cypress install` won't tell you if there's a caching problem — it will just silently redownload Cypress, which is inefficient. I totally agree that removing a `postinstall` script is great, but I feel it'd be safer to wait for an official migration path.

mockoocy:
> Looks like there were a few recent regressions with `--prefix`: https://github.com/pnpm/pnpm/releases/tag/v10.34.0 — I haven't seen any mention that it's been deprecated but you're right it's not mentioned in the docs any more and it looks like it's just an alias for `--dir`.

Oh, interesting. We use 10.33.2 in CI, so it is not an alias yet. I think I may keep `--dir` anyway in this PR then.

> Yeah, the problem is that cypress install won't tell you if there's a caching problem — it will just silently redownload Cypress, which is inefficient. I totally agree that removing a postinstall script is great, but I feel it'd be safer to wait for an official migration path.

So you mean that in case we break the caching in the future, we'd like the job to fail? Okay then

axelboc (approved):
Thanks! Maybe worth updating the docs with the `--dir` flag as well?

marcus-oscarsson:
Thanks !

PR #2129, opened 2026-06-22
Improved validation of sample name, to match front end code

marcus-oscarsson:
We noticed, with some help ;), that the sample name "validation" is any string (server side). So we improved validation of sample name, to match front end code.

### DIFF
--- a/mxcubeweb/core/models/adaptermodels.py
+++ b/mxcubeweb/core/models/adaptermodels.py
@@ -100,7 +100,7 @@ class FrontEndStackTraceModel(BaseModel):
 
 class SampleInputModel(BaseModel):
     location: str
-    sample_name: str = Field(default="", alias="sampleName")
+    sample_name: str = Field(default="", alias="sampleName", pattern=r"^[\w+\-:]*$")
     sample_id: str = Field(default="", alias="sampleID")
 
     class Config:


mockoocy (approved):

mockoocy:
Hmm... we gotta look at dependencies, CI job is unhappy.

```python
  ModuleNotFoundError

  No module named 'distlib'

  at ~/micromamba/envs/mxcubeweb/lib/python3.10/site-packages/virtualenv/seed/embed/via_app_data/pip_install/base.py:16 in <module>
       12│ from pathlib import Path
       13│ from tempfile import mkdtemp
       14│ from typing import TYPE_CHECKING
       15│ 
    →  16│ from distlib.scripts import ScriptMaker, enquote_executable
       17│ 
       18│ from virtualenv.util.path import safe_delete
       19│ 
       20│ if TYPE_CHECKING:

Cannot install flask-principal.


  ModuleNotFoundError

  No module named 'distlib'

  at ~/micromamba/envs/mxcubeweb/lib/python3.10/site-packages/virtualenv/seed/embed/via_app_data/pip_install/base.py:16 in <module>
       12│ from pathlib import Path
       13│ from tempfile import mkdtemp
       14│ from typing import TYPE_CHECKING
       15│ 
    →  16│ from distlib.scripts import ScriptMaker, enquote_executable
       17│ 
       18│ from virtualenv.util.path import safe_delete
       19│ 
       20│ if TYPE_CHECKING:

Cannot install f90nml.

```

marcus-oscarsson:
Yepp, I noticed that I don't know why we are getting that now, I have to investigate, but I'm in the middle of something else at the moment.

mockoocy:
I've fixed it. It seems like there was an issue with the micromamba environment. I have removed the cache entry created for it and re-ran the job. now the error is gone.
So let's treat it as a weird & random occurence.

marcus-oscarsson:
Excellent, well done @mockoocy

PR #2130, opened 2026-06-25
createSlice / Typescript for Reducers

beebyen:
Example of how the TypeScript migration should look for reducers. Typing the reducers will let us type the state in the components using the `RootStore`. We utilize `createSlice` to simplify the typing difficulties and reduce the amount of boilerplate. I think the `createSlice` refactor is doable across all reducers, as it is essentially just reorganizing the function code and exporting the actions, thus eliminating the need for handwriting all the actions. But some of the reducers are quite complicated and have a lot of missing types that would require more heavy lifting. 

I made a separate `ts-store.ts` so we can easily distinguish the TypeScript and JavaScript reducers without having to fix everything at once. Couldn't figure out how to get the Middleware typing to work, if anyone has suggestions on that. Let me know what you think :)


beebyen:
Also, if anyone has suggestions on how to test the components more thoroughly, please advise. All I've been doing is clicking around the UI  😄

ui/src/reducers/contextMenu.ts
```diff
@@ -0,0 +1,83 @@
...
+  pageY: 0,
+  sampleViewX: 0,
+  sampleViewY: 0,
+  genericContextMenu: {
+    id: '',
+    show: false,
+    x: 0,
+    y: 0,
+  },
+} satisfies ContextMenuState as ContextMenuState;
```
marcus-oscarsson: Do you really need the `as` here seems redundant ?

ui/src/reducers/contextMenu.ts
```diff
@@ -0,0 +1,83 @@
+import { createSlice, type PayloadAction } from '@reduxjs/toolkit';
+
+interface ContextMenuState {
+  show: boolean;
+  shape: { type: string };
```
marcus-oscarsson: I would also probably create a interface/type for the shape, to avoid the nesting and for `genericContextMenu` as well to avoid repetition. I think that should be doable ?

ui/src/reducers/contextMenu.ts
```diff
@@ -0,0 +1,83 @@
...
+      state,
+      action: PayloadAction<{
+        id: string;
+        show: boolean;
+        x: number;
+        y: number;
+      }>,
+    ) {
+      const genericContextMenu = {
+        ...state.genericContextMenu,
```
marcus-oscarsson: I don't think you need to copy this data, the way I understood is that `RTK` uses `immer` internally to handle the "immutability".

ui/tsconfig.json
```diff
@@ -11,7 +11,8 @@
     "incremental": true,
     "noImplicitOverride": true,
     "noImplicitReturns": true,
-    "noFallthroughCasesInSwitch": true
+    "noFallthroughCasesInSwitch": true,
+    "types": ["vite/client"]
```
mockoocy, line 15: How does that differ from that bit of `vite.config.js`?
https://github.com/mxcube/mxcubeweb/blob/9a7770671e88c25e183dbad6ff375199ae31ba1b/ui/vite.config.js#L1

I guess if they do the same thing, I'd be in favor of removing the weird comment directive from the vite config file and keeping it in tsconfig.

ui/package.json
```diff
@@ -64,7 +67,7 @@
     "prop-types": "15.8.1",
     "typescript": "6.0.3",
     "typescript-eslint": "8.59.0",
-    "vite": "6.4.3",
+    "vite": "6.4.2",
```
mockoocy, line 70: I am a bit confused, by the version change there, but I guess it some Work-In-Progress leftover?

ui/src/reducers/contextMenu.ts
beebyen: It does look strange but it was actually the recommendation from the [official docs](https://redux-toolkit.js.org/usage/usage-with-typescript#defining-the-initial-state-type)

ui/src/ts-store.ts
```diff
@@ -0,0 +1,23 @@
+import { configureStore } from '@reduxjs/toolkit';
```
marcus-oscarsson: So this creates a new "typed store", is this temporary for illustration or can this work together with the other store, we would need to somehow pass both stores ?

marcus-oscarsson: I'm not entirely sure how this works, as Ive only used TypeScript and RTK in one quite small project, and its some time ago. I'm sure you have a better idea :)

ui/src/reducers/contextMenu.ts
marcus-oscarsson: Not sure I understand what `as` is meant to do then, looks really weird. I would not add it if not necessary but its not very important at this stage

ui/src/ts-store.ts
```diff
@@ -0,0 +1,23 @@
...
+if (import.meta.hot) {
+  import.meta.hot.accept('./reducers/index.ts', (nextReducer) => {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+    store.replaceReducer(nextReducer?.default);
+  });
+}
+
+export type RootState = ReturnType<typeof store.getState>;
+export type AppDispatch = typeof store.dispatch;
+export const useAppSelector: TypedUseSelectorHook<RootState> = useSelector;
```
mockoocy: I wonder - can it be done gradually?


So in the case we have reduced containing all the slices;
i.e.
```ts
configureStore({
  reducer: {login, sampleView, contextMenu, ...}
})
```
would it work like so?
```ts
const show = useAppSelector(state => state.contextMenu.show) // ts infers boolean
const user = useAppSelector(state => state.login.user) // ts infers unknown
```
If it were to work like that, then it would make our lives much easier with the typescript stuff :)

ui/package.json
beebyen: Yeah, I didn't mean to, but it was annoying to fix so I've just left it since they are minor version upgrades...

ui/tsconfig.json
beebyen: I needed it to get the `import.meta.hot` typing: https://vite.dev/guide/api-hmr#intellisense-for-typescript, guess the other one wasn't covering it?

ui/src/reducers/contextMenu.ts
beebyen: Fair enough, I'll type it as usual since there are no issues with that.

ui/src/ts-store.ts
beebyen: I think I found a way that can reuse the single store. I agree, a single store is better.

ui/src/ts-store.ts
marcus-oscarsson: I think we have to do it like that, is there another way ? 

Then, I think that the perhaps easiest way to manage this  would be to create new reducers when we use RTK and keep the old ones intact and replace them in one piece. So that we don't have reducers that are partially RTK ?

marcus-oscarsson:
Really nice work @beebyen, I think we are getting close to have an idea how to handle this :)

ui/src/ts-store.ts
beebyen: What do you think about the solution I just pushed, where we just export the existing store in the `ts` file and then add the types that way? I think it will just have same `any` types for ones that haven't been migrated, but it will work for the typed ones.

ui/src/reducers/contextMenu.ts
beebyen: Nice catch, that's a very nice perk of `RTK`.

ui/src/ts-store.ts
marcus-oscarsson: Yes, I think that would work.

ui/tsconfig.json
axelboc: `"types": ["vite/client"]` seems to be the recommended approach now. Feel free to delete `vite.config.js` 👍

ui/src/components/GenericContextMenu/MXContextMenu.tsx
```diff
@@ -2,16 +2,17 @@ import { useEffect, useRef, useState } from 'react';
...
 
-import { showGenericContextMenu } from '../../actions/sampleGrid';
+import { showGenericContextMenu } from '../../reducers/contextMenu';
+import { type RootState } from '../../ts-store';
 import styles from './MXContextMenu.module.css';
 
-export default function MXContextMenu(props) {
+export default function MXContextMenu(props: React.PropsWithChildren) {
   const { children } = props;
   const { show, x, y, id } = useSelector(
```
mockoocy:
```suggestion
  const { show, x, y, id } = useAppSelector( state => state.contextMenu.genericContextMenu)
```
We should be using the typed hook instead of asserting the type of `state` each time

marcus-oscarsson:
It seems like I can't rebase this one. Could you rebase and then I think we can merge this one. If you have tested the context menu ?

mockoocy:
Oh, the pipeline broke. Hopefully pulling latest develop should fix it.
And btw, we prefer to rebase instead of merging branches (i.e. `git pull <remote> develop --rebase`) in this project (as per [contributing docs](https://www.mxcube.org/mxcubeweb/dev/contributing.html). It keeps the git history in a nicer state :)

beebyen:
Oops, I am not used to rebasing, sorry :(

marcus-oscarsson:
Don't worry, you will catch up ! thanks

marcus-oscarsson:
I will squash the commits on merge

marcus-oscarsson:
However you need to rebase again, because the last merge did not include the most recent changes.

Try something like this
```
git fetch --all
git checkout `contextmenu-reducer`
git rebase origin/develop
```

mockoocy (approved):
It's ok for me in the current shape.

One thing I wonder about is whether or not we could just rename `store.js -> store.ts`, but if TypeScript were to be problematic about it, we may keep `ts-store.ts` separate.

beebyen:
It causes import issues if they are named the same.

PR #2131, opened 2026-06-25
Improving queue serialization

marcus-oscarsson:
Same as #1987, just trying to get rid of the CI cache issues....

The idea is to extract the queue json serialization so that we can re-use this part and related functionality in `mxcubecore` at a later stage. It also provides means for validating the queue input in a much better way.

This will be done in at least two steps; first extraction of the current logic that can be moved "back" to `mxcubecore`.  

After that, improvement of the PyDantic models. The later will need a few iterations to reach the desired result. The current models more or less match whats used in `mxcubecore` and the flat structure returned by the `QueueModel.as_dict()` methods. This is consequently whats also used in the UI.

This logic will most likely be moved in a separate sub-module that can be easily transferred to `mxubecore` .

### DIFF
--- a/mxcubeweb/core/components/queue.py
+++ b/mxcubeweb/core/components/queue.py
@@ -1094,7 +1094,7 @@ class Queue(ComponentBase):
             "sample_node": sample_model,
         }
 
-    def queue_to_dict(self, node=None, include_lims_data=False):
+    def queue_to_dict(self, node=None):
         """Returns the dictionary representation of the queue.
 
         :param TaskNode node: list of Node objects to get representation for,
@@ -1120,7 +1120,7 @@ class Queue(ComponentBase):
         """
         return self._qs.queue_to_dict(node)
 
-    def queue_to_json(self, node=None, include_lims_data=False):
+    def queue_to_json(self, node=None):
         """Return the json representation of the queue.
 
         :param TaskNode node: list of Node objects to get representation for,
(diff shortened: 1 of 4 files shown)


mxcubeweb/core/components/queue.py
```diff
@@ -42,182 +42,299 @@
 ORIGIN_MX3 = "MX3"
 
 
-class Queue(ComponentBase):
-    def __init__(self, app, config):
-        super().__init__(app, config)
-        self.init_queue_settings()
+class TaskDataPathModel(BaseModel):
```
marcus-oscarsson: These models are still contained within `queue.py`. Next step will be to move these models and some logic into a submodule that can be easily moved in one go.

marcus-oscarsson:
This is now ready to be merged. This PR does not change or add any behavior other than modelling the data as its currently being passed. The PR adds validation and typing of the attributes/data passed but the structure is intact. As we have discussed the model needs to be worked on but thats another task ;)

mxcubeweb/core/components/queue.py
```diff
@@ -45,182 +49,592 @@
...
+    def queue_to_json(self, root_node=None) -> str:
+        return json.dumps(self.queue_to_dict(root_node), sort_keys=True, indent=4)
+
+    def pretty_print_queue(self, msg: str | None = None) -> None:
+        """Pretty print current queue state for debugging."""
+        try:
+            q = self.queue_to_dict()
+            # Prefer devtools.debug for nice printing of pydantic models
+            try:
+                from devtools import debug as _debug
```
mockoocy: why `_debug` instead of `debug`?

mockoocy (approved):
Definitely a step in the right direction 🥳

PR #2133, opened 2026-06-26
Change --prefix call to --dir in lint-tsc job.

mockoocy:
--dir is the more idiomatic way to specify pnpm workspace. Also; before pnpm 10.34 there are problems with resolution of pnpm-workspace.yml files if --prefix is used instead of --dir.

---
I've missed this one when changing CI jobs to use `--dir` previously, as the tsc job was not yer merged as of that time.

### DIFF
--- a/.github/workflows/ui.yaml
+++ b/.github/workflows/ui.yaml
@@ -78,10 +78,10 @@ jobs:
           install-poetry: "false"
 
       - name: Install UI dependencies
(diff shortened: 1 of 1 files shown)


mockoocy:
Python job fails, possibly same cache issue as before - weird.

mockoocy:
After rebasing with changes from https://github.com/mxcube/mxcubeweb/pull/2138
the pipeline got fixed. Very cool :partying_face:

PR #2137, opened 2026-07-01
[Harvester] Auto-send data collection to CRIMS on DC completion

jbflo:
## Summary

- `Harvester` component connects to `queue_entry_execute_finished` and
  automatically POSTs data collection info to CRIMS when a DC completes
  successfully 

- `_send_collection_to_crims` extracts `crystalUuid`, `sampleName`,
  `instrumentName` (beamline_name), `facilityName` (synchrotron_name),
  and `investigationID` (session_id) from the finished queue entry.

- Removes the manual "Send Data to Crims" button, its route, Redux
  action, and API call — superseded by the automatic trigger.

-  two required properties: `crims_upload_url` (full endpoint URL) and `crims_upload_key`.

## Coupled PR
Requires mxcubecore PR [LINK](https://github.com/mxcube/mxcubecore/pull/1632) which implements the updated
`send_data_collection_info_to_crims` function.

I might do a followup to convert the component to an adapter next.

### DIFF
--- a/mxcubeweb/core/components/harvester.py
+++ b/mxcubeweb/core/components/harvester.py
@@ -2,9 +2,11 @@ from __future__ import annotations
 
 import logging
 
+import gevent
 from mxcubecore import HardwareRepository as HWR
 from mxcubecore.HardwareObjects.abstract.sample_changer import Crims
 from mxcubecore.HardwareObjects.Harvester import HarvesterState
+from mxcubecore.queue_entry.data_collection import DataCollectionQueueEntry
 
 from mxcubeweb.core.components.component_base import ComponentBase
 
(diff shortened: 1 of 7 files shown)


mxcubeweb/core/components/harvester.py
```diff
@@ -150,45 +164,41 @@ def get_global_state(self):
...
-                    rest_token,
-                    crims_key,
+                    sample_name,
                 )
-            return True
         except Exception:
-            msg = "Could not send data collection to crims"
-            logging.getLogger("user_level_log").exception(msg)
-            return False
+            logging.getLogger("user_level_log").exception(
```
marcus-oscarsson: Pay attention to the logger used, we don't want to send `exception` logs over `user_level_log`. Its better to make one `exception` on `self.log.exception` and one informative log on `self.user_log.error`

marcus-oscarsson:
Thanks alot for these PRs @jbflo. We will merge the one on `mxcubecore` first and then we need to bump `mxcubecore` in `pyproject.toml`

It would also be great if as you mentioned before we could migrate this to an adapter, there is no reason for to be a separate component anymore.

jbflo:
> Thanks alot for these PRs @jbflo. We will merge the one on `mxcubecore` first and then we need to bump `mxcubecore` in `pyproject.toml`
> 

mxcubecore updated ` in `pyproject.toml

marcus-oscarsson:
Thanks, very nice. It would be wonderful if you could have a look at that adapter as well ;)

PR #2138, opened 2026-07-01
Put poetry into it's own virtual environment

### LINKED ISSUE #2134: Python test job fails sometimes
Example: https://github.com/mxcube/mxcubeweb/actions/runs/28243762802/job/83677142097?pr=2133

It has happened on a separate occasion there https://github.com/mxcube/mxcubeweb/pull/2129 - in that case removing the github actions cache entry worked.

What happens is that we first restore cached micromamba environment, then run `poetry install` to install packages, as specified in the `pyproject.toml` file.  During installation some packages cannot be installed, because of missing `distlib` dependency.
 In the logs of the `Install MXCuBE` step there is:
```
  - Downgrading distlib (0.4.3 /home/conda/feedstock_root/build_artifacts/bld/rattler-build_distlib_1781320269/work -> 0.3.9)
  ```
  
I will be probably looking into it later, but I hope someone has any ideas what may be wrong with the current one.
Maybe the `micromamba-setup` action we use is misconfigured?

### PR DESCRIPTION
**Pull request by mockoocy**
Poetry documentation strongly advises to use a separate environment for invoking poetry and installing the project.
https://python-poetry.org/docs/#ci-recommendations

Currently it is not the case, as poetry detects if it is running inside inside a virtual environment -- which is the case when using conda.

As a result poetry's dependencies may be updated during project installation and lead to hard to debug bugs.

---
I suspect this may fix https://github.com/mxcube/mxcubeweb/issues/2134

### DIFF
--- a/.github/actions/install_server/action.yml
+++ b/.github/actions/install_server/action.yml
@@ -1,0 +1,64 @@
+---
+name: Install MXCuBE server
+description: Sets up environments for Conda and Poetry, then install server app.
+
+inputs:
+  python-version:
+    description: Python version used by MXCuBE.
+    required: true
+    default: "3.11"
+  environment-file:
+    description: File specifying conda environments
+    default: conda-environment.yml
+  poetry-version:
+    description: Version of poetry to use
+    default: 2.4.1
+  micromamba-version:
+    description: Version of micromamba to use
+    default: 1.5.10-0
+  install-poetry:
+    description: >-
+      Install Poetry and the MXCuBE server package on top of the Conda env.
+      Set to "false" for jobs that only need Conda-provided tools (e.g. pnpm/node).
+    default: "true"
+
+runs:
+  using: composite
+  steps:
+    - name: Set up Conda environment
+      uses: mamba-org/setup-micromamba@v2
+      with:
+        create-args: >-
+          python=${{ inputs.python-version }}
+        micromamba-version: ${{ inputs.micromamba-version }}
+        environment-file: ${{ inputs.environment-file }}
+        cache-environment: true
+        post-cleanup: "all"
+        # We need to install poetry so that it is available from outside of the
+        # micromamba environment. Currently it is not feasible, as poetry does
+        # always choose ot use currently active environment, even if config flags are
+        # set to create a virtualenv inside project.
+        # https://github.com/python-poetry/poetry/issues/4055
+        # Using pipx is reccomended by poetry devs, as it creates a venv for each tool.
+    - name: Install Poetry
+      if: inputs.install-poetry == 'true'
+      shell: bash
+      run: |
+        pipx install "poetry==${{ inputs.poetry-version }}"
+
+    - name: Cache Poetry environment
+      if: inputs.install-poetry == 'true'
+      uses: actions/cache@v4
+      with:
+        path: .venv
+        key: >-
+          poetry-${{ runner.os }}-py${{ inputs.python-version }}-
+          ${{ hashFiles('poetry.lock', inputs.environment-file) }}
+
+    - name: Install MXCuBE
+      if: inputs.install-poetry == 'true'
+      shell: bash
+      run: |
+        CONDA_PY="$("${MAMBA_EXE}" run --name mxcubeweb bash -c 'command -v python')"
+        poetry env use "$CONDA_PY"
+        poetry install
--- a/.github/workflows/ui.yaml
+++ b/.github/workflows/ui.yaml
@@ -157,6 +157,11 @@ jobs:
           cache-environment: true
           post-cleanup: "all"
 
+      - name: Create Poetry virtualenv
+        run: >-
+          "${MAMBA_EXE}" run --name mxcubeweb sh -c
+          'poetry env use "$(command -v python)"'
+
       - name: Install MXCuBE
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by walesch-yan** [comment]
Thanks, that is a good catch! I'm just wondering as I see that multiple workflow files need to be updated for such an update, if it would also make sense for us to create a utility file with custom actions that we then only need to call inside the workflows. 

A custom action here could be the installation of conda environment, and poetry steps all in one, which reduces the changes, we have to do (and maybe miss) if a part of this would change

**Comment by mockoocy** [comment]
I will try to make this into a separate github action somehow then, sounds like a good idea actually.

File: .github/workflows/ui.yaml [hunk 1]
```diff
@@ -186,7 +127,7 @@ jobs:
       - name: Start MXCuBE-Web server (YAML config)
         run: |
           pkill -f mxcubeweb-server || true
-          ${MAMBA_EXE} run --name mxcubeweb mxcubeweb-server -r ./demo.yaml/ --static-folder $(pwd)/ui/build/ -L debug &
+          poetry run mxcubeweb-server -r ./demo.yaml/ --static-folder $(pwd)/ui/build/ -L debug &
```
mockoocy, line 136: Using poetry installed via pipx, not conda.
Perhaps we could stay on a safe side by not including poetry in the conda env? Although that would bring additional complexity 🙃

.github/actions/install_server/action.yml
mockoocy, line 1: The name is subject to change, I do not really like it, so if anyone has a better idea, I'm open to it ;)

.github/actions/install_server/action.yml
```diff
@@ -0,0 +1,62 @@
...
+        # micromamba environment. Currently it is not feasible, as poetry does
+        # always choose ot use currently active environment, even if config flags are
+        # set to create a virtualenv inside project.
+        # https://github.com/python-poetry/poetry/issues/4055
+        # Using pipx is reccomended by poetry devs, as it creates a venv for each tool.
+    - name: Install Poetry
+      shell: bash
+      run: |
+        pipx install "poetry==${{ inputs.poetry-version }}"
+
```
mockoocy, line 48: Turns out the original fix from this PR did not really work? The pipeline was green and nothing weird seemed to happen, but upon closer inspection I've found that poetry still would use conda's environment, even if instructed to create it's own.

I've linked a GitHub issue that is being related to that weird behaviour and it does not seem that anything would change in that regard in the near future.

.github/actions/install_server/action.yml
```diff
@@ -0,0 +1,62 @@
...
+        key: >-
+          poetry-${{ runner.os }}-py${{ inputs.python-version }}-
+          ${{ hashFiles('poetry.lock', inputs.environment-file) }}
+
+    - name: Install MXCuBE
+      shell: bash
+      run: |
+        CONDA_PY="$("${MAMBA_EXE}" run --name mxcubeweb bash -c 'command -v python')"
+        poetry env use "$CONDA_PY"
+        poetry install
```
mockoocy, line 64: Just to clarify: `poetry` itself is installed outside the conda environment.

`poetry env use` points Poetry to the Python interpreter from the `mxcubeweb` conda environment, so `poetry install` installs the dependencies into that same environment.

This is to prevent clash between dependencies of poetry and dependencies of mxcubeweb.

.github/actions/install_server/action.yml
marcus-oscarsson: Where did you find the issue/link with the same problem as us ?

marcus-oscarsson:
Thanks !

.github/actions/install_server/action.yml
mockoocy: I referred to the link that's in the comment that's in the code block I commented under.
It's about lack of option to specify for poetry to create it's own virtual environment if poetry is installed inside a conda enviornment. In that case it always installs the project into conda environment.

Using the same environment to install poetry and the project is a bad idea according to the [docs](https://python-poetry.org/docs/#installation) (there is a big red warning) and I believe that it may have lead to the issues we were experiencing before.

marcus-oscarsson: I see, ok I get it now. I thought it was an issue about the failing CI pipelines.

marcus-oscarsson: Thanks, nicely done !

fabcor-maxiv:
This problem has been known for a while (here <https://github.com/mxcube/mxcubecore/issues/1061#issuecomment-2440898688> among many other mentions). I gave it a lot of thought at the time, but never figured out a solution that would involve as little developer-facing changes as possible in order to satisfy the contributors who are less comfortable with this kind of tools.

I had a very quick look at this change, and the way I understand it Poetry is still installed in the Conda environment but not used, so it looks to me like it is still only avoiding some issues but not the whole range of issues caused by this Conda+Poetry setup.

I might have mentioned this somewhere: in the mean time after I last spent time working on this topic, [Pixi](https://pixi.prefix.dev) got more and more stable and added a whole bunch of new interesting features, and I now believe it might be a good (if not the best) solution to such cases where Conda and PyPI dependencies have to be mixed. But... all things considered, Pixi is still somewhat new and... as always... I do not know if the contributors would be OK with migrating to "yet another tool" and would have the patience to deal with the usual bumps during such a migration phase...

PR #2140, opened 2026-07-02
Guard drawGridPlugin on mouse movement updates and leftClick against non-primary mouse button events.

dominikatrojanowska:
After adding one line of logs to each of `rightClick` and `leftClick` methods: "leftClick/righClick", option.e.buttons), I realized that `leftClick` is invoked to often, actually after each mouse button event. 

Screenshot from 2026-07-02 10-15-05

Also in `onMouseMove` method the condition `options.e.buttons > 0` was a bit wired, causing together with above the possibility to draw grid with mouse wheel.

If that was not on purpose I propose to improve it a bit. Otherwise, the naming is a bit unfortunate.

### DIFF
--- a/ui/src/components/SampleView/SampleImage.jsx
+++ b/ui/src/components/SampleView/SampleImage.jsx
@@ -173,7 +173,7 @@ class SampleImage extends React.Component {
       this.canvas.add(this.centringHorizontalLine);
     }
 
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
Thanks !

PR #2141, opened 2026-07-03
Improved feedback mail utility

marcus-oscarsson:
Validating data sent passed feedback email


walesch-yan:
Quite a good (and fast) catch by the bot

PR #2145, opened 2026-07-08
Fix non-dragable GridForm

dominikatrojanowska:
Maybe other were broken as well but I just checked for the GridForm.  
Looks like some recent merge had broken this feature.  To reproduce:
1. Pull develop branch.
2. Remove `node_modules` and `build` packages
3.  `pnpm install && pnpm build`
4. Start mxcube `mxcube-server..`
5. Start develop `pnpm start --host`
6. Login --> Draw grid --> try to drag Grid Form and in the console you will see:
image

I based with my solution at: https://github.com/react-grid-layout/react-grid-layout/issues/2266. Sounds reasonable but I'm not very familiar with this topic, so if anybody have any better solution that's even better :)

### DIFF
--- a/ui/vite.config.js
+++ b/ui/vite.config.js
@@ -9,6 +9,14 @@ export default defineConfig({
     react(),
     { ...eslintPlugin(), apply: 'serve' }, // dev only to reduce build time
   ],
+  define: {
+    // Workaround for a known regression, present since react-draggable@4.6.0,
+    // where process.env is referenced in the browser bundle served by Vite's
+    // development server. It caused `ReferenceError: process is not defined`
+    // when clicking on draggable elements (e.g. GridForm in UI dev).
+    // https://github.com/react-grid-layout/react-draggable/issues/806
+    'process.env': '{}',
+  },
   server: {
     open: true, // open default browser on start
     strictPort: true, // fail if port already in use (must be 5173 to match server's CORS config)


mockoocy:
I can drag the grid just fine  on current develop :D

dominikatrojanowska:
Thanks @mockoocy - I see, it works fine at default UI server started with `mxcube-server...` but not on development one, I'm switching that to Draft until further investigation :)

marcus-oscarsson:
Tanks guys, interesting issue ! It seems to be working for me with the build as well. I have not tried with the development server.

mockoocy:
Okay, it broke for me after reinstalling packages. Maybe some recent update to one of the dependencies is at fault.

dominikatrojanowska:
~~seems like regression of vite from 6.4.3 to 6.4.2 broke this thing~~

mockoocy:
https://github.com/react-grid-layout/react-draggable/issues/806

This is related issue.
It seems that some developer has assumed that `process.env` works in NextJS and therefore is a web standard :-)

mockoocy:
> seems like regression of vite from 6.4.3 to 6.4.2 broke this thing
Indirectly yeah. The lockfile's contents have changed to include a newish version of react-draggable which contains the bug you've encountered

dominikatrojanowska:
Yep, thanks, I was close to this at the beginning :D so lter I was wrong with vite regression, it was rather react-draggable upgrade (form 4.5.0 to 4.7.0) a root cause. So I can pin lock files to the older version or keep this solution.

dominikatrojanowska:
> seems like regression of vite from 6.4.3 to 6.4.2 broke this thing
> Indirectly yeah. The lockfile's contents have changed to include a newish version of react-draggable which contains the bug you've encountered

As this is not a direct root cause to which this PR refers, I provided new one https://github.com/mxcube/mxcubeweb/pull/2148 :)

marcus-oscarsson:
Thanks, well done #2148 is merged. Ill close this one !

dominikatrojanowska:
But https://github.com/mxcube/mxcubeweb/pull/2148 does not solve the issue with draggable objects. Maybe I was not plain enough in the previous comment. 

To reproduce issue I mentioned:
1. Pull develop branch.
2. Remove `node_modules` and `build` packages
3.  `pnpm install && pnpm build`
4. Start mxcube `mxcube-server..`
5. Start develop `pnpm start --host`
6. Login --> Draw grid --> try to drag Grid Form and in console you will see:
image

Sorry for being super explicit but it is crucial to reinstall all packages and run development UI. This issue will not show up in the production mode of UI.

ui/vite.config.js
```diff
@@ -9,6 +9,15 @@ export default defineConfig({
...
   ],
+  define: {
+    // Some dependencies (e.g. react-draggable's internal logger reads
+    // process.env.DRAGGABLE_DEBUG) access process.env directly. The dev
+    // server doesn't provide a `process` global like the production build
+    // does, so any process.env.* access throws "process is not defined".
+    // Defining the whole object makes every process.env.X resolve to
+    // undefined instead of crashing.
+    'process.env': '{}',
+  },
```
mockoocy, line 19: I'd change the comment to reference the issue I've pasted in one of the previous comments and say that this is to circumvent the bug in `react-draggable` package.
It makes this more apparent that this is here to fix a particular bug rather than a deliberate choice.

Alternatively we could pin `react-draggable==4.5.0` in `package.json`, no strong feelings either way.

marcus-oscarsson: Hm, lets go with defining `process.env`, yes its a perhaps good idea making the comment underlining even more that its to fix an issue with `react-dragable`

marcus-oscarsson:
> But #2148 does not solve the issue with draggable objects. Maybe I was not plain enough in the previous comment.
> 

Ah, sorry, I did not follow that entirely and ended up mixing up the vite down grade and react-dragble version, I thought we pinned the later in the end. Just me being confused.  

I think this fix is perhaps better, as maybe there are other libraries that are/will be doing the same assumption and also we will not forget to unpin react-dragble later.

ui/vite.config.js
dominikatrojanowska: updated, let me know if that's fine

mockoocy (approved):

PR #2146, opened 2026-07-09
Fix & Refactor Context Menu

mockoocy:
Currently ContextMenu is not working, since some Redux actions have not been ported to the Redux Toolkit API.
Now it is required to pass the object containing action payload like so: `dispatch(action({arg1: ... ,arg2: ...}))`.

The obvious fix of wrapping all arguments in objects would make the types of the redux state diverge from the type definitions, unless all the fields were optional.
I've decided to use discriminated union pattern to model the valid states of the context menu and did a relatively small refactor to clean up the related state.

### DIFF
--- a/ui/src/components/GenericContextMenu/MXContextMenu.tsx
+++ b/ui/src/components/GenericContextMenu/MXContextMenu.tsx
@@ -2,15 +2,15 @@ import { useEffect, useRef, useState } from 'react';
 import { Dropdown } from 'react-bootstrap';
 import { useDispatch } from 'react-redux';
 
-import { showGenericContextMenu } from '../../reducers/contextMenu';
+import { hideMenu } from '../../reducers/contextMenu';
 import { useAppSelector } from '../../ts-store';
 import styles from './MXContextMenu.module.css';
 
 export default function MXContextMenu(props: React.PropsWithChildren) {
   const { children } = props;
-  const { show, x, y, id } = useAppSelector(
-    (state) => state.contextMenu.genericContextMenu,
-  );
+  const contextMenu = useAppSelector((state) => state.contextMenu);
+  const show = contextMenu.type === 'Generic';
+  const { id, x, y } = show ? contextMenu : { id: '', x: 0, y: 0 };
   const dispatch = useDispatch();
   const menuRef = useRef<HTMLDivElement>(null);
   const [position, setPosition] = useState({ x: 0, y: 0 });
@@ -31,12 +31,12 @@ export default function MXContextMenu(props: React.PropsWithChildren) {
   }, [show, x, y]);
 
   useEffect(() => {
-    function hideContextMenu() {
-      dispatch(showGenericContextMenu({ id: '', show: false, x: 0, y: 0 }));
+    function onDocumentClick() {
+      dispatch(hideMenu());
     }
-    document.addEventListener('click', hideContextMenu);
+    document.addEventListener('click', onDocumentClick);
     return () => {
-      document.removeEventListener('click', hideContextMenu);
+      document.removeEventListener('click', onDocumentClick);
     };
   }, [dispatch]);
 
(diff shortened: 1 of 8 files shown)


ui/src/reducers/contextMenu.ts
```diff
@@ -1,70 +1,54 @@
...
+  id: string;
+  x: number;
+  y: number;
+}
+
+type ContextMenuState = NoMenu | ShapeMenu | GenericMenu;
 
 const contextMenuSlice = createSlice({
   name: 'contextMenu',
-  initialState: INITIAL_STATE,
```
mockoocy, line 41: I made a factory function to ensure that TypeScript does properly infer `ContextMenuState`.

If it were to be 
```ts
const INITIAL_STATE: ContextMenuState = {type: 'None'}
```
then it would be narrowed to `NoMenu` type inside `createSlice` unless I do `as ContextMenuState` assert, which would make eslint mad

ui/src/components/GenericContextMenu/MXContextMenu.tsx
```diff
@@ -31,12 +31,12 @@ export default function MXContextMenu(props: React.PropsWithChildren) {
...
-    function hideContextMenu() {
-      dispatch(showGenericContextMenu({ id: '', show: false, x: 0, y: 0 }));
+    function onDocumentClick() {
+      dispatch(hideMenu());
     }
-    document.addEventListener('click', hideContextMenu);
+    document.addEventListener('click', onDocumentClick);
     return () => {
-      document.removeEventListener('click', hideContextMenu);
+      document.removeEventListener('click', onDocumentClick);
```
mockoocy, line 39: The function name change may be reverted, but I think it at least reflects the convention around event handlers.

marcus-oscarsson:
What do you think about adding Cypress tests when we do these kinds of changes ?

mockoocy:
> What do you think about adding Cypress tests when we do these kinds of changes ?

I suppose we should do more tests altogether :) 

I've added two simple cypress tests to check if context menu opens and closes properly.

beebyen:
The discriminated union works great here, love to see it

marcus-oscarsson:
Awesome, for the tests. Well done !

ui/src/containers/SampleGridTableContainer.jsx
```diff
@@ -131,15 +132,13 @@ export default function SampleGridTableContainer(props) {
...
-    (state) => state.contextMenu.genericContextMenu,
-  );
+  const contextMenu = useAppSelector((state) => state.contextMenu);
   const workflows = useSelector((state) => state.workflow.workflows);
   const defaultParameters = useSelector(
     (state) => state.taskForm.defaultParameters,
   );
   const availableMethods = new Set(Object.keys(defaultParameters));
-  const selectionRubberBand = document.querySelector('#selectionRubberBand');
+  const selectionRubberBandRef = useRef(null);
```
mockoocy, line 141: This fixes a problem I had in the new e2e tests. The underlying DOM element was initially not found and could not be accessed with the query Selector.
Anyway referencing DOM elements from React should be done with `useRef`.

ui/src/components/GenericContextMenu/MXContextMenu.tsx
```diff
@@ -2,15 +2,15 @@ import { useEffect, useRef, useState } from 'react';
...
 import styles from './MXContextMenu.module.css';
 
 export default function MXContextMenu(props: React.PropsWithChildren) {
   const { children } = props;
-  const { show, x, y, id } = useAppSelector(
-    (state) => state.contextMenu.genericContextMenu,
-  );
+  const contextMenu = useAppSelector((state) => state.contextMenu);
+  const show = contextMenu.type === 'Generic';
+  const { id, x, y } = show ? contextMenu : { id: '', x: 0, y: 0 };
```
axelboc, line 13: Looks good throughout 💯 — just a minor thing here:  it won't have any impact at all in this particular case but I would have maybe put the condition and fallback inside the selector to avoid re-renders when a non-generic context menu is shown.

```ts
const contextMenu = useAppSelector((state) => state.contextMenu.type === 'Generic' ? state.contextMenu : undefined);
```

Might be more relevant with other state slices, so I thought I'd mention in.

mockoocy: Oh yeah, it's better. I did not think of that, thank you!

PR #2148, opened 2026-07-09
Fix - revert accidental downgrading of vite form 6.4.3 to 6.4.2

dominikatrojanowska:
As discovered here: https://github.com/mxcube/mxcubeweb/pull/2145#issuecomment-4925980170  
This downgrade seems to be an overlook at some earlier PR.


marcus-oscarsson:
Perfect, well done ! Thanks alot !

PR #2149, opened 2026-07-10
Refactor Shapes reducer into an RTK Slice

mockoocy:
This is mostly a mechanical change, bar removal of `selectedShapes` state from `sampleView` as it is being derived from `shapes` object inside `shapes` reducer.

The behaviour behind adding / selecting / removing shapes has been guarded with an e2e test using cypress


ui/src/reducers/shapes.ts
```diff
@@ -0,0 +1,134 @@
...
+
+import {
+  createSelector,
+  createSlice,
+  type PayloadAction,
+} from '@reduxjs/toolkit';
+
+export type ShapeState = 'SAVED' | 'TMP' | 'HIDDEN';
+
+interface BaseShape {
```
mockoocy, line 11: Derived from [Shape](https://github.com/mxcube/mxcubecore/blob/7a397ed178962d90b88d3f22cbc14762fb048148/mxcubecore/HardwareObjects/SampleView.py#L939) objects at the mxcubecore side.

The goal is to reflect what the backend sends us for now.

ui/src/reducers/shapes.ts
```diff
@@ -0,0 +1,134 @@
...
+  beanPos: [number, number];
+  dxMm: number;
+  dyMm: number;
+  hideThreshold: number;
+  stepsX: number;
+  stepsY: number;
+  angle: number;
+  beamHeight: number;
+  x1: number;
+  y1: number;
```
mockoocy: I am quite certain that these all are unnecessary for the front-end, but I think it could be left that way for now.
It would be nice to define more strict schemas with pydantic for the shapes perhaps -- but that's for another time :)

ui/src/reducers/shapes.ts
```diff
@@ -0,0 +1,134 @@
...
+};
+
+export type LineShape = BaseShape & {
+  t: 'L';
+  // Both endpoints, flattened: [x1, y1, x2, y2].
+  refs: [string, string];
+  motorPositions: string; // stringified pair of Record<string, number>
+};
+
+export type GridCellCountFunction = 'zig-zag' | 'inverse-zig-zag';
```
matyson: are the other orientations being unused? if so, we could remove the count functions altogether

ui/src/reducers/shapes.ts
```diff
@@ -0,0 +1,134 @@
...
+  cellHSpace: number;
+  cellVSpace: number;
+  numCols: number;
+  numRows: number;
+  cellCountFun: GridCellCountFunction;
+  result: string | Record<string, number[]> | null;
+  resultDataPath?: string;
+  pixelsPerMm: [number, number];
+  motorPositions: { omega: number } & Record<string, number>;
+  beanPos: [number, number];
```
matyson: typo?

ui/src/reducers/shapes.ts
```diff
@@ -0,0 +1,134 @@
...
+  width: number;
+  height: number;
+  cellWidth: number;
+  cellHeight: number;
+  cellHSpace: number;
+  cellVSpace: number;
+  numCols: number;
+  numRows: number;
+  cellCountFun: GridCellCountFunction;
+  result: string | Record<string, number[]> | null;
```
matyson: type here is `Record<string,[number, number[]]>` or even `Record<string,[number,[number,number,number]]>` ( tho we should refactor it in the future)

ui/src/reducers/shapes.ts
mockoocy: Well -- I am not sure. There is code that relates to both at this moment.
I think it could be revisited later to see if this attribute and related [functions](https://github.com/mxcube/mxcubeweb/blob/afab8f5a26ff10af80c84b1027738de8c2817e6f/ui/src/components/SampleView/DrawGridPlugin.jsx#L712) can be removed.

ui/src/reducers/shapes.ts
mockoocy: I've fixed it.
I wonder if for the RGB results this could be narrowed even further into `Record<"RGB", [number, [number, number, number]]` -- albeit I am not sure how exactly this dict is being filled up.

matyson: No, each string correspond to the index position in the grid. But it is a very confusing data structure, could use a revisit

mockoocy: I have no clue how to interpret that then haha https://github.com/mxcube/mxcubeweb/blob/afab8f5a26ff10af80c84b1027738de8c2817e6f/ui/src/components/SampleView/DrawGridPlugin.jsx#L324

Although  `GridData.result` may correspond to a different shape perhaps, or that code is just wrong...

matyson: Oh, i haven't seen this, now i'm more confused

ui/src/reducers/shapes.ts
dominikatrojanowska: it comes from configuration file, I think I would rather implement [all](https://github.com/mxcube/mxcubeweb/blob/afab8f5a26ff10af80c84b1027738de8c2817e6f/ui/src/components/SampleView/DrawGridPlugin.jsx#L662) in this place as we don't know if other facilities use them 🤔

mockoocy: Ah, I've ommited two of these -- fixed :)

ui/src/reducers/shapes.ts
```diff
@@ -0,0 +1,134 @@
...
+};
+
+export type Shape = PointShape | TwoDPointShape | LineShape | GridShape;
+
+export type ShapesById = Record<string, Shape>;
+
+export interface ShapesState {
+  shapes: ShapesById;
+  overlayLevel?: number;
+}
```
dominikatrojanowska, line 79: I'm a bit confused with naming this object ShapesState. There is no strict relation to states of shapes.

ui/src/reducers/shapes.ts
dominikatrojanowska: I was already confused with userState, but thanks to you comment I know why there are two states. Maybe to avoid future confusion put this comment in the code? :)

ui/src/reducers/shapes.ts
```diff
@@ -0,0 +1,134 @@
...
+  id: string;
+  name: string;
+  state: ShapeState;
+  userState: ShapeState;
+  label: string;
+  selected: boolean;
+  screenCoord: [number, number];
+}
+
+export type PointShape = BaseShape & {
```
dominikatrojanowska: just for a curiosity ad I'm not very experienced with typescript: would `interface ... extends BaseShape {...}` be an equivalent of `type ... = BaseShape & {...}? Why did  you decide to use the second one? 🤔

ui/src/reducers/shapes.ts
mockoocy: This is comes from redux naming convention.
Shapes is the name of the slice, and state refers to what this slice can hold. Thus `ShapesState` is what _shapes_ slice can contain.
And I'd argue that there is a relation between this object and "state of shapes", state of a particular shape (hidden / visible / selected, etc.)  is defined there.

dominikatrojanowska: ofc there is but I meant strict, ok I didn't think about redux state in this context. You are completely right then. Thanks for the answer! :D

ui/src/reducers/shapes.ts
mockoocy: They are not generally equivalent, but in this case there is not much difference. I'd see a valid argument that interface could be a better fir here, but I am just more used to using type intersections; that's the reason they are here :)
The difference is how they handle the conflicts between properties in newly defined type and the base one.
If I had something like

```ts

interface BaseShape {
  // ...
  motorPositions: Record<string, number>
}

interface Line extends BaseShape {
  motorPositions: string | Record<string, number>
}
```
then I'd get a compilation error, because types of properties are incompatible.
In the case of intersection, `motorPosition` could be still set as `Record<string, number>` without any error.

And the biggest difference is that I can do intersection between any type or interface , while `extends` only works on interfaces.

ui/src/reducers/shapes.ts
mockoocy: left a comment

ui/src/reducers/shapes.ts
mockoocy: I changed these to use `extends` as it may be more understandable.

ui/src/components/SampleView/SampleImage.jsx
```diff
@@ -15,12 +15,12 @@ import {
   recordCentringClick,
   rotateToShape,
   setImageRatio,
-  setOverlay,
   toggleDrawGrid,
   updateShapes,
 } from '../../actions/sampleview.js';
 import { HW_STATE, QUEUE_RUNNING } from '../../constants';
 import { showShapeMenu } from '../../reducers/contextMenu.js';
+import { selectSelectedShapeIds, setOverlay } from '../../reducers/shapes.js';
```
dominikatrojanowska, line 23:
```suggestion
import { selectSelectedShapeIds, setOverlay } from '../../reducers/shapes.ts';
```

mockoocy: File extensions are a bit interesting here, we may use .js, .ts, or omit them altogether. That is thanks to Vite, which performs some magic. The generally most correct way seems to be using .js [as per docs](https://www.typescriptlang.org/docs/handbook/modules/reference.html#importing-and-exporting-typescript-specific-declarations).

I'd probably avoid using .ts extensions though -- at least for now --, as we don't have a [tsconfig rule](https://www.typescriptlang.org/tsconfig/#allowImportingTsExtensions) applied that allows for import of .ts extension files from inside other .ts files.

dominikatrojanowska:
LGTM :)

ui/src/reducers/shapes.ts
marcus-oscarsson: I don't think `resultType` is used anymore or was ever actually used. I think the idea is still valid. There can be several different "heat maps" as a result of a mesh scan. I.e one that outlines each individual crystal or the more traditional showing the "diffraction intensity" or some other metric. So `resultType` is basically what processing to display.

The result data is otherwise a base64 encoded string (PNG) or the RGB map. We are not using the later so I can't really tell you much more about it. It needs to work for the base64 encoded string version as well ! :)

marcus-oscarsson:
This is a very nice initiative, thanks for the nice initial work on this. We have not really discussed how to go about the TypeScript introduction/conversation yet but this is anyhow very welcome. I think starting with the stores and/or the API folder would make sense. 

I've been quite busy so I have not had the time to have a closer look at this, Ill try to get some time today. 

Needless to say Its extremely important that this is well tested, and we really appreciate that you added some Cypress tests.

ui/src/actions/sampleview.js
```diff
@@ -11,6 +11,11 @@ import {
   sendSetVideoSize,
   sendStartClickCentring,
 } from '../api/sampleview';
+import {
+  addShape as addShapeMutation,
```
marcus-oscarsson: I have to admit that I'm not fan of the `Mutation` suffix, is that common in the RTK context ?

ui/src/serverIO.js
```diff
@@ -167,7 +166,7 @@ class ServerIO {
     });
 
     this.hwrSocket.on('grid_result_available', (data) => {
-      dispatch(updateShapesAction([data.shape]));
+      dispatch(updateShapes([data.shape]));
```
marcus-oscarsson, line 169: What happens if `data.shape` is not on the `Shape` format here ? How is the error displayed and handled ?

mockoocy: There would not be an error. TypeScript types exist only at compile time and during compilation they are stripped in the resulting JavaScript. The behaviour should not differ from before.


To provide both type definitions and validation ( like e.g. Pydantic does) many people use some external libaries like [Zod](https://zod.dev/).

ui/src/actions/sampleview.js
mockoocy: I've seen similar naming scheme in some libraries (RTK Query, tanstack-query) but these probably imply server-side changes. I was not really sure about the suffix to use actually, but I've changed it to `Action` which is more meaningful there I guess.

ui/src/reducers/shapes.ts
mockoocy: We use RGB heat maps and I think `resultType` is what specifies the kind of result we want. I doubt that having two kinds of results at once really works.
I've left the `string` as a valid result shape, no worries.

The code snippet that has confused me seems like dead code to me.

ui/src/serverIO.js
marcus-oscarsson: I see, can we use `zod` in sort of warning only mode, so that we get a clear visual of the validation errors in the console ?

marcus-oscarsson: It would be nice if we could use something like this, somehow, in our API "boundary" to check against whats in the OpenAPI spec. Perhaps more at compile time then I guess.

marcus-oscarsson:
Really nice :).

ui/src/serverIO.js
mockoocy: Utilizing openAPI spec is something we could look into for sure. 

Zod is one option, and nice one as we can use it to also validate form data in some places, as well as we get typescript types from it. 
The other option that comes to my mind is using react json-schema form library that we already depend on, as it has some kind of json schema validation already. I am not sure how simple would that be though.

But none answer the problem that our type definitions could diverge. Nothing would really break then, but our type definitions would be less trustworthy. There are some tools to generate typescript types / zod schemas from OpenAPI, but I've never used these.

marcus-oscarsson: I see, yes its important that our definitions actually correspond to what we are using I thought that maybe using OpenAPI could help with that, but it can also become quite complicated.

Lets see what creative solutions we can come up with. I think this is already a wonderful start

PR #2151, opened 2026-07-15
Refactor Login reducer into typed RTK slice

mockoocy:


ui/src/reducers/login.ts
```diff
@@ -0,0 +1,129 @@
...
+import { createSlice, type PayloadAction } from '@reduxjs/toolkit';
+
+export type LoginType = 'User' | 'Proposal';
+
+export interface Lims {
+  name: string;
+  description: string;
+}
+
+export interface Session {
```
beebyen, line 12: How did you find the properties for this interface?

mockoocy: from there https://github.com/mxcube/mxcubecore/blob/ad5e5713a5c445123beccc416c35a49ae1ce5b8e/mxcubecore/HardwareObjects/Session.py#L23-L41

dominikatrojanowska: what do you think about adding [similar comment](https://github.com/mxcube/mxcubeweb/pull/2152#discussion_r3588027181)? :)

mockoocy: > what do you think about adding https://github.com/mxcube/mxcubeweb/pull/2152#discussion_r3588027181? :)

done

ui/src/reducers/login.ts
```diff
@@ -0,0 +1,130 @@
...
+  proposal_name: string;
+  sample_count?: string;
+  scheduled: string;
+  session_id: string;
+  start_date: string;
+  start_datetime: string;
+  start_time: string;
+  title: string;
+  user_portal_URL: string;
+  volumne?: string;
```
dominikatrojanowska:
```suggestion
  volume?: string;
```

ui/src/reducers/login.ts
```diff
@@ -0,0 +1,130 @@
...
+  proposal_id: string;
+  proposal_name: string;
+  sample_count?: string;
+  scheduled: string;
+  session_id: string;
+  start_date: string;
+  start_datetime: string;
+  start_time: string;
+  title: string;
+  user_portal_URL: string;
```
dominikatrojanowska:
```suggestion
  user_portal_URL?: string;
```

ui/src/reducers/login.ts
```diff
@@ -0,0 +1,130 @@
...
+  description: string;
+}
+export interface Session {
+  actual_end_date: string;
+  actual_end_time: string;
+  actual_start_date: string;
+  actual_start_time: string;
+  beamline_name: string;
+  code: string;
+  comments: string;
```
dominikatrojanowska:
```suggestion
  comments?: string;
```
in lims_session it is set as default to empty string, so I'm not sure about this suggestion

mockoocy: Let's follow the typing from the data model.
And I made sure and `Optional[str]` should be serialized into `string | null` rather than `string | undefined`, so the proper def will be `comment: string | null`.
I've applied some of these now.

ui/src/reducers/login.ts
```diff
@@ -0,0 +1,130 @@
...
+  session_id: string;
+  start_date: string;
+  start_datetime: string;
+  start_time: string;
+  title: string;
+  user_portal_URL: string;
+  volumne?: string;
+}
+// derived from mxcubecore/core/models/usermodels.py
+interface User {
```
dominikatrojanowska, line 44: missing field `username`?

ui/src/reducers/login.ts
```diff
@@ -0,0 +1,130 @@
...
+  start_time: string;
+  title: string;
+  user_portal_URL: string | null;
+  volume: string | null;
+}
+// derived from mxcubecore/core/models/usermodels.py
+interface User {
+  currentLoginAt: string;
+  email?: string;
+  requestsControl?: false;
```
dominikatrojanowska:
```suggestion
  requestsControl?: boolean;
```

ui/src/reducers/login.ts
```diff
@@ -0,0 +1,130 @@
...
+  loginType: LoginType;
+  limsName: Lims[];
+  proposalList: Session[];
+  rootPath: string;
+  selectedProposal: string;
+  selectedProposalID: string;
+  synchrotronName: string;
+  user: User;
+  useSSO: boolean;
+}
```
dominikatrojanowska: if it's not alphabetical I'd keep it in the same order to keep it readable :)

```suggestion
export interface LoggedInSliceState extends LoginBase {
  loggedIn: true;
  useSSO: boolean;
  loginType: LoginType;
  beamlineName: string;
  limsName: Lims[];
  proposalList: Session[];
  rootPath: string;
  selectedProposal: string;
  selectedProposalID: string;
  synchrotronName: string;
  user: User;
}
```

mockoocy: I've made it alphabetical, so that it is easier to inspect now. With the exception of the discriminant property (`loggedIn`) so that it is more obvious for which case these fields appear.

dominikatrojanowska: that's why I was confused, but it's still not alphabetical (s is after r, unless there is another hidden rule letter sizes) :P

mockoocy: Well, I've applied Ascii ordering (numbers rhen capitals then lowercase) :D

ui/src/reducers/login.ts
```diff
@@ -0,0 +1,131 @@
...
+    sessionRefreshInterval: 9000,
+  }),
+  reducers: {
+    setLoginInfo(
+      _state,
+      action: PayloadAction<LoginSliceState>,
+    ): LoginSliceState {
+      return { ...action.payload };
+    },
+    signOut(state): LoginSliceState {
```
dominikatrojanowska:
```suggestion
    signOut(state): LoginUnknownSliceState | LoginKnownSliceState {
```

ui/src/reducers/login.ts
```diff
@@ -0,0 +1,131 @@
...
+  proposalList: Session[];
+  rootPath: string;
+  selectedProposal: string;
+  selectedProposalID: string;
+  synchrotronName: string;
+  useSSO: boolean;
+  user: User;
+}
+
+type LoginSliceState =
```
marcus-oscarsson, line 87: Wow ;), that one is funny

ui/src/reducers/login.js
```diff
@@ -1,61 +0,0 @@
...
-        selectedProposal,
-        selectedProposalID,
-        loggedIn,
-        rootPath,
-        limsName,
-        useSSO,
-        sessionRefreshInterval,
-      } = action.loginInfo;
-      return {
-        ...state,
```
marcus-oscarsson, line 31: Pay attention to this `SET_LOGIN_INFO` kept what was already in the store not all of that information is always in `loginInfo`. `showProposalsForm` is a client side attribute/state

ui/src/reducers/login.ts
```diff
@@ -0,0 +1,131 @@
...
+    loggedIn: null,
+    showProposalsForm: false,
+    sessionRefreshInterval: 9000,
+  }),
+  reducers: {
+    setLoginInfo(
+      _state,
+      action: PayloadAction<LoginSliceState>,
+    ): LoginSliceState {
+      return { ...action.payload };
```
marcus-oscarsson: This might be a bit tricky, you may loose things that was in the store previously like `showProposalsForm` which is client side only state, right ? I made a comment above. I'm not certain it will have an impact.

ui/src/reducers/login.js
mockoocy: Yes -- I haven't noticed that `showProposalsForm` is the one attribute that does not come from the server. I've changed type defs a bit to reflect that and rewrote action a bit to preserve it.

marcus-oscarsson:
Yes a bit tricky that state ... Thanks !

ui/src/reducers/login.ts
```diff
@@ -0,0 +1,140 @@
...
+  username: string | null;
+}
+
+export interface LoginUnknownSliceState {
+  // before first login_info comes in -- login method may be unknown.
+  loggedIn: null;
+  showProposalsForm: boolean;
+}
+
+export interface LoginKnownSliceState {
```
axelboc, line 63: `LoggedOutSliceState` maybe?

axelboc (commented):
Just a quick suggestion as I peruse all the amazing PRs that I've missed while on holidays. So much typing 🤩

PR #2152, opened 2026-07-15
Refactor UIProperties into typed RTK slice

mockoocy:


ui/src/reducers/uiproperties.ts
```diff
@@ -0,0 +1,112 @@
+import { createSlice, type PayloadAction } from '@reduxjs/toolkit';
+
+// Types mirror the pydantic models in `mxcubeweb/core/models/configmodels.py`
```
mockoocy, line 3: For a while I've thought if there are some reasonable ways to generate these from pydantic models. 

But to be honest if someone were to forget to put a field there, it's no big deal anyway I suppose.

marcus-oscarsson: I guess you could from from PyDantic model to JSONSchema to TypeScript somehow ?

marcus-oscarsson: Well, the consequence will not be that bad if we make a typo or forgot a field but its a pity to make the effort and not get it right or have any use for it. I think we should try to use one of those validation libraries as soon as possible, as long as they come with "soft fail" validation so that we at-least can see the errors. We could perhaps even foresee to stream the back to the backend ui-log ...

marcus-oscarsson:
Looks good as far as I can see. And again thanks for the great initiative !

PR #2154, opened 2026-07-20
Refactor logger related code to use RTK

dominikatrojanowska:
After reviewing some RTK refactorings related PRs I felt like I want to try also :)


marcus-oscarsson:
Looks good, thanks !

PR #2156, opened 2026-07-20
Queue handling cleaup, Part 2, Decoupled the construction of the queue from the mxcubeweb Queue component

marcus-oscarsson:
This is Part 2 of the Queue handling cleanup and continuation of https://github.com/mxcube/mxcubeweb/pull/2131. 

The aim is to move as much logic as possible back to `mxcubecore` and simplify the queue handling on both `mxcubeweb` and `mxcubecore` sides.

This PR decouples the queue construction from the `mxcubeweb` component and places it into a `QueueBuilder` object. I know, I know, its not very appealing name, and it will perhaps (likely change). For now it just serves a logical place where the queue is constructed. 

Once this is done the idea is to identify the methods that really belong to `mxcubecore`, `QueueManager` and `QueueModel`. Once that is done `QueueBuilder` (introduced in this PR) and `QueueSerializer` (introduced in  https://github.com/mxcube/mxcubeweb/pull/2131) can be decoupled from `app` (the `mxcubeweb` object and only depend on `QueueManager` and `QueueModel`). This will be done in two or three additional follow up PRs.

Once this is done both `QueueBuilder` and `QueueSerializer` can be moved back to `mxcubecore`.

NB: This mostly only moves logic from the mxcubeweb Queue component to `QueueBuilder`. There is one important functional change and that is that `QueueBuilder` uses the `child_added` event to create `QueueEntry` objects instead of creating them manually, which is a considerable cleanup.

### DIFF
--- a/mxcubeweb/core/components/queue.py
+++ b/mxcubeweb/core/components/queue.py
@@ -8,7 +8,6 @@ from unittest.mock import Mock
 
 from mxcubecore import HardwareRepository as HWR
 from mxcubecore import queue_entry as qe
-from mxcubecore.HardwareObjects.Gphl import GphlQueueEntry
 from mxcubecore.model import queue_model_enumerables as qme
 from mxcubecore.model import queue_model_objects as qmo
 from mxcubecore.model.queue_model_enumerables import CENTRING_METHOD
@@ -1009,11 +1008,12 @@ class QueueSerializer:
         return self.queue_to_dict()
 
 
-class Queue(ComponentBase):
-    def __init__(self, app, config):
-        super().__init__(app, config)
-        self.init_queue_settings()
-        self._qs = QueueSerializer(app)
+class QueueBuilder:
+    """Creates queue models the creation of queue entries are handled
+    in queue_model_child_added event handler"""
+
+    def __init__(self, app):
+        self.app = app
 
     def build_prefix_path_dict(self, path_list):
         prefix_path_dict = {}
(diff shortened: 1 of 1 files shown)


mxcubeweb/core/components/queue.py
```diff
@@ -1052,433 +1052,132 @@ def get_run_number(self, pt):
...
 
-    def node_index(self, node):
-        """Get the position (index) in the queue, sample and node id of node <node>.
+    def add_sample(self, sample_id, item):
+        """Add a sample with sample id <sample_id> the queue.
 
-        :returns: dictionary on the form:
-                {'sample': sample, 'idx': index, 'queue_id': node_id}
+        :param str sample_id: Sample id (often sample changer location)
+        :returns: SampleQueueEntry
```
mockoocy: I think it's preferable to specify the argument / return types in the function declaration rather than a docstring.

mxcubeweb/core/components/queue.py
```diff
@@ -1750,106 +1451,721 @@ def _create_queue_entry(self, task, task_name):  # noqa: D417
...
+        :returns: The queue id of the data collection
+        :rtype: int
+        """
+        sample_model, _sample_entry = self.app.queue.get_entry(node_id)
+        dc_model = self._create_dc()
+
+        group_model, _group_entry = self._create_and_enqueue_task_group(sample_model)
+        dc_entry = self._attach_model_to_group(group_model, dc_model)
+        self.set_dc_params(dc_model, dc_entry, task, sample_model)
+
```
mockoocy, line 1634: I wonder a bit, does the order of operations matter here?

https://github.com/mxcube/mxcubeweb/pull/2156/changes#diff-881eac3320962966c742cc7ac2ce302272a9749ae301a359126914c5aec70940L1908-L1915

It seems that the order previously was to 
1. create dc
2. set prams
3. create and enqueue
4. attach to group
and now params are set after attaching to group.

PR #2164, opened 2026-07-22
Refactor sampleview reducer into typed RTK slice

mockoocy:
And another one :)


ui/src/reducers/sampleview.ts
```diff
@@ -0,0 +1,217 @@
...
+      format: 'MJPEG' | 'MPEG1';
+      videoSizes: [number, number][];
+      sourceIsScalable: boolean;
+      videoHash: string;
+      videoURL: string;
+      pixelsPerMm: [number, number];
+      scale: number;
+    };
+    beamInfo: {
+      apertureList: number[];
```
marcus-oscarsson: I think this is in reality a list of strings, for instance A10, A20 ....

ui/src/reducers/sampleview.ts
```diff
@@ -0,0 +1,217 @@
...
+      currentAperture: number;
+      position: [number, number];
+      shape: string;
+      size_x: number;
+      size_y: number;
+    };
+    diffractometer: {
+      phaseList: number[];
```
marcus-oscarsson: This is also a list of strings for instance "CENTRING", "COLLECTION" ....

marcus-oscarsson:
Looks good as far as I can see. Also a bit of cleanup of unused state, very nice :)

marcus-oscarsson:
I only had the two comments about the types for `phaseList` and `apertureList`

ui/src/components/SampleView/SampleImage.jsx
```diff
@@ -535,10 +534,10 @@ class SampleImage extends React.Component {
       );
     } else if (measureDistance) {
       this.canvas.selection = false; // Disable group selection
-      this.props.addDistancePoint(
-        option.e.layerX / imageRatio,
-        option.e.layerY / imageRatio,
-      );
+      this.props.addDistancePoint({
+        x: option.e.layerX / imageRatio,
+        y: option.e.layerY / imageRatio,
```
axelboc, line 539: I love that action arguments are now named object properties <3

PR #2166, opened 2026-07-24
Migrate WaitDialog reducer into typed RTK slice

mockoocy:


ui/src/containers/PleaseWaitDialog.tsx
dominikatrojanowska, line 1: this could be easily converted to tsx

ui/src/containers/PleaseWaitDialog.tsx
```diff
@@ -1,30 +1,29 @@
...
+import { useAppDispatch, useAppSelector } from '../ts-store';
 
 function PleaseWaitDialog() {
-  const dispatch = useDispatch();
+  const dispatch = useAppDispatch();
 
-  const { show, title, message, blocking, abortFun } = useSelector(
-    (state) => state.waitDialog,
-  );
+  const dialog = useAppSelector((state) => state.waitDialog);
```
dominikatrojanowska: maybe if dialog in null do not show modal at all (return empty jsx or something more appropriate) instead of checking dialog constantly? I mean that one global check should be possible

ui/src/containers/PleaseWaitDialog.tsx
mockoocy: done :)

ui/src/containers/PleaseWaitDialog.tsx
mockoocy: done, I've also changed the store a bit, so that `message` and `blocking` are no longer optional (with default values) so we have less null checks also.

ui/src/reducers/waitDialog.ts
```diff
@@ -0,0 +1,42 @@
...
+
+interface OpenDialogState {
+  title: string;
+  message: string;
+  blocking: boolean;
+  abortFun?: () => void;
+}
+
+// The dialog is either open (holding its content) or closed (nothing at all).
+type WaitDialogState = OpenDialogState | null;
```
dominikatrojanowska: could we smash it into one? Maybe it's only my personal feeling but **Open**DialogState is a bit confusing and since is not used anywhere else, what about this more straightforward version?
```suggestion
type waitDialogSlice = {
  title: string;
  message: string;
  blocking: boolean;
  abortFun?: () => void;
} | null
```

dominikatrojanowska:
just one comment more, everything else looks good to me :) :+1:

ui/src/reducers/waitDialog.ts
```diff
@@ -0,0 +1,42 @@
...
+    ): WaitDialogState => {
+      const {
+        title,
+        message = '',
+        blocking = false,
+        abortFun,
+      } = action.payload;
+      return { title, message, blocking, abortFun };
+    },
+    hideWaitDialog: (): WaitDialogState => null,
```
marcus-oscarsson: This one might be a bit tricky, we used to pass `show` to `Modal` which uses the flag internally to show or hide the dialog. We are now by-passing this by an early return which means that any animation or other internal state that of Modal that is dependent on `show ` will no longer be triggered. 

I don't think its a big deal, but we will lose any animation that `Modal` provides. I would add a `visible` or `show` state and toggle it to `false` or `true`

mockoocy:
Nice rermark about the animations.
I've changed the shape of the state now so that the dialog contents are still present in case a dialog is being closed, so that they are still visible during a "fade out'. 
The early return is still there, but only in case modal was never shown before.

marcus-oscarsson:
Well done !

PR #2169, opened 2026-07-27
Migrate remote-access actions to Redux Toolkit slice

dominikatrojanowska:
Route SET_RA_STATE/SET_ALLOW_REMOTE/ADD_CHAT_MESSAGE dispatches through the new reducers/remoteAccess slice actions instead of hand-rolled action objects, and export User/LoginSliceState types from login.ts so the slice can reference them. Also chatMessages is migrated to TypeScript and introduces two new types: Message and MessageEntry.


ui/src/reducers/remoteAccess.ts
```diff
@@ -0,0 +1,93 @@
+/* eslint-disable no-param-reassign */
+import { createSlice, type PayloadAction } from '@reduxjs/toolkit';
+
+import {
+  type Message,
+  type MessageEntry,
+  processChatMessageRecord,
+} from '../components/ChatComponent/chatMessages';
```
mockoocy: I think the types should be defined in the slice rather than the module contaning the component.

dominikatrojanowska: hm, I wanted to avoid circular imports, but as it's only type imported I can move it to slice.

ui/cypress/support.js
```diff
@@ -9,6 +9,19 @@ beforeEach(() => {
...
+  cy.get('body').then(($body) => {
+    if ($body.find('button:contains("Sign out")').length > 0) {
+      cy.findByRole('button', { name: /Sign out/u }).click();
+      cy.findByRole('button', { name: 'Sign in with proposal' }).should(
+        'be.visible',
+      );
+    }
+  });
+});
+
```
dominikatrojanowska, line 24: I included changes from [this PR](https://github.com/mxcube/mxcubeweb/pull/2171) here and else so it could be merged separately or within current PR. I found those dangling users problematic when testing remote access operations.

ui/src/reducers/remoteAccess.js
```diff
@@ -1,61 +0,0 @@
...
-      return {
-        ...state,
-        observers: action.data.observers,
-        allowRemote: action.data.allowRemote,
-        operator: action.data.operator,
-      };
-    }
-    case 'SET_MASTER': {
-      return { ...state, sid: action.sid };
-    }
```
dominikatrojanowska, line 26: not used - removed

marcus-oscarsson:
Are the tests still in WIP or did I misunderstand your comment ?

ui/src/reducers/remoteAccess.ts
```diff
@@ -0,0 +1,110 @@
...
+      });
+    },
+  },
+  extraReducers: (builder) => {
+    builder.addCase(
+      'SET_INITIAL_STATE',
+      (_state, action: SetInitialStateAction) => {
+        const { username } = action.data.login.user;
+        const messages = username
+          ? action.data.chatMessages.messages.map((record) =>
```
marcus-oscarsson: Watch out with this, we used to do `action.data.chatMessages?.messages?` are you sure that these can still not be undefined ?

ui/src/reducers/remoteAccess.js
```diff
@@ -1,61 +0,0 @@
...
-    case 'MARK_ALL_READ': {
-      return {
-        ...state,
-        messages: state.messages.map((msg) => ({ ...msg, read: true })),
-      };
-    }
-    case 'SET_INITIAL_STATE': {
-      const messages =
-        action.data.chatMessages?.messages?.map((entry) =>
```
marcus-oscarsson, line 44: See comment below ...

ui/src/components/SampleView/jsmpeg.min.js
```diff
@@ -823,7 +823,7 @@ JSMpeg.Source.WebSocket = (function () {
   WSSource.prototype.destroy = function () {
     clearTimeout(this.reconnectTimeoutId);
     this.shouldAttemptReconnect = false;
-    this.socket.close();
+    this.socket?.close();
```
marcus-oscarsson, line 826: Hm, not sure that this is related to this PR ?

marcus-oscarsson:
Thanks, well done @dominikatrojanowska. It looks good over all, just two comments for:
 
 - `action.data.chatMessages?.messages?` just make sure that none of these can actually be undefiend, otherwise use `?.`
 - For the jsmpeg fix, it looks like it fixes an issue but its also done in the minified/distributed  of JSMpeg version which I guess is not really a problem because we are "shipping" it with MXCuBE. Is this needed for the tests perhaps ?

Really nice to add the tests as well super :+1:. I'm happy to merge this if we can confirm the first point ?

ui/src/components/SampleView/jsmpeg.min.js
dominikatrojanowska: yes, this is needed for tests. After merging https://github.com/mxcube/mxcubeweb/pull/2171/changes I will rebase this PR and it will be more clear :)

dominikatrojanowska:
> Are the tests still in WIP or did I misunderstand your comment ?

Oh, I put it to draft first and marked that only tests are WIP, then removed draft flag and forget to update the description. I've just removed this WIP part.

dominikatrojanowska:
> * `action.data.chatMessages?.messages?` just make sure that none of these can actually be undefiend, otherwise use `?.`

The chatMessages is populated at the very first request to [/ra/chat/](https://github.com/mxcube/mxcubeweb/blob/develop/mxcubeweb/routes/ra.py#L178) endpoint during `SET_INITIAL_STATE` action at the login of the first user. Underlying [get_all_messages](https://github.com/mxcube/mxcubeweb/blob/develop/mxcubeweb/core/components/chat.py#L42) method returns `[]` by default.

image

The only possible chance to have undefined there is during fetch failure :thinking: I will add the `?` then :)

PR #2171, opened 2026-07-29
Add conditional clean up after each e2e spec to avoid dangling logged in users

dominikatrojanowska:


ui/cypress/support.js
```diff
@@ -9,6 +9,19 @@ beforeEach(() => {
   cy.visit('/');
 });
 
+afterEach(() => {
+  // Only log out if a session is actually active, so specs that never
```
marcus-oscarsson, line 13: I see ok,

marcus-oscarsson:
Looks good, great !. Is this the followup of https://github.com/mxcube/mxcubeweb/pull/2169 ?

dominikatrojanowska:
> Looks good, great !. Is this the followup of #2169 ?

rather a necessary step for tests in 2169, so should be merged before or within :)

PR #2173, opened 2026-07-31
Removed 'abortFun' from WaitDialog reducer

marcus-oscarsson:
To be merged on top of https://github.com/mxcube/mxcubeweb/pull/2166

As I don't have access to @mockoocy fork I can't use the branch `mockoocy:pm-typed-waitdialog-reducer` as a base for the PR. This simply suggests a way to get rid of `abortFun`, its not very elegant but it will work nicely as we are actually only using two abort functions: cancelControlRequest' and 'stopQueue'.

### DIFF
--- a/ui/src/containers/PleaseWaitDialog.tsx
+++ b/ui/src/containers/PleaseWaitDialog.tsx
@@ -1,7 +1,18 @@
 import { Button, Modal, ProgressBar } from 'react-bootstrap';
 
-import { hideWaitDialog } from '../reducers/waitDialog';
+import { stopQueue } from '../actions/queue';
+import { cancelControlRequest } from '../actions/remoteAccess';
+import {
+  hideWaitDialog,
+  type WaitDialogAbortAction,
+} from '../reducers/waitDialog';
 import { useAppDispatch, useAppSelector } from '../ts-store';
+
+// Set of well-known abort actions that can be triggered from the wait dialog.
+const ABORT_ACTIONS: Record<WaitDialogAbortAction, () => () => void> = {
+  cancelControlRequest,
+  stopQueue,
+};
 
 function PleaseWaitDialog() {
   const dispatch = useAppDispatch();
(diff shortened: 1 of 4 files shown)


ui/src/containers/PleaseWaitDialog.tsx
```diff
@@ -1,7 +1,19 @@
...
+  type WaitDialogAbortAction,
+} from '../reducers/waitDialog';
 import { useAppDispatch, useAppSelector } from '../ts-store';
+import { UnknownAction } from '@reduxjs/toolkit';
+
+// Set of well-known abort actions that can be triggered from the wait dialog.
+const ABORT_ACTIONS: Record<WaitDialogAbortAction, () => () => void> = {
+  cancelControlRequest,
+  stopQueue,
+};
```
mockoocy, line 15: I have changed the signature for values from `() => any` to `() => () => void` during rebase.

ui/src/reducers/waitDialog.ts
```diff
@@ -43,11 +50,13 @@ const waitDialogSlice = createSlice({
...
         message = '',
         blocking = false,
-        abortFun,
+        abortAction,
       } = action.payload;
-      return { show: true, dialog: { title, message, blocking, abortFun } };
+      return { show: true, dialog: { title, message, blocking, abortAction } };
+    },
+    hideWaitDialog(state) {
+      state.show = false;
```
mockoocy, line 58: In the previous PR we have discussed that we should preserve `dialog` state when the dialog is being closed, so that these are not empty during a "fade-out".


I've hijacked this PR and added one small commit with the above change :)

mockoocy (commented):
Since https://github.com/mxcube/mxcubeweb/pull/2166 is merged, I've made a rebase.

dominikatrojanowska:
LGTM :+1:

ui/src/reducers/waitDialog.ts
marcus-oscarsson: Sure, you did well !

marcus-oscarsson:
Thanks well done, looks good !

PR #2174, opened 2026-08-03
Migrate sample changer state to a typed Redux Toolkit slice

dominikatrojanowska:
Replaces the plain sampleChanger reducer/actions with sampleChanger.ts (RTK slice, typed PlateGrid/SampleChanger state) and splits global-state handling into actions/sampleChangerMaintenance.js. Also updates setSelectedWell to take a single { row, col } payload instead of two positional args, which RTK would have silently dropped.


ui/cypress/e2e/sampleChanger.cy.js
```diff
@@ -0,0 +1,115 @@
...
+    cy.wait(['@fetchSampleChangerContents', '@fetchLoadedSample']).then(
+      ([contentsCall]) => {
+        const contents = contentsCall.response.body;
+        expandAllDropdowns();
+
+        // Check every sample from the response is rendered
+        contents.children.forEach((container) => {
+          container.children.forEach((sample) => {
+            cy.findByText(`${sample.name} ${sample.id}`).should('exist');
+          });
```
walesch-yan, line 113: If you only check for the existence here, it does not matter if you expand the dropdowns or not before, because the element would exist behind/within the dropdown, and hence Cypress will always pass this test. (you can test this by commenting out the `expandAllDropdowns()` call before and you should see the tests still passing) 
However, If you would like to make sure, that the samples are visible to the user within the container (as I would recommend doing it as well) you could replace the content of the second `forEach` with:
```js
  cy.findByText(`${sample.name} ${sample.id}`)
    .should('exist')
    .scrollIntoView()
    .should('be.visible');
```

ui/src/reducers/sampleChanger.ts
```diff
@@ -0,0 +1,189 @@
...
+  dropHeight: number;
+  dropWidth: number;
+  rowTitle: string[];
+  colTitle: number[];
+  numberOfDrops: number;
+  type: string;
+  title: string;
+  wellOption: WellOption[];
+  rotation: number;
+}
```
walesch-yan, line 42: Maybe I'm missing something, but shouldn't `PlateGrid` also have a name field? As I understand it, the function `_setPlate` in `PlateManipulator.jsx` needs the plate name to be able to switch between the plates. If I'm correct, you should also add a name for each of the default plates

ui/src/components/Equipment/PlateManipulator.jsx
```diff
@@ -85,17 +83,13 @@ export default function PlateManipulator(props) {
       }
     });
 
-    dispatch(setPlate(plate_index));
+    dispatch(setCurrentPlate(plate_index));
   }
 
   useEffect(() => {
     _setPlate();
   });
```
walesch-yan: Just seeing this part of the component in the PR, maybe we should run the `useEffect` only when the `plate_label` changes. Also, because the name field is omitted in the `PlateGrid` interface and in the initial state values, shouldn't `_setPlate` always set the index to 0?

walesch-yan (commented):
Generally looks good to me, just added my two cents:)

ui/src/reducers/sampleChanger.ts
dominikatrojanowska: hm, I deliberately removed name from these objects... I will try to solve this other way, thanks for pointing :)

ui/src/components/Equipment/PlateManipulator.jsx
dominikatrojanowska: Ok, I will implement the first part to call `_setPlate` on the initial mount and after `plate_label` changes. But I'm not sure if I understood the second part... Tell me if this is correct after force pushing a new commit :)

walesch-yan: Yeah, my comment was a little confusing; basically, what I meant was:
Your PR removes the name from the plates. However, the old `_setPlate` function searches for the current plate using the value of the name, which should now always be `undefined`. Hence, the old version always returned the default index value, i.e., `0`, and set the plate at position `0`. So it is essentially just what I meant in the other comment as well

I think your new version should solve this issue; at least it looks good to me. Thanks! :)

walesch-yan (approved):
Thanks, this looks very nice, especially with the added tests!:)

I think there is only a merge conflict in the `serverIO.js` file to resolve, if you would like to do this?

PR #2175, opened 2026-08-03
Add dose estimation UI

mockoocy:
Requires https://github.com/mxcube/mxcubecore/pull/1643 to be merged first.

It adds UI for dose estimation, dose estimation logic is being handled entirely by derivatives of `AbstractDoseEstimator` hardware object on mxcubecore side.

### DIFF
--- a/demo.yaml/beamline.yaml
+++ b/demo.yaml/beamline.yaml
@@ -48,6 +48,7 @@ objects:
   - sample_view: sample_view.yaml
   - queue_manager: queue_manager.yaml
   - queue_model: queue_model.yaml
+  - dose_estimator: dose_estimator.yaml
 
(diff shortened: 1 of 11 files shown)


mockoocy:
Tests fail, but that's expected (it needs the mxcubecore PR to be mergedd first)

marcus-oscarsson:
Thanks, really nice that you guys made a PR with this. I/we think what you presented on the last MXCuBE meeting looked really promising. We would just need to have another look here, but as far as the concept goes I think we can merge this more or less as it is and then possibly work from there for any smaller adjustments.

marcus-oscarsson:
Would you mind attaching some screenshots on how it looks like in the mean time ?

PR #2176, opened 2026-08-04
Refactor SampleGrid Reducer into typed RTK slice

mockoocy:


ui/src/reducers/sampleGrid.ts
```diff
@@ -0,0 +1,357 @@
...
+  name: string | null;
+  originID?: number;
+  parameters: TaskParameters;
+  result?: unknown;
+  sampleID: string;
+  sampleQueueID: number | null;
+  taskIndex: number | null;
+}
+
+export interface Sample extends QueueNode {
```
mockoocy, line 32: these interfaces are based on models from mxcubecore, i.e.
https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/model/queue_model_objects.py

ui/src/reducers/queueGUI.ts
```diff
@@ -1,6 +1,8 @@
 /* eslint-disable no-param-reassign */
 import { createSlice, type PayloadAction } from '@reduxjs/toolkit';
 
+import { type SampleList, type TaskNode } from './sampleGrid';
+
```
mockoocy, line 5: so that they all are in the same place.

ui/src/reducers/sampleGrid.ts
```diff
@@ -0,0 +1,353 @@
...
+  sampleName: string;
+  sample_barcode?: string;
+  sc_state?: number;
+  smiles?: string | null;
+  tasks: TaskNode[];
+}
+
+export type SampleList = Record<string, Sample>;
+
+export interface Crystal {
```
mockoocy, line 69: same as above, it comes from mxcubecore.

it feels a bit awkward to it to be honest, since front-end does not really depend on most of these fields, but it feels a bit simpler to do it this way.

ui/src/reducers/sampleGrid.ts
```diff
@@ -0,0 +1,353 @@
...
+    ) {
+      state.sampleList[action.payload.sampleID].sc_state = action.payload.state;
+    },
+    setSampleAttribute(
+      state,
+      action: PayloadAction<{
+        sampleIDList: string[];
+        attr: 'checked';
+        value: boolean;
+      }>,
```
walesch-yan: It looks a little weird in this case to have a parameter whose intended value is basically fixed. Maybe we should change the reducer name to `setSampleChecked` and remove the parameter?

walesch-yan (approved):

ui/src/reducers/sampleGrid.ts
mockoocy: Oh yeah -- a good idea, done ;)

walesch-yan:
Thanks @mockoocy

PR #2179, opened 2026-08-06
Adapted Pydantic queue handling to match GPhL workflow, and a bug fix

rhfogh:
Changes to fix GPhL-workflow broken with new Pydantic-based queue handling.

NB: two change are not specific for GPhL:

The new validator for the label attribute of workflow parameters allows greek characters and parentheses in labels, which is used by GPhL. Claude promises that any characters with potentially dangerous implications are still banned.

The compression attribute of the PathData model is changed from str to bool. The relevant attribute is a boolean in all the code I could see.

### DIFF
--- a/mxcubeweb/core/components/queue.py
+++ b/mxcubeweb/core/components/queue.py
@@ -147,7 +147,7 @@ class TaskDataPathModel(BaseModel):
     precision: int = 0
     start_num: int = 0
     num_files: int = 0
-    compression: str = ""
+    compression: bool = False
     path: str = ""
     prefix: str = ""
 
--- a/mxcubeweb/core/components/queue.py
+++ b/mxcubeweb/core/components/queue.py
@@ -762,7 +762,7 @@ class QueueSerializer:
         parameters["path"] = parameters["directory"]
         parameters["subdir"] = self._subdir_from_path(parameters["path"])
         parameters["strategy_name"] = node.strategy_name
-        parameters["label"] = f"GΦL {parameters['strategy_name']}"
+        parameters["label"] = f"GPhL {parameters['strategy_name']}"
         parameters["shape"] = node.shape
         parameters["fileName"] = pt.get_image_file_name().replace(
             "%" + ("%sd" % str(pt.precision)), int(pt.precision) * "#"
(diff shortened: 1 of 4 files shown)


mxcubeweb/core/components/queue.py
```diff
@@ -69,6 +69,25 @@ def validate_safe_string(value: str | None, field_name: str) -> str | None:
...
+    """
+    if value is None:
+        return None
+
+    if not isinstance(value, str):
+        raise ValueError("label must be a string")
+
+    value = value.strip()
+
+    invalid_char = re.search(r"[^\w:+(). -]", value)
```
walesch-yan: Just wondering: How many different characters could appear in the strategy names? Is it only Φ, or others as well?

Generally, including all non-ASCII characters could introduce a spoofing opportunity for an attacker, so it is always nice to double-check, but I think in this case we should be fine

rhfogh: The only one currently is the Φ. And the parentheses, of course. Clearly I could replace GΦL by GPhL, and I could manage to get around the parentheses as well, so if people feel it is a problem this part is not crucial. 
What really threw me is that the illegal characters were simply ignored silently, so that I could not understand why the values did not appear where they should - it took Claude to inform me.

marcus-oscarsson: You mean that you never got the `ValueError(f"label contains invalid character:...` exception ?, thats strange in that case. The queue entry should not have been added at all.

I'm on the same track as @walesch-yan, I would not allow other than ASCII characters. We could add the letter  Φ and parenthesis explicitly if needed. Then again this is for a label so the risk of spoofing or that this ends up in a software that does not handle unicode are quite small.

We can go with these changes, If you think that Φ is very important in your label. But I'm curious why you did not get the exception.

rhfogh: I am also curious why I did not get the exception. I could have just missed it, I suppose. But given my personal relationship to Javascript this kind of behaviour did not really surprise me ;-).

Consensus seems to be with limiting ourselves to ASCII in labels, so I shall make the relevant change to the PR - just as soon as I am over the jetlag.

PR #2180, opened 2026-08-10
Fix: non-exstensible shape object:

dominikatrojanowska:
`selectShape` method is invoked from two different sources: canvas events (e.g. right-click on a shape) and the `GridForm` component's row selection. In the first case, the shape parameter is a fabric.js canvas object, which works fine. In the second case, it's a frozen Redux state object, which caused a crash at the `shape.active = ...` assignment — defining a new active property on a frozen object throws:
Uncaught TypeError: can't define property active: Object is not extensible

### DIFF
--- a/ui/src/components/SampleView/SampleImage.jsx
+++ b/ui/src/components/SampleView/SampleImage.jsx
@@ -666,10 +666,14 @@ class SampleImage extends React.Component {
 
       if (shapeData && include) {
         const selected = !shapeData.selected;
-        shape.active = selected;
+        if (Object.isExtensible(shape)) {
+          shape.active = selected;
+        }
         updatedShapes.push({ ...shapeData, selected });
       } else if (shapeData && !shapeData.selected) {
-        shape.active = true;
+        if (Object.isExtensible(shape)) {
+          shape.active = true;
+        }
         updatedShapes.push({ ...shapeData, selected: true });
       }
     });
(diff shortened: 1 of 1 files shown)


ui/src/components/SampleView/SampleImage.jsx
```diff
@@ -666,10 +666,14 @@ class SampleImage extends React.Component {
 
       if (shapeData && include) {
         const selected = !shapeData.selected;
-        shape.active = selected;
+        if (Object.isExtensible(shape)) {
+          shape.active = selected;
+        }
```
walesch-yan, line 671: I'm not sure that we even need to set the active field of the shapes here? Shouldn't it be enough to push the updateShapes, as `active` is changed for the fabric objects in `renderSampleView` and is later only ever used during a right-click event? However, since we update the Redux state in this function, `renderSampleView` should be triggered directly afterwards.

dominikatrojanowska: hm, not sure... `updateShapes` is an asynchronous function (eee, redux thunk, righ?) calling backend API, so there is a small period of time between selecting grid and updating a fabric objects / rerendering . Although it works fine with demo 🤔

dominikatrojanowska: maybe it's nothing to worry about - this property is only read by `rightClick` method afaic... So in the worse scenario the context menu won't be displayed properly but it should re-render correctly after second right click I think

walesch-yan: Yes, you are right, of course; I did not think that far. In a real environment, this time gap could be important

walesch-yan (approved):

PR #2184, opened 2026-08-13
Fix crash when ui_schema is undefined.

mockoocy:
Current code tries to foresee a situation where data_model.ui_schema is not defined. The fallback value `{}` does not work, as frontend expects a stringified json to parse.

### DIFF
--- a/mxcubeweb/core/components/queue.py
+++ b/mxcubeweb/core/components/queue.py
@@ -1,5 +1,6 @@
 import contextlib
 import itertools
+import json
 import logging
 import os
 import re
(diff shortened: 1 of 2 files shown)


mxcubeweb/core/components/queue.py
```diff
@@ -613,8 +614,6 @@ def pretty_print_queue(self, msg: str | None = None) -> None:
                 debug(q)
                 return
             except Exception:
-                import json
-
```
mockoocy, line 617: Okay, now I checked it again and here is a reason for the "missing" import.
`json` is conditionally imported (hence my IDE did not complain either). I think on my machine it just import `json` there. I've moved the import to the top of the file.

PR #2090, opened 2026-04-28
Remove unused TIMEOUT_GIVES_CONTROL feature.

dominikatrojanowska:
As agreed in discussion: https://github.com/mxcube/mxcubecore/pull/1115/changes#r1965220727.
Related:
* MR: https://github.com/mxcube/mxcubecore/pull/1598
* issue: https://github.com/mxcube/mxcubeweb/issues/1588

Should be merged after solving poetry dependencies: https://github.com/mxcube/mxcubeweb/pull/2091

PR #2186, opened 2026-08-14
refactor logging config options, change command line

epanepucci:
- ATTENTION: renamed short argparse flag from "-el" to "-e" as this is very unnatural and may eventually conflict in case we ever add a "-e" option since there is already a "-l"
- refactored logging handler names and map into the logging_handler consolidating the source of names into one place
- the "--enabled-loggers" has also been updated so it checks for typos by using "choices"

### DIFF
--- a/mxcubeweb/__init__.py
+++ b/mxcubeweb/__init__.py
@@ -18,7 +18,7 @@ from mxcubecore import HardwareRepository as HWR  # noqa: E402
 from mxcubeweb.app import MXCUBEApplication as mxcube  # noqa: E402
 from mxcubeweb.config import Config  # noqa: E402
 from mxcubeweb.core.models.configmodels import RuntimeOptions  # noqa: E402
-
+from mxcubeweb.logging_handler import LOGGER_NAMES  # noqa: E402
 from mxcubeweb.core.server.server import create_server  # noqa: E402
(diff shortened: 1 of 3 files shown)


mxcubeweb/__init__.py
```diff
@@ -96,6 +85,15 @@ def parse_args(argv):
         default=False,
     )
 
+    opt_parser.add_argument(
+        "-t",
+        "--ra-timeout",
```
marcus-oscarsson, line 90: This also seems to add --ra-timeout back, which was removed a few months ago (https://github.com/mxcube/mxcubeweb/pull/2090/changes). Do you finally need it ?

epanepucci: this argument is still being used on our branch

marcus-oscarsson: Hm, I see. Because we agreed to remove this option and the PR was made by @dominikatrojanowska (https://github.com/mxcube/mxcubeweb/pull/2090). 

Are you using the option ? The switch can perhaps otherwise be removed ?. As it is now this would only add the command line switch but the feature does not exist, as it was removed previously, so It wont actually do anything ;)

epanepucci: I see, it's because our mxcubeweb is behind upstream. I'll put it in draft and then update this once we're rebased.

PR #2188, opened 2026-09-03
Only show equipment tab for staff users

beebyen:
We have a request from MAXIV beamline to hide the equipment tab from users to prevent confusion / accidents. This is a simple update to allow only staff to see this page.


ui/src/components/MXNavbar/MXNavbar.jsx
```diff
@@ -11,7 +11,9 @@ import styles from './MXNavbar.module.css';
 function MXNavbar() {
   const isUserLogin = useSelector((state) => state.login.loginType === 'User');
   const selectedProposal = useSelector((state) => state.login.selectedProposal);
-  const { nickname, fullname } = useSelector((state) => state.login.user);
+  const { nickname, fullname, isstaff } = useSelector(
```
marcus-oscarsson, line 14: Hm, why not I'm not sure its something we would like. So in that case we would have to set all users to staff for this purpose which perhaps not a good idea long-term. Which makes me think that we should maybe add a third category of users. 

How are you populating the staff property ? from the configuration file ?

beebyen: It's a pre-existing property from the user model.

marcus-oscarsson:
> I will take a look at making it an option and update soon. Is there a pre-existing way that we add UI options like this?

Yes, there is away, you simply just have to extend the corresponding UI model and then use that attribute in the UI. 

However I don't think that the `equipment` view has a corresponding UI section. We would need to add a `equipment` section here I, https://github.com/mxcube/mxcubeweb/blob/37fdbe6a2c0a01677658f4cbc67d879205dc7e9e/demo/mxcube-web/ui.yaml#L5. Similar to `sample_view` or `beamline_setup` to start with it can simply have that one option, `enable_for_all_users`, or `enable_only_for_staff`. 

I think we can start with that. Let us know if you need any help with that ?

beebyen:
Can you take a look at what I added ? I think it aligns with what you said.

marcus-oscarsson:
> Can you take a look at what I added ? I think it aligns with what you said.

I'm sorry I thought the change was simply the comment. Sorry about that. Yes that looks good (great minds think a like:) )

marcus-oscarsson:
This looks fine, we are not testing this option in Cypress, I guess that could be an addition. And we should keep in mind that this will only hide the equipment view, access to the equipment will still be possible via the API as we are not checking the roles in the access control. I think this PR is fine if thats what's intended ? :)

beebyen:
I think that should be fine, I'll double check and then add the tests.

beebyen:
> I think that should be fine, I'll double check and then add the tests.

All good on our end. Also added the tests as requested!

marcus-oscarsson:
Very nice, thanks !

